PR #22, opened 2020-05-26
feat: add api common protos for future baseline test

xiaozhenliu-gg5:
1. add test-fixtures/protos/api-common-protos which has all the common protos
2. add showcase proto in the `test-fixtures/protos/google/showcase`
3. update develop.md

When we have needs to add baseline test, we can just add protos in `test-fixtures/protos/google` instead of downloading the whole `googleapis` (since we have common protos in place), and easily compare the outcome with baseline data (desired generated java code). we will have a baseline folder to store the generated client library, also baseline test script to run protoc command and do the comparison.


miraleung:
If these files are being copied from other GitHub repos, can we add them using [git submodules](https://git-scm.com/book/en/v2/Git-Tools-Submodules) instead? That way, we can avoid duplicating sources and keep up to sync with those repos as they get updated.

miraleung:
I'm also not sure we need these submodules. What's the motivation for adding them now? Generally deps are added on an as-needed basis.

xiaozhenliu-gg5:
I actually was not sure about the change, just wanted to set up the baseline test that does the file-diffing, also played with bazel. But definitely we can download the `googleapis/` folder in the current way, I will get some understanding of Bazel from this PR, and close it afterwards.

PR #27, opened 2020-05-28
[ggj][build] add header check to precommit hook

xiaozhenliu-gg5:
1. add license headers to all source files
2. add header check to precommit hook
3. add header check to circleCI workflow

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,3 +1,17 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 version: 2.1
 
 # ======================= JOBS =======================
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,3 +1,17 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 version: 2.1
 
 # ======================= JOBS =======================
@@ -26,7 +40,17 @@ jobs:
           name: Java Linter
           command: |
             bazel build //:google_java_format_verification
-
+  license-header:
+    docker:
+      -image: circleci/golang:1.13
+    working_directory: /home/circleci/workspace/gapic-generator-java
+    steps:
+      - checkout
+      - run:
+          name: License Header
+          command: |
+            go get -u github.com/google/addlicense
+            addlicense -c "Google LLC" -l apache .
 # ======================= WORKFLOWS =======================
 
 workflows:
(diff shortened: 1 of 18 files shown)


xiaozhenliu-gg5:
Two minor questions: 
1. Not sure we wanna header for circleCI/config.yaml though. 
2. do we expect to throw error if some source files do not have license or we can just directly add it for them

I tested locally that first removing the header of config.yml and run `git commit`, `pre-commit` will check and automatically add the license back. Thank you!

.githooks/pre-commit
```diff
@@ -101,4 +101,18 @@ then
...
+# Check Apache License header for source files.
+if [ $NUM_TOTAL_FILES_CHANGED -gt 0 ]
+then
+  echo_status "Checking Apache License Header ..."
+  go get -u github.com/google/addlicense
+  addlicense -c "Google LLC" -l apache .
+  FORMAT_STATUS=$?
+  if [ $FORMAT_STATUS != 0 ]
+  then
+    echo_error "License header check failed." "Please make sure adding the latest header to the source files."
```
miraleung, line 113: Nit: "Please ensure that all source files have a license header." or similar.

.githooks/pre-commit
```diff
@@ -101,4 +101,18 @@ then
   done
 fi
 
+# Check Apache License header for source files.
+if [ $NUM_TOTAL_FILES_CHANGED -gt 0 ]
+then
+  echo_status "Checking Apache License Header ..."
+  go get -u github.com/google/addlicense
+  addlicense -c "Google LLC" -l apache .
```
miraleung, line 109: Could you please confirm that this doesn't actually add the headers, just does a check?

miraleung, line 109: Can we run this check only on Java files?

.githooks/pre-commit
```diff
@@ -101,4 +101,18 @@ then
   done
 fi
 
+# Check Apache License header for source files.
+if [ $NUM_TOTAL_FILES_CHANGED -gt 0 ]
```
miraleung, line 105: $NUM_JAVA_FILES_CHANGED

.githooks/pre-commit
xiaozhenliu-gg5: thanks, done.

.githooks/pre-commit
xiaozhenliu-gg5: yes, `-check` enables the check mode. Thanks

.githooks/pre-commit
xiaozhenliu-gg5: done.

.githooks/pre-commit
```diff
@@ -101,4 +101,18 @@ then
   done
 fi
 
+# Check Apache License header for source files.
+if [ $NUM_JAVA_FILES_CHANGED -gt 0 ]
+then
+  echo_status "Checking Apache License Header ..."
+  go get -u github.com/google/addlicense
```
miraleung, line 108: Could we add some extra logic to do the following? We can pull this out into a separate method.

- Check whether the current system has a go path, e.g. ~/go/bin
- If not, print a link to set-up or installation steps and exit
- Otherwise, check if the go path is in the system path.
  - If not, add the go path to the system path.
- Check if addlicense exists and install addlicense only if it's not there.

.circleci/config.yml
```diff
@@ -26,7 +40,17 @@ jobs:
...
+    docker:
+      - image: circleci/golang:1.13
+    working_directory: /home/circleci/workspace/gapic-generator-java
+    steps:
+      - checkout
+      - run:
+          name: License Header
+          command: |
+            go get -u github.com/google/addlicense
+            addlicense -c "Google LLC" -l apache .
```
miraleung, line 53: This should be a check on Java-only files as well, so we don't accidentally modify files that shouldn't contain the header.

.githooks/pre-commit
```diff
@@ -101,4 +101,18 @@ then
...
+  addlicense -c "Google LLC" -l apache -check .
```
miraleung, line 109: I think this can be a Java-only check without pattern matching - please take a look at the [find](https://ss64.com/bash/find.html) command.

.githooks/pre-commit
xiaozhenliu-gg5: sure, working on it.

xiaozhenliu-gg5: done.

.githooks/pre-commit
xiaozhenliu-gg5: done, thanks for the information!

.circleci/config.yml
xiaozhenliu-gg5: done.

PR #29, opened 2020-06-10
[ggj][ast] chore!: add writer visitor and restructure packages to fix dep cycle

miraleung:
Adding the start of the visitor pattern for serialization. This also introduced some bazel cyclic dependencies, so I'm moving some files (which fit better in `ast` anyway) over to the `ast` package.


src/main/java/com/google/api/generator/engine/ast/BUILD.bazel
```diff
@@ -10,4 +10,13 @@ java_library(
     srcs = [
         ":ast_files",
     ],
+    deps = [
+        "//src/main/java/com/google/api/generator:autovalue",
+        "//src/main/java/com/google/api/generator/engine/lexicon",
+        "@com_google_auto_value_auto_value//jar",
+        "@com_google_auto_value_auto_value_annotations//jar",
```
xiaozhenliu-gg5, line 17: can I ask why we need `auto_value` dependencies here, since we have not used @auto_value in ../ast/*.java files? Thanks!

miraleung: We moved some files that use AutoValue over to the `ast` package due to the cyclic dependencies (and they make more sense here anyway).

miraleung:
Updated the comment.

PR #47, opened 2020-06-17
[ggj] feat: add Reference

miraleung:


src/main/java/com/google/api/generator/engine/ast/Reference.java
```diff
@@ -0,0 +1,92 @@
...
+        }
+      }
+      sb.append(">");
+    }
+    return sb.toString();
+  }
+
+  public Set<String> fullNames() {
+    Set<String> allNames =
+        Stream.<String>of(clazz().getCanonicalName()).collect(Collectors.toSet());
```
xiaozhenliu-gg5, line 57: For getting allNames, if I'm understanding correctly, can we do this?
```
Set<String> allNames = new HashSet<>(Arrays.asList(clazz().getCanonicalName())); 
```

miraleung: That is also an option. It turns out that this method is not needed, so I've removed `fullNames()`.

xiaozhenliu-gg5: 👍

PR #50, opened 2020-06-18
[ggj] fix: TypeNode integration of Reference, hashCode, cleanup

miraleung:


src/main/java/com/google/api/generator/engine/ast/TypeNode.java
```diff
@@ -60,22 +68,10 @@ public static Builder builder() {
...
   }
 
-  public static TypeNode createReferenceArrayType(Reference reference) {
-    return TypeNode.builder()
-        .setTypeKind(TypeKind.OBJECT)
-        .setReference(reference)
-        .setIsArray(true)
-        .build();
-  }
-
```
xiaozhenliu-gg5, line 74: curious why we want to remove this method `createReferenceArrayType `?  it can be used to initialize STRING_ARRAY? Thanks

miraleung: There is currently no frequent need (aside from one-offs) for array creation. Removing unused / one-off methods is part of API best practices to help maintainability.

PR #51, opened 2020-06-18
[ggj] feat: add ExprStatement

miraleung:


src/main/java/com/google/api/generator/engine/ast/Statement.java
```diff
@@ -0,0 +1,19 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.api.generator.engine.ast;
+
+public interface Statement extends AstNode {
+  void accept(AstNodeVisitor visitor);
+}
```
xiaozhenliu-gg5, line 19: For AstNode interface, it also has one accept method, can we reuse it? Thanks

miraleung: Statement is a subtype of AstNode, and re-specifying the accept method here will let implementing classes clearly see all the methods that need to be implemented.

PR #54, opened 2020-06-18
[ggj] feat: add TryCatchStatement

miraleung:


src/test/java/com/google/api/generator/engine/ast/TryCatchStatementTest.java
```diff
@@ -0,0 +1,143 @@
...
+    Reference exceptionReference = Reference.withClazz(IllegalArgumentException.class);
+    TypeNode type = TypeNode.withReference(exceptionReference);
+    VariableExpr variableExpr =
+        VariableExpr.builder().setVariable(createVariable("e", type)).setIsDecl(true).build();
+
+    TryCatchStatement tryCatch =
+        TryCatchStatement.builder()
+            .setTryBody(Arrays.asList(ExprStatement.withExpr(createAssignmentExpr())))
+            .setIsSampleCode(true)
+            .build();
```
xiaozhenliu-gg5, line 69: Here `isSampleCode` set to true to cover the case [here](https://github.com/googleapis/gapic-generator/blob/master/src/test/java/com/google/api/codegen/gapic/testdata/java/java_library.baseline#L2668) that we only have try block without catch block, right?

miraleung: Yup.

xiaozhenliu-gg5: 👍 Thanks, LGTM!

PR #56, opened 2020-06-19
feat: add MethodDefinition

miraleung:


src/test/java/com/google/api/generator/engine/ast/MethodDefinitionTest.java
```diff
@@ -0,0 +1,178 @@
...
+
+  @Test
+  public void validMethodDefinition_basic() {
+    MethodDefinition.builder()
+        .setName("close")
+        .setScope(ScopeNode.PUBLIC)
+        .setReturnType(TypeNode.VOID)
+        .setBody(Arrays.asList(ExprStatement.withExpr(createAssignmentExpr())))
+        .build();
+    // No exception thrown, we're good.
```
summer-ji-eng, line 33: Should we assert equal to something like "public void close() { int x = y; }"

summer-ji-eng:
Thanks for the effort. Looks great.

PR #61, opened 2020-06-19
feat: add string object value

xiaozhenliu-gg5:


src/main/java/com/google/api/generator/engine/ast/StringObjectValue.java
```diff
@@ -0,0 +1,39 @@
...
+@AutoValue
+public abstract class StringObjectValue implements ObjectValue {
+
+  @Override
+  public TypeNode type() {
+    return TypeNode.STRING;
+  }
+
+  public abstract String value();
+
```
miraleung, line 28: Since value() is the only thing callers will set, let's add a `withString` or `withValue` method.

src/test/java/com/google/api/generator/engine/writer/JavaWriterVisitorTest.java
```diff
@@ -73,6 +76,12 @@ public void writePrimitiveArrayType() {
 
   /** =============================== EXPRESSIONS =============================== */
   @Test
+  public void writeStringObjectValue() {
+    StringObjectValue s = StringObjectValue.builder().setValue("test").build();
+    assertThat(s.value()).isEqualTo("test");
```
miraleung, line 81: When writing this in Java source, this should be `"\"test\""`, otherwise it will show up as `String x = test`.

src/test/java/com/google/api/generator/engine/writer/JavaWriterVisitorTest.java
```diff
@@ -73,6 +76,12 @@ public void writePrimitiveArrayType() {
...
+    assertThat(s.type()).isEqualTo(TypeNode.STRING);
+  }
```
miraleung, line 83: Add an `AssignmentExpr` test so we can see this in larger context?

src/test/java/com/google/api/generator/engine/writer/JavaWriterVisitorTest.java
xiaozhenliu-gg5: done.

src/main/java/com/google/api/generator/engine/ast/StringObjectValue.java
xiaozhenliu-gg5: sure, done.

src/test/java/com/google/api/generator/engine/writer/JavaWriterVisitorTest.java
xiaozhenliu-gg5: thx for the suggestion, working on it!

xiaozhenliu-gg5: this is very helpful! turns out the class needs a `toString()` method to return just the value(), else it will use the default toString() which will fail the assignment expression!

src/main/java/com/google/api/generator/engine/ast/StringObjectValue.java
summer-ji-eng: For learning purpose, why do we need `withValue` to set? @miraleung

xiaozhenliu-gg5:
The unit test is added in `JavaWriterVisitortest` file @summer-ji-eng

src/main/java/com/google/api/generator/engine/ast/StringObjectValue.java
summer-ji-eng: I guess I know a little bit. So it allow users has short cut set method to set string value.
Use `StringObjectValue.withValue()` instead of `StringObjectValue.builder().setValue("\"test\"").build();`

miraleung: Yup, that's it. :)

src/test/java/com/google/api/generator/engine/writer/JavaWriterVisitorTest.java
```diff
@@ -88,6 +88,29 @@ public void writeAnnotation_withDescription() {
...
   /** =============================== EXPRESSIONS =============================== */
+  @Test
+  public void writeStringObjectValue() {
+    StringObjectValue s = StringObjectValue.builder().setValue("\"test\"").build();
+    assertThat(s.value()).isEqualTo("\"test\"");
+    assertThat(s.type()).isEqualTo(TypeNode.STRING);
+  }
+
+  @Test
+  public void writeAssignmentExpr_StringObjectValue() {
```
miraleung, line 99: Nit: `writeStringObjectValue_assignmentExpr`

src/test/java/com/google/api/generator/engine/writer/JavaWriterVisitorTest.java
```diff
@@ -88,6 +88,29 @@ public void writeAnnotation_withDescription() {
...
+    IdentifierNode identifier = IdentifierNode.builder().setName("x").build();
```
miraleung, line 100: May need to merge PR #64.

xiaozhenliu-gg5: LGTMed #64

src/test/java/com/google/api/generator/engine/writer/JavaWriterVisitorTest.java
xiaozhenliu-gg5: done.thanks!

src/test/java/com/google/api/generator/engine/writer/JavaWriterVisitorTest.java
```diff
@@ -89,6 +89,29 @@ public void writeAnnotation_withDescription() {
   }
 
   /** =============================== EXPRESSIONS =============================== */
+  @Test
+  public void writeStringObjectValue() {
+    StringObjectValue s = StringObjectValue.builder().setValue("\"test\"").build();
```
miraleung, line 94: Should have picked this up in the last review, but here we want to test that the writer visitor actually writes this string. The test here would go into `ast/StringObjectValueTest.java`. In the general case, the caller wouldn't pass in `"\"test\""` unless they wanted to print `\"test\"`.

src/main/java/com/google/api/generator/engine/ast/StringObjectValue.java
```diff
@@ -0,0 +1,48 @@
...
+  public static Builder builder() {
+    return new AutoValue_StringObjectValue.Builder();
+  }
+
+  public static StringObjectValue withValue(String value) {
+    return builder().setValue(value).build();
```
miraleung, line 34: The test case reminded me that we should also handle special and escaped characters, e.g. if someone passes a quote in here, or escaped characters.

src/test/java/com/google/api/generator/engine/writer/JavaWriterVisitorTest.java
```diff
@@ -89,6 +89,29 @@ public void writeAnnotation_withDescription() {
...
+  }
+
+  @Test
+  public void writeStringObjectValue_assignmentExpr() {
+    IdentifierNode identifier = IdentifierNode.builder().setName("x").build();
+    Variable variable = Variable.builder().setIdentifier(identifier).setType(TypeNode.STRING).build();
+    VariableExpr variableExpr =
+        VariableExpr.builder().setVariable(variable).setIsDecl(true).build();
+
+    Value value = StringObjectValue.withValue("\"test\"");
```
miraleung, line 106: In the expected behavior below, the caller should be doing `withValue("test")` IMO.

miraleung:
This needs fixing, let's not merge this yet.

miraleung:
> @summer-ji-eng Since `StringObjectValue` just holds objects, we're fine without a unit test. However, we should have unit tests for classes that have additional logic (e.g. type-checking). The writer visitor test is not intended as a replacement, and checks only for the serialization logic.
> 
> LGTM assuming the nit will be addressed.

On second thought, we *do* need a unit test for handling special characters.

xiaozhenliu-gg5:
what does `corresponding change for writer visitor` mean? Thanks, I will resolve the comments in a new PR. @miraleung

PR #67, opened 2020-06-23
[ggj] fix: use string tokens in writer visitor

miraleung:


summer-ji-eng:
Oops, sorry that mess up.

PR #71, opened 2020-06-24
[ggj] fix: sort types alphabetically for implements, throws, annotations

miraleung:


xiaozhenliu-gg5:
Curious that we can also sort a list, what's tha motivation for changing a list to set? Thanks!

miraleung:
Using a set ensures we can have only one of each TypeNode.

PR #77, opened 2020-06-25
build: add unit test variable and variable expr

summer-ji-eng:


src/main/java/com/google/api/generator/engine/ast/Variable.java
```diff
@@ -47,6 +47,7 @@ public Variable build() {
 
       Variable variable = autoBuild();
       Preconditions.checkState(!variable.type().typeKind().equals(TypeNode.TypeKind.VOID));
+      Preconditions.checkState(!variable.type().equals(TypeNode.NULL));
```
miraleung, line 50: Consider adding an error message, e,g. `checkState(condition, "Variable type cannot be null")`. Not part of this PR, but the VOID one above should have one as well.

summer-ji-eng: Let me add in message in this PR.

PR #78, opened 2020-06-26
[ggj] fix: use checkNotNull instead of checkState where appropriate

miraleung:


src/main/java/com/google/api/generator/engine/ast/TryCatchStatement.java
```diff
@@ -65,8 +65,8 @@ public static Builder builder() {
     public TryCatchStatement build() {
       TryCatchStatement tryCatchStatement = autoBuild();
       if (!tryCatchStatement.isSampleCode()) {
-        Preconditions.checkState(
-            tryCatchStatement.catchVariableExpr() != null,
+        Preconditions.checkNotNull(
+            tryCatchStatement.catchVariableExpr(),
```
xiaozhenliu-gg5, line 69: It can also be used to check `setDeprecated` in comment engine, thanks!

PR #86, opened 2020-06-27
[ggj] feat: new object expression

xiaozhenliu-gg5:


src/main/java/com/google/api/generator/engine/ast/NewObjectValue.java
```diff
@@ -0,0 +1,54 @@
...
+
+    //     // there should be reference in typeNode
+    // }
+    public abstract NewObjectValue build();
+  }
+
+  @Override
+  public String value() {
+    StringBuilder value = new StringBuilder();
+    value.append("new ").append(type().reference().name());
```
summer-ji-eng, line 47: what if the reference is null.

src/test/java/com/google/api/generator/engine/writer/JavaWriterVisitorTest.java
```diff
@@ -96,6 +97,16 @@ public void writeAnnotation_withDescription() {
...
+  @Test 
+  public void writeNewObjectValue_withArgs(){
+      Reference reference = Reference.withClazz(Integer.class);
+      TypeNode type = TypeNode.withReference(reference);
+      ValueExpr valueExpr = ValueExpr.builder().setValue(PrimitiveValue.builder().setType(TypeNode.INT).setValue("123").build()).build();
+      NewObjectValue newObjectValue = NewObjectValue.builder().setType(type).setArguments(Arrays.asList(valueExpr)).build();
+      newObjectValue.accept(writerVisitor);
+      assertEquals(writerVisitor.write(), "new Integer(123)");
+  }
+
```
summer-ji-eng, line 109: can we have assignment test like `List<Book> books = new ArrayList<Book>(); ` to build confidence of this feature.

src/main/java/com/google/api/generator/engine/ast/NewObjectValue.java
xiaozhenliu-gg5: This will be checked in the build() in NewObjectValue, thanks!

src/test/java/com/google/api/generator/engine/writer/JavaWriterVisitorTest.java
xiaozhenliu-gg5: yes sure, once the design is approved, I will add more tests like you mentioned. Thank you!

src/main/java/com/google/api/generator/engine/ast/NewObjectValue.java
xiaozhenliu-gg5: done.

src/test/java/com/google/api/generator/engine/writer/JavaWriterVisitorTest.java
xiaozhenliu-gg5: This kind of assignment is included in the unit test, thanks!

src/main/java/com/google/api/generator/engine/ast/NewObjectValue.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.Collections;
+import java.util.List;
+
+@AutoValue
+public abstract class NewObjectValue implements ObjectValue {
+  public abstract TypeNode type();
+
+  public abstract List<Expr> arguments();
+
+  abstract boolean isGeneric();
```
miraleung, line 28: Nit: `// Private.` comment please.

src/main/java/com/google/api/generator/engine/ast/NewObjectValue.java
```diff
@@ -0,0 +1,73 @@
...
+    public abstract Builder setArguments(List<Expr> arguments);
+
+    public abstract Builder setIsGeneric(boolean isGeneric);
+
+    abstract NewObjectValue autoBuild();
+
+    public NewObjectValue build() {
+      NewObjectValue newObjectValue = autoBuild();
+      Preconditions.checkState(
+          TypeNode.isReferenceType(newObjectValue.type()), "New Objects must be reference types.");
```
miraleung, line 55: Nit: "New Object". In general, let's stick with usual grammar rules (e.g. capitalization and punctuation) in comments, which we've previously discussed in prior PRs.

src/main/java/com/google/api/generator/engine/ast/NewObjectValue.java
```diff
@@ -0,0 +1,73 @@
...
+      return newObjectValue;
+    }
+  }
+
+  @Override
```
miraleung, line 60: Could we please group this together with the other classes above? 
go/java-style#s3.4.2-ordering-class-contents

src/main/java/com/google/api/generator/engine/writer/ImportWriterVisitor.java
```diff
@@ -98,6 +99,11 @@ public void visit(AnnotationNode annotation) {
     annotation.type().accept(this);
   }
 
+  @Override
+  public void visit(NewObjectValue newObjectValue) {
+    // Nothing to do.
```
miraleung, line 104: Why not?

src/main/java/com/google/api/generator/engine/writer/JavaWriterVisitor.java
```diff
@@ -124,6 +125,16 @@ public void visit(AnnotationNode annotation) {
     newline();
   }
 
+  @Override
+  public void visit(NewObjectValue newObjectValue){
+    buffer.append(newObjectValue.value()).append(LEFT_PAREN);
+    List<Expr> list = newObjectValue.arguments();
+    for(Expr expression : list){
```
miraleung, line 132: Could we please adjust the spacing, as mentioned several times in previous PRs?

src/test/java/com/google/api/generator/engine/ast/NewObjectValueTest.java
```diff
@@ -0,0 +1,63 @@
...
+import static junit.framework.Assert.assertEquals;
+
+import java.awt.List;
+import java.util.Arrays;
+import java.util.Collections;
+import org.junit.Test;
+
+public class NewObjectValueTest {
+  @Test
+  public void simpleNewObjectValue() {
```
miraleung, line 25: Could we please keep the test naming convention consistent? As per the following methods, this should be `newObjectValue_simple`.

src/main/java/com/google/api/generator/engine/ast/NewObjectValue.java
```diff
@@ -0,0 +1,73 @@
...
+  public String value() {
+    StringBuilder value = new StringBuilder();
+    value.append("new ").append(type().reference().name());
+    if (isGeneric() && type().reference().generics().isEmpty()) {
+      value.append("<>");
+    }
```
miraleung, line 66: More logic is needed here, please see the comments in the test below.

src/test/java/com/google/api/generator/engine/ast/NewObjectValueTest.java
```diff
@@ -0,0 +1,63 @@
...
+  public void newObjectValue_generics() {
+    ConcreteReference reference =
+        ConcreteReference.builder()
+            .setClazz(List.class)
+            .setGenerics(Arrays.asList(ConcreteReference.withClazz(String.class)))
+            .build();
+    TypeNode type = TypeNode.withReference(reference);
+    NewObjectValue newObjectValue = NewObjectValue.genericBuilder().setType(type).build();
+    assertEquals(newObjectValue.type(), type);
+    assertEquals(newObjectValue.value(), "new List<String>");
```
miraleung, line 47: If this were an ArrayList, would `new ArrayList<String>` be valid Java? Consider whether more stuff needs to be part of the returned `value()`.

src/test/java/com/google/api/generator/engine/ast/NewObjectValueTest.java
```diff
@@ -0,0 +1,63 @@
...
+    ConcreteReference reference = ConcreteReference.withClazz(Integer.class);
+    TypeNode type = TypeNode.withReference(reference);
+    ValueExpr valueExpr =
+        ValueExpr.builder()
+            .setValue(PrimitiveValue.builder().setType(TypeNode.INT).setValue("123").build())
+            .build();
+    NewObjectValue newObjectValue =
+        NewObjectValue.builder().setType(type).setArguments(Arrays.asList(valueExpr)).build();
+    assertEquals(newObjectValue.type(), type);
+    assertEquals(newObjectValue.value(), "new Integer");
```
miraleung, line 61: Same here, what about the arguments?

src/test/java/com/google/api/generator/engine/ast/NewObjectValueTest.java
```diff
@@ -0,0 +1,63 @@
...
+  }
+}
```
miraleung, line 63: Could we please add a test with more arguments?

src/main/java/com/google/api/generator/engine/writer/JavaWriterVisitor.java
```diff
@@ -124,6 +125,16 @@ public void visit(AnnotationNode annotation) {
     newline();
   }
 
+  @Override
+  public void visit(NewObjectValue newObjectValue){
+    buffer.append(newObjectValue.value()).append(LEFT_PAREN);
```
miraleung, line 130: This should be part of NewObjectValue, since it would not be complete without the arguments.

src/test/java/com/google/api/generator/engine/ast/NewObjectValueTest.java
miraleung: Also, a test that exercises the failure case.

miraleung: Also, tests that *fully* exercise the `isGeneric() && generics().isEmpty()` logic.

xiaozhenliu-gg5:
Thanks for the comments! @miraleung sorry for the misorder of space and test method name, thought the design doc would be approved first. But I will surely adjust them since the design looks good to you 😀

src/test/java/com/google/api/generator/engine/ast/NewObjectValueTest.java
xiaozhenliu-gg5: sure, the test with args is in JavaWriterVisitorTest, and a failure test is added to the NewObjectValue test, because of the conflict setting of `isGeneric()` and `type().reference.generics()`.

src/main/java/com/google/api/generator/engine/ast/NewObjectValue.java
miraleung: Thanks for making this change. It helped me rethink the design, and it looks like `ObjectValue` may not be the best surface for `NewObject` to implement. Would you mind trying out (in the doc) what `NewObjectExpr` would look like? That way, we can do all the writing in `JavaWriterVisitor` so `value()` won't be needed.

src/test/java/com/google/api/generator/engine/ast/NewObjectValueTest.java
miraleung: Rather, I meant that we should write the `()` all together. I think `NewObjectExpr` will allow us to do that, please see the comment above.

src/test/java/com/google/api/generator/engine/ast/NewObjectValueTest.java
miraleung: Resolving in favor of the `NewObjectExpr` comments.

src/main/java/com/google/api/generator/engine/writer/JavaWriterVisitor.java
miraleung: Resolving in favor of the `NewObjectExpr` comments.

src/main/java/com/google/api/generator/engine/ast/NewObjectValue.java
```diff
@@ -0,0 +1,81 @@
...
+    // Private.
+    abstract Builder setIsGeneric(boolean isGeneric);
+
+    abstract NewObjectValue autoBuild();
+
+    public NewObjectValue build() {
+      NewObjectValue newObjectValue = autoBuild();
+      // Check the object is reference type.
+      Preconditions.checkState(
+          TypeNode.isReferenceType(newObjectValue.type()), "New Object must be reference types.");
```
miraleung, line 71: Nit: Let's stick with regular grammar rules in error messages, as mentioned in prior PR reviews.

src/test/java/com/google/api/generator/engine/ast/NewObjectValueTest.java
miraleung: On reading the usage of the API in this test, can you see anything that might look confusing to a user? How many edge cases does that method's logic have, and what are the side effects?

As a user, I would expect that this usage should not error out. Could you please write more unit tests that *fully* exercise the edge case logic? This can help us find unexpected behavior and change the implementation.

miraleung: Alternatively, you may leave a TODO here for me and I can fix it if that helps.

src/test/java/com/google/api/generator/engine/writer/ImportWriterVisitorTest.java
```diff
@@ -46,6 +48,19 @@ public void setUp() {
...
+    ConcreteReference reference =
+        ConcreteReference.builder()
+            .setClazz(List.class)
+            .setGenerics(Collections.emptyList())
+            .build();
+    TypeNode type = TypeNode.withReference(reference);
+    NewObjectValue newObjectValue = NewObjectValue.genericBuilder().setType(type).build();
+    newObjectValue.accept(writerVisitor);
+    assertEquals(writerVisitor.write(), "import java.util.List;\n\n");
+  }
```
miraleung, line 62: Could you please add more tests for a non-zero list of generics and arguments?

src/test/java/com/google/api/generator/engine/writer/JavaWriterVisitorTest.java
```diff
@@ -98,6 +100,33 @@ public void writeAnnotation_withDescription() {
...
+    ConcreteReference reference =
+        ConcreteReference.builder()
+            .setClazz(LinkedList.class)
+            .setGenerics(Arrays.asList(ConcreteReference.withClazz(Object.class)))
+            .build();
+    TypeNode type = TypeNode.withReference(reference);
+    NewObjectValue newObjectValue = NewObjectValue.genericBuilder().setType(type).build();
+    newObjectValue.accept(writerVisitor);
+    assertEquals(writerVisitor.write(), "new LinkedList<Object>()");
+  }
```
miraleung, line 128: Could we please add a test case that uses both generics (more than one generic) and arguments (more than one, and not just ValueExpr, e.g. a method invocation expr or variable)?

xiaozhenliu-gg5:
1. design for new object expression is updated, only one builder with a public `isGeneric()` flag is developed. 
2. more unit tests with generics and arguments for `ImportWriterVisitor` and `JavaWriterVisitor` are added.

src/main/java/com/google/api/generator/engine/ast/NewObjectExpr.java
```diff
@@ -38,29 +38,24 @@ public static Builder builder() {
...
 
     public NewObjectExpr build() {
       NewObjectExpr newObjectExpr = autoBuild();
-      // Check the object is reference type.
+      // 1. New object expression should be reference type.
       Preconditions.checkState(
-          TypeNode.isReferenceType(newObjectExpr.type()), "New object cannot be reference types.");
-      // Check if there is a conflict between isGeneric() setting and generics() setting.
+          TypeNode.isReferenceType(newObjectExpr.type()),
+          "New object expression should be reference types.");
```
summer-ji-eng: I think we could add one more precondition to check if the type is null type. Similar to
```
Preconditions.checkState(
            !newObjectExpr.type().equals(TypeNode.NULL), "New object expression should not be null type");
```

src/main/java/com/google/api/generator/engine/writer/ImportWriterVisitor.java
```diff
@@ -107,6 +108,12 @@ public void visit(AnnotationNode annotation) {
     annotation.type().accept(this);
   }
 
+  @Override
+  public void visit(NewObjectExpr newObjectExpr) {
+    newObjectExpr.type().accept(this);
+    expressions(newObjectExpr.arguments());
+  }
+
```
summer-ji-eng, line 116: nit: move under EXPRESSIONS down below

src/main/java/com/google/api/generator/engine/ast/AstNodeVisitor.java
```diff
@@ -24,6 +24,7 @@
 
   public void visit(AnnotationNode annotation);
 
+  public void visit(NewObjectExpr newObjectExpr);
```
summer-ji-eng, line 27: nit: move under `EXPRESSIONS` session

src/main/java/com/google/api/generator/engine/writer/JavaWriterVisitor.java
```diff
@@ -139,6 +140,27 @@ public void visit(AnnotationNode annotation) {
     newline();
   }
 
+  @Override
+  public void visit(NewObjectExpr newObjectValue) {
```
summer-ji-eng, line 144: nit: I suggest name the variable as `newObjectExpr`

src/main/java/com/google/api/generator/engine/ast/AstNodeVisitor.java
miraleung: +1

src/main/java/com/google/api/generator/engine/ast/NewObjectExpr.java
```diff
@@ -0,0 +1,69 @@
...
+
+    // Private accessor.
+    abstract TypeNode type();
+
+    abstract boolean isGeneric();
+
+    abstract NewObjectExpr autoBuild();
+
+    public NewObjectExpr build() {
+      // 1. New object expression should be reference type.
```
miraleung, line 58: The line below is self-explanatory, could we please remove this comment?

src/main/java/com/google/api/generator/engine/ast/NewObjectExpr.java
```diff
@@ -0,0 +1,69 @@
...
+      Preconditions.checkState(
+          TypeNode.isReferenceType(type()), "New object expression should be reference types.");
+      // 2. Edge case: isGeneric is false, but type().generics() is not empty
```
miraleung, line 61: Can we shorten this to something like, "type().generics() is not empty, so we'll set isGeneric to true."

src/main/java/com/google/api/generator/engine/ast/NewObjectExpr.java
```diff
@@ -0,0 +1,69 @@
...
+      // We will set isGeneric to be true for the users.
+      if (!isGeneric() && !type().reference().generics().isEmpty()) {
```
miraleung, line 63: We can simplify this to `setIsGeneric(!type().reference().generics().isEmpty())`. Could you please also add a comment on why the new, simplified logic is valid?

src/main/java/com/google/api/generator/engine/writer/ImportWriterVisitor.java
miraleung: +1

src/main/java/com/google/api/generator/engine/writer/JavaWriterVisitor.java
miraleung: +1

src/main/java/com/google/api/generator/engine/writer/JavaWriterVisitor.java
```diff
@@ -139,6 +140,27 @@ public void visit(AnnotationNode annotation) {
...
+    buffer.append(NEW);
+    space();
+    newObjectValue.type().accept(this);
+    // If isGeneric() is true, but generic list is empty, we will append `<>` to the buffer.
+    if (newObjectValue.isGeneric() && newObjectValue.type().reference().generics().isEmpty()) {
+      buffer.append(LEFT_ANGLE).append(RIGHT_ANGLE);
+    }
+    buffer.append(LEFT_PAREN);
```
miraleung, line 152: `leftParen()`, same below.

src/main/java/com/google/api/generator/engine/writer/JavaWriterVisitor.java
```diff
@@ -139,6 +140,27 @@ public void visit(AnnotationNode annotation) {
...
+    int size = newObjectValue.arguments().size();
```
miraleung, line 153: Nit: `numArguments` instead of `size` which is a rather generic name.

src/test/java/com/google/api/generator/engine/ast/NewObjectExprTest.java
```diff
@@ -0,0 +1,124 @@
...
+    NewObjectExpr newObjectExpr =
+        NewObjectExpr.builder().setType(type).setArguments(Arrays.asList(valueExpr)).build();
+    assertEquals(newObjectExpr.type(), type);
+    assertEquals(newObjectExpr.isGeneric(), false);
+  }
+
+  @Test
+  public void validNewObjectExpr_emptyGeneric() {
+    // IsGeneric() is true, but generics() is empty.
+    // [Constructing] `new LinedList<>()` no exception should be thrown.
```
miraleung, line 76: Nit: `LinkedList`

src/test/java/com/google/api/generator/engine/ast/NewObjectExprTest.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+import org.junit.Test;
+
+public class NewObjectExprTest {
+  @Test
+  public void validNewObjectValue_basic() {
+    // IsGeneric() is true and generics() is not empty.
```
miraleung, line 30: Nit: `isGeneric()` lowerCamelCase, same below.

src/test/java/com/google/api/generator/engine/writer/ImportWriterVisitorTest.java
```diff
@@ -54,6 +55,71 @@ public void setUp() {
     writerVisitor = new ImportWriterVisitor(CURRENT_PACKAGE);
   }
 
+  @Test
```
miraleung, line 58: Add a test for something like `new ArrayList<>()`?

src/test/java/com/google/api/generator/engine/writer/ImportWriterVisitorTest.java
```diff
@@ -54,6 +55,71 @@ public void setUp() {
...
+  public void writeNewObjectExprImports_genericsAndVariableArgs() {
+    // [Constructing] new HashMap<List<String>, IOException>>(int initialCapacity, float loadFactor)
```
miraleung, line 60: Consider a test where the argument types need to be imported. Feel free to either modify this test or make another one. 

Also, our sample code should strive to be valid Java if possible. For instance, passing an exception type to the constructor of this HashMap would not be valid Java, so we should avoid doing that in tests.

src/main/java/com/google/api/generator/engine/writer/ImportWriterVisitor.java
xiaozhenliu-gg5: good catch! done.

src/main/java/com/google/api/generator/engine/ast/AstNodeVisitor.java
xiaozhenliu-gg5: done.

src/main/java/com/google/api/generator/engine/ast/NewObjectExpr.java
xiaozhenliu-gg5: done.

src/main/java/com/google/api/generator/engine/ast/NewObjectExpr.java
xiaozhenliu-gg5: done.

src/main/java/com/google/api/generator/engine/ast/NewObjectExpr.java
xiaozhenliu-gg5: Thanks for the advice, but the simplified logic might not work here. For the case that `type().generics()` is empty,  we do not want to set `isGeneric()` to be false. `isGeneric` controls whether to print out `<>`. Since we have use case like `new List<>()` here the `type().generics()` is empty but `isGeneric()` is still true. Does that sound reasonable to you?

src/main/java/com/google/api/generator/engine/writer/JavaWriterVisitor.java
xiaozhenliu-gg5: got it, thanks! Done.

src/test/java/com/google/api/generator/engine/ast/NewObjectExprTest.java
xiaozhenliu-gg5: thanks!

src/main/java/com/google/api/generator/engine/ast/NewObjectExpr.java
xiaozhenliu-gg5: I think there's a small chance for users to set the new object expression as null type, but added the check for safety. And also a unit test.

src/test/java/com/google/api/generator/engine/writer/ImportWriterVisitorTest.java
xiaozhenliu-gg5: sure, added a basic unit test for it.

src/test/java/com/google/api/generator/engine/writer/ImportWriterVisitorTest.java
xiaozhenliu-gg5: sure, the test below `writeNewObjectExprImports_methodExprArg` covers the case that `new IOException(message, cause(mapArg))` where the second arg `cause(mapArg)` is a method which takes a `hashmap` and return a `throwable`, and `HashMap` is asserted to be imported for this case.

Sounds reasonable to me, changed the generics for HashMap to be `<List<String>, Integer>`

src/main/java/com/google/api/generator/engine/ast/NewObjectExpr.java
miraleung: Have you explored other alternatives such as `setIsGeneric(isGeneric() || !type().reference().generics().isEmpty())`?

src/test/java/com/google/api/generator/engine/writer/ImportWriterVisitorTest.java
```diff
@@ -54,6 +55,79 @@ public void setUp() {
...
+            .setIsGeneric(true)
+            .setType(TypeNode.withReference(ConcreteReference.withClazz(ArrayList.class)))
+            .build();
+    newObjectExpr.accept(writerVisitor);
+    assertEquals(writerVisitor.write(), "import java.util.ArrayList;\n\n");
+  }
+
+  @Test
+  public void writeNewObjectExprImports_genericsAndVariableArgs() {
+    // [Constructing] new HashMap<List<String>, Integer>>(int initialCapacity, float loadFactor)
```
miraleung, line 72: As before, could we please add a test where the argument types need to be imported as well?

src/test/java/com/google/api/generator/engine/writer/JavaWriterVisitorTest.java
```diff
@@ -106,6 +107,70 @@ public void writeAnnotation_withDescription() {
...
+        ConcreteReference.builder()
+            .setClazz(HashMap.class)
+            .setGenerics(Arrays.asList(ConcreteReference.withClazz(String.class), listRef))
+            .build();
+    TypeNode type = TypeNode.withReference(mapRef);
+    MethodInvocationExpr methodExpr =
+        MethodInvocationExpr.builder()
+            .setMethodName("foobar")
+            .setReturnType(TypeNode.INT)
+            .setStaticReferenceName("SomeClass")
```
miraleung, line 159: This will need to be merged with the recent staticRefType change in PR #122.

src/main/java/com/google/api/generator/engine/ast/NewObjectExpr.java
xiaozhenliu-gg5: Thanks for the suggestion!

src/test/java/com/google/api/generator/engine/writer/ImportWriterVisitorTest.java
xiaozhenliu-gg5: cool, added a more straightforward real use case from baseline : `new FileOutputStream(File file)` where `FileOutputStream` and `File` both need to be imported. Thanks

src/test/java/com/google/api/generator/engine/writer/JavaWriterVisitorTest.java
xiaozhenliu-gg5: sure, approved the PR and will make a change after it's merged.

xiaozhenliu-gg5: done.

src/main/java/com/google/api/generator/engine/ast/NewObjectExpr.java
```diff
@@ -0,0 +1,68 @@
...
+    abstract TypeNode type();
+    // Private accessor.
+    abstract boolean isGeneric();
+
+    abstract NewObjectExpr autoBuild();
+
+    public NewObjectExpr build() {
+      Preconditions.checkState(
+          TypeNode.isReferenceType(type()), "New object expression should be reference types.");
+      Preconditions.checkState(
```
miraleung, line 60: We can probably remove this with the null-check integrated into `isReferenceType`.

xiaozhenliu-gg5: got it, done.

PR #94, opened 2020-07-07
feat: add static import toggling

miraleung:
This will allow us to distinguish between static imports or nested references. 
- Static import case: `import java.util.Map.Entry` and references will use only `Entry`.
- Non-static: `import java.util.Map` and references will use `Map.Entry`.

### DIFF
--- a/src/test/java/com/google/api/generator/engine/ast/ConcreteReferenceTest.java
+++ b/src/test/java/com/google/api/generator/engine/ast/ConcreteReferenceTest.java
@@ -30,6 +30,30 @@ public class ConcreteReferenceTest {
   public void basicConcreteReference() {
     Reference reference = ConcreteReference.builder().setClazz(Integer.class).build();
     assertEquals(reference.name(), Integer.class.getSimpleName());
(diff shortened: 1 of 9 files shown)


src/test/java/com/google/api/generator/engine/ast/ConcreteReferenceTest.java
```diff
@@ -30,6 +30,30 @@
...
     assertEquals(reference.name(), Integer.class.getSimpleName());
+    assertFalse(reference.isStaticImport());
+  }
+
+  @Test
+  public void basicConcreteReference_setIsStaticImport() {
+    Reference reference =
+        ConcreteReference.builder().setClazz(Integer.class).setIsStaticImport(true).build();
+    assertEquals(reference.name(), Integer.class.getSimpleName());
+    assertFalse(reference.isStaticImport());
```
summer-ji-eng, line 41: Sorry, I confused the value of `isStaticImport`, why should be `false` since `setIsStaticImport(true)` at line 39

PR #95, opened 2020-07-07
build: integrate null value in method

summer-ji-eng:
Add check null test in MethodDefinition
Add unit test in MethodInvocationExpr

### DIFF
--- a/src/main/java/com/google/api/generator/engine/ast/MethodDefinition.java
+++ b/src/main/java/com/google/api/generator/engine/ast/MethodDefinition.java
@@ -129,7 +129,7 @@ public abstract class MethodDefinition implements AstNode {
       if (isAbstract()) {
         Preconditions.checkState(
             !isFinal() && !isStatic() && !scope().equals(ScopeNode.PRIVATE),
-            "Abstract mehtods cannot be static, final, or private");
+            "Abstract methods cannot be static, final, or private");
       }
 
       // If this method overrides another, ensure that the Override annotaiton is the last one.
@@ -138,6 +138,9 @@ public abstract class MethodDefinition implements AstNode {
       }
 
       MethodDefinition method = autoBuild();
+
+      Preconditions.checkState(
+          !method.returnType().equals(TypeNode.NULL), "Method returnType can not be null");
 
       if (!method.returnType().equals(TypeNode.VOID)) {
         Preconditions.checkNotNull(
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/api/generator/engine/ast/MethodDefinition.java
```diff
@@ -129,7 +129,7 @@ public MethodDefinition build() {
       if (isAbstract()) {
         Preconditions.checkState(
             !isFinal() && !isStatic() && !scope().equals(ScopeNode.PRIVATE),
-            "Abstract mehtods cannot be static, final, or private");
```
miraleung, line 132: Thank you for catching things like these along the way.

src/main/java/com/google/api/generator/engine/ast/MethodDefinition.java
```diff
@@ -139,6 +139,9 @@ public MethodDefinition build() {
 
       MethodDefinition method = autoBuild();
 
+      Preconditions.checkState(
+          !method.returnType().equals(TypeNode.NULL), "Method returnType can not be null");
```
miraleung, line 143: Nit: "return type"

src/main/java/com/google/api/generator/engine/ast/MethodInvocationExpr.java
```diff
@@ -110,6 +110,11 @@ public MethodInvocationExpr build() {
       }
 
       MethodInvocationExpr methodInvocationExpr = autoBuild();
+
+      Preconditions.checkState(
+          !methodInvocationExpr.type().equals(TypeNode.NULL),
+          "MethodInvocationExpr type can not be null");
```
miraleung, line 116: Nit: Consider rephrasing this such that someone who doesn't know this API understands what they did wrong. E.g. "Methods cannot be invoked on null types."

src/main/java/com/google/api/generator/engine/ast/MethodDefinition.java
miraleung: Also, consider rephrasing this as per the comment below. For example, readers might wonder why they cannot return null, because `return null` is valid Java. However, what we mean to say is something like, "Null is not a valid method return type."

summer-ji-eng: agree. Thanks.

src/main/java/com/google/api/generator/engine/ast/MethodInvocationExpr.java
summer-ji-eng: sg

PR #96, opened 2020-07-08
build: integrate null type in Class Definition

summer-ji-eng:
A class cannot extends NullType because it is an interface.

But a class can implements an NullType, so I put two cases in `classdefinitiontest`
Here is an example to implement NullType, where it is imported from `javax.lang.model.type.NullType`
```
class Summer implements NullType {

  @Override
  public TypeKind getKind() {
    return null;
  }

  @Override
  public <R, P> R accept(TypeVisitor<R, P> v, P p) {
    return null;
  }

  @Override
  public List<? extends AnnotationMirror> getAnnotationMirrors() {
    return null;
  }

  @Override
  public <A extends Annotation> A getAnnotation(Class<A> annotationType) {
    return null;
  }

  @Override
  public <A extends Annotation> A[] getAnnotationsByType(Class<A> annotationType) {
    return null;
  }
}
```

### DIFF
--- a/src/test/java/com/google/api/generator/engine/ast/ClassDefinitionTest.java
+++ b/src/test/java/com/google/api/generator/engine/ast/ClassDefinitionTest.java
@@ -18,7 +18,6 @@ import static org.junit.Assert.assertThrows;
 
 import java.util.Arrays;
 import java.util.List;
-import javax.lang.model.type.NullType;
 import org.junit.Test;
 
 public class ClassDefinitionTest {
@@ -110,18 +109,6 @@ public class ClassDefinitionTest {
 
   @Test
   public void invalidClassDefinition_implementsNullType() {
-    assertThrows(
-        IllegalStateException.class,
-        () -> {
-          ClassDefinition.builder()
-              .setPackageString("com.google.example.library.v1.stub")
-              .setName("LibraryServiceStub")
-              .setScope(ScopeNode.PUBLIC)
-              .setImplementsTypes(
-                  Arrays.asList(
-                      TypeNode.withReference(ConcreteReference.withClazz(NullType.class))))
-              .build();
-        });
     assertThrows(
         IllegalStateException.class,
         () -> {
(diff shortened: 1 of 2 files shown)


miraleung:
Not related to this PR, but this reminded me that we will need to ensure that `NullType` does not get imported when ImportWriterVisitor visits NullObjectValue. Please feel free to add this work to our tracker as well.

Also, if we end up having to generate code that extends or implements the actual `NullType` class, we can refactor this to use our own `ast.NullType` class as the underlying representative class.

summer-ji-eng:
> Since we likely won't be generating code that `implements NullType`, I think it's OK to prevent that case as well, so we can ensure no one does `implements null`.

Cool. Thanks comment out.

summer-ji-eng:
> Not related to this PR, but this reminded me that we will need to ensure that `NullType` does not get imported when ImportWriterVisitor visits NullObjectValue. Please feel free to add this work to our tracker as well.
> 
> Also, if we end up having to generate code that extends or implements the actual `NullType` class, we can refactor this to use our own `ast.NullType` class as the underlying representative class.

summer-ji-eng:
> Not related to this PR, but this reminded me that we will need to ensure that `NullType` does not get imported when ImportWriterVisitor visits NullObjectValue. Please feel free to add this work to our tracker as well.

Cool. Added in our tracksheet.

src/test/java/com/google/api/generator/engine/ast/ClassDefinitionTest.java
```diff
@@ -107,6 +108,32 @@ public void validClassDefinition_statementsAndMethods() {
...
+    assertThrows(
+        IllegalStateException.class,
+        () -> {
+          ClassDefinition.builder()
+              .setPackageString("com.google.example.library.v1.stub")
+              .setName("LibraryServiceStub")
+              .setScope(ScopeNode.PUBLIC)
+              .setImplementsTypes(
+                  Arrays.asList(
+                      TypeNode.withReference(ConcreteReference.withClazz(NullType.class))))
```
miraleung, line 122: Did you mean `TypeNode.NULL`?

summer-ji-eng: I try to cover two cases here. But I guess we do not need cover this one right now. I will remove it.

PR #113, opened 2020-07-14
[ggj] feat: add CastExpr

miraleung:


src/main/java/com/google/api/generator/engine/ast/CastExpr.java
```diff
@@ -0,0 +1,54 @@
...
+    abstract CastExpr autoBuild();
+
+    public CastExpr build() {
+      CastExpr castExpr = autoBuild();
+      Preconditions.checkState(
+          TypeNode.isReferenceType(castExpr.type()), "Casting can only be done to reference types");
+      Preconditions.checkState(
+          TypeNode.isReferenceType(castExpr.expr().type()),
+          "Casting can only be done on reference types");
+      return castExpr;
```
xiaozhenliu-gg5, line 51: Question on the two precondition check: how about the primitive type casting? like `int result = (int)(x / y);` Thanks!

miraleung: Good catch, fixed.

PR #119, opened 2020-07-20
[ggj][comment-engine] feat: add comment escaper for comment classes

xiaozhenliu-gg5:


xiaozhenliu-gg5:
1. CommentEscaper is similar as Formatter to StringObjectValue, but `"` is not escaped any more, because the comment won't be surrounded by `"`. But for BlockComment/JavaDocComment `*/` is special and should be escaped. Currently it's replaced by `*[space]/`, please LMK if there's a better way to handle that. @miraleung Thanks!
2. `escaper` package is added since it will be shared by comment classes.
3. unit tests for Block/LineComment are not separately added since they are comparatively simple and they are tested in `JavaWriterVisitorTest`.

src/main/java/com/google/api/generator/engine/ast/JavaDocComment.java
```diff
@@ -115,6 +115,11 @@ public JavaDocComment build() {
       if (!Strings.isNullOrEmpty(deprecated)) {
         componentsList.add(String.format("@deprecated %s", deprecated));
       }
+      // Escape component in list one by one, because we will join the components by `\n`
+      // `\n` will be taken as escape character by the comment escaper.
+      for (int i = 0; i < componentsList.size(); i++) {
+        componentsList.set(i, CommentEscaper.escape(componentsList.get(i)));
```
miraleung, line 121: Inline with `stream()`?

src/main/java/com/google/api/generator/engine/ast/LineComment.java
```diff
@@ -32,8 +33,15 @@ public static LineComment withComment(String comment) {
...
   public abstract static class Builder {
     public abstract Builder setComment(String comment);
+    // Private accessor.
+    abstract String comment();
 
-    public abstract LineComment build();
+    public abstract LineComment autoBuild();
+
+    public LineComment build() {
+      setComment(CommentEscaper.escape(comment()));
```
miraleung, line 42: Does this need escaping?

src/test/java/com/google/api/generator/engine/ast/JavaDocCommentTest.java
```diff
@@ -29,6 +29,21 @@ public void createJavaDocComment_basic() {
...
+            .setThrows("RunTimeException", "This is an unexpected block end */")
+            .build();
+    String expected =
+        "Service comment may include special characters: <>&\"`'@\n"
+            + "<p> title: GetBigBook: 'War and Peace'\n"
+            + "@throws RunTimeException This is an unexpected block end * /";
+    assertEquals(javaDocComment.comment(), expected);
+  }
+
   @Test
```
miraleung, line 47: What about some test cases for comments that would have failed before?

miraleung: Or comments with `/**` or `*/` in the middle?

src/main/java/com/google/api/generator/engine/escaper/CommentEscaper.java
```diff
@@ -0,0 +1,44 @@
...
+            .addEscape('\n', "\\n")
+            .addEscape('\\', "\\\\")
+            .build();
+
+    private SpecialEscaper() {}
+
+    @Override
+    public String escape(String sourceString) {
+      // Replace unexpected block end `*/` with `* /`.
+      return escaper.escape(sourceString.replace("*/", "* /"));
```
miraleung, line 37: Could you please provide more context for this logic? I'm not sure that replacing `*/` with `* /` is the appropriate approach because this edits the comment.

src/main/java/com/google/api/generator/engine/escaper/CommentEscaper.java
```diff
@@ -0,0 +1,44 @@
...
+package com.google.api.generator.engine.escaper;
+
+import com.google.common.escape.Escaper;
+import com.google.common.escape.Escapers;
+
+public class CommentEscaper {
+  private static class SpecialEscaper extends Escaper {
+    private static final Escaper escaper =
+        Escapers.builder()
+            .addEscape('\t', "\\t")
```
miraleung, line 24: I think I see what you're doing, could you please add a comment (on what is happening here) for future readers?

src/main/java/com/google/api/generator/engine/escaper/CommentEscaper.java
```diff
@@ -0,0 +1,44 @@
...
+
+package com.google.api.generator.engine.escaper;
+
+import com.google.common.escape.Escaper;
+import com.google.common.escape.Escapers;
+
+public class CommentEscaper {
+  private static class SpecialEscaper extends Escaper {
+    private static final Escaper escaper =
+        Escapers.builder()
```
miraleung, line 23: What is the reference list for these escape characters, can we add a link?

Any reason for handling only a subset of Java's escape characters?

src/main/java/com/google/api/generator/engine/ast/LineComment.java
xiaozhenliu-gg5: yes, there is also case that LineComment contains special characters \t\b\r etc, if they are not escaped, when we call java formatter, it will throw error of FormatException.

src/main/java/com/google/api/generator/engine/escaper/CommentEscaper.java
xiaozhenliu-gg5: The escape [characters](https://stackoverflow.com/questions/1367322/what-are-all-the-escape-characters) in comment would mess the JavaFormmater, so we handle them here. `"`  is overlooked because unlike the string object value, the comment won't be surrounded by `"`, so `\"` in input string is enough.

src/main/java/com/google/api/generator/engine/escaper/CommentEscaper.java
xiaozhenliu-gg5: sure! done.

src/main/java/com/google/api/generator/engine/escaper/CommentEscaper.java
xiaozhenliu-gg5: sure, this covers the case that we have `*/` in the middle of comment string, that will break Block/JavaDoc comment since it's an unexpected end.
For example (unit test also shows this case):
```
/** this is a block comment with unexpected end */ we should replace it. */
```

src/test/java/com/google/api/generator/engine/ast/JavaDocCommentTest.java
xiaozhenliu-gg5: added more special characteres `\t\b\r\n` which would have failed without the escaper.

For `/**` and `*/`, `/**` is not special because it will not break us, `/** this is a test /** comment` is valid. `*/` is tested in throw description, the whole java doc comment will be like:
```
/**
* Service comment may include special characters: <>\t\b\r&\"\n`'@
* <p> title: GetBigBook: 'War and Peace'
* @throws RunTimeException This is an unexpected block end * /
*/
```

src/main/java/com/google/api/generator/engine/ast/JavaDocComment.java
xiaozhenliu-gg5: sure, TWIL `list.stream().forEach()` only iterates the values but cannot change them, while `list.stream().map` can modify the value but the return value of the method must be used (reassigning to `componentList`).

src/main/java/com/google/api/generator/engine/escaper/CommentEscaper.java
miraleung: As above, could we please add a link (to Oracle's docs) in the comments of the source code? Also, let's add a comment on why we won't handle certain characters like quotes.

src/main/java/com/google/api/generator/engine/escaper/CommentEscaper.java
miraleung: A cursory search shows that [HTML escaping](https://stackoverflow.com/questions/631817/how-to-quote-in-javadocs) is one possible approach, which I think we should take. Alternatively, would it be valid if we escaped the `*` with a backslash? Adding a space essentially changes the comment's contents, which is undesirable.

If we pick the HTML escaper route (and I think we should), feel free to add a TODO and leave it for the next PR (below).

As a side note (not related to this PR), I've since noticed that the baseline file intentionally performs HTML escaping on special characters such as `<` and `>`. (We briefly talked about these chars in the original comment PR and thought that rendering them as-is was the intended behavior.) As a separate task (which I've added to the work tracker sheet), could you please add an HTML escaper for JavaDocComment in a separate PR?

src/test/java/com/google/api/generator/engine/ast/JavaDocCommentTest.java
miraleung: Note that we _may_ need to replace those with different characters if we perform HTML escaping (see next comment). PTAL at the baseline file to investigate how we handled these characters, if at all.

src/main/java/com/google/api/generator/engine/escaper/CommentEscaper.java
xiaozhenliu-gg5: of course, added the [link](https://docs.oracle.com/javase/tutorial/java/data/characters.html) for reference, and explained why we do not escape `"` or `'` here. Thanks

src/main/java/com/google/api/generator/engine/escaper/CommentEscaper.java
xiaozhenliu-gg5: Sounds cool, that explains why we have `&lt;&gt;&amp;` etc in the comment, which is for rendering `<>` correctly in browser. Thanks for the advice! Removed the html part in tests and added a TODO for the HTML escaper, and `*/` test case will be added by that implementation.

src/test/java/com/google/api/generator/engine/ast/JavaDocCommentTest.java
xiaozhenliu-gg5: Thanks, I will take investigation of the HTML characters escaping and make a separate PR after this.

PR #120, opened 2020-07-21
[ggj] feat: add comment statement class

xiaozhenliu-gg5:


src/main/java/com/google/api/generator/engine/ast/CommentStatement.java
```diff
@@ -0,0 +1,60 @@
...
+package com.google.api.generator.engine.ast;
+
+import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableList;
+import javax.annotation.Nullable;
+
+@AutoValue
+public abstract class CommentStatement implements Statement {
+
+  public abstract ImmutableList<LineComment> lineComments();
```
miraleung, line 24: As discussed today and previous PRs, `CommentStatement` is a thin wrapper for `Comment`. See my comment in PR #30: "In the future, there should be a CommentStatement that lightly wraps Comment."

xiaozhenliu-gg5: I got the point and updated the API/unit tests thank you!

src/main/java/com/google/api/generator/engine/ast/CommentStatement.java
```diff
@@ -0,0 +1,42 @@
...
+  public static CommentStatement withComment(Comment comment) {
+    return builder().setComment(comment).build();
+  }
+
+  @Override
+  public void accept(AstNodeVisitor visitor) {
+    visitor.visit(this);
+  }
+
+  public static Builder builder() {
```
miraleung, line 32: We can probably make this method and the Builder class "private".

src/test/java/com/google/api/generator/engine/ast/CommentStatementTest.java
```diff
@@ -0,0 +1,53 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.api.generator.engine.ast;
+
+import static junit.framework.Assert.assertEquals;
+
+import org.junit.Test;
+
+public class CommentStatementTest {
```
miraleung, line 21: Since the CommentStatement class has no extra logic beyond setting and getting values, we don't need this test.

src/test/java/com/google/api/generator/engine/writer/JavaWriterVisitorTest.java
```diff
@@ -381,6 +382,46 @@ public void writeJavaDocComment_specialChar() {
     assertEquals(writerVisitor.write(), expected);
   }
 
+  @Test
```
miraleung, line 385: To reduce redundancy (and since CommentStatement has no special logic), can we combine these tests with the existing BlockComment, LineComment, or JavaDocComment ones? Up to you which way to merge these tests.

src/test/java/com/google/api/generator/engine/ast/CommentStatementTest.java
xiaozhenliu-gg5: got it, removed this unit test file.

src/main/java/com/google/api/generator/engine/ast/CommentStatement.java
xiaozhenliu-gg5: For consistency, should we keep this builder public since there might be chance to call `CommentStatement.builder()` as other classes. But also open to change this, thanks.

src/test/java/com/google/api/generator/engine/writer/JavaWriterVisitorTest.java
xiaozhenliu-gg5: Sure, I merged basic line/block/javadoc comment test with the commentStatement class, and keep the special character tests isolated to the line/block/javadoc comment class.

src/main/java/com/google/api/generator/engine/ast/CommentStatement.java
miraleung: The `builder()` method and `Builder` class are redundant because we only need `withComment`, could we please make them "private"?

xiaozhenliu-gg5: Done, thanks! `builder` method is made `private`, while `Builder` class cannot be private because
```
 error: Builder has private access in CommentStatement
  static final class Builder extends CommentStatement.Builder {
```
so it's scoped in package level.

PR #125, opened 2020-07-23
[ggj]feat: add html escaper for comments

xiaozhenliu-gg5:


xiaozhenliu-gg5:
Based on investigation, only `JavaDocComment` needs a `htmlEscaper`, which takes care of escaping `<>&*@` (same as [monolith](https://github.com/googleapis/gapic-generator/blob/master/src/main/java/com/google/api/codegen/util/java/JavaCommentReformatter.java#L23)). And among javaDocComment components, only comment and sample code can contain html escape characters. `<p> <ol> <ul>` are [hard-coded](https://github.com/googleapis/gapic-generator/blob/master/src/main/resources/com/google/api/codegen/java/settings.snip#L30) in monolith generator templates.

src/main/java/com/google/api/generator/engine/escaper/CommentEscaper.java
```diff
@@ -18,7 +18,7 @@
 import com.google.common.escape.Escapers;
 
 public class CommentEscaper {
-  private static class SpecialEscaper extends Escaper {
+  private static class SpecialCharEscaper extends Escaper {
```
miraleung, line 21: I would recommend splitting SpecialCharEscaper and HtmlEscaper into separate public classes, since there isn't anything inherently "comment"-y about them.

src/main/java/com/google/api/generator/engine/escaper/CommentEscaper.java
```diff
@@ -32,16 +32,41 @@
...
     }
   }
+  // HtmlEscaper is separately defined from SpecialCharEscaper even if they extend the same Escaper,
+  // because HtmlEscaper would be only called by JavaDocComment.
+  private static class HtmlEscaper extends Escaper {
+    // Based on the observation of the generated java files, we escape the following
+    // five characters by html escaper. We do not directly use guava HtmlEscapers here because
+    // it only escapes`<>&\"'` as specified by HTML 4.01.
+    private static final Escaper escaper =
+        Escapers.builder()
```
miraleung, line 49: Curious, did you find any patterns for `<p>` and list (e.g. `<ol>`) escaping?

src/main/java/com/google/api/generator/engine/escaper/CommentEscaper.java
miraleung: Also, let's rename this particular escaper to something like "MetacharEscaper" or similar.

xiaozhenliu-gg5: Thanks for the advice, done.

src/main/java/com/google/api/generator/engine/escaper/CommentEscaper.java
xiaozhenliu-gg5: At least in baseline file, all `<p> <ol>` components are hard-coded in monolith (for example: [settings.snip](https://github.com/googleapis/gapic-generator/blob/master/src/main/resources/com/google/api/codegen/java/settings.snip#L30) I iterated all of them, and can find corresponding template in monolith generator) so I think we do not have escape them here. But I do find in `java-pubsub` that some `<p>` which are neither in the monolith template or proto files. I suspect it comes from samplegen, but double check that.

src/main/java/com/google/api/generator/engine/escaper/HtmlEscaper.java
```diff
@@ -0,0 +1,43 @@
...
+// limitations under the License.
+
+package com.google.api.generator.engine.escaper;
+
+import com.google.common.escape.Escaper;
+import com.google.common.escape.Escapers;
+
+public class HtmlEscaper extends Escaper {
+  // Based on the observation of the generated java files, we escape the following
+  // five characters by html escaper. We do not directly use guava HtmlEscapers here because
```
miraleung, line 22: Good note.

xiaozhenliu-gg5: Thanks!

PR #128, opened 2020-07-23
[ggj][wip]feat: support multi-line comment (file header)

xiaozhenliu-gg5:
TODO: integrate html escaper when it's in place.

### DIFF
--- a/src/main/java/com/google/api/generator/engine/ast/AstNodeVisitor.java
+++ b/src/main/java/com/google/api/generator/engine/ast/AstNodeVisitor.java
@@ -50,6 +50,8 @@ public interface AstNodeVisitor {
 
   public void visit(JavaDocComment javaDocComment);
 
+  public void visit(MultiLineComment multiLineComment);
+
   /** =============================== STATEMENTS =============================== */
   public void visit(ExprStatement exprStatement);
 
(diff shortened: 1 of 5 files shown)


xiaozhenliu-gg5:
Request for a high-level review to get the idea how we wanna support the file header comment. @miraleung  Thanks!

src/main/java/com/google/api/generator/engine/ast/AstNodeVisitor.java
```diff
@@ -50,6 +50,8 @@
 
   public void visit(JavaDocComment javaDocComment);
 
+  public void visit(MultiLineComment multiLineComment);
```
miraleung, line 53: Why do we need a new MultiLineComment? As discussed earlier today, we should reuse CommentStatement and add this to ClassDefinition.

src/test/java/com/google/api/generator/engine/writer/JavaWriterVisitorTest.java
```diff
@@ -380,6 +381,72 @@ public void writeJavaDocComment_specialChar() {
...
+    assertEquals(writerVisitor.write(), expected);
+  }
+
+  @Test
+  public void writeMultiLineComment_specialChar() {
+    // TODO(xiaozhenliu): enable the check after html escaper is in place.
+    MultiLineComment multiLineComment =
+        MultiLineComment.builder()
+            .addComment("This may contain special characters like \t\b\r\n\f\"")
+            // .addComment("Also */ or <> & which should be handled by htmlEscaper.")
```
miraleung, line 437: I see this is a WIP PR, let's make a mental note of this to ensure we don't check in commented-out code.

PR #130, opened 2020-07-24
[ggj] feat: boxed and primitive value equality in cast expression

xiaozhenliu-gg5:
[1/6] support boxed/primitive type equality in cast expression.


src/main/java/com/google/api/generator/engine/ast/TypeNode.java
```diff
@@ -37,13 +38,34 @@
...
-  public static final TypeNode FLOAT = builder().setTypeKind(TypeKind.FLOAT).build();
+  public static final TypeNode BYTE = builder().setTypeKind(TypeKind.BYTE).build();
+  public static final TypeNode CHAR = builder().setTypeKind(TypeKind.CHAR).build();
   public static final TypeNode DOUBLE = builder().setTypeKind(TypeKind.DOUBLE).build();
+  public static final TypeNode FLOAT = builder().setTypeKind(TypeKind.FLOAT).build();
+  public static final TypeNode INT = builder().setTypeKind(TypeKind.INT).build();
   public static final TypeNode LONG = builder().setTypeKind(TypeKind.LONG).build();
+  public static final TypeNode SHORT = builder().setTypeKind(TypeKind.SHORT).build();
+
+  public static final TypeNode BOOLEAN_WRAPPER =
```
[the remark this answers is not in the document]
xiaozhenliu-gg5: Sounds great, done.

src/main/java/com/google/api/generator/engine/ast/CastExpr.java
```diff
@@ -52,16 +52,15 @@ public CastExpr build() {
...
             "Numeric and boolean types are not inter-castable");
-      } else {
+      } else if (castType.isPrimitiveType() || exprType.isPrimitiveType()) {
+        // Check that if one type is primitive, and the other is its wrapper class type.
+        // For example: `Integer` and `int` they can be castable.
         Preconditions.checkState(
-            TypeNode.isReferenceType(castExpr.type())
-                && TypeNode.isReferenceType(castExpr.expr().type()),
-            "Reference types can only be casted to reference types");
+            TypeNode.boxedPrimitiveEquality(castType, exprType),
```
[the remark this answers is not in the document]
xiaozhenliu-gg5: Cool, thanks! Done.

src/main/java/com/google/api/generator/engine/ast/TypeNode.java
```diff
@@ -97,6 +118,15 @@ public static boolean isReferenceType(TypeNode type) {
         && type.reference() != null;
   }
 
+  public static boolean isBoxedTypeEquals(TypeNode type1, TypeNode type2) {
+    if (type2.isPrimitiveType()) {
+      TypeNode type = type1;
+      type1 = type2;
+      type2 = type;
```
summer-ji-eng, line 125: Sorry, I am little confused here. Why do we need to do a swap here?

xiaozhenliu-gg5: the swap ensures that type1 is primitive and type2 is boxed type, then we do not have to worry about the order when passing in the arguments. For example, this method can be used to check `isBoxedTypeEquals(int, Integer)` or `isBoxedTypeEquals(Integer, int)`. Thanks

src/main/java/com/google/api/generator/engine/ast/CastExpr.java
```diff
@@ -52,16 +52,15 @@ public CastExpr build() {
...
-      } else {
+      } else if (castType.isPrimitiveType() || exprType.isPrimitiveType()) {
+        // Check that if one type is primitive, and the other is its wrapper class type.
+        // For example: `Integer` and `int` they can be castable.
         Preconditions.checkState(
-            TypeNode.isReferenceType(castExpr.type())
-                && TypeNode.isReferenceType(castExpr.expr().type()),
-            "Reference types can only be casted to reference types");
+            TypeNode.isBoxedTypeEquals(castType, exprType),
+            "Primitive types can only be cast to their corresponding boxed types.");
```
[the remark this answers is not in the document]
xiaozhenliu-gg5: this expression is valid, updated the message to `Primitive types can only be cast to their corresponding boxed types and vice versa.`

src/main/java/com/google/api/generator/engine/ast/TypeNode.java
xiaozhenliu-gg5: done.

src/test/java/com/google/api/generator/engine/ast/TypeNodeTest.java
```diff
@@ -0,0 +1,31 @@
...
+package com.google.api.generator.engine.ast;
+
+import static junit.framework.Assert.assertEquals;
+
+import org.junit.Test;
+
+public class TypeNodeTest {
+  @Test
+  public void isBoxedTypeEquals() {
+    assertEquals(TypeNode.isBoxedTypeEquals(TypeNode.INT, TypeNode.INT_OBJECT), true);
```
miraleung, line 24: Why not `assertTrue`?

src/main/java/com/google/api/generator/engine/ast/TypeNode.java
```diff
@@ -155,4 +187,17 @@ private static TypeNode createPrimitiveArrayType(TypeKind typeKind) {
   private static boolean isPrimitiveType(TypeKind typeKind) {
     return !typeKind.equals(TypeKind.OBJECT);
   }
+
+  private static HashMap<TypeNode, TypeNode> createBoxedTypeMap() {
```
miraleung, line 191: We generally return the superinterface when it comes to collections, can we return a Map here instead?

src/test/java/com/google/api/generator/engine/ast/CastExprTest.java
```diff
@@ -73,6 +73,49 @@ public void validCastExpr_basicPrimitiveWidening() {
 
   }
 
+  @Test
+  public void validCastExpr_castPrimitiveToBoxedType() {
+    PrimitiveValue intValue = PrimitiveValue.builder().setValue("3").setType(TypeNode.INT).build();
+    ValueExpr valueExpr = ValueExpr.withValue(intValue);
+    CastExpr expr = CastExpr.builder().setType(TypeNode.INT_OBJECT).setExpr(valueExpr).build();
```
miraleung, line 80: Could we remove `CastExpr expr` since it's unused? Same below.

src/test/java/com/google/api/generator/engine/ast/CastExprTest.java
```diff
@@ -73,6 +73,49 @@ public void validCastExpr_basicPrimitiveWidening() {
...
+  }
+
+  @Test
+  public void validCastExpr_castBoxedTypeToPrimitive() {
+    Variable variable = Variable.builder().setName("x").setType(TypeNode.DOUBLE_OBJECT).build();
+    VariableExpr variableExpr = VariableExpr.builder().setVariable(variable).build();
+    CastExpr expr = CastExpr.builder().setType(TypeNode.DOUBLE).setExpr(variableExpr).build();
+  }
+
+  @Test
```
miraleung, line 90: Great to see that these edge cases were considered. In this case, both `nestedPrimitiveBoxedType` tests are already covered by the tests for casting primitives to objects and vice versa, so we can remove them.

src/test/java/com/google/api/generator/engine/writer/JavaWriterVisitorTest.java
```diff
@@ -727,6 +727,20 @@ public void writeCastExpr_nested() {
     assertEquals(writerVisitor.write(), "((String) ((Object) str))");
   }
 
+  @Test
```
miraleung, line 730: We won't need this test either, since we can assume that CastExpr will behave correctly and there hasn't been a change to `JavaWriterVisitor`.

src/test/java/com/google/api/generator/engine/ast/TypeNodeTest.java
xiaozhenliu-gg5: done.

src/test/java/com/google/api/generator/engine/ast/CastExprTest.java
xiaozhenliu-gg5: done.

src/test/java/com/google/api/generator/engine/ast/CastExprTest.java
xiaozhenliu-gg5: done. Thanks!

src/main/java/com/google/api/generator/engine/ast/TypeNode.java
xiaozhenliu-gg5: noted, done.

src/test/java/com/google/api/generator/engine/writer/JavaWriterVisitorTest.java
xiaozhenliu-gg5: got it, done.

src/main/java/com/google/api/generator/engine/ast/CastExpr.java
```diff
@@ -52,11 +52,15 @@ public CastExpr build() {
       // Only numeric types can be casted around.
       if (castType.isPrimitiveType() && exprType.isPrimitiveType()) {
         // Check that the types are either both booleans or both not booleans.
-        // TODO(xiaozhenliu): This is one of the places where we need to check the equality of boxed
-        // and primitive types.
         Preconditions.checkState(
             !(castType.equals(TypeNode.BOOLEAN) ^ exprType.equals(TypeNode.BOOLEAN)),
```
miraleung, line 56: Could you also please check if `boolean x = (Boolean) true` and the reverse are valid?

xiaozhenliu-gg5: Yes, verified in local files and unit tests, cast expression below are valid.
```
        boolean flag = (Boolean) true;
        Boolean flag1 = (boolean) true;
```
but since similar test case is covered by `validCastExpr_castBoxedTypeToPrimitive` and `validCastExpr_castPrimitiveToBoxedType`, we won't add extra test for this. Thanks

miraleung: The test case won't catch this, because the original check looks specifically at the primitive type. Could we please add a test case? 

With the `equals()` change, the logic above won't need to be changed either.

src/main/java/com/google/api/generator/engine/ast/CastExpr.java
```diff
@@ -52,11 +52,15 @@ public CastExpr build() {
...
             "Numeric and boolean types are not inter-castable");
+      } else if (castType.isPrimitiveType() || exprType.isPrimitiveType()) {
+        // Check that if one type is primitive, and the other is its wrapper class type.
+        // For example: `Integer` and `int` they can be castable.
+        Preconditions.checkState(
+            TypeNode.isBoxedTypeEquals(castType, exprType),
```
miraleung, line 62: As per my comments in the doc, it may be best to alter `TypeNode.equals()` and use that instead of having an explicit `isBoxedTypeEquals` check here.

xiaozhenliu-gg5: sure, updated the `equals()` method and the condition check. Open to the error message throwing for the second check. Thanks!

src/main/java/com/google/api/generator/engine/ast/CastExpr.java
xiaozhenliu-gg5: I'm a little bit confused here, the test cases would be like below if I'm understanding correctly which are already covered:
```
  @Test
  public void validCastExpr_castPrimitiveToBoxedType() {
    // [Constructing] `(Boolean) true` where `true` is a primitive value that is `boolean` type.
    PrimitiveValue booleanValue = PrimitiveValue.builder().setValue("true").setType(TypeNode.BOOLEAN).build();
    ValueExpr valueExpr = ValueExpr.withValue(booleanValue);
    CastExpr.builder().setType(TypeNode.BOOLEAN_OBJECT).setExpr(valueExpr).build();
  }

  @Test
  public void validCastExpr_castBoxedTypeToPrimitive() {
     // [Constructing] `(boolean) flag` where `flag` is a object value that is `Boolean` type.
    Variable variable = Variable.builder().setName("flag").setType(TypeNode.BOOLEAN_OBJECT).build();
    VariableExpr variableExpr = VariableExpr.builder().setVariable(variable).build();
    CastExpr.builder().setType(TypeNode.BOOLEAN).setExpr(variableExpr).build();
  }
```
are we talking about the whole assignment expression? Thanks

xiaozhenliu-gg5: a draft PR for reference: https://github.com/googleapis/gapic-generator-java/pull/140/files#diff-59533acd4d8a07afff86d7de2cd8606dR106

src/main/java/com/google/api/generator/engine/ast/TypeNode.java
```diff
@@ -124,9 +146,10 @@ public boolean equals(Object o) {
     }
 
     TypeNode type = (TypeNode) o;
-    return typeKind().equals(type.typeKind())
-        && (isArray() == type.isArray())
-        && Objects.equals(reference(), type.reference());
+    return isBoxedTypeEquals(type, this)
```
miraleung, line 149: Could we please add parentheses around logic groups for clarity? Also, what if one of the boxed types are arrays (which needs a test as well)?

src/main/java/com/google/api/generator/engine/ast/CastExpr.java
xiaozhenliu-gg5: done.

src/main/java/com/google/api/generator/engine/ast/TypeNode.java
xiaozhenliu-gg5: of course, good point. Wrapped the strict equality check to `strictEquals()` method. 
`int[]` cannot be cast to `Integer[]` so boxed type array is not equal to its primitive type array. This check is included in `isBoxedTypeEquals()` method, since we the TypeNode we put in the map is without array setting. Added a unit test to verify that. Thanks

src/main/java/com/google/api/generator/engine/ast/CastExpr.java
```diff
@@ -52,17 +52,16 @@ public CastExpr build() {
...
         Preconditions.checkState(
             !(castType.equals(TypeNode.BOOLEAN) ^ exprType.equals(TypeNode.BOOLEAN)),
             "Numeric and boolean types are not inter-castable");
       } else {
         Preconditions.checkState(
-            TypeNode.isReferenceType(castExpr.type())
-                && (castExpr.expr().type().equals(TypeNode.NULL)
-                    || TypeNode.isReferenceType(castExpr.expr().type())),
-            "Reference types can only be casted to reference types");
+            castExpr.type().equals(castExpr.expr().type())
```
miraleung, line 60: Could we add some parentheses to clarify the order? This currently reads `( cond1 || cond2 && (cond3))`, so the intent of the first three aren't clear.

src/main/java/com/google/api/generator/engine/ast/TypeNode.java
```diff
@@ -138,6 +158,24 @@ public int hashCode() {
...
+  private boolean strictEquals(TypeNode other) {
+    return typeKind().equals(other.typeKind())
+        && (isArray() == other.isArray())
+        && Objects.equals(reference(), other.reference());
+  }
+
+  private boolean isBoxedTypeEquals(TypeNode other) {
+    // If both types are primitive/reference type, return false.
+    // Array of boxed/primitive type is not considered equal.
+    if (isPrimitiveType() == other.isPrimitiveType() || isArray() || other.isArray()) {
```
miraleung, line 170: We probably don't need the `isPrimitiveType` check, since the map equality check will take care of that.

src/main/java/com/google/api/generator/engine/ast/TypeNode.java
```diff
@@ -138,6 +158,24 @@ public int hashCode() {
     return hash;
   }
 
+  private boolean strictEquals(TypeNode other) {
```
miraleung, line 161: Consider making this (and `isBoxedTypeEquals`) un-private, and add a `@VisibleForTesting` annotation. That will allow us to add separate tests for clarity.

src/main/java/com/google/api/generator/engine/ast/TypeNode.java
xiaozhenliu-gg5: Removed the duplicate check and the comments above. thanks for catching this.

src/main/java/com/google/api/generator/engine/ast/TypeNode.java
xiaozhenliu-gg5: sure, sounds reasonable to me, made the two methods as `protected` and `@VisibleForTesting`. This helps me to tested the two methods separately.

src/main/java/com/google/api/generator/engine/ast/CastExpr.java
miraleung: Since this is a long condition, could we please put sub-condition into named variables to the overall one is easier to read?

xiaozhenliu-gg5: Sounds cool, defined a `isReferenceTypeEquals` to check the condition when castType is a reference type.

src/main/java/com/google/api/generator/engine/ast/CastExpr.java
```diff
@@ -52,17 +52,16 @@ public CastExpr build() {
       // Only numeric types can be casted around.
       if (castType.isPrimitiveType() && exprType.isPrimitiveType()) {
         // Check that the types are either both booleans or both not booleans.
-        // TODO(xiaozhenliu): This is one of the places where we need to check the equality of boxed
-        // and primitive types.
         Preconditions.checkState(
             !(castType.equals(TypeNode.BOOLEAN) ^ exprType.equals(TypeNode.BOOLEAN)),
             "Numeric and boolean types are not inter-castable");
       } else {
+        boolean isReferenceTypeEquals =
```
miraleung, line 59: This logic reads to me as "The cast type is a reference and the expression type is either null or a reference type." Could we please update the name to something that better reflects this? And/or break it down further so we have a variable for the expression type check only?

Also, what is the rationale for the current expression type check instead of `!isPrimitiveType && !equals(TypeNode.VOID)`?

xiaozhenliu-gg5: Broke the condition check further down to `exprType_referenceOrNull` and `validReferenceTypeCast`. ReferenceTypeCast is only valid when castType is referenceType AND exprType is NULL or reference.

`!TypeNode.isPrimitiveType(exprType) && !equals(TypeNode.VOID)` only ensures the TypeKind is `Object`, but misses the check for `exprType.reference()` is not null.

src/test/java/com/google/api/generator/engine/ast/CastExprTest.java
```diff
@@ -84,6 +84,52 @@ public void validCastExpr_basicPrimitiveWidening() {
...
+            .setReference(ConcreteReference.withClazz(Integer.class))
+            .setTypeKind(TypeNode.TypeKind.OBJECT)
+            .build();
+    Variable variable = Variable.builder().setName("intArray").setType(intArray).build();
+    VariableExpr variableExpr = VariableExpr.builder().setVariable(variable).build();
+    assertThrows(
+        IllegalStateException.class,
+        () -> CastExpr.builder().setType(integerArray).setExpr(variableExpr).build());
+  }
+
```
miraleung, line 132: Add a cast to Void type?

src/main/java/com/google/api/generator/engine/ast/CastExpr.java
miraleung: Thanks, this is clearer. `isValidPrimitiveTypeCast` seems to imply that `void` should be disallowed, could we add a test to ensure this is the case?

xiaozhenliu-gg5: yes, `isValidPrimitiveTypeCast` ensures that void type is not allowed here.
 While we cannot create a variable with `void` type since we have condition check in Variable: `!variable.type().typeKind().equals(TypeNode.TypeKind.VOID), "Variable type cannot be void"`.

src/test/java/com/google/api/generator/engine/ast/CastExprTest.java
xiaozhenliu-gg5: Done.

src/test/java/com/google/api/generator/engine/ast/CastExprTest.java
```diff
@@ -123,4 +169,13 @@ public void invalidCastExpr_castNumericToBoolean() {
         IllegalStateException.class,
         () -> CastExpr.builder().setType(TypeNode.BOOLEAN).setExpr(variableExpr).build());
   }
+
+  @Test
+  public void invalidCastExpr_castToVoidType() {
+    Variable variable = Variable.builder().setName("x").setType(TypeNode.VOID_OBJECT).build();
```
miraleung, line 175: This looks like it tests `(void) x` where `x` is a void variable. Casting an already-void-typed variable will fail, but what about casting an int-typed variable?

src/main/java/com/google/api/generator/engine/ast/CastExpr.java
miraleung: It doesn't though, because this doesn't look like it prevents `int x; (void) x`.

xiaozhenliu-gg5: Actually we take `void` as a primitive type in `isPrimitiveType()` for TypeNode, so we won't disallow void here. But the assertion above prevents that:
```
      Preconditions.checkState(
          !castType.equals(TypeNode.VOID) && !exprType.equals(TypeNode.VOID),
          "Casting cannot be done on or to void types");
```

src/test/java/com/google/api/generator/engine/ast/CastExprTest.java
xiaozhenliu-gg5: Updated the unit test to `int x =(void) x` . This fails for the first condition check, which does not allow `void` in castType or exprType.

src/main/java/com/google/api/generator/engine/ast/CastExpr.java
```diff
@@ -52,17 +52,18 @@ public CastExpr build() {
...
-        // and primitive types.
         Preconditions.checkState(
             !(castType.equals(TypeNode.BOOLEAN) ^ exprType.equals(TypeNode.BOOLEAN)),
             "Numeric and boolean types are not inter-castable");
       } else {
+        boolean isExprTypeReferenceOrNull =
+            exprType.equals(TypeNode.NULL) || TypeNode.isReferenceType(exprType);
+        boolean isValidReferenceTypeCast =
+            TypeNode.isReferenceType(castType) && isExprTypeReferenceOrNull;
+        boolean isValidPrimitiveTypeCast = castType.equals(exprType);
```
miraleung, line 63: Thanks for adding the test. This variable is still confusing to the reader, can we either rename it or add a check that both types are primitive to this condition?

xiaozhenliu-gg5: confused here, the above condition `if (castType.isPrimitiveType() && exprType.isPrimitiveType())` checks the condition when both types are primitive. Here `isValidPrimitiveTypeCast` means when either exprType or castType is primitive, then if they are equal (strictEqual || boxedTypeEqual) then it's a valid cast.

Updated the name `isValidPrimitiveTypeCast ` to `isEqualType` if it helps. 
The whole condition check is:
- if both types are primitive: they should be both boolean, or both not boolean.
- if not: 1. one type is primitive, another one is reference: check boxed/primitive type equality (`isEqualType`)
              2. both reference type: one reference can be cast to null/reference type  (`isValidReferenceTypeCast`)

PR #137, opened 2020-07-28
feat: add ThisObjectValue

summer-ji-eng:
fullfill `ThisObjectValue`
- Refer instance variable
- Invoke class’s method with or without arguments
- Return the referenced class

### DIFF
--- a/src/main/java/com/google/api/generator/engine/ast/ThisObjectValue.java
+++ b/src/main/java/com/google/api/generator/engine/ast/ThisObjectValue.java
@@ -46,7 +46,8 @@ public abstract class ThisObjectValue implements ObjectValue {
     private ThisObjectValue build() {
       ThisObjectValue thisObjectValue = autoBuild();
       Preconditions.checkState(
-          TypeNode.isReferenceType(thisObjectValue.type()), "The \"this\" object can only refer to object types");
+          TypeNode.isReferenceType(thisObjectValue.type()),
+          "The \"this\" object can only refer to object types");
       return thisObjectValue;
     }
   }
--- a/src/main/java/com/google/api/generator/engine/ast/ThisObjectValue.java
+++ b/src/main/java/com/google/api/generator/engine/ast/ThisObjectValue.java
@@ -21,6 +21,7 @@ import com.google.common.base.Preconditions;
 public abstract class ThisObjectValue implements ObjectValue {
   private static final String THIS_VALUE = "this";
 
+  @Override
   public abstract TypeNode type();
 
   @Override
--- a/src/main/java/com/google/api/generator/engine/ast/ThisObjectValue.java
+++ b/src/main/java/com/google/api/generator/engine/ast/ThisObjectValue.java
@@ -20,6 +20,7 @@ import com.google.common.base.Preconditions;
 @AutoValue
 public abstract class ThisObjectValue implements ObjectValue {
   private static final String THIS_VALUE = "this";
+  private static final String PKG_JAVA_LANG = "java.lang";
 
   public abstract TypeNode type();
 
@@ -47,7 +48,7 @@ public abstract class ThisObjectValue implements ObjectValue {
       Preconditions.checkState(
           TypeNode.isReferenceType(thisObjectValue.type()), "this can only refer to object types");
       Preconditions.checkState(
-          !TypeNode.isJavaLangObject(thisObjectValue.type()),
+          !thisObjectValue.type().reference().isFromPackage(PKG_JAVA_LANG),
           "The class type should belongs to custom object");
       return thisObjectValue;
     }
--- a/src/main/java/com/google/api/generator/engine/ast/ThisObjectValue.java
+++ b/src/main/java/com/google/api/generator/engine/ast/ThisObjectValue.java
@@ -6,10 +6,13 @@ import com.google.common.base.Preconditions;
 @AutoValue
 public abstract class ThisObjectValue implements ObjectValue {
   private static final String THIS_VALUE = "this";
+
   public abstract TypeNode type();
 
   @Override
-  public String value() {return THIS_VALUE;}
+  public String value() {
+    return THIS_VALUE;
+  }
 
   public static ThisObjectValue withType(TypeNode type) {
     return builder().setType(type).build();
@@ -28,13 +31,10 @@ public abstract class ThisObjectValue implements ObjectValue {
     public ThisObjectValue build() {
       ThisObjectValue thisObjectValue = autoBuild();
       Preconditions.checkState(
-          TypeNode.isReferenceType(thisObjectValue.type()),
-          "this can only refer to object types"
-      );
+          TypeNode.isReferenceType(thisObjectValue.type()), "this can only refer to object types");
       Preconditions.checkState(
           !TypeNode.isJavaLang(thisObjectValue.type()),
-          "The class type should belongs to custom object"
-      );
+          "The class type should belongs to custom object");
       return thisObjectValue;
     }
   }
--- a/src/main/java/com/google/api/generator/engine/ast/ThisObjectValue.java
+++ b/src/main/java/com/google/api/generator/engine/ast/ThisObjectValue.java
@@ -1,3 +1,17 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package com.google.api.generator.engine.ast;
 
 import com.google.auto.value.AutoValue;
--- a/src/main/java/com/google/api/generator/engine/ast/ThisObjectValue.java
+++ b/src/main/java/com/google/api/generator/engine/ast/ThisObjectValue.java
@@ -32,8 +32,8 @@ public abstract class ThisObjectValue implements ObjectValue {
           "this can only refer to object types"
       );
       Preconditions.checkState(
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/api/generator/engine/ast/ThisObjectValue.java
```diff
@@ -0,0 +1,55 @@
...
+  public abstract static class Builder {
+    public abstract Builder setType(TypeNode type);
+
+    public abstract ThisObjectValue autoBuild();
+
+    public ThisObjectValue build() {
+      ThisObjectValue thisObjectValue = autoBuild();
+      Preconditions.checkState(
+          TypeNode.isReferenceType(thisObjectValue.type()), "this can only refer to object types");
+      Preconditions.checkState(
```
summer-ji-eng, line 49: I add this new check here to make sure the `this` type is belongs to custom class object. I see all the case of using this is inside of class definition, like `public Class Student {}`. This check is to exclude java build-in class like `java.lang.*`. [example](https://github.com/googleapis/gapic-generator-java/pull/137/files#diff-eb673d6d6c1d0400da66863c9b86eddfR39) [implementation](https://github.com/googleapis/gapic-generator-java/pull/137/files#diff-3eb77b740c6a645ccd1ef7aa836d7348R108)

But I am not pretty sure about this. Open to your advice and suggestion @miraleung @xiaozhenliu-gg5. Thanks a lot.

miraleung: I think we don't need this check because `this` can be misused in many non-`java.lang` classes, which span all Java libs or other deps. Since checking the entire space is not feasible, we may as well not have this check.

src/test/java/com/google/api/generator/engine/ast/ThisObjectValueTest.java
```diff
@@ -0,0 +1,66 @@
...
+    ClassDefinition classDefinition =
+        ClassDefinition.builder()
+            .setPackageString("com.google.example.library.v1.stub")
+            .setName("LibraryServiceStub")
+            .setScope(ScopeNode.PUBLIC)
+            .build();
+    ConcreteReference ref = ConcreteReference.withClazz(classDefinition.getClass());
+    TypeNode typeNode = TypeNode.withReference(ref);
+    ThisObjectValue thisObjectValue = ThisObjectValue.withType(TypeNode.withReference(ref));
+    Assert.assertEquals(thisObjectValue.value(), "this");
```
miraleung, line 48: Nit: Since we hardcoded `"this"` as the return value for `value()`, we probably don't need to test this condition.

src/main/java/com/google/api/generator/engine/ast/ThisObjectValue.java
```diff
@@ -0,0 +1,56 @@
...
+  public static ThisObjectValue withType(TypeNode type) {
+    return builder().setType(type).build();
+  }
+
+  private static Builder builder() {
+    return new AutoValue_ThisObjectValue.Builder();
+  }
+
+  @AutoValue.Builder
+  public abstract static class Builder {
```
miraleung, line 41: Let's remove `public` to make this "private."

src/test/java/com/google/api/generator/engine/ast/ThisObjectValueTest.java
```diff
@@ -0,0 +1,66 @@
...
+package com.google.api.generator.engine.ast;
+
+import static org.junit.Assert.assertThrows;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class ThisObjectValueTest {
+  @Test
+  public void testThisObjectValueTest() {
```
miraleung, line 24: Nit: Let's follow the existing `validFoobar_reasonOrEdgeCase` convention. In g3, folks are migrating away from having "test" as a suffix, since this is already a test class.

src/test/java/com/google/api/generator/engine/ast/ThisObjectValueTest.java
```diff
@@ -0,0 +1,66 @@
...
+            .build();
+    TypeNode typeNode = TypeNode.withReference(ref);
+    ThisObjectValue thisObjectValue = ThisObjectValue.withType(TypeNode.withReference(ref));
+
+    Assert.assertEquals(thisObjectValue.value(), "this");
+    Assert.assertEquals(thisObjectValue.type(), typeNode);
+  }
+
+  @Test
+  public void testThisObjectValueTest_invalid() {
```
miraleung, line 38: The test case has the name "invalid" but the behavior looks like it succeeds?

src/test/java/com/google/api/generator/engine/ast/ThisObjectValueTest.java
```diff
@@ -0,0 +1,66 @@
...
+            .build();
+    ConcreteReference ref = ConcreteReference.withClazz(classDefinition.getClass());
+    TypeNode typeNode = TypeNode.withReference(ref);
+    ThisObjectValue thisObjectValue = ThisObjectValue.withType(TypeNode.withReference(ref));
+    Assert.assertEquals(thisObjectValue.value(), "this");
+    Assert.assertEquals(thisObjectValue.type(), typeNode);
+  }
+
+  @Test
+  public void testThisObjectValueTest_invalid_type() {
```
miraleung, line 53: Can we make this name more specific, e.g. `invalidThisObjectValue_primitiveType`?

src/test/java/com/google/api/generator/engine/ast/ThisObjectValueTest.java
```diff
@@ -0,0 +1,66 @@
...
+        () -> {
+          ThisObjectValue.withType(TypeNode.withReference(ref));
+        });
+    assertThrows(
+        IllegalStateException.class,
+        () -> {
+          ThisObjectValue.withType(TypeNode.BOOLEAN);
+        });
+  }
+}
```
miraleung, line 66: Add a test for `this` on a null type?

src/test/java/com/google/api/generator/engine/writer/JavaWriterVisitorTest.java
```diff
@@ -1654,6 +1655,56 @@ public void writeClassDefinition_statementsAndMethods() {
             "}\n"));
   }
 
+  @Test
+  public void writeMethodReturnThisObjectValue() {
```
miraleung, line 1659: Nit: The name of this test is valid, but since we're aiming to test `this`, let's call it something like `writeThisObjectValue_methodReturn`. Same below.

src/test/java/com/google/api/generator/engine/writer/JavaWriterVisitorTest.java
```diff
@@ -1654,6 +1655,56 @@ public void writeClassDefinition_statementsAndMethods() {
...
+            .build();
+    methodDefinition.accept(writerVisitor);
+    assertEquals(
+        writerVisitor.write(),
+        String.format("public Student apply() {\n" + "return this;\n" + "}\n"));
+  }
+
+  @Test
+  public void writeExprStatementForThisObjectValue() {
+    // Example of ThisObjectValue access its fields
```
miraleung, line 1679: Nit: Usual grammar rules apply to comments, e.g. Example of ThisObjectValue accessing its fields. <-
Same below please.

src/main/java/com/google/api/generator/engine/ast/ThisObjectValue.java
summer-ji-eng: Sounds reasonable. removed. Thanks.

src/test/java/com/google/api/generator/engine/ast/ThisObjectValueTest.java
summer-ji-eng: Gotcha. I don't have change to make changes on g3 yet :( 👍

src/test/java/com/google/api/generator/engine/ast/ThisObjectValueTest.java
summer-ji-eng: I will remove the check to make consistent with other tests.

src/test/java/com/google/api/generator/engine/ast/ThisObjectValueTest.java
summer-ji-eng: agree

src/main/java/com/google/api/generator/engine/ast/ThisObjectValue.java
```diff
@@ -0,0 +1,52 @@
...
+package com.google.api.generator.engine.ast;
+
+import com.google.auto.value.AutoValue;
+import com.google.common.base.Preconditions;
+
+@AutoValue
+public abstract class ThisObjectValue implements ObjectValue {
+  private static final String THIS_VALUE = "this";
+
+  public abstract TypeNode type();
```
miraleung, line 24: `@Override` for `ObjectValue`.

src/test/java/com/google/api/generator/engine/ast/ThisObjectValueTest.java
```diff
@@ -0,0 +1,51 @@
...
+            .setName("Student")
+            .setPakkage("com.google.example.examples.v1")
+            .build();
+    TypeNode typeNode = TypeNode.withReference(ref);
+    ThisObjectValue.withType(typeNode);
+    // No exception thrown, we're good.
+  }
+
+  @Test
+  public void invalidThisObjectValue_NonReferenceType() {
```
miraleung, line 35: Nit: "_nonReferenceType", same with _nullType below.

src/main/java/com/google/api/generator/engine/ast/ThisObjectValue.java
```diff
@@ -0,0 +1,53 @@
...
+  @AutoValue.Builder
+  abstract static class Builder {
+    abstract Builder setType(TypeNode type);
+
+    abstract ThisObjectValue autoBuild();
+
+    private ThisObjectValue build() {
+      ThisObjectValue thisObjectValue = autoBuild();
+      Preconditions.checkState(
+          TypeNode.isReferenceType(thisObjectValue.type()), "this can only refer to object types");
```
miraleung, line 49: Nit: Rephrase to make it clear that "this" is `this`, e.g. `The "this" object can only ...`

PR #138, opened 2020-07-28
[ggj] fix: clean up composer methods

miraleung:
Remove redundant wrappers around helper classes.


src/main/java/com/google/api/generator/gapic/composer/Composer.java
```diff
@@ -48,7 +48,7 @@
   public static List<GapicClass> generateStubClasses(
       Service service, Map<String, Message> messageTypes) {
     List<GapicClass> clazzes = new ArrayList<>();
-    clazzes.add(generateStubServiceStub(service, messageTypes));
+    clazzes.add(ServiceStubClassComposer.instance().generate(service, messageTypes));
```
summer-ji-eng, line 51: I wonder what is the purpose of making this change. From my point of view, function name `generatestubServiceStub` is more readable than this way.
And if we wanna this change, to keep consistency, should we also need to change `generateStubServiceSettings` by using this way.
But I prefer the old way.

miraleung: Having an extra wrapper around `ServiceStubClassComposer` is redundant, since that method is only used once. The rest will be changed as they are implemented, so this will eventually be consistent.

miraleung: Having an extra wrapper around `ServiceStubClassComposer` is redundant, since that method is only used once. The rest will be changed as they are implemented, so this will eventually be consistent.

summer-ji-eng: Gotcha. I think we could try to put a logic into one PR to get more context.

miraleung: Done.

PR #141, opened 2020-07-28
[ggj] fix: enable casting for null expressions

miraleung:


miraleung:
Merging to fix the build.

PR #144, opened 2020-07-29
feat: Add ReferenceConstructorExpr

summer-ji-eng:
include 
- invoke `this` constructor
- invoke `super` constructor
example:
```
public class Rectangle {
    private int x, y;
    private int width, height;
        
    public Rectangle() {
        this(0, 0, 1, 1);
    }
    public Rectangle(int x, int y, int width, int height) {
        this.x = x;
        this.y = y;
        this.width = width;
        this.height = height;
    }
    ...
}
```


src/main/java/com/google/api/generator/engine/ast/ReferenceConstructorExpr.java
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  public static Builder superBuilder() {
+    return new AutoValue_ReferenceConstructorExpr.Builder()
+        .setArguments(Collections.emptyList())
+        .setKeywordKind(KeywordKind.SUPER);
+  }
+
+  @AutoValue.Builder
+  public abstract static class Builder {
```
miraleung, line 54: Let's see if we can remove `public`.

src/test/java/com/google/api/generator/engine/ast/ReferenceConstructorExprTest.java
```diff
@@ -0,0 +1,64 @@
...
+  public void invalidReferenceConstructorExpr_nonReferenceType() {
+    assertThrows(
+        IllegalStateException.class,
+        () -> {
+          ReferenceConstructorExpr.thisBuilder().setType(TypeNode.BOOLEAN).build();
+        });
+  }
+
+  @Test
+  public void invalidReferenceConstructorExpr_NullType() {
```
miraleung, line 57: Nit: _nullType

src/main/java/com/google/api/generator/engine/ast/ReferenceConstructorExpr.java
summer-ji-eng: Tried. We will have compiler error because we still need to access `setArguments, setType` inside of Builder.
Screen Shot 2020-07-30 at 8 57 22 PM

src/main/java/com/google/api/generator/engine/writer/ImportWriterVisitor.java
```diff
@@ -186,6 +187,12 @@ public void visit(EnumRefExpr enumRefExpr) {
     enumRefExpr.type().accept(this);
   }
 
+  @Override
+  public void visit(ReferenceConstructorExpr referenceConstructorExpr) {
+    referenceConstructorExpr.type().accept(this);
```
miraleung, line 192: Add a test?

summer-ji-eng: Agree. Added.

src/main/java/com/google/api/generator/engine/ast/ReferenceConstructorExpr.java
```diff
@@ -0,0 +1,71 @@
...
+    public abstract Builder setType(TypeNode node);
+
+    public abstract Builder setArguments(List<Expr> arguments);
+
+    abstract Builder setKeywordKind(KeywordKind keywordKind);
```
miraleung, line 59: Nit: Please add a `// Private.` comment for clarity.

src/test/java/com/google/api/generator/engine/ast/ReferenceConstructorExprTest.java
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  @Test
+  public void invalidReferenceConstructorExpr_nonReferenceType() {
+    assertThrows(
+        IllegalStateException.class,
+        () -> {
+          ReferenceConstructorExpr.thisBuilder().setType(TypeNode.BOOLEAN).build();
+        });
+  }
```
miraleung, line 54: Add a test for `Boolean` or `Integer` object, e.g. for classes that may inherit from those objects.

summer-ji-eng: Sorry, I misunderstand here. First `Boolean` and `Integer` are final class which can't be inherited.

Do you mean add test which setType(TypeNode.Integer)? For now, it will not throw error.
In `ReferenceConstructorExpr.java`, only checked the type `isReferencedType`, however, `Integer, String` are referenceType but they are belongs to `java.lang`

If we want the check, we could add 
```
      Preconditions.checkState(referenceConstructorExpr.type().reference().isFromPackage(PKG_JAVA_LANG),
          "ReferenceConstructExpr can only refer to custom object. ");
``

Let me know what you think. @miraleung thanks.

miraleung: SG, let's leave it as is then.

PR #146, opened 2020-08-03
[ggj] feat: integrate comment statement with method definition

xiaozhenliu-gg5:


src/main/java/com/google/api/generator/engine/ast/MethodDefinition.java
```diff
@@ -34,7 +34,9 @@
   // Required.
   public abstract IdentifierNode methodIdentifier();
 
-  // TODO(xiahzhenliu): Add a Nullable JavaDoc here.
+  @Nullable
+  public abstract ImmutableList<CommentStatement> commentStatements();
```
miraleung, line 38: Let's call this `headerCommentStatements` or similar.

xiaozhenliu-gg5: Done. Thanks

src/main/java/com/google/api/generator/engine/ast/MethodDefinition.java
```diff
@@ -34,7 +34,9 @@
   // Required.
   public abstract IdentifierNode methodIdentifier();
 
-  // TODO(xiahzhenliu): Add a Nullable JavaDoc here.
+  @Nullable
```
miraleung, line 37: Why `@Nullable`?

xiaozhenliu-gg5: Removed it since we have default value for it.

src/test/java/com/google/api/generator/engine/ast/MethodDefinitionTest.java
```diff
@@ -661,4 +662,14 @@ private static AssignmentExpr createAssignmentExpr() {
...
   }
+
+  private static List<CommentStatement> createCommentStatements() {
+    JavaDocComment javaDocComment =
+        JavaDocComment.builder()
+            .addComment("Constructs an instance of GrpcMyProtoStub, using the given settings.")
+            .addComment(
+                "This is protected so that it is easy to make a subclass, but otherwise, the static factory methods should be preferred.")
+            .build();
+    return new ArrayList<>(Arrays.asList(CommentStatement.withComment(javaDocComment)));
```
miraleung, line 673: Why do we need a new `ArrayList` wrapping a List?

xiaozhenliu-gg5: nice catch, `Arrays.asList()` only should work, thanks.

PR #157, opened 2020-08-06
[ggj] feat: support file header (apache license) for outer class definition

xiaozhenliu-gg5:


src/main/java/com/google/api/generator/engine/ast/ClassDefinition.java
```diff
@@ -23,6 +23,9 @@
 
 @AutoValue
 public abstract class ClassDefinition implements AstNode {
+  // Required for outer classes.
+  @Nullable
```
miraleung, line 27: Why `@Nullable`?

src/main/java/com/google/api/generator/engine/ast/ClassDefinition.java
```diff
@@ -122,8 +127,9 @@ public ClassDefinition build() {
 
       ClassDefinition classDef = autoBuild();
 
-      // Only nested classes can forego having a package.
+      // Only nested classes can forego having a package and file header.
       if (!classDef.isNested()) {
+        Preconditions.checkNotNull(classDef.fileHeader(), "Outer classes must have a file header.");
```
miraleung, line 132: Could we please remove this check? Having a file header shouldn't be mandatory.

src/main/java/com/google/api/generator/gapic/composer/FileHeader.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public class FileHeader {
+  public static final List<CommentStatement> APACHE_LICENSE = createApacheLicense();
+
+  private static List<CommentStatement> createApacheLicense() {
+    String[] fileHeadeStrings = {
+      "Copyright 2020 Google LLC",
```
miraleung, line 28: Can this be a single string instead?

src/main/java/com/google/api/generator/gapic/composer/FileHeader.java
```diff
@@ -0,0 +1,51 @@
...
+      "WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.",
+      "See the License for the specific language governing permissions and",
+      "limitations under the License."
+    };
+
+    List<CommentStatement> apacheFileHeader = new ArrayList<>();
+    Arrays.stream(fileHeadeStrings)
+        .forEach(
+            s -> {
+              apacheFileHeader.add(CommentStatement.withComment(LineComment.withComment(s)));
```
miraleung, line 47: Clients ([example](https://github.com/googleapis/java-dataproc/blob/master/google-cloud-dataproc/src/main/java/com/google/cloud/dataproc/v1/stub/AutoscalingPolicyServiceStubSettings.java))currently have this in a BlockComment, so let's use that instead.

src/main/java/com/google/api/generator/gapic/composer/FileHeader.java
```diff
@@ -0,0 +1,51 @@
...
+
+package com.google.api.generator.gapic.composer;
+
+import com.google.api.generator.engine.ast.CommentStatement;
+import com.google.api.generator.engine.ast.LineComment;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public class FileHeader {
```
miraleung, line 23: Can we rename this to something more specific, such as "ApacheLicenseComposer"?

src/main/java/com/google/api/generator/gapic/composer/FileHeader.java
```diff
@@ -0,0 +1,51 @@
...
+  public static final List<CommentStatement> APACHE_LICENSE = createApacheLicense();
```
miraleung, line 24: Can we set the value directly?

src/main/java/com/google/api/generator/gapic/composer/FileHeader.java
```diff
@@ -0,0 +1,51 @@
...
+
```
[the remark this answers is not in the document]
miraleung: Let's remove the licenses from ClassDefinition and JavaWriterVisitor (except for one longer test), since that should not be mandatory.

src/main/java/com/google/api/generator/gapic/composer/GrpcServiceCallableFactoryClassComposer.java
```diff
@@ -72,6 +72,7 @@ public GapicClass generate(Service service, Map<String, Message> ignore) {
 
     ClassDefinition classDef =
         ClassDefinition.builder()
+            .setFileHeader(FileHeader.APACHE_LICENSE)
```
miraleung, line 75: Instead of changing this in each class, let's change all the ClassDefinitions at once in Composer.

src/test/java/com/google/api/generator/engine/ast/ClassDefinitionTest.java
```diff
@@ -132,12 +136,26 @@ public void invalidClassDefinition_implementsNullType() {
         });
   }
 
+  @Test
+  public void invalidClassDefinition_outerClassMissingFileHeader() {
```
miraleung, line 140: Can we ensure that missing a file header is valid?

src/test/java/com/google/api/generator/engine/ast/ClassDefinitionTest.java
```diff
@@ -64,6 +65,7 @@ public void validClassDefinition_nestedBasic() {
   @Test
   public void validClassDefinition_withAnnotationsExtendsAndImplements() {
     ClassDefinition.builder()
+        .setFileHeader(createFileHeader())
```
miraleung, line 68: Let's remove this line, same below.

src/test/java/com/google/api/generator/gapic/composer/GrpcServiceStubClassComposerTest.java
```diff
@@ -55,7 +55,21 @@ public void generateServiceClasses() {
 
   // TODO(miraleung): Update this when a file-diffing test mechanism is in place.
   private static final String EXPECTED_CLASS_STRING =
-      "package com.google.showcase.v1beta1.stub;\n"
+      "// Copyright 2020 Google LLC\n"
```
miraleung, line 58: We can remove these blocks when we migrate the header addition to Composer.java.

src/main/java/com/google/api/generator/gapic/composer/FileHeader.java
xiaozhenliu-gg5: fix for BlockComment is here: https://github.com/googleapis/gapic-generator-java/pull/168
It cannot be wrapped in the current blockComment because the formatting will be messed up, more details are in #168 PR description. Thanks

src/main/java/com/google/api/generator/gapic/composer/GrpcServiceCallableFactoryClassComposer.java
xiaozhenliu-gg5: Not sure we can move the header addition to Composer, because in Composer class, we only `generateGenericClass` there, for class definition like `GrpcServiceCallableFactoryClassComposer`, it's [defined](https://github.com/googleapis/gapic-generator-java/blob/master/src/main/java/com/google/api/generator/gapic/composer/GrpcServiceCallableFactoryClassComposer.java#L73) in its own class.

src/test/java/com/google/api/generator/gapic/composer/GrpcServiceStubClassComposerTest.java
xiaozhenliu-gg5: It's removed since it's tested in `ComposerTest`

xiaozhenliu-gg5:
Left some questions in the unresolved comments @miraleung Summarized them here for clarity:
1. altered BlockComment class for supporting file header composer.
2. all file header addition to *Composer classes are removed since it's optional, but add the file header to `Composer.java` itself and test it in new `ComposerTest.java`.

src/main/java/com/google/api/generator/gapic/composer/FileHeader.java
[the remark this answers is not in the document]
xiaozhenliu-gg5: Talked offline to understand this point. (unresolved though)

src/main/java/com/google/api/generator/gapic/composer/ApacheLicenseComposer.java
```diff
@@ -0,0 +1,37 @@
...
+          + "you may not use this file except in compliance with the License.\n"
+          + "You may obtain a copy of the License at\n\n"
+          + "     https://www.apache.org/licenses/LICENSE-2.0\n\n"
+          + "Unless required by applicable law or agreed to in writing, software\n"
+          + "distributed under the License is distributed on an \"AS IS\" BASIS,\n"
+          + "WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.\n"
+          + "See the License for the specific language governing permissions and\n"
+          + "limitations under the License.";
+
+  public static final List<CommentStatement> APACHE_LICENSE =
```
miraleung, line 35: Nit: APACHE_LICENSE_COMMENT_STATEMENT

src/main/java/com/google/api/generator/gapic/composer/ApacheLicenseComposer.java
```diff
@@ -0,0 +1,37 @@
...
+// limitations under the License.
+
+package com.google.api.generator.gapic.composer;
+
+import com.google.api.generator.engine.ast.BlockComment;
+import com.google.api.generator.engine.ast.CommentStatement;
+import java.util.Arrays;
+import java.util.List;
+
+public class ApacheLicenseComposer {
```
miraleung, line 22: Looking at this again, maybe call this class "ApacheLicense" instead, since it doesn't do any composing?

src/test/java/com/google/api/generator/gapic/composer/ComposerTest.java
```diff
@@ -0,0 +1,80 @@
...
+  public void generateStubServiceSettings() {
+    Map<String, Message> messageTypes = Parser.parseMessages(echoFileDescriptor);
+    Map<String, ResourceName> resourceNames = Parser.parseResourceNames(echoFileDescriptor);
+    Set<ResourceName> outputResourceNames = new HashSet<>();
+
+    List<Service> services =
+        Parser.parseService(echoFileDescriptor, messageTypes, resourceNames, outputResourceNames);
+    Service echoProtoService = services.get(0);
+    JavaWriterVisitor visitor = new JavaWriterVisitor();
+    Composer.generateStubServiceSettings(echoProtoService).classDefinition().accept(visitor);
```
miraleung, line 55: This class will be changing soon, can we make the test more robust?

src/main/java/com/google/api/generator/gapic/composer/Composer.java
```diff
@@ -100,6 +102,7 @@ private static GapicClass generateGenericClass(Kind kind, String name, Service s
 
     ClassDefinition classDef =
         ClassDefinition.builder()
+            .setFileHeader(ApacheLicenseComposer.APACHE_LICENSE)
```
miraleung, line 105: We currently generate many classes, and this method is a placeholder that will be going away. Can you think of ways to restructure this such that we can add a license to a class and test that in isolation? How would we incorporate that into the existing classes?

src/main/java/com/google/api/generator/gapic/composer/ApacheLicenseComposer.java
xiaozhenliu-gg5: since it's not a composer, how about moving it to `utils`?

src/main/java/com/google/api/generator/gapic/composer/Composer.java
xiaozhenliu-gg5: sure, good point. Since we are creating different GapicClasses in the composers, I think a central place to add a file header is `GapicClass` class.  Call `GapicClass.create().addApacheLicnese()` for adding license to existing class. 3 reasons are in consideration:

1. ClassDefinition is another alternative place to have this helper, but in the end, it's gapic class that needs/wraps it, so putting the helper in gapic class might make more sense.
2. The way to incorporate header to the existing class is `GapicClass.create(kind, classDef).addApacheLicense();` 
3. it can be independently tested in `GapicClassTest` without altering composer files.

Alternatives: 
1. tried to have the helper `addApacheLicense(GapicClass)` in `Composer` class, but in that way we will call it by `clazzes.add(**Composer.instance().generate().addApacheLicense())` I think adding file header is more suitable in a class level, instead of composer level.
2. add the helper to `ApacheLicense` as a util, but `ApacheLicense.addLicenseToGapicClass()` method should makes more sense to directly be called from the class side `GapicClass.addApacheLicense()`
Let me know your thoughts @miraleung Thanks

miraleung: GapicClass has no extra logic beyond holding data, so let's keep it that way. Let's add a helper method in Composer.java that adds license headers to each GapicClass instead.

xiaozhenliu-gg5: updated the `Composer` class and add `ComposerTest` to test file header addition in isolation.

src/main/java/com/google/api/generator/engine/writer/JavaWriterVisitor.java
```diff
@@ -62,14 +62,15 @@
 
   private static final String COLON = ":";
   private static final String COMMA = ",";
-  private static final String BLOCK_COMMENT_START = "/**";
```
miraleung, line 65: Please separate the changes in #168 from this PR.

src/main/java/com/google/api/generator/gapic/composer/Composer.java
```diff
@@ -106,4 +108,23 @@ private static GapicClass generateGenericClass(Kind kind, String name, Service s
             .build();
     return GapicClass.create(kind, classDef);
   }
+
+  @VisibleForTesting
+  protected static List<GapicClass> addApacheLicenseToGapicClassList(
+      List<GapicClass> gapicClassList) {
+    gapicClassList =
+        gapicClassList.stream()
```
miraleung, line 116: Let's return the value instead of reassigning to the variable.

src/main/java/com/google/api/generator/engine/writer/JavaWriterVisitor.java
xiaozhenliu-gg5: Hold this PR until #168 is merged.

src/main/java/com/google/api/generator/engine/ast/ClassDefinition.java
```diff
@@ -78,8 +81,12 @@ public static Builder builder() {
         .setNestedClasses(Collections.emptyList());
   }
 
+  public abstract Builder toBuilder();
+
   @AutoValue.Builder
   public abstract static class Builder {
+    public abstract Builder setFileHeader(List<CommentStatement> fileHeader);
```
miraleung, line 88: Does it make sense to error-out in `build()` if this is a nested class?

src/main/java/com/google/api/generator/gapic/composer/Composer.java
```diff
@@ -106,4 +108,21 @@ private static GapicClass generateGenericClass(Kind kind, String name, Service s
             .build();
     return GapicClass.create(kind, classDef);
   }
+
+  @VisibleForTesting
+  protected static List<GapicClass> addApacheLicenseToGapicClassList(
```
miraleung, line 113: For consistency with other similar methods, just `addApacheLicense` so we can change the argument type without changing the method name.

src/test/java/com/google/api/generator/engine/ast/ClassDefinitionTest.java
```diff
@@ -347,4 +348,8 @@ private static ForStatement createForStatement() {
         .setBody(Arrays.asList(ExprStatement.withExpr(createAssignmentExpr())))
         .build();
   }
+  // Create a simple block comment to stand for the Apache License header.
+  private static List<CommentStatement> createFileHeader() {
```
miraleung, line 352: If this is just used once, do we need a helper method for it? Would prefer to just add the logic to the method where it's used.

src/main/java/com/google/api/generator/engine/ast/ClassDefinition.java
xiaozhenliu-gg5: yes, added one condition check in `build` to confirm nested classes do not have file headers.

src/main/java/com/google/api/generator/gapic/composer/Composer.java
xiaozhenliu-gg5: Done.

src/test/java/com/google/api/generator/engine/ast/ClassDefinitionTest.java
xiaozhenliu-gg5: Updated with a unit test (nestClassWithFileHeader) which also uses this. So I guess we can keep it. But this similar helper is only called once in JavaWriterVisitorTest, so removed it there. Thanks

src/main/java/com/google/api/generator/engine/ast/ClassDefinition.java
```diff
@@ -129,6 +136,9 @@ public ClassDefinition build() {
         Preconditions.checkState(!classDef.isStatic(), "Outer classes cannot be static");
         Preconditions.checkState(
             !classDef.scope().equals(ScopeNode.PRIVATE), "Outer classes cannot be private");
+      } else {
+        Preconditions.checkState(
+            classDef.fileHeader().isEmpty(), "Nested classes cannot have file header");
```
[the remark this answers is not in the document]
xiaozhenliu-gg5: Fixed.

src/main/java/com/google/api/generator/gapic/utils/ApacheLicense.java
```diff
@@ -0,0 +1,37 @@
...
+
+package com.google.api.generator.gapic.utils;
+
+import com.google.api.generator.engine.ast.BlockComment;
+import com.google.api.generator.engine.ast.CommentStatement;
+import java.util.Arrays;
+import java.util.List;
+
+public class ApacheLicense {
+  private static final String fileHeadeString =
```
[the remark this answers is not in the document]
xiaozhenliu-gg5: Fixed

PR #145, opened 2020-07-29
feat: Add SuperObjectValue and unit test

summer-ji-eng:
fullfill super

- access parent class's fields
- access parent class's method

PR #159, opened 2020-08-07
feat: Add SuperObjectValue component

summer-ji-eng:
Correct #145 merge problem
#145 closed


src/main/java/com/google/api/generator/engine/ast/SuperObjectValue.java
```diff
@@ -0,0 +1,54 @@
...
+
+import com.google.auto.value.AutoValue;
+import com.google.common.base.Preconditions;
+
+@AutoValue
+public abstract class SuperObjectValue implements ObjectValue {
+  private static final String SUPER_VALUE = "super";
+
+  @Override
+  public abstract TypeNode type();
```
miraleung, line 25: Please add a comment here (or at withType below) that this will be the type of the "super" object, and not the child type of the supertype.

summer-ji-eng: Done.

PR #172, opened 2020-08-10
[ggj] feat: ResourceName codegen fields

miraleung:


src/test/java/com/google/api/generator/gapic/composer/ResourceNameHelperClassComposerTest.java
```diff
@@ -0,0 +1,170 @@
...
+    assertEquals(2, tokenHierarchies.size());
+    assertThat(tokenHierarchies.get(0)).containsExactly("project", "session");
+    assertThat(tokenHierarchies.get(1))
+        .containsExactly("project", "environment", "user", "session");
+  }
+
+  @Test
+  public void parseTokenHierarchy_invalidPatterns() {
+    List<String> patterns =
+        Arrays.asList(
```
xiaozhenliu-gg5, line 83: For confirmation, we do not want throw errors even if this pattern is invalid, right?

Also how about the pattern with wildcard in token like `project/{project=**}`? Or this is not in our consideration yet. Thanks

src/main/java/com/google/api/generator/gapic/composer/ResourceNameHelperClassComposer.java
```diff
@@ -96,4 +204,61 @@ public GapicClass generate(ResourceName resourceName) {
                 c -> c.getSimpleName(),
                 c -> TypeNode.withReference(ConcreteReference.withClazz(c))));
   }
+
+  private static Map<String, VariableExpr> createFixedClassMemberVariables() {
+    Map<String, TypeNode> memberVars = new HashMap<>();
```
xiaozhenliu-gg5, line 209: Curious the reason why we call it `fixedClassMemberVariable`, seems it has String as key and its corresponding TypeNode as value. Will we append more key-values here? Thanks! 
And some comments would be helpful for this helper if possible

src/test/java/com/google/api/generator/gapic/composer/ResourceNameHelperClassComposerTest.java
miraleung: No, we don't want to throw errors here. The wildcard patterns will be handled by L237 in the corresponding class.

src/main/java/com/google/api/generator/gapic/composer/ResourceNameHelperClassComposer.java
miraleung: The "fixed" part refers to the independent nature of these variables, in the sense that they don't depend on the underlying protobuf.

PR #176, opened 2020-08-11
[ggj] feat: add ctors to ResourceName class codegen

miraleung:


src/test/java/com/google/api/generator/gapic/composer/ResourceNameHelperClassComposerTest.java
```diff
@@ -166,5 +167,26 @@ public void generateResourceNameClass() {
           + "  private String foobar;\n"
           + "  private String variant;\n"
           + "  private String project;\n"
+          + "\n"
+          + "  @Deprecated\n"
+          + "  private FoobarName() {}\n"
+          + "\n"
```
xiaozhenliu-gg5, line 173: Use [java-speech](https://github.com/googleapis/java-speech/blob/master/proto-google-cloud-speech-v1p1beta1/src/main/java/com/google/cloud/speech/v1p1beta1/PhraseSetName.java#L61) as an example, trying to understand why we have a deprecated private constructor here. Thanks

PR #177, opened 2020-08-11
[ggj] feat: add of and format methods to ResName codegen

miraleung:


src/test/java/com/google/api/generator/gapic/composer/ResourceNameHelperClassComposerTest.java
```diff
@@ -227,5 +227,62 @@ public void generateResourceNameClass() {
           + "  public Builder toBuilder() {\n"
           + "    return new Builder(this);\n"
           + "  }\n"
+          + "\n"
+          + "  public static FoobarName of(String project, String foobar) {\n"
+          + "    return newProjectFoobarBuilder().setProject(project).setFoobar(foobar).build();\n"
+          + "  }\n"
```
xiaozhenliu-gg5, line 233: this `newProjectFoobarBuilder` seems not defined above, there is a `ProjectFoobarBuilder` instead, should it aslo start with `new` name? 

Also `ProjectFoobarBuilder` and `newBuilder` both return `new Builder()` is this intended? Thanks.

src/test/java/com/google/api/generator/gapic/composer/ResourceNameHelperClassComposerTest.java
```diff
@@ -227,5 +227,62 @@ public void generateResourceNameClass() {
...
+          + "\n"
+          + "  @BetaApi(\"The static create methods are not stable yet and may be changed in the"
+          + " future.\")\n"
+          + "  public static FoobarName ofProjectFoobarBuilder() {\n"
+          + "    return newProjectFoobarBuilder().setProject(project).setFoobar(foobar).build();\n"
```
xiaozhenliu-gg5, line 238: below `of***Builder` method all has arguments of the setter tokens, is this intended here that we will construct the Builder from the global variable `project` and `foobar` instead of passing them in?

src/test/java/com/google/api/generator/gapic/composer/ResourceNameHelperClassComposerTest.java
miraleung: Fixed `newProjectFoobarBuilder`. Yes, returning `new Builder()` for both is intended, e.g. in [this class](https://github.com/googleapis/java-dataproc/blob/master/proto-google-cloud-dataproc-v1/src/main/java/com/google/cloud/dataproc/v1/AutoscalingPolicyName.java).

src/test/java/com/google/api/generator/gapic/composer/ResourceNameHelperClassComposerTest.java
miraleung: Fixed.

PR #180, opened 2020-08-12
[ggj] feat: add ReturnExpr to AST

miraleung:


src/main/java/com/google/api/generator/engine/ast/AstNodeVisitor.java
```diff
@@ -47,6 +47,8 @@
 
   public void visit(EnumRefExpr enumRefExpr);
 
+  public void visit(ReturnExpr enumRefExpr);
+
```
xiaozhenliu-gg5, line 51: nit: `returnExpr`

src/main/java/com/google/api/generator/engine/ast/MethodDefinition.java
```diff
@@ -275,6 +275,10 @@ public MethodDefinition build() {
         }
 
         if (method.returnExpr() != null && !isLastStatementThrowExpr) {
+          // TODO(miraleung): Refactor this to use ReturnExpr under the covers instead.
+          Preconditions.checkState(
+              !(method.returnExpr() instanceof ReturnExpr),
+              "A method's return expression can only consist of non-ReturnExpr expressions");
           if (method.returnType().isPrimitiveType()) {
```
xiaozhenliu-gg5, line 282: Curious about the motivation of adding this new class `ReturnExpr`, it has keyword `return` with an expression. And I can see it's one of the valid expression statement, method's return expr cannot be a ReturnType. Is this only for method return expr checking? 
For TODO comment ` Refactor this to use ReturnExpr under the covers instead.` what does this mean? Thanks!

miraleung: The motivation is to enable generating code like `if (true) { return x; }`. We prevent the returnExpr in MethodDefinition from being a ReturnExpr so that we don't do something like `return return x`. The TODO means that in the future, I will clean this up to make MethodDefinition use ReturnExpr in the implementation instead of its current approach.

src/main/java/com/google/api/generator/engine/ast/AstNodeVisitor.java
miraleung: Done.

PR #187, opened 2020-08-12
[ggj] fix: fix condition expression type in ternary expression

xiaozhenliu-gg5:


xiaozhenliu-gg5:
This PR is independent, only for fixing the condition check (condition expression should be boolean type), and also removed unused lines in unit test. it's nothing related to boxed/primitive check. One unit test is added where the condition expression is not boolean. @miraleung Thanks.

src/main/java/com/google/api/generator/engine/ast/TernaryExpr.java
```diff
@@ -54,15 +54,21 @@ public static Builder builder() {
...
     abstract TernaryExpr autoBuild();
 
     public TernaryExpr build() {
+      Preconditions.checkState(
+          conditionExpr().type().equals(TypeNode.BOOLEAN),
+          "Ternary condition must be a boolean-typed expression.");
+
       if (!thenExpr().type().equals(elseExpr().type())) {
-        // Not both primitives, and no boxed equality, and one of them is null.
+        // If the types are not equal, then they should be assignable to the other.
```
miraleung, line 71: Removing the comment should be fine, since the check is self-documenting now.

src/test/java/com/google/api/generator/engine/ast/TernaryExprTest.java
```diff
@@ -41,9 +38,9 @@ public void validTernaryExpr_primitiveType() {
             .setThenExpr(thenExpr)
             .setElseExpr(elseExpr)
             .build();
-    assertThat(ternaryExpr.conditionExpr().type()).isEqualTo(TypeNode.BOOLEAN);
-    assertThat(ternaryExpr.thenExpr().type()).isEqualTo(ternaryExpr.elseExpr().type());
-    assertThat(ternaryExpr.type()).isEqualTo(ternaryExpr.thenExpr().type());
+    assertTrue(ternaryExpr.conditionExpr().type().equals(TypeNode.BOOLEAN));
```
miraleung, line 41: `assertEquals`?

src/test/java/com/google/api/generator/engine/ast/TernaryExprTest.java
```diff
@@ -66,9 +63,9 @@ public void validTernaryExpr_objectType() {
...
```
miraleung, line 66: Ditto, same below.

PR #189, opened 2020-08-13
[ggj] feat: boxed/primitive type equality check for while statement

xiaozhenliu-gg5:
[5/6] support boxed/primitive type equality in while statement.


src/test/java/com/google/api/generator/engine/ast/WhileStatementTest.java
```diff
@@ -0,0 +1,90 @@
...
+    // The condition expression type can be boolean or its boxed type.
+    VariableExpr condExpr =
+        VariableExpr.withVariable(
+            Variable.builder().setName("condition").setType(TypeNode.BOOLEAN_OBJECT).build());
+    WhileStatement whileStatement =
+        WhileStatement.builder()
+            .setConditionExpr(condExpr)
+            .setBody(createAssignmentExprList())
+            .build();
+    assertTrue(whileStatement.conditionExpr().type().equals(TypeNode.BOOLEAN));
```
xiaozhenliu-gg5, line 46: The only place that needs boxed/primitive check is the condition expression, it should allow boolean type or its boxed type. And added unit test for better coverage.

miraleung, line 46: `assertEquals`? Same elsewhere.

xiaozhenliu-gg5: This assertTrue is intended, since it does not have to be `boolean`, this check should also allow `Boolean` object, same elsewhere, let me know which one you prefer. And I will make it consistent. Thanks

miraleung: Why wouldn't `assertEquals(TypeNode.BOOLEAN, whileStatement.conditionExpr().type())` check for `Boolean`?

xiaozhenliu-gg5: Got the point, fixed it.

PR #190, opened 2020-08-13
[ggj] feat: boxed/primitive equality check for method definition class

xiaozhenliu-gg5:
[6/6] support boxed/primitive type equality in method definition class.


src/main/java/com/google/api/generator/engine/ast/MethodDefinition.java
```diff
@@ -275,18 +275,11 @@ public MethodDefinition build() {
...
-            Preconditions.checkState(
-                !method.returnExpr().type().isPrimitiveType()
-                    && method.returnType().isSupertypeOrEquals(method.returnExpr().type()),
-                "Method reference return type is not a subtype of the return expression type");
-          }
+          TypeNode returnExprType = method.returnExpr().type();
+          TypeNode returnType = method.returnType();
+          Preconditions.checkState(
+              returnExprType.equals(returnType) || returnType.isSupertypeOrEquals(returnExprType),
+              "Method return type should be equal to the return expression type, or it should be a super type of the return expression type.");
```
xiaozhenliu-gg5, line 282: Updated the error message since return type should be a `super type` of the return expression type, not `subtype`

src/main/java/com/google/api/generator/engine/ast/MethodDefinition.java
```diff
@@ -275,18 +275,11 @@ public MethodDefinition build() {
...
-                    && method.returnType().equals((method.returnExpr().type())),
-                "Method primitive return type does not match the return expression type");
-
-          } else {
-            Preconditions.checkState(
-                !method.returnExpr().type().isPrimitiveType()
-                    && method.returnType().isSupertypeOrEquals(method.returnExpr().type()),
-                "Method reference return type is not a subtype of the return expression type");
-          }
+          TypeNode returnExprType = method.returnExpr().type();
```
miraleung, line 278: What's the motivation for this change? Keeping the checks separate makes them clearer, though we can remove the redundant `!isPrimitiveType()` check in the else-block above.

xiaozhenliu-gg5: Cool, kept the original checks to make them clearer. 
1. removed redundant `isPrimitiveType()` in the first block, since returnExpr type can be reference type for boxed/primitive type equality.
2. removed `isPrimitiveType()` check in else block, but `isSupertypeOrEquals ` only checks when both types are reference types. So added `equals()` condition to allow boxed/primitive types.

src/main/java/com/google/api/generator/engine/ast/MethodDefinition.java
```diff
@@ -277,15 +277,14 @@ public MethodDefinition build() {
...
+                "Method primitive return type is not equal to the return expression type");
 
           } else {
             Preconditions.checkState(
-                !method.returnExpr().type().isPrimitiveType()
-                    && method.returnType().isSupertypeOrEquals(method.returnExpr().type()),
-                "Method reference return type is not a subtype of the return expression type");
+                method.returnType().equals(method.returnExpr().type())
+                    || method.returnType().isSupertypeOrEquals(method.returnExpr().type()),
+                "Method reference return type is not a super/boxed type of the return expression type");
```
xiaozhenliu-gg5, line 287: Cool, kept the original checks to make them clearer.

1. removed redundant `isPrimitiveType()` in the first block, since `returnExpr` type can be reference type for boxed/primitive type equality.
2. removed `isPrimitiveType()` check in else block, but `isSupertypeOrEquals` only checks when both types are reference types. So added `equals()` condition to allow boxed/primitive types.

src/main/java/com/google/api/generator/engine/ast/MethodDefinition.java
```diff
@@ -277,15 +277,14 @@ public MethodDefinition build() {
         if (method.returnExpr() != null && !isLastStatementThrowExpr) {
           if (method.returnType().isPrimitiveType()) {
             Preconditions.checkState(
-                method.returnExpr().type().isPrimitiveType()
-                    && method.returnType().equals((method.returnExpr().type())),
-                "Method primitive return type does not match the return expression type");
+                method.returnType().equals((method.returnExpr().type())),
+                "Method primitive return type is not equal to the return expression type");
```
miraleung, line 281: Would prefer keeping the original message, due to the actual semantics of `equals()`.

src/main/java/com/google/api/generator/engine/ast/MethodDefinition.java
miraleung: The `equals()` check is now partly redundant. Instead, let's make the first primitive-check condition `if (method.returnType().isPrimitiveType() || method.returnExpr().type().isPrimitiveType())` and keep the equals check there. 

That will also let us keep the original message here.

src/main/java/com/google/api/generator/engine/ast/MethodDefinition.java
xiaozhenliu-gg5: Thanks, the advice below sounds reasonable to me, changed it back.

src/main/java/com/google/api/generator/engine/ast/MethodDefinition.java
xiaozhenliu-gg5: Yes, the `equals` check in second block is only for boxed/primitive type equality which is partly redundant. That's a great idea, moved the primitive types check to the first block and reverted back the message change.

PR #225, opened 2020-08-26
[ggj][ast] fix: prevent duplicate MethodDefinition annotations

miraleung:


src/main/java/com/google/api/generator/engine/ast/MethodDefinition.java
```diff
@@ -122,10 +123,7 @@ public static Builder constructorBuilder() {
     public abstract Builder setHeaderCommentStatements(
         List<CommentStatement> headeCommentStatements);
 
-    public Builder setAnnotations(List<AnnotationNode> annotations) {
-      annotationsBuilder().addAll(annotations);
-      return this;
-    }
```
xiaozhenliu-gg5, line 128: Just a thought that if we can have the duplicates removal here when setting Annotations, so the adding @override logic stays the same. Or we may have more than one @override in the annotation list, that's why we wanna remove duplicates after all annotations are added. Thanks

miraleung: It's more efficient to do the removal in `autoBuild()`.

PR #232, opened 2020-08-27
[ggj][codegen] feat: initial param block, RetrySettingsComposer, test

miraleung:


src/main/java/com/google/api/generator/gapic/model/GapicServiceConfig.java
```diff
@@ -77,7 +79,8 @@ public static GapicServiceConfig create(ServiceConfig serviceConfig) {
             Collectors.toMap(
                 m -> getRetryCodeName(service, m),
                 m -> retryCodesLookup(service, m),
-                (l1, l2) -> l2));
+                (l1, l2) -> l2,
+                LinkedHashMap::new));
```
xiaozhenliu-gg5, line 83: TWIL :)

PR #255, opened 2020-08-30
[javanaut]feat: Add operation expr interface and operator kind enum

summer-ji-eng:
Prepare common interface operation expr for
- Arithmetic Operation Expr
- Unary Operation Expr
- Relational Operation Expr
- Logical Operation Expr

### DIFF
--- a/src/main/java/com/google/api/generator/engine/ast/OperationExpr.java
+++ b/src/main/java/com/google/api/generator/engine/ast/OperationExpr.java
@@ -1,3 +1,5 @@
+// Copyright 2020 Google LLC
+//
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
--- a/src/main/java/com/google/api/generator/engine/ast/OperationExpr.java
+++ b/src/main/java/com/google/api/generator/engine/ast/OperationExpr.java
@@ -17,4 +17,8 @@ package com.google.api.generator.engine.ast;
 public interface OperationExpr extends Expr {
 
   OperatorKind operatorKind();
+
+  TypeNode type();
+
+  void accept(AstNodeVisitor visitor);
 }
--- a/src/main/java/com/google/api/generator/engine/ast/OperationExpr.java
+++ b/src/main/java/com/google/api/generator/engine/ast/OperationExpr.java
@@ -1,0 +1,18 @@
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.api.generator.engine.ast;
+
+public interface OperationExpr extends Expr {
+
+  OperatorKind operatorKind();
+}
--- a/src/main/java/com/google/api/generator/engine/ast/OperatorKind.java
+++ b/src/main/java/com/google/api/generator/engine/ast/OperatorKind.java
@@ -18,15 +18,15 @@ import java.util.EnumSet;
 
 public enum OperatorKind {
   ARITHMETIC_ADDITION,
-  UNARY_POST_INCREMENT,
-  LOGICAL_NOT,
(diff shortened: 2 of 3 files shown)


src/main/java/com/google/api/generator/engine/ast/OperationExpr.java
```diff
@@ -0,0 +1,20 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.api.generator.engine.ast;
+
+public interface OperationExpr extends Expr {
+
+  OperatorKind operatorKind();
```
miraleung, line 19: Could we add `type()` here as well? Although this is technically fine, adding `type()` here will help us see all the methods that need to be implemented in subclasses.

src/main/java/com/google/api/generator/engine/ast/OperatorKind.java
```diff
@@ -0,0 +1,33 @@
...
+  UNARY_POST_INCREMENT(true);
+
+  private final boolean isPostfixOperator;
+
+  OperatorKind(final boolean isPostfixOperator) {
+    this.isPostfixOperator = isPostfixOperator;
+  }
+
+  public boolean isPostfixOperator() {
+    return this.isPostfixOperator;
```
miraleung, line 31: Instead of passing in a boolean to each enum, what about just doing `return !this.equals(UNARY_LOGICAL_NOT)`?

miraleung: Also, consider whether the two-expression operators should return true or false here.

src/main/java/com/google/api/generator/engine/writer/JavaWriterVisitor.java
```diff
@@ -100,6 +101,13 @@
   private static final String VOLATILE = "volatile";
   private static final String WHILE = "while";
 
+  // Operators
+  private static final String ADDITION = "+";
```
miraleung, line 105: Nit: Consider using `OPERATOR_ADDITION` here, similar for the rest.

src/main/java/com/google/api/generator/engine/ast/OperationExpr.java
summer-ji-eng: sg

src/main/java/com/google/api/generator/engine/ast/OperatorKind.java
summer-ji-eng: Got your point. Had a little think on this, prefix operators are(!) (++) (--) (+) (-), instead of use this.equals(UNARY_LOGICAL_NOT) && this.equals(UNARY_PREFIX_INCREMENT)...
I added a enumset for looking up the operator.
LMK if you have concern. Thanks.

src/main/java/com/google/api/generator/engine/writer/JavaWriterVisitor.java
summer-ji-eng: sg

src/main/java/com/google/api/generator/engine/ast/OperationExpr.java
summer-ji-eng: Since our purpose is to make all methods need to be implemented visible, I also added `accept`. LMK if you have other opinion. :)

miraleung: Yes, thanks for doing that.

src/main/java/com/google/api/generator/engine/ast/OperatorKind.java
```diff
@@ -0,0 +1,35 @@
...
+public enum OperatorKind {
+  ARITHMETIC_ADDITION,
+  RELATIONAL_EQUAL_TO,
+  RELATIONAL_NOT_EQUAL_TO,
+  UNARY_LOGICAL_NOT,
+  UNARY_POST_INCREMENT;
+
+  private static final EnumSet<OperatorKind> PREFIX_OPERATORS_SET;
+
+  static {
```
miraleung, line 28: Have you tried just assigning this outside of the static block?

summer-ji-eng: Assigned outside of static block.

Previously, It showed the compiler error  "constructor can't access static field" if I declare `private EnumSet<OperatorKind> PREFIX_OPERATORS_SET = EnumSet.of(..)`. So I put static block. 

But Turns out if I declare it as static variable will fix that compiler error.

PR #260, opened 2020-09-01
[javanaut] feat: Add arithmetic operation expr with unit test

summer-ji-eng:
Add arithmetic expr and its unit test:
- importvisitor test
- javawritevisitor test

### DIFF
--- a/src/main/java/com/google/api/generator/engine/ast/ArithmeticOperationExpr.java
+++ b/src/main/java/com/google/api/generator/engine/ast/ArithmeticOperationExpr.java
@@ -49,34 +49,29 @@ public abstract class ArithmeticOperationExpr implements OperationExpr {
   @AutoValue.Builder
   public abstract static class Builder {
 
-    // private, required
+    // Private setter
     abstract Builder setLhsExpr(Expr expr);
 
-    // private, required
+    // Private setter
     abstract Builder setRhsExpr(Expr expr);
 
-    // private, required
+    // Private setter
     abstract Builder setOperatorKind(OperatorKind operator);
 
-    // private, required
+    // Private setter
     abstract Builder setType(TypeNode type);
 
     abstract ArithmeticOperationExpr autoBuild();
 
     private ArithmeticOperationExpr build() {
       ArithmeticOperationExpr arithmeticOperationExpr = autoBuild();
-      Expr lhsExpr = arithmeticOperationExpr.lhsExpr();
-      Expr rhsExpr = arithmeticOperationExpr.rhsExpr();
-      TypeNode lhsExprType = lhsExpr.type();
-      TypeNode rhsExprType = rhsExpr.type();
+      TypeNode lhsExprType = arithmeticOperationExpr.lhsExpr().type();
+      TypeNode rhsExprType = arithmeticOperationExpr.rhsExpr().type();
       OperatorKind operator = arithmeticOperationExpr.operatorKind();
       final String errorMsg =
-          "Arithmetic operator "
-              + operator
-              + " can not be applied to "
-              + lhsExprType.toString()
-              + ", "
-              + rhsExprType.toString();
+          String.format(
+              "Arithmetic operator %s can not be applied to %s, %s.",
+              operator, lhsExprType.toString(), rhsExprType.toString());
 
       // None of expression should be void type
       Preconditions.checkState(
@@ -91,7 +86,7 @@ public abstract class ArithmeticOperationExpr implements OperationExpr {
     }
 
     private boolean isValidConcatTypes(TypeNode lhsType, TypeNode rhsType) {
-      // concat require at least one expression type is String
+      // concat requires at least one String-typed expression
       return lhsType.equals(TypeNode.STRING) || rhsType.equals(TypeNode.STRING);
     }
   }
--- a/src/main/java/com/google/api/generator/engine/ast/ArithmeticOperationExpr.java
+++ b/src/main/java/com/google/api/generator/engine/ast/ArithmeticOperationExpr.java
@@ -1,0 +1,98 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.api.generator.engine.ast;
+
+import com.google.auto.value.AutoValue;
+import com.google.common.base.Preconditions;
+
+@AutoValue
+public abstract class ArithmeticOperationExpr implements OperationExpr {
+
+  public abstract Expr lhsExpr();
+
+  public abstract Expr rhsExpr();
+
+  public abstract OperatorKind operatorKind();
+
+  public abstract TypeNode type();
+
+  @Override
+  public void accept(AstNodeVisitor visitor) {
+    visitor.visit(this);
+  }
+
+  public static ArithmeticOperationExpr concatWithExprs(Expr lhsExpr, Expr rhsExpr) {
+    return builder()
+        .setLhsExpr(lhsExpr)
+        .setRhsExpr(rhsExpr)
+        .setOperatorKind(OperatorKind.ARITHMETIC_ADDITION)
+        .setType(TypeNode.STRING)
+        .build();
+  }
+
+  private static Builder builder() {
+    return new AutoValue_ArithmeticOperationExpr.Builder();
+  }
+
+  @AutoValue.Builder
+  public abstract static class Builder {
+
+    // private, required
+    abstract Builder setLhsExpr(Expr expr);
+
+    // private, required
+    abstract Builder setRhsExpr(Expr expr);
+
+    // private, required
+    abstract Builder setOperatorKind(OperatorKind operator);
+
+    // private, required
+    abstract Builder setType(TypeNode type);
+
+    abstract ArithmeticOperationExpr autoBuild();
+
+    private ArithmeticOperationExpr build() {
+      ArithmeticOperationExpr arithmeticOperationExpr = autoBuild();
+      Expr lhsExpr = arithmeticOperationExpr.lhsExpr();
+      Expr rhsExpr = arithmeticOperationExpr.rhsExpr();
+      TypeNode lhsExprType = lhsExpr.type();
+      TypeNode rhsExprType = rhsExpr.type();
+      OperatorKind operator = arithmeticOperationExpr.operatorKind();
+      final String errorMsg =
+          "Arithmetic operator "
+              + operator
+              + " can not be applied to "
+              + lhsExprType.toString()
+              + ", "
+              + rhsExprType.toString();
+
+      // None of expression should be void type
+      Preconditions.checkState(
+          !lhsExprType.equals(TypeNode.VOID) && !rhsExprType.equals(TypeNode.VOID), errorMsg);
+
+      // Concat operator type checks
+      if (operator.equals(OperatorKind.ARITHMETIC_ADDITION)) {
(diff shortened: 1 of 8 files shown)


src/test/java/com/google/api/generator/engine/writer/JavaWriterVisitorTest.java
```diff
@@ -351,6 +352,34 @@ public void writeVariableExpr_nestedReference() {
...
+    MethodInvocationExpr rhsExpr =
+        MethodInvocationExpr.builder()
+            .setExprReferenceExpr(methodInvocationExpr)
+            .setMethodName("getString")
+            .setReturnType(TypeNode.STRING)
+            .build();
+    ArithmeticOperationExpr arithmeticOperationExpr =
+        ArithmeticOperationExpr.concatWithExprs(lhsExpr, rhsExpr);
+    arithmeticOperationExpr.accept(writerVisitor);
+    assertThat(writerVisitor.write()).isEqualTo("\"someWord\" + getMethod().getString()");
```
summer-ji-eng, line 369: A question regarding to StringValueObject: StringValueObject is with escaper "\", but in the sample code we see "someWord + getMethod().getString()" without escaper.

Should we remove this escaper in concat case? @miraleung @xiaozhenliu-gg5

src/main/java/com/google/api/generator/engine/ast/ArithmeticOperationExpr.java
```diff
@@ -0,0 +1,98 @@
...
+    abstract Builder setType(TypeNode type);
+
+    abstract ArithmeticOperationExpr autoBuild();
+
+    private ArithmeticOperationExpr build() {
+      ArithmeticOperationExpr arithmeticOperationExpr = autoBuild();
+      Expr lhsExpr = arithmeticOperationExpr.lhsExpr();
+      Expr rhsExpr = arithmeticOperationExpr.rhsExpr();
+      TypeNode lhsExprType = lhsExpr.type();
+      TypeNode rhsExprType = rhsExpr.type();
```
xiaozhenliu-gg5, line 71: since we only use `lhs/rhsType` below, can we combine lhsExpr into lhsType assignment line?

src/main/java/com/google/api/generator/engine/ast/ArithmeticOperationExpr.java
```diff
@@ -0,0 +1,98 @@
...
+      OperatorKind operator = arithmeticOperationExpr.operatorKind();
+      final String errorMsg =
+          "Arithmetic operator "
+              + operator
+              + " can not be applied to "
+              + lhsExprType.toString()
+              + ", "
+              + rhsExprType.toString();
+
```
xiaozhenliu-gg5, line 80: we probably can make this clearer by String.format().

src/main/java/com/google/api/generator/engine/ast/ArithmeticOperationExpr.java
```diff
@@ -0,0 +1,98 @@
...
+      // Concat operator type checks
+      if (operator.equals(OperatorKind.ARITHMETIC_ADDITION)) {
+        Preconditions.checkState(isValidConcatTypes(lhsExprType, rhsExprType), errorMsg);
+      }
+
+      return arithmeticOperationExpr;
+    }
+
+    private boolean isValidConcatTypes(TypeNode lhsType, TypeNode rhsType) {
+      // concat require at least one expression type is String
```
xiaozhenliu-gg5, line 94: nit: "Concat requires at least one String-typed expression" ?

src/main/java/com/google/api/generator/engine/ast/ArithmeticOperationExpr.java
```diff
@@ -0,0 +1,98 @@
...
+
+  private static Builder builder() {
+    return new AutoValue_ArithmeticOperationExpr.Builder();
+  }
+
+  @AutoValue.Builder
+  public abstract static class Builder {
+
+    // private, required
+    abstract Builder setLhsExpr(Expr expr);
```
xiaozhenliu-gg5, line 53: nit "Private and required"? same below.

src/test/java/com/google/api/generator/engine/writer/JavaWriterVisitorTest.java
xiaozhenliu-gg5: I'm a little bit confused, what do you mean by sample code? For my understanding `String x = "someWord" + getMethod.getString();` as the test case shows is exactly we want?

summer-ji-eng: Oh, you are right. Thanks for pointing out. I am misunderstand the outside double quote.

summer-ji-eng: we are good here 👍

src/main/java/com/google/api/generator/engine/ast/ArithmeticOperationExpr.java
summer-ji-eng: sg

src/main/java/com/google/api/generator/engine/ast/ArithmeticOperationExpr.java
summer-ji-eng: sg

src/main/java/com/google/api/generator/engine/ast/ArithmeticOperationExpr.java
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  private static Builder builder() {
+    return new AutoValue_ArithmeticOperationExpr.Builder();
+  }
+
+  @AutoValue.Builder
+  public abstract static class Builder {
+
+    // Private setter
```
miraleung, line 52: Nit: "." at the end of comment sentences, same elsewhere.

src/main/java/com/google/api/generator/engine/ast/ArithmeticOperationExpr.java
```diff
@@ -0,0 +1,93 @@
...
+      TypeNode rhsExprType = arithmeticOperationExpr.rhsExpr().type();
+      OperatorKind operator = arithmeticOperationExpr.operatorKind();
+      final String errorMsg =
+          String.format(
+              "Arithmetic operator %s can not be applied to %s, %s.",
+              operator, lhsExprType.toString(), rhsExprType.toString());
+
+      // None of expression should be void type
+      Preconditions.checkState(
+          !lhsExprType.equals(TypeNode.VOID) && !rhsExprType.equals(TypeNode.VOID), errorMsg);
```
miraleung, line 78: Instead of checking for `void`, what about checking that they are valid types instead? Note that `int + String` is not valid but `String + int` (or most things on the RHS, I think, please double-check this) is fine.

If we want to (optionally) support numeric types, we can check that here too, and then call `setType` with the appropriately widened type. This is a P2 though, so feel free to leave it for a separate PR.

src/main/java/com/google/api/generator/engine/ast/ArithmeticOperationExpr.java
```diff
@@ -0,0 +1,93 @@
...
+  public abstract OperatorKind operatorKind();
+
+  public abstract TypeNode type();
+
+  @Override
+  public void accept(AstNodeVisitor visitor) {
+    visitor.visit(this);
+  }
+
+  public static ArithmeticOperationExpr concatWithExprs(Expr lhsExpr, Expr rhsExpr) {
```
miraleung, line 36: Nit: `withExprs` since there aren't other kinds of behaviors.

src/main/java/com/google/api/generator/engine/ast/ArithmeticOperationExpr.java
```diff
@@ -0,0 +1,93 @@
...
+        .setType(TypeNode.STRING)
+        .build();
+  }
+
+  private static Builder builder() {
+    return new AutoValue_ArithmeticOperationExpr.Builder();
+  }
+
+  @AutoValue.Builder
+  public abstract static class Builder {
```
miraleung, line 50: Can `public` be removed?

src/test/java/com/google/api/generator/engine/ast/ArithmeticOperationExprTest.java
```diff
@@ -0,0 +1,71 @@
...
+package com.google.api.generator.engine.ast;
+
+import static org.junit.Assert.assertThrows;
+
+import org.junit.Test;
+
+public class ArithmeticOperationExprTest {
+  @Test
+  public void concatOperator_validBasic() {
+    VariableExpr lhsExpr =
```
miraleung, line 24: Could you please confirm if `x + getSomeString()` is valid?

src/test/java/com/google/api/generator/engine/writer/ImportWriterVisitorTest.java
```diff
@@ -778,6 +780,21 @@ public void writeReferenceConstructorExprImports_withArgs() {
...
 
+  @Test
+  public void writeArithmeticOperationExprImports() {
+    MethodInvocationExpr lhsExpr =
+        MethodInvocationExpr.builder()
+            .setStaticReferenceType(TypeNode.withReference(ConcreteReference.withClazz(Expr.class)))
+            .setMethodName("getSomething")
+            .setReturnType(TypeNode.STRING)
+            .build();
+    ValueExpr rhsExpr = ValueExpr.withValue(NullObjectValue.create());
```
miraleung, line 791: Let's make this an expression that requires an import (and `null` is already tested elsewhere).

src/main/java/com/google/api/generator/engine/ast/ArithmeticOperationExpr.java
summer-ji-eng: sg

src/main/java/com/google/api/generator/engine/ast/ArithmeticOperationExpr.java
summer-ji-eng: For concat case, one of expr is string-type is good regardless the position and none of expr is void type. Saying like `5 + "abc"` and `"abc" + 5` are both valid.

Regarding to `void`, for all arithmetic operation (+,-,*,/,%) the expression type on lhs and rhs have to be non-void type. So I think it is worth to take it out for all operators. 

In my original design thinking, we could separate addition operator usage into concat and plus. Because concat always return String, plus return type is based on expression type and related to data-type precedence, like
```
int a = 5; double b = 5.5;
double c = a + b;
```
when we need to support numeric plus usage, we might have another convenience method like `plusWithExprs`, different type-checking and `setType` for plus usage.

Therefore, void type-checking for all operators, and for each operator/usage has its own type-checking. LMK if you have other opinion. Thanks :)

src/test/java/com/google/api/generator/engine/ast/ArithmeticOperationExprTest.java
summer-ji-eng: added.

src/main/java/com/google/api/generator/engine/ast/ArithmeticOperationExpr.java
summer-ji-eng: sg

src/test/java/com/google/api/generator/engine/ast/ArithmeticOperationExprTest.java
summer-ji-eng: I am confused what need to confirm for `x + getSomeString()` here. It is valid in Java, 
```
  private static String getString() {
    return "ab";
  }
int x = 5;
String c = x + getString();
```
I have added a writer test [here](https://github.com/googleapis/gapic-generator-java/pull/260/files#diff-9dd064308f18929ebfe29d140d59466eR369)

src/test/java/com/google/api/generator/engine/writer/ImportWriterVisitorTest.java
summer-ji-eng: In java, `String c = "ac" + null;`, it is worthy to add this test case here.
sg, I will add a test case that requires an import.

src/test/java/com/google/api/generator/engine/ast/ArithmeticOperationExprTest.java
miraleung: Thanks for checking.

miraleung (approved):

PR #272, opened 2020-09-03
[ggj][engx] fix: use singleton in HtmlEscaper

miraleung:


xiaozhenliu-gg5 (approved):

xiaozhenliu-gg5:
Curious that why we did not catch the build failure in local, but it seems just a name change `charEscaper`. Thanks for fixing this.

PR #273, opened 2020-09-03
[ggj][engx] fix: use singleton in MetacharEscaper

miraleung:


xiaozhenliu-gg5 (approved):
LGTM! the build failure is from the name change `charEscaper` instead of `escaper` (original name). Thanks for fixing this.

PR #279, opened 2020-09-08
[ggj][infra][4/5]feat: move the expected string in unit test to golden files 

xiaozhenliu-gg5:
Find more details in design doc: go/java-micro-file-diff-infra

Move the expected codegen string into goldens files, no logic for updating goldens is included.
update_golden bazel rules for each unit test will be coming once #314 merged in.


src/test/java/com/google/api/generator/engine/JavaCodeGeneratorTest.java
```diff
@@ -96,7 +101,7 @@
   private static final Variable bookKindVar = createVarFromVaporRef(bookKindRef, "bookKind");
 
   @Test
-  public void validJavaClass() {
+  public void validJavaClass() throws IOException {
     // Create outer class variableDecls.
```
xiaozhenliu-gg5, line 105: This IOException will be moved once DiffUtils and AssertUtils wrapper are in place.

src/test/java/com/google/api/generator/gapic/composer/GrpcServiceCallableFactoryClassComposerTest.java
```diff
@@ -35,6 +38,10 @@
 public class GrpcServiceCallableFactoryClassComposerTest {
   private ServiceDescriptor echoService;
   private FileDescriptor echoFileDescriptor;
+  private static final String GOLDENFILES_DIRECTORY =
+      "src/test/java/com/google/api/generator/gapic/composer/goldens/";
+  private static final String GOLDENFILES_NAME =
```
miraleung, line 43: Would prefer for each filename to be local to each test method, since this is a per-instance type of data.

src/test/java/com/google/api/generator/gapic/composer/GrpcServiceCallableFactoryClassComposerTest.java
```diff
@@ -35,6 +38,10 @@
...
+      "GrpcServiceCallableFactoryClassComposerTest.golden";
```
miraleung, line 44: For the `*ClassComposerTest`s, can we give these the same name as the generated class?

miraleung (changes requested):

src/test/java/com/google/api/generator/gapic/composer/GrpcServiceCallableFactoryClassComposerTest.java
```diff
@@ -35,6 +38,10 @@
 public class GrpcServiceCallableFactoryClassComposerTest {
   private ServiceDescriptor echoService;
   private FileDescriptor echoFileDescriptor;
+  private static final String GOLDENFILES_DIRECTORY =
```
miraleung, line 41: Consider having a `ComposerConstants` class in this package, and putting this constant there instead.

miraleung (changes requested):

src/test/java/com/google/api/generator/gapic/composer/GrpcServiceCallableFactoryClassComposerTest.java
xiaozhenliu-gg5: Sure, renamed the goldens as the generated class name.

src/test/java/com/google/api/generator/gapic/composer/GrpcServiceCallableFactoryClassComposerTest.java
xiaozhenliu-gg5: moved the golden file name as a local string instead of global constant.

src/test/java/com/google/api/generator/gapic/composer/GrpcServiceCallableFactoryClassComposerTest.java
xiaozhenliu-gg5: Done, also moved `TESTFILES_DIRECTORY` there for share use.

src/test/java/com/google/api/generator/gapic/composer/BUILD.bazel
```diff
@@ -37,8 +37,9 @@ java_proto_library(
 
 [java_test(
     name = test_name,
-    srcs = ["{0}.java".format(test_name)],
+    srcs = glob(["*.java"]),
     data = [
```
xiaozhenliu-gg5, line 41: This change is for including `ComposerConstants.java` as src, not only the tests files.

src/test/java/com/google/api/generator/gapic/composer/BUILD.bazel
```diff
@@ -18,7 +18,7 @@ TESTS = [
 
 filegroup(
     name = "composer_files",
-    srcs = ["{0}.java".format(f) for f in TESTS],
+    srcs = glob(["*.java"]),
 )
```
xiaozhenliu-gg5, line 22: This is for including all files for formatter

src/test/java/com/google/api/generator/gapic/composer/BUILD.bazel
miraleung: How about:
```
srcs = [
    "{0}.java".format(test_name)", 
    "ComposerConstants.java",
]
```

miraleung (changes requested):
LGTM with `BUILD` file comment.

\o/ looks great!

src/test/java/com/google/api/generator/gapic/composer/BUILD.bazel
xiaozhenliu-gg5: Thanks! That should also work, while if we add more helpers, those files have to be explicitly included. By using `glob(["*.java"])` we can avoid modifying this even if more files are added. WDYT?

xiaozhenliu-gg5: Done! Talked offline and for suggested option, we can avoid including other test files.

miraleung (approved):

PR #289, opened 2020-09-10
[javanaut][1/2]feat: Add RelationalOperationExpr to enable equality operators

summer-ji-eng:
Enable equality operators (==) and (!=) and add unit test, JavaWriteVisitorTest, ImportWriterVisitorTest

Upcoming a separated PR for convenience wrapper "lessThanWithExprs" enable (<) operator


src/main/java/com/google/api/generator/engine/ast/RelationalOperationExpr.java
```diff
@@ -0,0 +1,154 @@
...
+    // operator (!=).
+    private boolean isValidEqualityType(TypeNode lhsType, TypeNode rhsType) {
+      // The Operators can not be applied on void type.
+      if (lhsType.equals(TypeNode.VOID) || rhsType.equals(TypeNode.VOID)) {
+        return false;
+      }
+      // If the expressions' type are array, the types should be array and matched, or either is
+      // null type;
+      if (lhsType.isArray() || rhsType.isArray()) {
+        if (lhsType.equals(TypeNode.NULL) || rhsType.equals(TypeNode.NULL)) {
```
miraleung, line 111: If I understand correctly, this section reads, "either one is null, or both are the same type," and the second is taken care of by the change above. Can we simplify the return statement to `return lhsType.equals(TypeNode.NULL) || rhsType.equals(TypeNode.NULL)`?

src/main/java/com/google/api/generator/engine/ast/RelationalOperationExpr.java
```diff
@@ -0,0 +1,154 @@
...
+          return false;
+        }
+        return lhsType.equals(rhsType);
+      }
+      // If lhs expression type is boolean or its boxed type, rhs should be boolean or boxed or null
+      // or new Object
+      if (lhsType.equals(TypeNode.BOOLEAN)) {
+        return rhsType.equals(lhsType)
+            || rhsType.equals(TypeNode.NULL)
+            || rhsType.equals(TypeNode.OBJECT);
```
miraleung, line 125: We can remove this, it will be taken care of by `equals`.

src/main/java/com/google/api/generator/engine/ast/RelationalOperationExpr.java
```diff
@@ -0,0 +1,154 @@
...
+            || (!lhsType.isArray() && rhsType.isArray())) {
+          return false;
+        }
+        return lhsType.equals(rhsType);
+      }
+      // If lhs expression type is boolean or its boxed type, rhs should be boolean or boxed or null
+      // or new Object
+      if (lhsType.equals(TypeNode.BOOLEAN)) {
+        return rhsType.equals(lhsType)
+            || rhsType.equals(TypeNode.NULL)
```
miraleung, line 124: The following was invalid for me: `boolean a = false; assertFalse(a == null);`.

src/main/java/com/google/api/generator/engine/ast/RelationalOperationExpr.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    // isValidEqualityType checks expressions' type for equality operator (==) and non-equality
+    // operator (!=).
+    private boolean isValidEqualityType(TypeNode lhsType, TypeNode rhsType) {
+      // The Operators can not be applied on void type.
+      if (lhsType.equals(TypeNode.VOID) || rhsType.equals(TypeNode.VOID)) {
+        return false;
+      }
+      // If the expressions' type are array, the types should be array and matched, or either is
```
miraleung, line 108: Let's add a `if (lhsType.equals(rhsType)) { return true; }` here, which will simplify a lot of the logic below.

src/main/java/com/google/api/generator/engine/ast/RelationalOperationExpr.java
```diff
@@ -0,0 +1,154 @@
...
+        if ((lhsType.isArray() && !rhsType.isArray())
+            || (!lhsType.isArray() && rhsType.isArray())) {
+          return false;
+        }
+        return lhsType.equals(rhsType);
+      }
+      // If lhs expression type is boolean or its boxed type, rhs should be boolean or boxed or null
+      // or new Object
+      if (lhsType.equals(TypeNode.BOOLEAN)) {
+        return rhsType.equals(lhsType)
```
miraleung, line 123: We can remove this with the first equality check above.

src/main/java/com/google/api/generator/engine/ast/RelationalOperationExpr.java
```diff
@@ -0,0 +1,154 @@
...
+            || rhsType.equals(TypeNode.NULL)
+            || rhsType.equals(TypeNode.OBJECT);
+      }
+      // If lhs expression type is numeric type (char, byte, short, int, long, double), the rhs
+      // expression type should be any numeric type or any numeric boxed type
+      // if lhs is boxed numeric type, rhs could be null or Object;
+      if (TypeNode.isNumericType(lhsType)) {
```
miraleung, line 130: What about adding `isNumericType(lhsType) && isNumericType(rhsType)` to the first early-return equality check?

src/main/java/com/google/api/generator/engine/ast/RelationalOperationExpr.java
```diff
@@ -0,0 +1,154 @@
...
+        if (TypeNode.isNumericType(rhsType)) {
+          return true;
+        }
+        return TypeNode.isBoxedType(lhsType)
```
miraleung, line 134: I think we won't need this with the Object check below.

src/main/java/com/google/api/generator/engine/ast/RelationalOperationExpr.java
```diff
@@ -0,0 +1,154 @@
...
+            && (rhsType.equals(TypeNode.NULL) || rhsType.equals(TypeNode.OBJECT));
+      }
+      // If lhs expression type is new Object or null, the rhs type should be a reference type or
+      // null or boxed type;
+      if (TypeNode.OBJECT.equals(lhsType) || TypeNode.NULL.equals(lhsType)) {
```
miraleung, line 139: We can simplify this by checking that, if either LHS or RHS are Objects, the other must be an Object too.

src/test/java/com/google/api/generator/engine/ast/RelationalOperationExprTest.java
```diff
@@ -0,0 +1,431 @@
...
+// limitations under the License.
+
+package com.google.api.generator.engine.ast;
+
+import static org.junit.Assert.assertThrows;
+
+import com.google.api.generator.engine.ast.TypeNode.TypeKind;
+import org.junit.Test;
+
+public class RelationalOperationExprTest {
```
miraleung, line 22: This is nice and thorough! Some more scenarios to consider:
- Numeric object and some Object
- For each test, flip the RHS and LHS values (in the same test) so we can ensure we're not just looking at the LHS types.
- Any type and null.
- null and null.

miraleung (changes requested):

src/main/java/com/google/api/generator/engine/ast/RelationalOperationExpr.java
```diff
@@ -0,0 +1,154 @@
...
+    return TypeNode.BOOLEAN;
+  }
+
+  @Override
+  public void accept(AstNodeVisitor visitor) {
+    visitor.visit(this);
+  }
+
+  // Convenience wrapper.
+  public static RelationalOperationExpr equalToWithExprs(Expr lhsExpr, Expr rhsExpr) {
```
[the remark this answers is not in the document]
summer-ji-eng: emm, I have different opinion. In general both `equals()` and “==” operator in Java are used to compare objects to check equality. And here, we try to express `lhs == rhs` using equal to operator. Regarding to `equals()`, we use `MethodInvocation` to express. Therefore, I prefer use operator name here. Or maybe name like `equalToOpertorWithExprs`? LMK what you think. 😀

src/main/java/com/google/api/generator/engine/ast/TypeNode.java
```diff
@@ -148,6 +148,11 @@ public static boolean isNumericType(TypeNode type) {
         || type.equals(TypeNode.CHAR);
   }
 
+  public static boolean isBoxedType(TypeNode type) {
+    return type.typeKind().equals(TypeKind.OBJECT)
```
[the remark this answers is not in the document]
summer-ji-eng: I used to try `BOXED_TYPE_MAP.values().contains(type)` and `BOXED_TYPE_MAP.containsValue(type)`, but it break [test case](https://github.com/googleapis/gapic-generator-java/pull/289/files#diff-eb49f476bfd2402d83757f2bffd6bb6cR417) invalid for Object equal to boolean, valid for Object equal to New Boolean.
`contains()` compares TypeNode.BOOLEAN with TypeNode.BOOLEAN_OBJECT and return true.

This way is able to distinguish TypeNode.BOOLEAN and TypeNode.BOOLEAN_OBJECT.

src/main/java/com/google/api/generator/engine/ast/RelationalOperationExpr.java
summer-ji-eng: sg. Thanks for pointing out. It makes the logical more simple and clear. 🙏

src/main/java/com/google/api/generator/engine/ast/RelationalOperationExpr.java
summer-ji-eng: sg

src/main/java/com/google/api/generator/engine/ast/RelationalOperationExpr.java
summer-ji-eng: My bad. Null and new Object is only valid for boxed type. I should add condition for boxed boolean type. I will add a unit test case for this as well.
```
  @Test
  public void equalToOperationExpr_invalidBooleanWithNullType() {
    // LHS: boolean type, RHS: null type
    VariableExpr lhsExpr = createVariableExpr(TypeNode.BOOLEAN, "x");
    ValueExpr rhsExpr = ValueExpr.withValue(NullObjectValue.create());
    assertThrows(
        IllegalStateException.class,
        () -> RelationalOperationExpr.equalToWithExprs(lhsExpr, rhsExpr));
  }

  @Test
  public void equalToOperationExpr_invalidBooleanWithObjectType() {
    // LHS: boolean type, RHS: object type
    VariableExpr lhsExpr = createVariableExpr(TypeNode.BOOLEAN, "x");
    VariableExpr rhsExpr = createVariableExpr(TypeNode.OBJECT, "y");
    assertThrows(
        IllegalStateException.class,
        () -> RelationalOperationExpr.equalToWithExprs(lhsExpr, rhsExpr));
  }
```

src/main/java/com/google/api/generator/engine/ast/RelationalOperationExpr.java
summer-ji-eng: Thanks for suggestion. I move `if (lhsType.equals(rhsType)) return true;` to the top, it simplify the logic. Please see the update.

src/main/java/com/google/api/generator/engine/ast/RelationalOperationExpr.java
summer-ji-eng: See updates.

src/main/java/com/google/api/generator/engine/ast/RelationalOperationExpr.java
summer-ji-eng: updated.

src/main/java/com/google/api/generator/engine/ast/RelationalOperationExpr.java
summer-ji-eng: updated.

src/test/java/com/google/api/generator/engine/ast/RelationalOperationExprTest.java
summer-ji-eng: Thanks your suggestions. I added more test cases.
- Numeric object and some Object (See LHS Numeric Boxed type, compare with Object, String ) [code](https://github.com/googleapis/gapic-generator-java/pull/289/files#diff-eb49f476bfd2402d83757f2bffd6bb6cR132-R219) 
- For each test, flip the RHS and LHS values (in the same test) so we can ensure we're not just looking at the LHS types. (Added missed flip test case and comment in each test)
- Any type and null. (Added in each type)
- null and null (add the [test](https://github.com/googleapis/gapic-generator-java/pull/289/files#diff-eb49f476bfd2402d83757f2bffd6bb6cR632-R639)).

src/main/java/com/google/api/generator/engine/ast/RelationalOperationExpr.java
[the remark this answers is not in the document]
miraleung: SGTM.

src/main/java/com/google/api/generator/engine/ast/TypeNode.java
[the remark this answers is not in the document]
summer-ji-eng: Updated by 
```
  public static boolean isBoxedType(TypeNode type) {
    return isReferenceType(type) && BOXED_TYPE_MAP.containsValue(type);
  }
```

src/main/java/com/google/api/generator/engine/ast/RelationalOperationExpr.java
```diff
@@ -0,0 +1,135 @@
...
+      }
+
+      return relationalOperationExpr;
+    }
+
+    // isValidEqualityType checks expressions' type for equality operator (==) and non-equality
+    // operator (!=).
+    private boolean isValidEqualityType(TypeNode lhsType, TypeNode rhsType) {
+      // If the expression's types are matched, return true
+      if (lhsType.equals(rhsType)) return true;
```
miraleung, line 105: The linter should have caught this; we generally use newlines and braces for all blocks, even if there's only one line. Context [here](https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used).

src/main/java/com/google/api/generator/engine/ast/RelationalOperationExpr.java
```diff
@@ -0,0 +1,135 @@
...
+
+      // If the expressions' type are array, the types should be array and matched, or either is
+      // null type;
+      if (lhsType.isArray() || rhsType.isArray()) {
+        return lhsType.equals(TypeNode.NULL) || rhsType.equals(TypeNode.NULL);
+      }
+
+      // If lhs expression type is numeric type (char, byte, short, int, long, double), the rhs
+      // expression type should be any numeric type or any numeric boxed type
+      if (TypeNode.isNumericType(lhsType) && TypeNode.isNumericType(rhsType)) return true;
```
miraleung, line 115: Ditto.

src/main/java/com/google/api/generator/engine/ast/RelationalOperationExpr.java
```diff
@@ -0,0 +1,135 @@
...
+
+      // If lhs expression type is new Object or null, the rhs type should be a reference type or
+      // null or boxed type;
+      if (lhsType.equals(TypeNode.OBJECT) || lhsType.equals(TypeNode.NULL)) {
+        return TypeNode.isReferenceType(rhsType)
+            || rhsType.equals(TypeNode.OBJECT)
+            || rhsType.equals(TypeNode.NULL)
+            || TypeNode.isBoxedType(rhsType);
```
miraleung, line 123: This line should be taken care of by the `rhs.equals(TypeNode.OBJECT)` check?

src/main/java/com/google/api/generator/engine/ast/RelationalOperationExpr.java
```diff
@@ -0,0 +1,135 @@
...
+      TypeNode rhsExprType = relationalOperationExpr.rhsExpr().type();
+      OperatorKind operator = relationalOperationExpr.operatorKind();
+      final String errorMsg =
+          String.format(
+              "Relational operator %s can not be applied to %s, %s.",
+              operator, lhsExprType.toString(), rhsExprType.toString());
+
+      if (operator.equals(OperatorKind.RELATIONAL_EQUAL_TO)
+          || operator.equals(OperatorKind.RELATIONAL_NOT_EQUAL_TO)) {
+        Preconditions.checkState(isValidEqualityType(lhsExprType, rhsExprType), errorMsg);
```
miraleung, line 95: Consider simplifying `isValidEqualityType` further to do only a one-way check (i.e. assume `lhs` and `rhs` are always in those  positions). Then, we can just call `isValidEqualityType(rhs, lhs)` to check the opposite condition.

xiaozhenliu-gg5 (approved):

src/test/java/com/google/api/generator/engine/ast/RelationalOperationExprTest.java
miraleung: Thanks for the detailed test cases.

miraleung (changes requested):

src/main/java/com/google/api/generator/engine/ast/RelationalOperationExpr.java
```diff
@@ -0,0 +1,135 @@
...
+    abstract RelationalOperationExpr autoBuild();
+
+    private RelationalOperationExpr build() {
+      RelationalOperationExpr relationalOperationExpr = autoBuild();
+      TypeNode lhsExprType = relationalOperationExpr.lhsExpr().type();
+      TypeNode rhsExprType = relationalOperationExpr.rhsExpr().type();
+      OperatorKind operator = relationalOperationExpr.operatorKind();
+      final String errorMsg =
+          String.format(
+              "Relational operator %s can not be applied to %s, %s.",
```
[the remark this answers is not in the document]
summer-ji-eng: Screen Shot 2020-09-15 at 1 18 54 PM

I am following the Java compiler error message style for operator type-checking. Unlike logical operator, the type-checking is comparatively simple than relational, we could ask for users given boolean type expression. In theory, we could rephrase the error message base on input type condition, like if lhs is numeric type, rhs can only be applied on any numeric type or any numeric boxed type. I wonder if we should follow the Java compiler error message to be consistent with general Java rule.

However, if you prefer to specify the type-checking with detailed error message, we could open a separated ticket and refactor all other type-checking error messages.

src/main/java/com/google/api/generator/engine/ast/RelationalOperationExpr.java
summer-ji-eng: Good to know. Thanks for sharing. Updated.

src/main/java/com/google/api/generator/engine/ast/RelationalOperationExpr.java
summer-ji-eng: Yes. Misunderstand last time you comment here.
Removed `TypeNode.isBoxedType(rhsType);`

src/main/java/com/google/api/generator/engine/ast/RelationalOperationExpr.java
summer-ji-eng: Can you clarify a bit more on the benefit of doing further simplification? Currently the equality is symetric, meaning if A = B is valid, then B = A will pass the compiler as well.

src/main/java/com/google/api/generator/engine/ast/RelationalOperationExpr.java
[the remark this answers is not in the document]
miraleung: SG, thanks for the background.

src/main/java/com/google/api/generator/engine/ast/RelationalOperationExpr.java
miraleung: It's great that the current logic is implemented in a symmetric way. Benefits of simplification include easier readability and maintainability. For example, someone reading the method doesn't need to check that the symmetry is implemented, and can be confident that symmetry will come from the second invocation. Not having to implement symmetry will also make it easier for someone to modify one-way logic or add checks.

Marking this as optional for a separate PR.

miraleung (approved):

src/main/java/com/google/api/generator/engine/ast/RelationalOperationExpr.java
summer-ji-eng: Good discussion here. I created a ticket for refactoring type-checking.

PR #292, opened 2020-09-10
[ggj][composer][1/2]feat: add pagedCallable getter in serviceClient for list method

xiaozhenliu-gg5:
Find more details in go/gapic-java-list-methods

For list method, except for the usual UnaryCallable getter, we need an extra “[PagedCallable()](https://github.com/googleapis/gapic-generator/blob/master/src/test/java/com/google/api/codegen/gapic/testdata/java/java_library.baseline#L2878)” getter which returns UnaryCallable  with PagedResponse.

TODO[2/2] update the return type and callable in service list method. #294


src/main/java/com/google/api/generator/gapic/composer/ServiceClientClassComposer.java
```diff
@@ -664,14 +680,24 @@ private static MethodDefinition createCallableMethod(
                             method.inputType().reference(),
                             method.lro().responseType().reference(),
                             method.lro().metadataType().reference())
-                        : Arrays.asList(
-                            method.inputType().reference(), method.outputType().reference())));
+                        : isPaged
```
miraleung, line 683: Let's use a private helper that takes the enum above and returns the appropriate name, instead of a two-stage ternary expression.

src/main/java/com/google/api/generator/gapic/composer/ServiceClientClassComposer.java
```diff
@@ -664,14 +680,24 @@ private static MethodDefinition createCallableMethod(
...
+                                    .reference())
+                            : Arrays.asList(
+                                method.inputType().reference(), method.outputType().reference())));
 
     String rawMethodName = JavaStyle.toLowerCamelCase(method.name());
     String methodName =
         isLroCallable
-            ? String.format("%sOperationCallabke", rawMethodName)
-            : String.format("%sCallable", rawMethodName);
+            ? String.format(OPERATION_CALLABLE_NAME_PATTERN, rawMethodName)
```
miraleung, line 697: Ditto.

src/main/java/com/google/api/generator/gapic/composer/ServiceClientClassComposer.java
```diff
@@ -622,16 +629,25 @@ private static MethodDefinition createLroAsyncMethod(
...
+    return createCallableMethod(serviceName, method, types, false, true);
   }
 
   private static MethodDefinition createCallableMethod(
-      String serviceName, Method method, Map<String, TypeNode> types, boolean isLroCallable) {
+      String serviceName,
+      Method method,
+      Map<String, TypeNode> types,
+      boolean isLroCallable,
+      boolean isPaged) {
```
miraleung, line 650: Instead of adding a boolean, can we have this take an enum?

`enum CallableMethodKind { REGULAR, LRO, PAGED }`

miraleung (changes requested):

src/main/java/com/google/api/generator/gapic/composer/ServiceClientClassComposer.java
xiaozhenliu-gg5: SG!

src/main/java/com/google/api/generator/gapic/composer/ServiceClientClassComposer.java
```diff
@@ -73,6 +74,16 @@
 
 public class ServiceClientClassComposer implements ClassComposer {
   private static final ServiceClientClassComposer INSTANCE = new ServiceClientClassComposer();
+  private static final String PAGED_RESPONSE_TYPE_NAME_PATTERN = "%sPagedResponse";
+  private static final String CALLABLE_NAME_PATTERN = "%sCallable";
+  private static final String PAGED_CALLABLE_NAME_PATTERN = "%sPagedCallable";
+  private static final String OPERATION_CALLABLE_NAME_PATTERN = "%sOperationCallable";
+
+  enum CallableMethodKind {
```
miraleung, line 82: `private`?

src/main/java/com/google/api/generator/gapic/composer/ServiceClientClassComposer.java
```diff
@@ -905,4 +932,34 @@ private static MethodDefinition createCallableMethod(
...
     return Variable.builder().setName(name).setType(type).build();
   }
+
+  private static String getClientClassName(String serviceName) {
+    return String.format("%sClient", serviceName);
+  }
+
+  private static List<Reference> getGenericsForCallable(
+      CallableMethodKind kind, Method method, Map<String, TypeNode> types) {
+    if (kind.equals(CallableMethodKind.LRO)) {
```
miraleung, line 942: Just FYI that `switch` would also work here.

miraleung (approved):

src/main/java/com/google/api/generator/gapic/composer/ServiceClientClassComposer.java
xiaozhenliu-gg5: sure, was considering about `switch` but since there are only three conditions, probably just `if` is more clearer. But open to refactor this later if more condition is added.

src/main/java/com/google/api/generator/gapic/composer/ServiceClientClassComposer.java
xiaozhenliu-gg5: SG!

PR #294, opened 2020-09-10
[ggj][composer][2/2] feat: update gapic list methods in service client

xiaozhenliu-gg5:
Find more details in go/gapic-java-list-methods

1. The return types of a List Java method should be the “Paged” type of which is specified in the protobufs. (e.g. `ListBooksPagedResponse` instead of `ListBooksResponse`).
2. For the Java method which takes `request`, it should call “PagedCallable()” to call the request.


src/test/java/com/google/api/generator/gapic/composer/ServiceClientClassComposerTest.java
```diff
@@ -223,17 +223,16 @@ public void generateServiceClasses() {
...
           + "  }\n"
           + "\n"
           + "  public final UnaryCallable<PagedExpandRequest, PagedExpandPagedResponse>\n"
           + "      pagedExpandPagedCallable() {\n"
           + "    return stub.pagedExpandPagedCallable();\n"
           + "  }\n"
           + "\n"
-          + "  public final UnaryCallable<PagedExpandRequest, PagedExpandResponse>"
-          + " pagedExpandCallable() {\n"
+          + "  public final UnaryCallable<PagedExpandRequest, PagedExpandResponse> pagedExpandCallable() {\n"
```
summer-ji-eng, line 235: This looks wired to me. 
Should it looks like
```
+ "  public final UnaryCallable<PagedExpandRequest, PagedExpandResponse>"
+ " pagedExpandCallable() {\n"```

xiaozhenliu-gg5: I think that's because of the formatting tool, without this change, the unit test fails. :(

miraleung (approved):

PR #296, opened 2020-09-10
[ggj][bazel] fix: rename protoc plugin binary

miraleung:
... for consistency with the other languages' microgenerators.


summer-ji-eng (commented):
Code LGTM, maybe I missed something here. I am curious why do we have this name change here.

miraleung:
The name change will help bring us in alignment with the other microgenerators, which follow the `foo_gapic` convention (e.g. [Python](https://github.com/googleapis/gapic-generator-python/blob/master/rules_python_gapic/py_gapic.bzl#L39), [Go](https://github.com/googleapis/gapic-generator-go/blob/master/rules_go_gapic/go_gapic.bzl#L121)).

PR #311, opened 2020-09-16
[ggj][codegen][ast] fix: solidify codegen method order with AST Comparables

miraleung:


xiaozhenliu-gg5 (approved):

src/main/java/com/google/api/generator/engine/ast/TypeNode.java
```diff
@@ -88,6 +88,31 @@
...
+      }
+      // b is a reference type or null, so a < b.
+      return -1;
+    }
+
+    if (this.equals(TypeNode.NULL)) {
+      // Can't self-compare, so we don't need to check whether the other one is TypeNode.NULL.
+      return other.isPrimitiveType() ? 1 : -1;
+    }
+
```
summer-ji-eng, line 106: Do we need to consider the VOID type comparsion?

summer-ji-eng (approved):
LGTM with a question of comparison  type.

src/main/java/com/google/api/generator/engine/ast/TypeNode.java
miraleung: No, it doesn't need to be handled separately.

PR #314, opened 2020-09-18
[ggj][infra][3/5] feat: implement update golden bazel rules for dummy test

xiaozhenliu-gg5:
Find more details in design doc: go/java-micro-file-diff-infra

1. Added `rules_bazel/java_diff_test.bzl` where `update_golden` bazel rule is defined, including two parts: running JUnit test and overwritting goldens.
2. Added `SingleJUnitTestRunner` which helps us to manually trigger the JUnit test, it takes `test_class_name` as arg and emits test result if any failure is detected.
3. Added `Utils.java` in `test/framework` which helps to save the generated code from JUnit test to tmp file in folder `$TEST_OUTPUT_HOME`. System environment is set to interact with update golden bazel rules, the folder will get deleted once the bazel rules finish.
4. `update_golden` rule is called from `BUILD.bazel` in dummy test, and `junit_runner_binary` is defined for running the JUnit test when updating goldens.

How we use it (check out this branch and modify the dummy/goldens/*.golden)
-----------------
```
> bazel test //...
Tests have failures: [classWithHeader(com.google.api.generator.gapic.dummy.FileDiffInfraDummyTest): Differences found: 
--- golden
+++ codegen
@@ -17,3 +17,3 @@
 package com.google.showcase.v1beta1.stub;
 
-public class  {}
+public class EchoStubSettings {}]
//src/test/java/com/google/api/generator/gapic/dummy:FileDiffInfraDummyTest FAILED in 5.6s
  /private/var/tmp/_bazel_xiaozhenliu/27d546162d4fdfcb8b9c40b4c5bf4971/execroot/com_google_api_codegen/bazel-out/darwin-fastbuild/testlogs/src/test/java/com/google/api/generator/gapic/dummy/FileDiffInfraDummyTest/test.log


> bazel run //src/test/java/com/google/api/generator/gapic/dummy:FileDiffInfraDummyTest_update
> bazel test //...
.....
Executed 53 out of 66 tests: 66 tests pass.
```

Next step would be:
1. Move the existing expected class string to file-diff infra, including 
(a)gapic/composer/*Test.
(b)engine/astIntegrationTest
2. Update developer doc to include this instruction for updating goldens.


dependencies.properties
```diff
@@ -32,5 +32,6 @@ maven.org_threeten_threetenbp=org.threeten:threetenbp:1.3.3
 
 # Testing.
 maven.junit_junit=junit:junit:4.13
+maven.org_hamcrest_hamcrest_core=org.hamcrest:hamcrest-core:1.3
```
xiaozhenliu-gg5, line 35: This dependency is for running JUnit test manually, before JUnit 4.11 it's wrapped along with junit package. Now it has to be explicitly added.

src/test/java/com/google/api/generator/test/framework/Utils.java
```diff
@@ -0,0 +1,56 @@
...
+package com.google.api.generator.test.framework;
+
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+public class Utils {
+
```
xiaozhenliu-gg5, line 24: `Utils` and `SingleJUnitTestRunner` are placed in `test/framework` folder because they will be shared by folders `engine/` and `gapic/`

dependencies.properties
miraleung: Could you please add that note as a comment?

rules_bazel/java/java_diff_test.bzl
```diff
@@ -0,0 +1,85 @@
...
+    inputs = ctx.files.srcs
+    output = ctx.outputs.output
+    test_runner = ctx.executable.test_runner
+
+    command = """
+    mkdir local_tmp  
+    TEST_OUTPUT_HOME="$(pwd)/local_tmp" \
+    {test_runner_path} $@ &&
+    cd local_tmp 
+    zip -r ../{output} .
```
miraleung, line 12: `zip -r -j {output} local_tmp/*` instead of `cd`?

src/test/java/com/google/api/generator/test/framework/Utils.java
```diff
@@ -0,0 +1,56 @@
...
+  public static void saveCodegenToFile(Class clazz, String fileName, String codegen) {
```
miraleung, line 25: Add a javadoc and sample usage?

miraleung (changes requested):
Took a first pass.

dependencies.properties
xiaozhenliu-gg5: Sure, done.

rules_bazel/java/java_diff_test.bzl
xiaozhenliu-gg5: Unfortunately in `local_tmp` we have `com/google/api/generator/.../File1.golden` and `-j` option will flatten the path and only zip `File1.golden` file without the folders.

src/test/java/com/google/api/generator/test/framework/Utils.java
xiaozhenliu-gg5: sure, done!

rules_bazel/java/java_diff_test.bzl
```diff
@@ -0,0 +1,85 @@
+def _junit_output_impl(ctx):
+    test_class_name = ctx.attr.test_class_name
+    inputs = ctx.files.srcs
+    output = ctx.outputs.output
+    test_runner = ctx.executable.test_runner
+
+    command = """
```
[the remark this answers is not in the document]
xiaozhenliu-gg5: Got it, done.

src/test/java/com/google/api/generator/test/framework/Utils.java
```diff
@@ -0,0 +1,67 @@
...
+    // This system environment variable `TEST_OUTPUT_HOME` is used to specify a folder
+    // which contains generated output from JUnit test.
+    // It will be set when running `bazel run testTarget.update` command.
+    String testOutputHome = System.getenv("TEST_OUTPUT_HOME");
+    String relativeGoldenDir = getTestoutGoldenDir(clazz);
+    Path testOutputDir = Paths.get(testOutputHome, relativeGoldenDir);
+    testOutputDir.toFile().mkdirs();
+    try {
+      File testOutputFile = Paths.get(testOutputHome, relativeGoldenDir, fileName).toFile();
+      FileWriter myWriter = new FileWriter(testOutputFile);
```
vam-google, line 47: Please consider using [try with resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) block for this

src/test/java/com/google/api/generator/test/framework/Utils.java
```diff
@@ -0,0 +1,67 @@
...
+      myWriter.write(codegen);
+      myWriter.flush();
+      myWriter.close();
+    } catch (IOException e) {
+      throw new SaveCodegenToFileException(
```
vam-google, line 52: Since this file is literally about writing to a file, throwing IOException directly form it looks correct (i.e. no need to wrap it in an unchecked exception).

src/test/java/com/google/api/generator/test/framework/Utils.java
```diff
@@ -0,0 +1,67 @@
...
+          String.format(
+              "Error occured when saving codegen to file %s/%s", relativeGoldenDir, fileName));
+    }
+  }
+
+  private static String getTestoutGoldenDir(Class clazz) {
+    return clazz.getPackage().getName().replace(".", "/") + "/goldens/";
+  }
+
+  private static class SaveCodegenToFileException extends RuntimeException {
```
vam-google, line 62: If you choose to keep the wrapper exception here, please declare it public. Keepng it private makes it impossible to catch it in a reasonable way.

src/test/java/com/google/api/generator/test/framework/Utils.java
```diff
@@ -0,0 +1,67 @@
...
+    public SaveCodegenToFileException(String errorMessage) {
```
vam-google, line 63: Plase add the constructor with the "cause" argument and use that for wrapping the IOException. 

Note, the idiomatic way of declaring java exception is to declare all 4 "standard" public exception constructors in it: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Exception.html.

vam-google (commented):
It looks good. I left quite a few comments, but most of them are minor and/or optional. 

My only real concern is the general structure of the BUILD.bazel files (if we don't be careful here, we may end up in a very verbose layout, with many BUILD files everywhere and many targets being redeclared in each BUILD file, which will be very difficult to maintain).

src/test/java/com/google/api/generator/gapic/dummy/BUILD.bazel
```diff
@@ -21,3 +23,39 @@ filegroup(
...
+        "@org_hamcrest_hamcrest_core//jar",
+    ],
+)
+
+TEST_CLASS_NAME = "com.google.api.generator.gapic.dummy.FileDiffInfraDummyTest"
+
+# Run `bazel run src/test/java/com/google/api/generator/gapic/dummy:FileDiffInfraDummyTest.update`
+# to update goldens as expected generated code.
+update_golden(
+    name = "FileDiffInfraDummyTest.update",
```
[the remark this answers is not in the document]
xiaozhenliu-gg5: Thanks! My consideration is for keeping the consistency of the test name `FileDiffInfraDummyTest`, so that when we run `bazel run //src/test/..../dummy: FileDiffInfraDummyTest ` test fails then the command `bazel run //src/test/.../dummy:FileDiffInfraDummyTest.update` can help us update the goldens. But I'm open to modify this to `FileDiffInfraDummyTest_update` or `fileDiffInfraSummyTest_update` for matching lower_snake_case. Let me know your opinion on this.

rules_bazel/java/java_diff_test.bzl
```diff
@@ -0,0 +1,86 @@
...
+        "srcs": attr.label_list(
+            allow_files = True,
+            mandatory = True,
+        ),
+        "test_runner": attr.label(
+            mandatory = True,
+            executable = True,
+            cfg = "host",
+        ),
+        "output_suffix": attr.string(mandatory = False, default = ".zip"),
```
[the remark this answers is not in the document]
xiaozhenliu-gg5: sounds good to me, removed the argument and hard coded on the `output` attr. Thanks

rules_bazel/java/java_diff_test.bzl
```diff
@@ -0,0 +1,86 @@
...
+    },
+    outputs = {
+        "bin": "%{name}.sh",
+    },
+    executable = True,
+    implementation = _overwrite_golden_impl,
+)
+
+def update_golden(name, test_class_name, test_runner, srcs):
+    jnuit_output_name = "%s_output" % name
```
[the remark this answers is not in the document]
xiaozhenliu-gg5: Done, thanks for catching.

src/test/java/com/google/api/generator/gapic/dummy/BUILD.bazel
```diff
@@ -21,3 +23,39 @@ filegroup(
         "@junit_junit//jar",
     ],
 ) for test_name in TESTS]
+
+# JUnit runner binary, this is used to generate test output for updating goldens files.
+java_binary(
+    name = "junit_runner_binary",
```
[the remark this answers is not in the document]
xiaozhenliu-gg5: SG! Done.

src/test/java/com/google/api/generator/gapic/dummy/BUILD.bazel
```diff
@@ -1,3 +1,5 @@
+load("//:rules_bazel/java/java_diff_test.bzl", "update_golden")
```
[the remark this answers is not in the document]
xiaozhenliu-gg5: That makes sense to me, this `junit_runner` will be called for updating every goldens files, I moved it to `root/BUILD.bazel` so that `rules_bazel/java/java_diff_bzl` can directly use it (hard-code the runner) instead of declaring this binary for each package.

src/test/java/com/google/api/generator/gapic/dummy/BUILD.bazel
```diff
@@ -21,3 +23,39 @@ filegroup(
...
+# to update goldens as expected generated code.
+update_golden(
+    name = "FileDiffInfraDummyTest.update",
+    srcs = [
+        ":dummy_files",
+        "//src/test/java/com/google/api/generator/gapic/dummy/goldens:goldens_files",
+        "//src/test/java/com/google/api/generator/test/framework:framework_files",
+    ],
+    test_class_name = TEST_CLASS_NAME,
+    test_runner = "//src/test/java/com/google/api/generator/gapic/dummy:junit_runner_binary",
```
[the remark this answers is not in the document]
xiaozhenliu-gg5: SG, done!

src/test/java/com/google/api/generator/gapic/dummy/BUILD.bazel
```diff
@@ -21,3 +23,39 @@ filegroup(
...
 ) for test_name in TESTS]
+
+# JUnit runner binary, this is used to generate test output for updating goldens files.
+java_binary(
+    name = "junit_runner_binary",
+    srcs = [
+        ":dummy_files",
+        "//src/test/java/com/google/api/generator/test/framework:framework_files",
+    ],
+    data = ["//src/test/java/com/google/api/generator/gapic/dummy/goldens:goldens_files"],
```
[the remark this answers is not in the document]
xiaozhenliu-gg5: Since I moved this to root `BUILD.bazel` the `goldens_files` data stays the same, while I changed the goldens files attr in `java_test` to `glob(["goldens/*.golden"])`

src/test/java/com/google/api/generator/gapic/dummy/BUILD.bazel
```diff
@@ -21,3 +23,39 @@ filegroup(
...
+        "@junit_junit//jar",
+        "@org_hamcrest_hamcrest_core//jar",
+    ],
+)
+
+TEST_CLASS_NAME = "com.google.api.generator.gapic.dummy.FileDiffInfraDummyTest"
+
+# Run `bazel run src/test/java/com/google/api/generator/gapic/dummy:FileDiffInfraDummyTest.update`
+# to update goldens as expected generated code.
+update_golden(
```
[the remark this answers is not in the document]
xiaozhenliu-gg5: Thanks for the context, renamed the target to `updated_golden` and also `overwritten_golden`.

src/test/java/com/google/api/generator/gapic/dummy/FileDiffInfraDummyTest.java
```diff
@@ -36,6 +37,9 @@
   // TODO(xiaozhenliu): remove this test class once the file-diff infra is in place and well-tested.
   private static final String GOLDENFILES_DIRECTORY =
       "src/test/java/com/google/api/generator/gapic/dummy/goldens/";
+  private static final String GOLDENFILES_SIMPLE_CLASS = "FileDiffInfraDummyTestSimpleClass.golden";
```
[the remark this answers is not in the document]
xiaozhenliu-gg5: Currently the `GOLDENFILES_SIMPLE_CLASS ` is static variable and `this.getClass().getSimpleName()` for extracting class name needs non-static context. Removed `static` for the variable since it's only used in this class. Thanks

src/test/java/com/google/api/generator/gapic/dummy/FileDiffInfraDummyTest.java
```diff
@@ -36,6 +37,9 @@
   // TODO(xiaozhenliu): remove this test class once the file-diff infra is in place and well-tested.
   private static final String GOLDENFILES_DIRECTORY =
       "src/test/java/com/google/api/generator/gapic/dummy/goldens/";
```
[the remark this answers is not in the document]
xiaozhenliu-gg5: Thanks, added private helper for deriving this from package name.

src/test/java/com/google/api/generator/test/framework/Utils.java
xiaozhenliu-gg5: SG! This helps me remove the `flush` and `close` methods.

src/test/java/com/google/api/generator/test/framework/SingleJUnitTestRunner.java
```diff
@@ -0,0 +1,55 @@
...
+      throw new JUnitClassNotFoundException(
+          String.format("JUnit test class %s is not found.", className));
+    }
+    Result result = new JUnitCore().run(Request.aClass(clazz));
+    if (!result.wasSuccessful()) {
+      System.out.println("Tests have failures: " + result.getFailures());
+    }
+  }
+
+  private static class JUnitClassNotFoundException extends RuntimeException {
```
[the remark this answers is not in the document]
xiaozhenliu-gg5: For the private Exception for wrapping the checked exception: I think we do not want to throw it in the main class and let outer class to handle this (for `SaveCodegenToFileException `, no outer class for this JunitRunner case), the exception is **private** since they are only used here. Same for exceptions of [differ class ](https://github.com/googleapis/gapic-generator-java/blob/master/src/test/java/com/google/api/generator/test/framework/Differ.java#L60)when we find the differences between goldens and codegen. Generally RuntimeExceptions are exceptions that can be prevented programmatically, and we do control the arguments passing to this class, so I guess uncheckedException might be fine here. But loop @miraleung for some inputs.I will leave the remaining exception-related comments unresloved until we make the call here. Thanks!

dependencies.properties
xiaozhenliu-gg5: Not sure about `gradle` side, but yes if this is only needed from bazel build, I can move this to the `repositoreis.bzl`. Done.

miraleung (commented):
> It looks good. I left quite a few comments, but most of them are minor and/or optional.
> 
> My only real concern is the general structure of the BUILD.bazel files (if we don't be careful here, we may end up in a very verbose layout, with many BUILD files everywhere and many targets being redeclared in each BUILD file, which will be very difficult to maintain).

The motivation for the BUILD file structure in this repo is to keep individual repos self-contained (following google3 style), so the proliferation of files is fine.

Otherwise +1 to Vadym's comments below.

src/test/java/com/google/api/generator/gapic/dummy/BUILD.bazel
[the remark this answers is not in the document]
miraleung: +1 to using `lower_snake_case`.

src/test/java/com/google/api/generator/test/framework/SingleJUnitTestRunner.java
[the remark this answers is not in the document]
miraleung: We wrap our exceptions to prevent propagating `throws` everywhere. Declaring this as protected or public SGTM though.

src/test/java/com/google/api/generator/test/framework/Utils.java
miraleung: +1

src/test/java/com/google/api/generator/test/framework/Utils.java
miraleung: Would prefer to keep the wrapping for consistency.

src/test/java/com/google/api/generator/test/framework/Utils.java
miraleung: +1

src/test/java/com/google/api/generator/gapic/dummy/BUILD.bazel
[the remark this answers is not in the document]
xiaozhenliu-gg5: Renamed to `FileDiffInfraDummyTest_update ` reserved the test name for consistency of test target name, and append `_update` for bazel target name convention. Thanks

src/test/java/com/google/api/generator/test/framework/SingleJUnitTestRunner.java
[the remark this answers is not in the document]
xiaozhenliu-gg5: Updated the nested exception to `public`. Thanks

src/test/java/com/google/api/generator/test/framework/Utils.java
xiaozhenliu-gg5: Done, changed it to `public`.

rules_bazel/java/java_diff_test.bzl
```diff
@@ -0,0 +1,86 @@
...
+
+
+overwrite_golden = rule(
+    attrs = {
+        "unit_test_results": attr.label(
+            mandatory = True,
+            allow_single_file = True),
+    },
+    outputs = {
+        "bin": "%{name}.sh",
```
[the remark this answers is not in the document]
xiaozhenliu-gg5: Yes, to be an executable, the `overwritten_golden` has to output some executable file (sh for this case), which is generated by `_overwritten_golden_impl`. ([ref](must produce an executable output file))

rules_bazel/java/java_diff_test.bzl
```diff
@@ -0,0 +1,86 @@
...
+    inputs = ctx.files.srcs
+    output = ctx.outputs.output
+    test_runner = ctx.executable.test_runner
+
+    command = """
+    mkdir local_tmp  
+    TEST_OUTPUT_HOME="$(pwd)/local_tmp" \
+    {test_runner_path} $@
+    cd local_tmp 
+    zip -r ../{output} .
```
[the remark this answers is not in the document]
xiaozhenliu-gg5: Here in `FileDiffInfraDummyTest` we have two unit tests that would generate golden files (mimic the test like `ResourceNameHelperClassComposerTest`) so the zip contains more than one `.golden` file. `zip`. file might be the best way to transport multiple files.

src/test/java/com/google/api/generator/test/framework/Utils.java
```diff
@@ -0,0 +1,67 @@
...
+
+package com.google.api.generator.test.framework;
+
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+public class Utils {
```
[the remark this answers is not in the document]
xiaozhenliu-gg5: Renamed it to `SaveCodegen` which has single method `saveCodegenToFile()`

rules_bazel/java/java_diff_test.bzl
xiaozhenliu-gg5: Done.

src/test/java/com/google/api/generator/test/framework/Utils.java
xiaozhenliu-gg5: Added one more constructor with `cause` argument since we only use the constructor with error message.

dependencies.properties
xiaozhenliu-gg5: moved it back based on offline talk that we may want to refactor the dependencies all together in future.

rules_bazel/java/java_diff_test.bzl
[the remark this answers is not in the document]
xiaozhenliu-gg5: Added comments to illustrate that.

src/test/java/com/google/api/generator/test/framework/Utils.java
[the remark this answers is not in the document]
xiaozhenliu-gg5: Changed the name back to `Utils` since I moved the helper `getGoldenDir()` and `getClassName()` to this class. The two helpers will be shared among unit tests.

dependencies.properties
miraleung: Thanks.

rules_bazel/java/java_diff_test.bzl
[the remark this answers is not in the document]
miraleung: Thanks, that helps.

rules_bazel/java/java_diff_test.bzl
```diff
@@ -0,0 +1,87 @@
...
+            allow_single_file = True),
+    },
+    outputs = {
+        "bin": "%{name}.sh",
+    },
+    executable = True,
+    implementation = _overwritten_golden_impl,
+)
+
+def updated_golden(name, test_class_name, srcs):
```
miraleung, line 76: Nit: Consider naming this `update_golden` to better reflect the imperative-like usage.

rules_bazel/java/java_diff_test.bzl
```diff
@@ -0,0 +1,87 @@
...
+    )
+    ctx.actions.write(
+        output = ctx.outputs.bin,
+        content = script_content,
+        is_executable = True,
+    )
+    return [DefaultInfo(executable = ctx.outputs.bin)]
+
+
+overwritten_golden = rule(
```
miraleung, line 63: Nit: Consider naming this `overwrite_golden`.

miraleung (approved):

rules_bazel/java/java_diff_test.bzl
xiaozhenliu-gg5: > Bazel rules and macros name are supposed to be nouns, naming what the artifact type that the rule produces. Note, even though the rules and macros look like python funcations (which are named as verbs) they are not such. Here is would be something like "updated_golden" or something like that.

rules_bazel/java/java_diff_test.bzl
xiaozhenliu-gg5: Open to rename it in a refactor PR thanks!

miraleung: Done in #361.

miraleung:
From a clean client, I tried running this update command, and got the error below, plus lots of warnings. Seems that this isn't working and requires each incremental target to be built first, whereas it should all just work with `bazel run`. I'll fix these issues too.

```
bazel run //src/test/java/com/google/api/generator/gapic/dummy:FileDiffInfraDummyTest_update

Target //src/test/java/com/google/api/generator/gapic/dummy:FileDiffInfraDummyTest_update up-to-date:
  bazel-bin/src/test/java/com/google/api/generator/gapic/dummy/FileDiffInfraDummyTest_update.sh
INFO: Elapsed time: 8.211s, Critical Path: 0.06s
INFO: 0 processes.
INFO: Build completed successfully, 4 total actions
INFO: Build completed successfully, 4 total actions
unzip:  cannot find or open bazel-out/k8-fastbuild/bin/src/test/java/com/google/api/generator/gapic/dummy/FileDiffInfraDummyTest_update_output.zip, bazel-out/k8-fastbuild/bin/src/test/java/com/google/api/generator/gapic/dummy/FileDiffInfraDummyTest_update_output.zip.zip or bazel-out/k8-fastbuild/bin/src/test/java/com/google/api/generator/gapic/dummy/FileDiffInfraDummyTest_update_output.zip.ZIP.

```

miraleung:
Another one I found: In a clean client, I get this error, even though it updates fine:
```
Error occurs when reading golden file
```

miraleung:
Fixed all of these in #363.

rules_bazel/java/java_diff_test.bzl
xiaozhenliu-gg5: and #357

xiaozhenliu-gg5:
This implementation was based on the prerequisites of users running `bazel test ` or `bazel build ` first before updating any goldens files. The usage was described in the description, but it does not support directly `bazel run testTarget_update` for updating goldens before any test run.  
`Error occurs when reading golden file` only happens when not finding/reading the goldens files, I did not have chance to repro that error, but thanks for the refactor to allow the direct usage of update_goldens.

PR #355, opened 2020-09-29
[ggj][ast]feat!: Integrate OperationExpr into GeneralForStatement

summer-ji-eng:
Integrate operator expr "<", "++" into index-base for-loop.


miraleung (changes requested):
Will review this after you've had a chance to apply changes from the doc review.

src/main/java/com/google/api/generator/engine/ast/GeneralForStatement.java
```diff
@@ -61,22 +65,24 @@ public static Builder builder() {
...
-
-    public abstract Builder setBody(List<Statement> body);
-
-    // Private.
-    abstract Builder setLocalVariableExpr(VariableExpr variableExpr);
-
-    abstract Builder setMaxSizeExpr(Expr maxSizeExpr);
-
-    abstract VariableExpr localVariableExpr();
+    // Private setter.
```
miraleung, line 68: Are we trying to make these private, or just visible for testing? In the first case, this Builder class should be "private" too, and we wouldn't test the other methods. Otherwise, it should be fine to expose these.

src/main/java/com/google/api/generator/engine/ast/GeneralForStatement.java
```diff
@@ -87,16 +93,18 @@ public GeneralForStatement build() {
...
-      setLocalVariableExpr(varExpr.toBuilder().setIsDecl(false).build());
+      Preconditions.checkState(
+          terminationExpr.type().equals(TypeNode.BOOLEAN),
+          "Terminal expression %s must be boolean-type expression.");
+      Preconditions.checkState(
+          (incrementExpr instanceof MethodInvocationExpr)
+              || (incrementExpr instanceof AssignmentExpr)
+              || (incrementExpr instanceof AssignmentOperationExpr)
+              // TODO(developer): Currently we only support postIncrement (i++), please add
+              // postDecrement, prefixIncrement, prefixIncrement if needs.
```
miraleung, line 104: Nit: "needed."

src/main/java/com/google/api/generator/engine/ast/GeneralForStatement.java
```diff
@@ -87,16 +93,18 @@ public GeneralForStatement build() {
...
-      // This temporary workaround will be removed soon, so it doesn't need a test.
-      setLocalVariableExpr(varExpr.toBuilder().setIsDecl(false).build());
+      Preconditions.checkState(
+          terminationExpr.type().equals(TypeNode.BOOLEAN),
+          "Terminal expression %s must be boolean-type expression.");
+      Preconditions.checkState(
+          (incrementExpr instanceof MethodInvocationExpr)
+              || (incrementExpr instanceof AssignmentExpr)
+              || (incrementExpr instanceof AssignmentOperationExpr)
+              // TODO(developer): Currently we only support postIncrement (i++), please add
```
miraleung, line 103: `TODO(unsupported)` for consistency?

src/test/java/com/google/api/generator/engine/ast/GeneralForStatementTest.java
```diff
@@ -21,57 +21,226 @@
...
+        variableExpr, initValue, maxSizeExpr, Collections.emptyList());
   }
 
   @Test
-  public void invalidForStatement() {
-    Variable variable = Variable.builder().setName("str").setType(TypeNode.STRING).build();
+  public void invalidForStatement_privateVariable() {
+    Variable variable = Variable.builder().setName("i").setType(TypeNode.INT).build();
     VariableExpr variableExpr =
         VariableExpr.builder().setVariable(variable).setScope(ScopeNode.PRIVATE).build();
```
miraleung, line 74: A privately-scoped variable should work, could you see if this compiles?

```
private int i  = 0;
void foobar() {
  for (i = 0; i < 10; i++) {
    System.out.println("i = " + i);
  }
}
```
}

src/main/java/com/google/api/generator/engine/ast/GeneralForStatement.java
```diff
@@ -61,22 +65,24 @@ public static Builder builder() {
...
     // Type-checking will be done in the sub-expressions.
     public GeneralForStatement build() {
-      VariableExpr varExpr = localVariableExpr();
+      GeneralForStatement generalForStatement = autoBuild();
+      AssignmentExpr initializationExpr = generalForStatement.initializationExpr();
+      Expr terminationExpr = generalForStatement.terminationExpr();
+      Expr incrementExpr = generalForStatement.incrementExpr();
+      VariableExpr varExpr = initializationExpr.variableExpr();
       Preconditions.checkState(
           varExpr.scope().equals(ScopeNode.LOCAL),
```
miraleung, line 87: This should have been a `final` check only, could you please fix this as well?

miraleung (changes requested):

src/main/java/com/google/api/generator/engine/ast/GeneralForStatement.java
summer-ji-eng: sure

src/main/java/com/google/api/generator/engine/ast/GeneralForStatement.java
summer-ji-eng: We should try to make these private setter at this stage, only expose convenience wrapper to user.  My thought is to check type for initialization, termination, increment expression. So I also try to add unit test for the type-checking. I am not sure which one to choose, need your opinion:
1. if we make Builder private,  and we wouldn't test the methods,  do we still need to check the expressions' types?
2. or we make these tests VisibleForTesting

src/test/java/com/google/api/generator/engine/ast/GeneralForStatementTest.java
summer-ji-eng: Confirm. This code is valid and pass compile. Any suggestion on how should we do check for this case?

src/main/java/com/google/api/generator/engine/ast/GeneralForStatement.java
summer-ji-eng: I am not sure if I get the point here. Variable declared in initialization should be a local variable, variable scope private, public, protected, static, are not allowed, but final is allowed in below case with i++ invalid of "cannot assign a value to final variable 'i'"
```
    for (final int i = 0; i < 10; i++) {
      System.out.println("i = " + i);
    } 
```

src/test/java/com/google/api/generator/engine/ast/GeneralForStatementTest.java
summer-ji-eng: Okay, I spend some time on variable check. Let me organize in the design doc. I will let you know once I finished.

src/main/java/com/google/api/generator/engine/ast/GeneralForStatement.java
summer-ji-eng: I got the point. I will update the doc with variable type checking.

summer-ji-eng:
@miraleung ready for second round review. Thanks

summer-ji-eng:
Merge dependencies fix. It is ready to review @miraleung @xiaozhenliu-gg5 
Thanks for reviewing :)

src/main/java/com/google/api/generator/engine/ast/GeneralForStatement.java
```diff
@@ -43,60 +35,85 @@ public void accept(AstNodeVisitor visitor) {
...
+      // Declare a variable inside for-loop initialization expression.
+      if (localVarExpr.isDecl()) {
+        Preconditions.checkState(
+            localVarExpr.scope().equals(ScopeNode.LOCAL),
+            String.format(
+                "Variable %s declare in a general for-loop cannot have a non-local scope",
+                localVarExpr.variable().identifier().name()));
+        Preconditions.checkState(!localVarExpr.isStatic(), "Modifier 'static' not allow here.");
+      }
+      // TODO (unsupport): Uncomment the following code if public setter for the initialization,
```
miraleung, line 92: In general, checking in commented-out unused code is discouraged because it may suffer "code rot" as the codebase evolves. Instead, could you please convert these into a brief summary on the conditions to check instead?

miraleung (changes requested):
LGTM with the commented-out code addressed.

src/main/java/com/google/api/generator/engine/ast/GeneralForStatement.java
summer-ji-eng: Make sense. Updated the comments.

src/main/java/com/google/api/generator/engine/ast/GeneralForStatement.java
```diff
@@ -22,19 +22,11 @@
 
 @AutoValue
 public abstract class GeneralForStatement implements Statement {
-  public abstract Expr initializationExpr();
+  public abstract AssignmentExpr initializationExpr();
```
miraleung, line 25: Could you please confirm if this must be an AssignmentExpr (e.g. what about passing `foobar()` here)?

miraleung (approved):

src/main/java/com/google/api/generator/engine/ast/GeneralForStatement.java
summer-ji-eng: Actually, it should not be limited to assignmentExpr, it should be expr which belongs to statmentExprList(https://docs.oracle.com/javase/specs/jls/se10/html/jls-14.html#jls).
My original though is to simplify the way to get variable. After consideration of back compatible, it should change back to expr. I will make change on this.

summer-ji-eng: @miraleung changed the interface use `Expr` for initialization. Before I merge the code, could you take one more look. Thanks.

src/main/java/com/google/api/generator/engine/ast/GeneralForStatement.java
```diff
@@ -43,60 +35,69 @@ public void accept(AstNodeVisitor visitor) {
...
+    GeneralForStatement build() {
+      GeneralForStatement generalForStatement = autoBuild();
+      Expr initExpr = generalForStatement.initializationExpr();
+      if (initExpr instanceof AssignmentExpr) {
+        VariableExpr localVarExpr = ((AssignmentExpr) initExpr).variableExpr();
+        // Declare a variable inside for-loop initialization expression.
+        if (localVarExpr.isDecl()) {
+          Preconditions.checkState(
+              localVarExpr.scope().equals(ScopeNode.LOCAL),
+              String.format(
```
miraleung, line 89: Any non-final variable should work here, I think?

miraleung (changes requested):

miraleung (approved):

summer-ji-eng:
Confirmed in Chat.

PR #363, opened 2020-10-03
[ggj][bazel][goldens] fix: refactor and fix bazel golden_update rules

miraleung:
Fixes the following issues:
- `bazel run testTarget_update` did not work when starting from a clean cache. This refactoring ensures that all the dependent targets are built when the former is run.
- Removes all the Bazel direct-import warnings.
- Separates per-test targets from the test runner binary, so that we don't need to glob all sources and all goldens into a single `java_binary` target.
- Fixes file-update errors for tests that use multiple goldens.

### DIFF
--- a/rules_bazel/java/java_diff_test.bzl
+++ b/rules_bazel/java/java_diff_test.bzl
@@ -44,7 +44,7 @@ junit_output_zip = rule(
     implementation = _junit_output_impl,  
 )
     
-def _overwritten_golden_impl(ctx):
+def _overwrite_golden_impl(ctx):
     script_content = """
     #!/bin/bash
     cd ${{BUILD_WORKSPACE_DIRECTORY}}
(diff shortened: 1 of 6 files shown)


xiaozhenliu-gg5 (approved):

rules_bazel/java/java_diff_test.bzl
```diff
@@ -1,87 +1,84 @@
...
+        data = [],
+        deps = []):
+    golden_junit_runner_name = "%s_junit_runner" % name
+    native.java_binary(
+        name = golden_junit_runner_name,
         srcs = srcs,
+        data = data,
+        main_class = "com.google.api.generator.test.framework.SingleJUnitTestRunner",
+        deps = ["//src/test/java/com/google/api/generator/test/framework:junit_runner"] + deps,
     )
```
xiaozhenliu-gg5, line 77: Can you explain a little bit why the `def junit_output_zip ` is replaced by the `native.java_binary` and it will make the build running before running overwrite_golden?  Thanks

PR #366, opened 2020-10-04
feat: add final keyword check on assignment expr

summer-ji-eng:
Cannot assign a value to final variable.

Screen Shot 2020-10-03 at 10 35 38 PM


src/main/java/com/google/api/generator/engine/ast/AssignmentExpr.java
```diff
@@ -67,6 +68,15 @@ public AssignmentExpr build() {
                   lhsType.reference().name(), rhsType.reference().name()));
         }
       }
+
+      if (!assignmentExpr.variableExpr().isDecl()) {
```
miraleung, line 72: What if the decl is final too? That is, I think we can remove the outer check.

miraleung (changes requested):

PR #378, opened 2020-10-08
[ggj][infra][1/5] integration-test: add goldens files for Redis client library

xiaozhenliu-gg5:
Find more details in design doc: go/java-micro-integration-test-infra

1. add client library files generated from monolith generator (3 intermediary srcjars from [rule](https://github.com/googleapis/gapic-generator-java/blob/master/rules_java_gapic/java_gapic.bzl#L37)) to goldens folder.
2. add a bazel file for the goldens/redis folder.

Next Step:

[2/5] Add integration test bazel rules which will run Redis client library generation and compare Java source code with the added goldens files.


miraleung (approved):
Could we add a README to `test/integration/goldens` so that future readers will know that these are actual source Java files, but just goldens?

miraleung (changes requested):
On second thought, let's update these goldens with the upcoming microgenerator rules.

xiaozhenliu-gg5:
On hold until the microgenerator rules are ready and update Redis goldens. @miraleung

test/integration/goldens/redis/BUILD.bazel
```diff
@@ -0,0 +1,6 @@
+package(default_visibility = ["//visibility:public"])
+
+filegroup(
+    name = "redis_goldens_files",
```
miraleung, line 4: Just `goldens_files` should be sufficient, since this is in the `redis` directory.

miraleung (approved):
LGTM with BUILD target comment.

test/integration/goldens/redis/BUILD.bazel
xiaozhenliu-gg5: SG, renamed

PR #393, opened 2020-10-13
[ggj][infra][2/5] feat: bazel rules support for integration test

xiaozhenliu-gg5:
Find more details in design doc: go/java-micro-integration-test-infra

Support bazel rules to run integration test against Redis API, which includes 1) code generation by micro-generator 2) compare codegen with goldens in `test/integration/goldens/redis`.

Run the below command to trigger the integration test, and the test failure (diff between codegen and goldens) can be found directly in test.log 

```
bazel test //test/integration:redis

FAIL: //test/integration:redis (see /usr/local/google/home/xiaozhenliu/.cache/bazel/_bazel_xiaozhenliu/96b19089c047889e0621b0e06e850da1/execroot/gapic_generator_java/bazel-out/k8-fastbuild/testlogs/test/integration/redis/test.log)
Target //test/integration:redis up-to-date:
  bazel-bin/test/integration/diff_output.txt
  bazel-bin/test/integration/check_diff_script.sh
INFO: Elapsed time: 155.658s, Critical Path: 64.83s
INFO: 487 processes: 35 internal, 375 linux-sandbox, 77 worker.
INFO: Build completed, 1 test FAILED, 487 total actions
//test/integration:redis                                                 FAILED in 0.1s
  /usr/local/google/home/xiaozhenliu/.cache/bazel/_bazel_xiaozhenliu/96b19089c047889e0621b0e06e850da1/execroot/gapic_generator_java/bazel-out/k8-fastbuild/testlogs/test/integration/redis/test.log



============= test.log ===============
exec ${PAGER:-/usr/bin/less} "$0" || exit 1
Executing tests from //test/integration:redis
-----------------------------------------------------------------------------
diff codegen_tmp/CloudRedisClient.java test/integration/goldens/redis/CloudRedisClient.java
17,18d16
< package com.google.cloud.redis.v1;
<
```


rules_bazel/java/java_integration_test.bzl
```diff
@@ -0,0 +1,82 @@
...
+            input[1][JavaInfo].source_jars[0], 
+            input[2][JavaInfo].source_jars[0],
+        ],
+        outputs = [diff_output],
+        command = script,
+    )
+
+
+    check_diff_script_content = """
+    cd ${{BUILD_WORKSPACE_DIRECTORY}}
```
xiaozhenliu-gg5, line 37: Can I get some input here? @miraleung @vam-google 

The path of `diff_output` is `bazel-out/k8-fastbuild/bin/test/integration/redis_diff.txt`, but seems we won't get access to `bazel-out` folders in this environment. `${{BUILD_WORKSPACE_DIRECTORY}}` does not work well here, it complains about `HOME not set`. 

Some available paths according to this [doc](https://docs.bazel.build/versions/master/test-encyclopedia.html#initial-conditions)
```
>> PWD: /usr/local/google/home/xiaozhenliu/.cache/bazel/_bazel_xiaozhenliu/96b19089c047889e0621b0e06e850da1/sandbox/linux-sandbox/1/execroot/gapic_generator_java/bazel-out/k8-fastbuild/bin/test/integration/codegen_comparison_script.sh.runfiles/gapic_generator_java

>> TEST_TMPDIR: /usr/local/google/home/xiaozhenliu/.cache/bazel/_bazel_xiaozhenliu/96b19089c047889e0621b0e06e850da1/sandbox/linux-sandbox/1/execroot/gapic_generator_java/_tmp/9ce7325f79eb1f0710ae560ae2ac77ca

>> TEST_WORKSPACE: gapic_generator_java

>> TEST_SRCDIR: /usr/local/google/home/xiaozhenliu/.cache/bazel/_bazel_xiaozhenliu/96b19089c047889e0621b0e06e850da1/sandbox/linux-sandbox/1/execroot/gapic_generator_java/bazel-out/k8-fastbuild/bin/test/integration/
codegen_comparison_script.sh.runfiles
```

rules_bazel/java/java_integration_test.bzl
```diff
@@ -0,0 +1,82 @@
...
+    unzip -j {input_resource_name} -d  codegen_tmp
+    unzip -j {input_test} -d codegen_tmp
+    cd codegen_tmp
+    # Remove unneeded non-Java files, like MANIFEST
+    rm -rf $(find . -type f ! -name "*.java")
+    cd ..
+    diff codegen_tmp test/integration/goldens/{api_name}/ > {diff_output}
+    """.format(
+        diff_output = diff_output.path,
+        input = input[0][JavaInfo].source_jars[0].path,
```
vam-google, line 20: If you prefer accessing each individual source jar by its index here, then pelase consider just splitting  this one api_target list argument into 3 different arguments, each of which with a descriptive name, something like:
`gapic_library`
`resource_name_library`
`test_library`

rules_bazel/java/java_integration_test.bzl
vam-google: When the `BUILD_WORKSPACE_DIRECTORY` env variable is not present, in which context do you run your command: `bazel build`, `bazel run`, `bazel test`? I believe `BUILD_WORKSPACE_DIRECTORY` is there only for bazel run.

rules_bazel/java/java_integration_test.bzl
```diff
@@ -0,0 +1,82 @@
...
+        outputs = [diff_output],
+        command = script,
+    )
+
+
+    check_diff_script_content = """
+    cd ${{BUILD_WORKSPACE_DIRECTORY}}
+    if [ -s {diff_output} ]
+    then
+        cat {diff_output}
```
vam-google, line 40: Is it a final variant or a debugging artifacts, which will be removed before submitting?

xiaozhenliu-gg5: This is a final variant, `diff_output` is a text file that contains all the diffs between goldens and codegen, here we are checking if the `diff_output` is empty. Empty means no difference then test passes, but if it;s not empty, `exit 1` will help us throw error, and `cat {diff_output}` prints the diff information to the generated `sh` file.

xiaozhenliu-gg5:
> When the BUILD_WORKSPACE_DIRECTORY env variable is not present, in which context do you run your command: bazel build, bazel run, bazel test? I believe BUILD_WORKSPACE_DIRECTORY is there only for bazel run.

we use `bazel test //test/integration: redis` to trigger the integration test of Redis API. More info is in description. Thanks! @vam-google If `BUILD_WORKSPACE_DIRECTORY` is only for bazel run, then is there an equivalent env that can enable us to have access to the generated `diff_output` folder?

rules_bazel/java/java_integration_test.bzl
xiaozhenliu-gg5: Sure, I think that would make the input more descriptive. Renamed them, thank you!

rules_bazel/java/java_integration_test.bzl
```diff
@@ -0,0 +1,84 @@
+def _compare_with_goldens_test_impl(ctx):
```
miraleung, line 1: Just `integration_test` should be sufficient, since this is in the `java` directory.

Also, could we rename this method to something more "imperative-verb"-like, such as `diff_integration_goldens`?

miraleung (changes requested):

rules_bazel/java/java_integration_test.bzl
```diff
@@ -0,0 +1,84 @@
...
+    )
+
+
+    check_diff_script_content = """
+    cd ${{BUILD_WORKSPACE_DIRECTORY}}
+    if [ -s {diff_output} ]
+    then
+        cat {diff_output}
+        exit 1 
+    else
```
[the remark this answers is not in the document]
xiaozhenliu-gg5: Sure, removed.

rules_bazel/java/java_integration_test.bzl
xiaozhenliu-gg5: 1) In the goldens files rules, we still use `actions.write` [here](https://github.com/googleapis/gapic-generator-java/blob/master/rules_bazel/java/java_diff_test.bzl#L36) I think last time it needs extra `bazel build` because the test runner (code generation) is separate from the goldens comparison step. Please correct me if wrong.  And This case is similar to goldens files and `actions.run_shell` will not work because as an executable rule, we need to return a `[DefaultInfo(executable =..)]`, `actions.write()` is for writing such a shell script which can be executed when using running `bazel test //test/integration:redis`.

2) This `actions.write()` is trying to write a script and return executable. The script will check if the `diff_output.txt` is empty or not: empty means no diff, then test passes, while if the diff file is not empty, fail the test. I'm trying to define a way to tell the test succeed or not.

Now the main question is that when entering the second `actions.write()`, the bazel enters in test mode, it's running in a isolated container, so the generated srcjar or diff file can not be passed in. If we run this rule by `bazel run //test/integration:redis` it will work fine, because in `bazel run` mode, we have access to the `BUILD_WORKSPACE_DIRECTORY ` and all generated files which is located at `bazel-out/k8-fastbuild/bin/test/integration/redis_diff.txt` @miraleung @vam-google

rules_bazel/java/java_integration_test.bzl
```diff
@@ -0,0 +1,84 @@
...
+    outputs = {
+        "diff_output": "%{name}_diff.txt",
+        "check_diff_script": "check_diff_script.sh",
+    },
+    implementation = _compare_with_goldens_test_impl,
+    test = True,
+)
+
+
+def integration_test(name, target, data):
```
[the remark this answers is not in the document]
xiaozhenliu-gg5: sure, this wrapper works mainly for taking the existing target and explicitly name the three srcjar libraries (gapic_library, resource-name_library, test_library), if we remove the wrapper and directly takes one target, then this one target would contain 3 generated srcjars, in the ` _compare_with_goldens_test_impl` macro, we will have to take those jars by `input[0], input[1]`...like the previous [commit](https://github.com/googleapis/gapic-generator-java/pull/393/commits/5e65c2d50b00b31e7271eeb1f32347e634dbe7ef). Let me know your thoughts or there's better way to do it, thanks

rules_bazel/java/java_integration_test.bzl
xiaozhenliu-gg5: @miraleung previously we have only 1 arg (target) which we can remove the wrapper function of `integration_test`, while adding a wrapper to call `compare_with_goldens_test` can help us split the list arguments and give a descriptive name. WDYT?

rules_bazel/java/java_integration_test.bzl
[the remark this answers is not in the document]
xiaozhenliu-gg5: SG, added comments for clarity, thanks

xiaozhenliu-gg5: renamed the file to `integration_test.bzl` and the method name to `diff_integration_goldens_test` (bazel test requires the name to end with `_test`)

rules_bazel/java/java_integration_test.bzl
xiaozhenliu-gg5: This is resolved, the data we want to use in the test rules can be passed by `runfiles` attribute.

miraleung (approved):

PR #399, opened 2020-10-21
[ggj][engx]chore: update pre-commit

xiaozhenliu-gg5:


.githooks/pre-commit
```diff
@@ -4,36 +4,34 @@
...
+# Constants.
+BOLD="\e[1m"
+UNSET="\e[0m"
+WHITE="\e[97m"
+RED="\e[91m"
+BACK_MAGENTA="\e[45m"
+BACK_BLUE="\e[44m"
+BACK_RED="\e[41m"
+BACK_GREEN="\e[42m"
+# Check only the staged files.
```
miraleung, line 16: Nit: Would prefer to keep the "number of files" section in its original spot so that it's closer to its invocation site.

miraleung (approved):

.githooks/pre-commit
xiaozhenliu-gg5: Moved it back, thanks!

PR #427, opened 2020-10-29
[ggj][engx][test] test: add gradle assembly rules for logging and redis

miraleung:


xiaozhenliu-gg5 (commented):
Seems we have more changes than adding assembly rules. Will review it once the extra changes are merged.

PR #430, opened 2020-10-29
[ggj][engx] test: add asset API gradle pkg rules

miraleung:


xiaozhenliu-gg5 (commented):
Wondering the reason we add this rule, do we need it in the integration test somewhere? Thanks~

miraleung:
This rule will make it easy to bazel-build the final client library, and then Gradle-build it for development and testing. In the future, we may want to add that Gradle build to the integration tests.

PR #471, opened 2020-11-10
[composer][SampleCode]Utils function formatter for Sample Code

summer-ji-eng:
input: List<Statements>
output: Formatted String.

1. wrap the statements in a fake class so that the block of code is valid for google-java-format.
2. trim the first and last two lines
3. remove the first 4 indent each line


src/main/java/com/google/api/generator/gapic/utils/SampleCodeJavaFormatter.java
```diff
@@ -0,0 +1,64 @@
...
+      String.format("void fakeMethod() %s%s", LEFT_BRACE, NEWLINE);
+  private static final String FAKE_METHOD_CLOSE = String.format("%s%s", RIGHT_BRACE, NEWLINE);
+  private static final String FAKE_CLASS_CLOSE = String.format("%s", RIGHT_BRACE);
+
+  /**
+   * format utilize google-java-format to format the sample code statements where these are wrapped
+   * in a FakeClass to pretend as a valid source code. Because google-java-format is a program that
+   * reformats Java source code.
+   */
+  public static String format(List<Statement> statements) {
```
miraleung, line 42: Can we simplify this by having it take a String instead of a list of statements? For example, one should be able to call `format(writer.write())` and get the appropriately formatted code.

src/main/java/com/google/api/generator/gapic/utils/SampleCodeJavaFormatter.java
```diff
@@ -0,0 +1,64 @@
...
+
+  private static final String FAKE_CLASS_TITLE =
+      String.format("public class FakeClass %s%s", LEFT_BRACE, NEWLINE);
+  private static final String FAKE_METHOD_TITLE =
+      String.format("void fakeMethod() %s%s", LEFT_BRACE, NEWLINE);
+  private static final String FAKE_METHOD_CLOSE = String.format("%s%s", RIGHT_BRACE, NEWLINE);
+  private static final String FAKE_CLASS_CLOSE = String.format("%s", RIGHT_BRACE);
+
+  /**
+   * format utilize google-java-format to format the sample code statements where these are wrapped
```
miraleung, line 38: In general, the class comment should only describe a high-level summary of the expected behavior, and we should avoid mentioning implementation details here.

miraleung (changes requested):

src/main/java/com/google/api/generator/gapic/utils/SampleCodeJavaFormatter.java
summer-ji-eng: Great point. Updated the code.

src/main/java/com/google/api/generator/gapic/utils/SampleCodeJavaFormatter.java
summer-ji-eng: My purpose is to explain the why we need to wrap the code into a class for google-java-format. I moved the comment inside of `format` method then. 
What do you suggest on this comment? Do you think it is necessary? Or maybe we just simply remove the comment. LMK what you think. Thanks

src/main/java/com/google/api/generator/gapic/utils/SampleCodeJavaFormatter.java
```diff
@@ -0,0 +1,71 @@
...
+
+  private static final String FAKE_CLASS_TITLE =
+      String.format("public class FakeClass %s%s", LEFT_BRACE, NEWLINE);
+  private static final String FAKE_METHOD_TITLE =
+      String.format("void fakeMethod() %s%s", LEFT_BRACE, NEWLINE);
+  private static final String FAKE_METHOD_CLOSE = String.format("%s%s", RIGHT_BRACE, NEWLINE);
+  private static final String FAKE_CLASS_CLOSE = String.format("%s", RIGHT_BRACE);
+
+  /**
+   * This method is used to format sample code by utilizing google-java-format to format the Java
```
miraleung, line 38: As before, could we please avoid mentioning implementation-specific details in the class comment? That is, this comment should not contain `google-java-format` or related details.

src/main/java/com/google/api/generator/gapic/utils/SampleCodeJavaFormatter.java
```diff
@@ -0,0 +1,71 @@
...
+
+    String formattedString = null;
+    try {
+      formattedString = FORMATTER.formatSource(buffer.toString());
+    } catch (FormatterException e) {
+      throw new FormatException(
+          String.format("The sample code should be string where is composed by statements; %s", e));
+    }
+    return formattedString
+        .replaceAll("^([^\n]*\n){2}|([^\n]*\n){2}$", "")
```
miraleung, line 60: Add some comments to summarize what these regexes are doing?

src/main/java/com/google/api/generator/gapic/utils/SampleCodeJavaFormatter.java
```diff
@@ -0,0 +1,71 @@
...
+        .replaceAll("(?m)^ {4}", "")
+        .trim();
+  }
+
+  @VisibleForTesting
```
miraleung, line 65: What about just [rethrowing](https://stackoverflow.com/questions/1097527/rethrowing-exceptions-in-java-without-losing-the-stack-trace) the exception instead of wrapping it, since this is the only location it's used?

src/test/java/com/google/api/generator/gapic/utils/SampleCodeJavaFormatterTest.java
```diff
@@ -0,0 +1,148 @@
...
+  public void validFormatSampleCode_tryCatchStatement() {
+    List<Statement> statements = Arrays.asList(createTryCatchSampleCode());
+    String result = SampleCodeJavaFormatter.format(writeStatements(statements));
+    String expected =
+        String.format(createLines(3), "try (boolean condition = false) {\n", "  int x = 3;\n", "}");
+    assertEquals(expected, result);
+  }
+
+  @Test
+  public void validFormatSampleCode_longLineStatement() {
```
miraleung, line 50: Since we're just testing the formatter, let's simplify this test by directly writing strings instead of constructing code via the AST.

miraleung (changes requested):

miraleung (changes requested):
Also, let's move these files out of `utils` and into `composer/samplecode` instead.

src/main/java/com/google/api/generator/gapic/utils/SampleCodeJavaFormatter.java
summer-ji-eng: Update the comment.

src/main/java/com/google/api/generator/gapic/utils/SampleCodeJavaFormatter.java
summer-ji-eng: If we rethrowing the exception here, we will have to catch every time use `SampleCodeFormatter.format`. And it will be use in each composeSampleCode function, (composeClassHeaderMethodSampleCode, composeClassHeaderCredentialSampleCode, composeClassHeaderEndpointSampleCode, etc)

src/main/java/com/google/api/generator/gapic/utils/SampleCodeJavaFormatter.java
summer-ji-eng: Added.

src/main/java/com/google/api/generator/gapic/composer/samplecode/SampleCodeJavaFormatter.java
```diff
@@ -0,0 +1,75 @@
...
+
+  private static final String FAKE_CLASS_TITLE =
+      String.format("public class FakeClass %s%s", LEFT_BRACE, NEWLINE);
+  private static final String FAKE_METHOD_TITLE =
+      String.format("void fakeMethod() %s%s", LEFT_BRACE, NEWLINE);
+  private static final String FAKE_METHOD_CLOSE = String.format("%s%s", RIGHT_BRACE, NEWLINE);
+  private static final String FAKE_CLASS_CLOSE = String.format("%s", RIGHT_BRACE);
+
+  /**
+   * This method is used to format sample code string where composed by statements based on Google
```
miraleung, line 38: Can we remove the "composed by statements" part?

src/main/java/com/google/api/generator/gapic/composer/samplecode/SampleCodeJavaFormatter.java
```diff
@@ -0,0 +1,75 @@
...
+   * Java style guide.
+   *
+   * @param sampleCode A string is composed by statements.
+   * @return String Formatted sample code string based on google java style.
+   */
+  public static String format(String sampleCode) {
+    final StringBuffer buffer = new StringBuffer();
+    buffer.append(FAKE_CLASS_TITLE);
```
miraleung, line 46: Add a comment to explain why we append the fake class and method?

src/main/java/com/google/api/generator/gapic/composer/samplecode/SampleCodeJavaFormatter.java
```diff
@@ -0,0 +1,75 @@
...
+  private static final Formatter FORMATTER = new Formatter();
+  private static final String RIGHT_BRACE = "}";
+  private static final String LEFT_BRACE = "{";
+  private static final String NEWLINE = "\n";
+
+  private static final String FAKE_CLASS_TITLE =
+      String.format("public class FakeClass %s%s", LEFT_BRACE, NEWLINE);
+  private static final String FAKE_METHOD_TITLE =
+      String.format("void fakeMethod() %s%s", LEFT_BRACE, NEWLINE);
+  private static final String FAKE_METHOD_CLOSE = String.format("%s%s", RIGHT_BRACE, NEWLINE);
```
miraleung, line 34: We can just use two `LEFT_BRACE`s insetad of this variable, could we please remove it?

src/test/java/com/google/api/generator/gapic/composer/samplecode/SampleCodeJavaFormatterTest.java
```diff
@@ -0,0 +1,81 @@
...
+            "        echoSettingsBuilder\n",
+            "            .echoSettings()\n",
+            "            .getRetrySettings()\n",
+            "            .toBuilder()\n",
+            "            .setTotalTimeout(Duration.ofSeconds(30))\n",
+            "            .build());");
+    assertEquals(expected, result);
+  }
+
+  @Test(expected = FormatException.class)
```
miraleung, line 72: Can we use `assertThrows` for consistency with our other tests?

miraleung (changes requested):

src/main/java/com/google/api/generator/gapic/utils/SampleCodeJavaFormatter.java
miraleung: Moving the comment into the method SGTM.

src/main/java/com/google/api/generator/gapic/composer/samplecode/SampleCodeJavaFormatter.java
summer-ji-eng: Done

src/main/java/com/google/api/generator/gapic/composer/samplecode/SampleCodeJavaFormatter.java
summer-ji-eng: Done

src/main/java/com/google/api/generator/gapic/composer/samplecode/SampleCodeJavaFormatter.java
summer-ji-eng: SG

src/test/java/com/google/api/generator/gapic/composer/samplecode/SampleCodeJavaFormatterTest.java
summer-ji-eng: Gotcha.

miraleung (approved):

PR #500, opened 2020-11-17
[engine][ast]Implement Break Statement

summer-ji-eng:
To support write break statement:
Use case in composing sample code like:
```
   * try (EchoClient echoClient = EchoClient.create()) {
   *   PagedExpandRequest request = PagedExpandRequest.newBuilder().build();
   *   while (true) {
   *     PagedExpandResponse response = EchoClient.pagedExpandCallable().call(request);
   *     for (PagedExpandResponse element : response.getResponsesList()) {
   *       // doThingsWith(element);
   *     }
   *     String nextPageToken = response.getNextPageToken();
   *     if (!String.isNullOrEmpty(nextPageToken)) {
   *       request = PagedExpandRequest.toBuilder().setPageToken(nextPageToken).build();
   *     } else {
   *       break;
   *     }
   *   }
   * }
```

### DIFF
--- a/src/main/java/com/google/api/generator/engine/ast/BreakStatement.java
+++ b/src/main/java/com/google/api/generator/engine/ast/BreakStatement.java
@@ -1,3 +1,17 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/api/generator/engine/ast/BreakStatement.java
```diff
@@ -0,0 +1,28 @@
+// Copyright 2020 Google LLC
```
miraleung, line 1: Why not a `BreakExpr`? That would be more consistent with the rest of the AST.

miraleung (changes requested):

miraleung:
Could we please update the sample sniippet in the description as well?

xiaozhenliu-gg5 (approved):

miraleung (approved):

PR #512, opened 2020-11-18
[composer][2/5]Implement pure RPC method's sample code

summer-ji-eng:
SampleCode
```
 * try (SubscriptionAdminClient subscriptionAdminClient = SubscriptionAdminClient.create()) {
 *   ProjectSubscriptionName name = ProjectSubscriptionName.of("[PROJECT]", "[SUBSCRIPTION]");
 *   TopicName topic = TopicName.ofProjectTopicName("[PROJECT]", "[TOPIC]");
 *   PushConfig pushConfig = PushConfig.newBuilder().build();
 *   int ackDeadlineSeconds = 0;
 *   Subscription response = subscriptionAdminClient.createSubscription(name, topic, pushConfig, ackDeadlineSeconds);
 * }
```


src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
```diff
@@ -50,13 +55,30 @@ public static TryCatchStatement composeRpcMethodSampleCode(
 
   private static TryCatchStatement composeUnaryRpcMethodSampleCode(
       Method method, List<MethodArgument> arguments, TypeNode clientType) {
-    // TODO(summerji): compose sample code for unary rpc method.
+    // Assign each method arguments with default value.
```
miraleung, line 58: Can we please add TODOs for adding unit tests for each private method here?

src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
```diff
@@ -50,13 +55,30 @@ public static TryCatchStatement composeRpcMethodSampleCode(
...
+    List<Statement> bodyStatements =
+        arguments.stream()
+            .map(
+                methodArg ->
+                    ExprStatement.withExpr(assignMethodArgumentWithDefaultValue(methodArg)))
+            .collect(Collectors.toList());
+    // Invoke current method based on return type.
+    // e.g. if return void, echoClient.echo(..); or,
+    // e.g. if return other type, EchoResponse response = echoClient.echo(...);
+    Expr invokeMethodExpr =
```
miraleung, line 68: This is quite large, so perhaps use an if-else instead?

src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
```diff
@@ -50,13 +55,30 @@ public static TryCatchStatement composeRpcMethodSampleCode(
...
+        method.outputType().equals(TypeNode.VOID)
+            ? MethodInvocationExpr.builder()
+                .setExprReferenceExpr(createVariableDeclExpr(getClientName(clientType), clientType))
+                .setMethodName(method.name())
+                .setReturnType(clientType)
+                .build()
+            : createAssignExprForVariableWithClientMethod(
+                RESPONSE_VAR_NAME, method.outputType(), clientType, method.name(), arguments);
+    bodyStatements.add(ExprStatement.withExpr(invokeMethodExpr));
+
```
miraleung, line 78: Could you please confirm if more sample code will be coming?

miraleung (changes requested):

src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
summer-ji-eng: Sure. Added.

src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
summer-ji-eng: Cool. Refactored.

src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
summer-ji-eng: Sorry, I am not sure if I am understanding correctly.

This PR is aim to produce the sample code for pure unary Rpc method kind. There are more coming for other kinds  methods:
- paged unary rpc method
- lro unary rpc method
- unary default rpc method
and rpc callable methods where includes:
- stream rpc method
- paged rpc callable method
- lro rpc callable method
- unary callable method

src/main/java/com/google/api/generator/gapic/composer/BUILD.bazel
```diff
@@ -15,9 +15,11 @@ java_library(
     deps = [
         "//:service_config_java_proto",
         "//src/main/java/com/google/api/generator/engine/ast",
+        "//src/main/java/com/google/api/generator/engine/writer",
```
miraleung, line 18: Wasn't this diff already part of the base branch? Can we rebase this such that difss from base PRs go away?

src/main/java/com/google/api/generator/gapic/composer/ClassComposer.java
```diff
@@ -16,9 +16,11 @@
...
 import com.google.api.generator.gapic.model.GapicClass;
 import com.google.api.generator.gapic.model.Message;
+import com.google.api.generator.gapic.model.ResourceName;
 import com.google.api.generator.gapic.model.Service;
 import java.util.Map;
 
 public interface ClassComposer {
-  GapicClass generate(Service service, Map<String, Message> messageTypes);
+  GapicClass generate(
+      Service service, Map<String, ResourceName> resourceNames, Map<String, Message> messageTypes);
```
miraleung, line 25: Can we avoid adding another parameter (that isn't used in most places) to this interface? I would prefer making the Composer (that actually uses this) not inherit from ClassComposer instead. Please add a TODO on me to clean up the hierarchy afterwards.

src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
```diff
@@ -0,0 +1,223 @@
...
+import com.google.api.generator.gapic.model.MethodArgument;
+import com.google.api.generator.gapic.model.ResourceName;
+import com.google.api.generator.gapic.utils.JavaStyle;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public final class SampleCodeHelperComposer {
+  private static String RESPONSE_VAR_NAME = "response";
```
miraleung, line 38: I thought I saw this in another PR already, can we make the diffs from base PRs go away, so I can identify what I should be reviewing?

Will review the rest of the code afterwards.

miraleung (changes requested):

src/test/java/com/google/api/generator/gapic/composer/goldens/EchoClient.golden
```diff
@@ -63,7 +63,20 @@ import javax.annotation.Generated;
  *
  * <p>To customize credentials:
  *
+ * <pre>{@code
```
miraleung, line 66: Wasn't this already checked in? Can we sync it so that the diffs from prior PRs don't show up here?

src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
```diff
@@ -0,0 +1,223 @@
...
+        .build();
+  }
+
+  private static Expr createAssignExprForVariableWithClientMethod(
+      String variableName,
+      TypeNode variableType,
+      TypeNode clientType,
+      String methodName,
+      List<MethodArgument> arguments) {
+    VariableExpr varExpr = createVariableExpr(variableName, variableType);
```
miraleung, line 176: Likewise, with the refactoring above, we won't need to recreate the VariableExprs here.

src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
```diff
@@ -0,0 +1,223 @@
...
+      TypeNode clientType,
+      Map<String, ResourceName> resourceNames) {
+    // TODO(summerji): Add unit tests.
+    // Assign each method arguments with default value.
+    List<Statement> bodyStatements =
+        arguments.stream()
+            .map(
+                methodArg ->
+                    ExprStatement.withExpr(
+                        assignMethodArgumentWithDefaultValue(methodArg, resourceNames)))
```
miraleung, line 73: Left this comment in subsequent PRs. We should generally reuse the VariableExprs instead of recreating them multiple times. For instance, this could be a list of Exprs instead, which we would then map into AssignmentExprs -> Statements. Then we can pass the list of Exprs into the method invocation.

Same re-creation comment applies throughout.

miraleung (changes requested):

src/main/java/com/google/api/generator/gapic/composer/ClassComposer.java
summer-ji-eng: Added. Thanks for doing that.

src/test/java/com/google/api/generator/gapic/composer/goldens/EchoClient.golden
summer-ji-eng: Will do.

src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
summer-ji-eng: Done.

src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
summer-ji-eng: emm, if we use list of Exprs as body, it makes hard for insert a CommentStament, or breakstatement in the middle.
For example, when we need a comment in the middle
```
   * try (AutoscalingPolicyServiceClient autoscalingPolicyServiceClient = AutoscalingPolicyServiceClient.create()) {
   *   RegionName parent = RegionName.of("[PROJECT]", "[REGION]");
   *   ListAutoscalingPoliciesRequest request = ListAutoscalingPoliciesRequest.newBuilder()
   *     .setParent(parent.toString())
   *     .build();
   *   while (true) {
   *     ListAutoscalingPoliciesResponse response = autoscalingPolicyServiceClient.listAutoscalingPoliciesCallable().call(request);
   *     for (AutoscalingPolicy element : response.getPoliciesList()) {
   *       // doThingsWith(element);
   *     }
   *
   *     String nextPageToken = response.getNextPageToken();
   *     if (!Strings.isNullOrEmpty(nextPageToken)) {
   *       request = request.toBuilder().setPageToken(nextPageToken).build();
   *     } else {
   *       break;
   *     }
   *   }
   * }
```
For this situation, in the WhileStatement, the composing order is assignmentExpr, ForStatement, EmptyLineStatement, AssignmentExpr, IfStatement.
How do we handle the mixed situation?

miraleung: Keep a list of statements, and add to it iteratively. Example in pseudocode:

```
VariableExpr responseVarExpr; // Construct and assign value here.
Expr responseValueExpr; // Construct and assign value here.
AssignmentExpr responseAssignExpr = 
    AssignmentExpr.builder().setVariableExpr(responseVarExpr).setValueExpr(responseValueExpr).build();

List<Statement> whileBody = new ArrayList<>();
whileBody.add(ExprStatement.withExpr(responseAssignExpr));

// Construct for loop
whileBody.add(doThingsForStatement);
whileBody.add(EmptyLineStatement.create());

// etc.
```

src/main/java/com/google/api/generator/gapic/composer/ClassComposer.java
miraleung: To say this a different way: Could we avoid adding the `resourceNames` parameter to this interface? The TODO for me to clean this up would go into the classes that need the parameter, which would now no longer implement `ClassComposer`.

src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
```diff
@@ -45,26 +54,49 @@ public static TryCatchStatement composeRpcMethodSampleCode(
...
     VariableExpr clientVarExpr = createVariableExpr(getClientName(clientType), clientType);
+    // Assign each method arguments with default value.
+    List<Expr> bodyExpr =
+        arguments.stream()
+            .map(methodArg -> assignMethodArgumentWithDefaultValue(methodArg, resourceNames))
+            .collect(Collectors.toList());
+    // Invoke current method based on return type.
+    // e.g. if return void, echoClient.echo(..); or,
+    // e.g. if return other type, EchoResponse response = echoClient.echo(...);
+    if (method.outputType().equals(TypeNode.VOID)) {
```
miraleung, line 75: PTAL at #534. The proto equivalent of `void` is `Empty`, and #534 has examples of how to detect such types.

miraleung: Also, `deleteExclusion` is a good output example to check for this case.

src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
```diff
@@ -45,26 +54,49 @@ public static TryCatchStatement composeRpcMethodSampleCode(
...
+      Method method,
+      List<MethodArgument> arguments,
+      TypeNode clientType,
+      Map<String, ResourceName> resourceNames) {
+    // TODO(summerji): Add unit tests.
     VariableExpr clientVarExpr = createVariableExpr(getClientName(clientType), clientType);
+    // Assign each method arguments with default value.
+    List<Expr> bodyExpr =
+        arguments.stream()
+            .map(methodArg -> assignMethodArgumentWithDefaultValue(methodArg, resourceNames))
```
miraleung, line 70: Note that if `assignMethodArgumentWithDefaultValue` will only ever be called from a stream, it may be more efficient to make it take `arguments` instead.

src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
```diff
@@ -124,6 +158,41 @@ private static AssignmentExpr assignClientVariableWithCreateMethodExpr(
...
+            .setReturnType(variableType)
+            .build();
+    return AssignmentExpr.builder()
+        .setVariableExpr(varExpr.toBuilder().setIsDecl(true).build())
+        .setValueExpr(clientMethodInvocationExpr)
+        .build();
+  }
+
+  private static List<Expr> mapMethodArgumentsToVariableExprs(List<MethodArgument> arguments) {
+    return arguments.stream()
```
miraleung, line 191: As previously mentioned in some reviews, this is another instance where we're creating the `VariableExpr` twice - once for the assignment of the default value, and another time for the method argument. Could we please refactor this such that we only create them once?

One way is to create the set of variables in the original method, then pass the variable into `DefaultValueComposer`. Also, we can add a `VariableExpr` variant to `DefaultValueComposer` as well.

miraleung (changes requested):

miraleung (changes requested):
Since this is no longer blocking the migration, let's add the unit tests as well.

src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
miraleung: Could you please address the variable creation in the rest of this PR as well?

src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
summer-ji-eng: Gotcha, Thanks showing me example.

src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
summer-ji-eng: sg

src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
summer-ji-eng: Thanks, very useful. Updated code.

src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
summer-ji-eng: Also update variable creation. Thanks.

src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
summer-ji-eng: Very useful. Updated the code.

src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
```diff
@@ -26,17 +28,26 @@
...
 
 public final class SampleCodeHelperComposer {
+  private static String RESPONSE = "response";
 
   public static TryCatchStatement composeRpcMethodSampleCode(
       Method method,
       List<MethodArgument> arguments,
       TypeNode clientType,
       Map<String, ResourceName> resourceNames) {
+    Preconditions.checkState(
```
summer-ji-eng, line 46: @miraleung could you give me feedback here. I am not sure. Do we need to check if the arguments pass in is empty?
Any possible for a methodSignatures contain an empty list `List<List<MethodArgument>>` ?

src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
```diff
@@ -71,14 +82,45 @@ private static TryCatchStatement composeUnaryRpcMethodSampleCode(
...
-    // TODO(summerji): compose sample code for unary rpc method.
     VariableExpr clientVarExpr = createVariableExpr(getClientName(clientType), clientType);
+    // Assign each method arguments with its default value.
+    Map<String, VariableExpr> methodArgVarExprMap = mapMethodArgumentsToVariableExprs(arguments);
+    List<Expr> methodArgumentsAssignmentExpr =
+        assignMethodArgumentsWithDefaultValues(arguments, methodArgVarExprMap, resourceNames);
+    List<Expr> methodVarExprs =
+        arguments.stream()
+            .map(arg -> methodArgVarExprMap.get(arg.name()))
+            .collect(Collectors.toList());
```
summer-ji-eng, line 93: Regarding to create variable expr once issue, for method arguments, what I did here is to create a Map<String, VariableExpr> for method arguments, where key is argument's name and value is its variable expr.
Then I fetch the values from map to make a `List<Expr>` and reused in the below session. 
What do you think of this approach. Welcome any suggestion. Thanks.

summer-ji-eng:
@miraleung Can you take another look at latest update. Mostly addressed your previous comment (thanks a lot).

Regarding to the string method signature refer to resource reference, I have created a bugizer cc you and @chingor13, I will follow up to get the feedback on default value.

src/main/java/com/google/api/generator/gapic/composer/ClassComposer.java
```diff
@@ -22,5 +22,7 @@
 
 public interface ClassComposer {
   GapicClass generate(
+      // TODO(miraleung): clean up the hierarchy to avoid pass another parameter (resourceNames is
```
miraleung, line 25: As already mentioned [here](https://github.com/googleapis/gapic-generator-java/pull/512#discussion_r528046098), please remove `resourceNames` from this interface and have the class(es) that need it not inherit from ClassComposer. The TODO on me should go into those classes, not this one.

miraleung (changes requested):

summer-ji-eng:
Address comments and refactor in #573

PR #517, opened 2020-11-18
[draft][5/5]Implement default unary rpc method's sample code

summer-ji-eng:
Default method generated from [here](https://github.com/googleapis/gapic-generator-java/blob/d3dce0538a527e0214a1a72bc27e8418b0e2905e/src/main/java/com/google/api/generator/gapic/composer/ServiceClientClassComposer.java#L474)

There are 3 kinds sample code for default unary rpc:
1. pure unary rpc's sample code
2. paged unary rpc's sample code
3. lro unary rpc's sample code.


src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
```diff
@@ -206,6 +267,37 @@ private static AssignmentExpr assignClientVariableWithCreateMethodExpr(TypeNode
...
+            .setMethodName("build")
+            .setReturnType(requestType)
+            .build();
+    return AssignmentExpr.builder()
+        .setVariableExpr(createVariableDeclExpr("request", requestType))
+        .setValueExpr(requestBuildExpr)
+        .build();
+  }
+
   private static Expr assignMethodArgumentWithDefaultValue(
```
miraleung, line 301: Can we refactor this so that we use the variable expr throughout (like one would do with Java programming) instead of recreating it several times?

src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
```diff
@@ -174,18 +175,78 @@ private static TryCatchStatement composePagedUnaryRpcMethodSampleCode(
...
+                        assignMethodArgumentWithDefaultValue(methodArg, resourceNames)))
+            .collect(Collectors.toList());
+    // Assign request variables with set argument attributes.
+    // e.g EchoRequest
+    bodyStatements.add(
+        ExprStatement.withExpr(createRequestBuilderExpr(method.inputType(), arguments)));
+
+    if (method.isPaged()) {
+      // For loop on invoke client method's iterator with comment.
+      // e.g. for (EchoClient echoClient : echoClient.PagedExpand(request).iterateAll()) {//..}
```
miraleung, line 201: This line looks off, I'm not sure that `iterateAll` on that method would return a list of `echoClient`s.

miraleung (changes requested):

summer-ji-eng:
refactor in #594

PR #519, opened 2020-11-18
[draft][2/7]Implement stream.client rpc callable sample code

summer-ji-eng:
```
   * try (EchoClient echoClient = EchoClient.create()) {
   *   ApiStreamObserver<EchoResponse> responseObserver =
   *       new ApiStreamObserver<EchoResponse>() {
   *         @Override
   *         public void onNext(EchoResponse response) {
   *           // Do something when receive a response.
   *         }
   *
   *         @Override
   *         public void onError(Throwable t) {
   *           // Add error-handling
   *         }
   *
   *         @Override
   *         public void onCompleted() {
   *           // Do something when complete.
   *         }
   *       };
   *   ApiStreamObserver<EchoRequest> requestObserver =
   *       echoClient.collectCallable().clientStreamingCall(responseObserver);
   *
   *   String content = "content951530617";
   *   EchoRequest request = EchoRequest.newBuilder().setContent(content).build();
   *   requestObserver.onNext(request);
   * }
```

TODO: Add escape for `@` in `<pre>{@code ..}</pre>`

### DIFF
--- a/src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
+++ b/src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
@@ -14,13 +14,19 @@
 
 package com.google.api.generator.gapic.composer;
 
+import com.google.api.gax.rpc.ApiStreamObserver;
+import com.google.api.generator.engine.ast.AnonymousClassExpr;
 import com.google.api.generator.engine.ast.AssignmentExpr;
 import com.google.api.generator.engine.ast.CommentStatement;
+import com.google.api.generator.engine.ast.ConcreteReference;
+import com.google.api.generator.engine.ast.EmptyLineStatement;
 import com.google.api.generator.engine.ast.Expr;
 import com.google.api.generator.engine.ast.ExprStatement;
 import com.google.api.generator.engine.ast.ForStatement;
 import com.google.api.generator.engine.ast.LineComment;
+import com.google.api.generator.engine.ast.MethodDefinition;
 import com.google.api.generator.engine.ast.MethodInvocationExpr;
+import com.google.api.generator.engine.ast.ScopeNode;
 import com.google.api.generator.engine.ast.Statement;
 import com.google.api.generator.engine.ast.TryCatchStatement;
 import com.google.api.generator.engine.ast.TypeNode;
@@ -31,6 +37,7 @@ import com.google.api.generator.gapic.model.Method.Stream;
 import com.google.api.generator.gapic.model.MethodArgument;
 import com.google.api.generator.gapic.model.ResourceName;
 import com.google.api.generator.gapic.utils.JavaStyle;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
```diff
@@ -284,13 +293,46 @@ private static TryCatchStatement composeStreamRpcCallableMethodSampleCode(
...
+        MethodInvocationExpr.builder()
+            .setExprReferenceExpr(createStreamObserverVarExpr(method, false))
+            .setMethodName("onNext")
+            .setArguments(createVariableExpr(REQUEST_VAR_NAME, method.inputType()))
+            .build();
+
+    List<Statement> bodyStatements = new ArrayList<>();
+    // Create StreamObserver response variable with anonymous class.
+    bodyStatements.add(
+        ExprStatement.withExpr(createAssignStreamObserverResponseAnonymousClassExpr(method)));
```
miraleung, line 314: This is currently creating the variable twice: Once here, and once in wcreateAssignStreamObserverRequestExpr`. In general, we should be treating `VariableExpr`s like real, Java variables. Could we refactor this so that it is only created once, and passed around?

src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
```diff
@@ -284,13 +293,46 @@ private static TryCatchStatement composeStreamRpcCallableMethodSampleCode(
...
+    // If variant method signatures exists, use the first one.
+    List<MethodArgument> arguments =
+        method.methodSignatures().isEmpty()
+            ? Collections.emptyList()
+            : method.methodSignatures().get(0);
+    MethodInvocationExpr onNextRequestMethodExpr =
+        MethodInvocationExpr.builder()
+            .setExprReferenceExpr(createStreamObserverVarExpr(method, false))
+            .setMethodName("onNext")
+            .setArguments(createVariableExpr(REQUEST_VAR_NAME, method.inputType()))
```
miraleung, line 308: Same issue here as below, let's create just one request variable and pass it around.

src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
```diff
@@ -468,6 +592,14 @@ private static String getLroMethodName(String methodName) {
     return JavaStyle.toLowerCamelCase(String.format(ASYNC_NAME_PATTERN, methodName));
   }
 
+  private static String getObserverVariableName(String variableName) {
```
miraleung, line 595: This is unnecessary since there are only two static names. Can we use `REQUEST_OBSERVER_VAR_NAME` and `RESPNOSE_OBSERVER_VAR_NAME` instead?

src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
miraleung: For these types of locally-scoped variables, I would prefer using a local String instead, since (a) it is used only in this method, (b) it helps enforce a single point of creation, and (c) a local String won't be more expensive than the global constant.

src/main/java/com/google/api/generator/gapic/composer/SampleCodeHelperComposer.java
miraleung: Also, PTAL at other such issues in this PR, and the other PRs as well. Will look at the downstream PRs when this has been addressed.

miraleung (changes requested):

summer-ji-eng:
refactor in #635

PR #569, opened 2020-11-25
[samplecode][Prepare][1/2]Add ResourceName map into ClassComposer prepare for the samplecode implementation

summer-ji-eng:
In order to compose default value in the sample code. We require ResourceName maps pass down to the SampleCodeComposer. See the reference at #573 [use case](https://github.com/googleapis/gapic-generator-java/pull/573/files#diff-130316a1e625f84e85d173ec57d1c698a88af9764638206b004cff4347c7ab9fR40)

### DIFF
--- a/src/main/java/com/google/api/generator/gapic/composer/ClassComposer.java
+++ b/src/main/java/com/google/api/generator/gapic/composer/ClassComposer.java
@@ -16,13 +16,9 @@ package com.google.api.generator.gapic.composer;
 
 import com.google.api.generator.gapic.model.GapicClass;
 import com.google.api.generator.gapic.model.Message;
-import com.google.api.generator.gapic.model.ResourceName;
 import com.google.api.generator.gapic.model.Service;
 import java.util.Map;
 
 public interface ClassComposer {
-  // TODO(b/174257081): Clean up the hierarchy to avoid pass another parameter (resourceNames is
-  // only used for composing sample code).
-  GapicClass generate(
-      Service service, Map<String, Message> messageTypes, Map<String, ResourceName> resourceNames);
+  GapicClass generate(Service service, Map<String, Message> messageTypes);
 }
--- a/src/main/java/com/google/api/generator/gapic/composer/ClassComposer.java
+++ b/src/main/java/com/google/api/generator/gapic/composer/ClassComposer.java
@@ -16,9 +16,13 @@ package com.google.api.generator.gapic.composer;
 
 import com.google.api.generator.gapic.model.GapicClass;
 import com.google.api.generator.gapic.model.Message;
+import com.google.api.generator.gapic.model.ResourceName;
 import com.google.api.generator.gapic.model.Service;
 import java.util.Map;
 
 public interface ClassComposer {
-  GapicClass generate(Service service, Map<String, Message> messageTypes);
+  // TODO(b/174257081): Clean up the hierarchy to avoid pass another parameter (resourceNames is
+  // only used for composing sample code).
+  GapicClass generate(
+      Service service, Map<String, Message> messageTypes, Map<String, ResourceName> resourceNames);
 }
--- a/src/main/java/com/google/api/generator/gapic/composer/ServiceClientClassComposer.java
+++ b/src/main/java/com/google/api/generator/gapic/composer/ServiceClientClassComposer.java
@@ -62,6 +62,7 @@ import com.google.api.generator.gapic.model.Message;
 import com.google.api.generator.gapic.model.Method;
 import com.google.api.generator.gapic.model.Method.Stream;
 import com.google.api.generator.gapic.model.MethodArgument;
+import com.google.api.generator.gapic.model.ResourceName;
 import com.google.api.generator.gapic.model.Service;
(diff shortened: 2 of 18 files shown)


src/main/java/com/google/api/generator/gapic/composer/ClassComposer.java
```diff
@@ -16,9 +16,13 @@
...
+import com.google.api.generator.gapic.model.ResourceName;
 import com.google.api.generator.gapic.model.Service;
 import java.util.Map;
 
 public interface ClassComposer {
-  GapicClass generate(Service service, Map<String, Message> messageTypes);
+  // TODO(b/174257081): Clean up the hierarchy to avoid pass another parameter (resourceNames is
+  // only used for composing sample code).
+  GapicClass generate(
+      Service service, Map<String, Message> messageTypes, Map<String, ResourceName> resourceNames);
```
miraleung, line 27: As mentioned [here](https://github.com/googleapis/gapic-generator-java/pull/512#discussion_r530645801) and [here](https://github.com/googleapis/gapic-generator-java/pull/512#discussion_r528046098), let's not put `resourceNames` in the interface.

miraleung (changes requested):

src/main/java/com/google/api/generator/gapic/composer/ClassComposer.java
summer-ji-eng: Remove the resource name from `ClassComposer` interface.
Move `TODO` under `GapicClass`

Used `TODO(username)` instead.

src/main/java/com/google/api/generator/gapic/composer/ServiceClientClassComposer.java
```diff
@@ -108,8 +109,10 @@ public static ServiceClientClassComposer instance() {
     return INSTANCE;
   }
 
-  @Override
-  public GapicClass generate(Service service, Map<String, Message> messageTypes) {
+  public GapicClass generate(
+      Service service, Map<String, Message> messageTypes, Map<String, ResourceName> resourceNames) {
+    // TODO(miraleung): Clean up the hierarchy to avoid pass another parameter (resourceNames is
+    // only used for composing sample code). See b/174257081.
```
miraleung, line 115: As noted in #512, could we please keep this public codebase clean of internal bug numbers? While we're here, could we please update the bug's content as well?

miraleung (changes requested):

src/main/java/com/google/api/generator/gapic/composer/ServiceClientClassComposer.java
summer-ji-eng: sg

miraleung (changes requested):
Is this PR still needed, since we're no longer adding the map to the interface?

summer-ji-eng:
> Is this PR still needed, since we're no longer adding the map to the interface?

In my opinion, yes. This PR is mainly for add `resourceNames` map into `GapicClass` where it is prepare work for the rest implementation.

src/main/java/com/google/api/generator/gapic/composer/ServiceClientClassComposer.java
```diff
@@ -108,8 +109,10 @@ public static ServiceClientClassComposer instance() {
     return INSTANCE;
   }
 
-  @Override
-  public GapicClass generate(Service service, Map<String, Message> messageTypes) {
+  public GapicClass generate(
+      Service service, Map<String, Message> messageTypes, Map<String, ResourceName> resourceNames) {
+    // TODO(miraleung): Clean up the hierarchy to avoid pass another parameter (resourceNames is
```
miraleung, line 114: Could we please move this TODO to the class level, since that's where the `implements` would go?

miraleung (changes requested):
Could we please change the PR's title? Also, can we include this changes with the PR that needs it, since the motivation for this change will not be clear to future readers?

src/main/java/com/google/api/generator/gapic/composer/ServiceClientClassComposer.java
summer-ji-eng: Sure.

summer-ji-eng:
> Could we please change the PR's title? Also, can we include this changes with the PR that needs it, since the motivation for this change will not be clear to future readers?

I changed the PR's title. But feel free to change if you have better naming. Thanks a lot.

miraleung (changes requested):
> Also, can we include this changes with the PR that needs it, since the motivation for this change will not be clear to future readers?

Could we merge this into the PR that actually uses the resource name map? Would prefer this since the refactoring is relatively straightforward, and the usage will help provide the appropriate motivation  for future readers.

summer-ji-eng:
> > Also, can we include this changes with the PR that needs it, since the motivation for this change will not be clear to future readers?
> 
> Could we merge this into the PR that actually uses the resource name map? Would prefer this since the refactoring is relatively straightforward, and the usage will help provide the appropriate motivation for future readers.

All right. Close this PR, #573 will cover this. Thanks.

PR #573, opened 2020-11-26
[samplecode][1/3]Implement Pure Unary RPC sample code

summer-ji-eng:
`createMethodVariant` -> sample code and pass into the `ServiceClientCommentComposer`

unit tests: MethodSampleCodeHelperComposerTest

integration test: Echo.proto, Identity.proto and golden files in /test/integration`


src/main/java/com/google/api/generator/gapic/composer/MethodSampleCodeHelperComposer.java
```diff
@@ -0,0 +1,167 @@
...
+import com.google.api.generator.gapic.model.MethodArgument;
+import com.google.api.generator.gapic.model.ResourceName;
+import com.google.api.generator.gapic.utils.JavaStyle;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class MethodSampleCodeHelperComposer {
+  private static String RESPONSE = "response";
```
miraleung, line 35: This can just be a local literal, since it's only used once.

src/main/java/com/google/api/generator/gapic/composer/MethodSampleCodeHelperComposer.java
```diff
@@ -0,0 +1,167 @@
...
+
+  public static TryCatchStatement composeUnaryRpcMethodSampleCode(
+      Method method,
+      List<MethodArgument> arguments,
+      TypeNode clientType,
+      Map<String, ResourceName> resourceNames) {
+    VariableExpr clientVarExpr = createVariableExpr(getClientName(clientType), clientType);
+    // Assign each method arguments with its default value.
+    Map<String, VariableExpr> methodArgVarExprMap = createMethodArgumentsVariableExprs(arguments);
```
miraleung, line 44: This seems to be doing a lot of extra work: (a) creating the VarExprs from the method args, (b) coupling the ordering of the VarExprs and MethodArguments to create AssignExprs, and then (c) gathering the VarExprs from the result. 

Instead, can we do the following, which would be much simpler:
- Add a VarExpr variant to DefaultValueComposer (and tests)
- Do (a), but in a one-liner not inside the wrapper method, and into a List
- Use the new method in DefaultValueComposer to do (b) directly, in a one-liner that's not wrapped in a method

Then the list of VarExprs can be passed directly into the method invocation builder.

src/main/java/com/google/api/generator/gapic/composer/MethodSampleCodeHelperComposer.java
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  private static AssignmentExpr createAssignmentExpr(VariableExpr variableExpr, Expr valueExpr) {
+    return AssignmentExpr.builder()
+        .setVariableExpr(variableExpr.toBuilder().setIsDecl(true).build())
+        .setValueExpr(valueExpr)
+        .build();
+  }
+
+  private static VariableExpr createVariableExpr(String variableName, TypeNode type) {
```
miraleung, line 162: I'd prefer not to have this in the helper, even if it's used in other methods in future PRs, because (a) the logic is straightforward, and (b) it helps future readers see immediately what/how the VariableExpr is being created. The other, more complex settings for VariableExpr are also the reason we don't have a simpler wrapper in that class.

src/main/java/com/google/api/generator/gapic/composer/MethodSampleCodeHelperComposer.java
```diff
@@ -0,0 +1,167 @@
...
+  private static String getClientName(TypeNode clientType) {
+    return JavaStyle.toLowerCamelCase(clientType.reference().name());
+  }
+
+  private static boolean isProtoEmptyType(TypeNode type) {
+    return type.reference().pakkage().equals("com.google.protobuf")
+        && type.reference().name().equals("Empty");
+  }
+
+  private static AssignmentExpr createAssignmentExpr(VariableExpr variableExpr, Expr valueExpr) {
```
miraleung, line 155: Same as the comment below.

src/main/java/com/google/api/generator/gapic/composer/MethodSampleCodeHelperComposer.java
```diff
@@ -0,0 +1,167 @@
...
+            .setArguments(argumentsVarExprs)
+            .setReturnType(variableExpr.variable().type())
+            .build();
+    return AssignmentExpr.builder()
+        .setVariableExpr(variableExpr.toBuilder().setIsDecl(true).build())
+        .setValueExpr(clientMethodInvocationExpr)
+        .build();
+  }
+
+  private static String getClientName(TypeNode clientType) {
```
miraleung, line 146: This wouldn't be the client name, since any type could be passed in, and it's simply converting that to lowercase. Could we inline this logic instead of adding a helper method, since it's straightforward?

src/main/java/com/google/api/generator/gapic/composer/MethodSampleCodeHelperComposer.java
```diff
@@ -0,0 +1,167 @@
...
+        assignMethodArgumentsWithDefaultValues(arguments, methodArgVarExprMap, resourceNames);
+    List<Expr> methodVarExprs =
+        arguments.stream()
+            .map(arg -> methodArgVarExprMap.get(arg.name()))
+            .collect(Collectors.toList());
+    // Invoke current method based on return type.
+    // e.g. if return void, echoClient.echo(..); or,
+    // e.g. if return other type, EchoResponse response = echoClient.echo(...);
+    boolean returnsVoid = isProtoEmptyType(method.outputType());
+    Expr responseExpr = null;
```
miraleung, line 55: This can be removed, please see the next comment.

src/main/java/com/google/api/generator/gapic/composer/MethodSampleCodeHelperComposer.java
```diff
@@ -0,0 +1,167 @@
...
+      responseExpr =
+          createAssignExprForVariableWithClientMethod(
+              createVariableExpr(RESPONSE, method.outputType()),
+              clientVarExpr,
+              JavaStyle.toLowerCamelCase(method.name()),
+              methodVarExprs);
+    }
+
+    List<Expr> bodyExpr = new ArrayList<>();
+    bodyExpr.addAll(methodArgumentsAssignmentExpr);
```
miraleung, line 74: Can we move this right after the list of assignment exprs are defined, then we can add the responseExprs directly instead of assigning to a null value first?

src/main/java/com/google/api/generator/gapic/composer/ServiceClientCommentComposer.java
```diff
@@ -162,16 +162,18 @@ static CommentStatement createCreateMethodStubArgComment(
...
     JavaDocComment.Builder methodJavadocBuilder = JavaDocComment.builder();
 
     if (method.hasDescription()) {
       methodJavadocBuilder =
           processProtobufComment(method.description(), methodJavadocBuilder, null);
     }
 
-    // methodJavadocBuilder.addParagraph(METHOD_DESCRIPTION_SAMPLE_CODE_SUMMARY_STRING);
-    // TODO(summerji): Add sample code here and uncomment the above.
+    if (!sampleCode.isEmpty()) {
```
miraleung, line 173: Prefer using `Strings.isNullOrEmpty` instead, please see existing examples in the codebase.

src/main/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposer.java
```diff
@@ -161,6 +165,16 @@ public static String composeClassHeaderEndpointSampleCode(
             ExprStatement.withExpr(initClientVarExpr)));
   }
 
+  public static String composeRpcMethodHeaderSampleCode(
+      Method method,
+      List<MethodArgument> arguments,
+      TypeNode clientType,
+      Map<String, ResourceName> resourceNames) {
+    return SampleCodeWriter.write(
+        MethodSampleCodeHelperComposer.composeUnaryRpcMethodSampleCode(
```
miraleung, line 174: Could we please inline this logic into this class since it's targeted only for the ServiceClient, rather than creating a separate MethodSampleCodeHelperComposer class?

src/test/java/com/google/api/generator/gapic/composer/BUILD.bazel
```diff
@@ -38,6 +39,7 @@ TEST_DEPS = [
     "//src/main/java/com/google/api/generator/gapic/protoparser",
     "//src/test/java/com/google/api/generator/gapic/testdata:showcase_java_proto",
     "//src/test/java/com/google/api/generator/gapic/testdata:testgapic_java_proto",
+    "@com_google_api_api_common//jar",
```
miraleung, line 42: What is this new dep used for?

src/test/java/com/google/api/generator/gapic/composer/MethodSampleCodeHelperComposerTest.java
```diff
@@ -0,0 +1,454 @@
...
+import com.google.api.generator.gapic.model.ResourceReference;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+public class MethodSampleCodeHelperComposerTest {
```
miraleung, line 37: Will review this once the other changes are done.

miraleung (changes requested):

src/main/java/com/google/api/generator/gapic/composer/MethodSampleCodeHelperComposer.java
summer-ji-eng: This is saved for future PR use.

src/main/java/com/google/api/generator/gapic/composer/MethodSampleCodeHelperComposer.java
summer-ji-eng: Not sure if I am understand correctly.
`Add a VarExpr variant to DefaultValueComposer (and tests)` -> if varExpr refer to a method argument variable expr, it only contains method argument's name and type, however to get a default value it require the method argument's `field`. I guess I didn't get the point of add VarExpr into DefaultValueComposer. Could you explain it more with details? Thanks

src/main/java/com/google/api/generator/gapic/composer/MethodSampleCodeHelperComposer.java
summer-ji-eng: Okay.

src/main/java/com/google/api/generator/gapic/composer/MethodSampleCodeHelperComposer.java
summer-ji-eng: Same as `createVariableExpr`

src/main/java/com/google/api/generator/gapic/composer/MethodSampleCodeHelperComposer.java
summer-ji-eng: Similar to `createVariableExpr`
Flatten the code
```
    VariableExpr clientVarExpr =
        VariableExpr.withVariable(
            Variable.builder()
                .setName(JavaStyle.toLowerCamelCase(clientType.reference().name()))
                .setType(clientType)
                .build());
```
vs.
current code
```
VariableExpr clientVarExpr = createVariableExpr(getClientName(clientType), clientType);
```
Why not just use wrapper method where indicate the function to make the code easy to read?

src/main/java/com/google/api/generator/gapic/composer/MethodSampleCodeHelperComposer.java
summer-ji-eng: Sure

src/main/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposer.java
summer-ji-eng: The purpose of creating a separate `MethodSampleCodeHelperComposer` is to make it reusable and isolate from `ServiceClientSampleCodeCompose`. There are three callers:
1. Service Client Methods --> MethodSampleCodeHelperComposer
2. Service Client Class --> MethodSampleCodeHelperComposer
3. PackgeInfoComposer -> MethodSampleCodeHelperComposer

src/test/java/com/google/api/generator/gapic/composer/MethodSampleCodeHelperComposerTest.java
summer-ji-eng: `ConcreteReference.withClazz(com.google.api.resourcenames.ResourceName.class)`
This is used in unit tests. To create resource name  type.

src/main/java/com/google/api/generator/gapic/composer/MethodSampleCodeHelperComposer.java
miraleung: PTAL at the `field` fields that DefaultValueComposer is actually using, and refactor them into separate methods (making any existing ones empty wrappers instead). For example:
- `private Expr createDefaultValue(TypeNode type, String varName)` would replace any non-`Field`-specific logic in the current `createDefaultValue(Field f)` method.
- `private Expr createDefaultValue(VariableExpr variableExpr)`
- `createDefaultValue(Field f)` would call the first one where possible.

src/main/java/com/google/api/generator/gapic/composer/MethodSampleCodeHelperComposer.java
miraleung: Let's stick with using the full builder expression, because the reader would need to look elsewhere to see how `createVariableExpr` is setting up the variable. It's not immediately clear whether it has a decl, or a reference expression, etc. 

See also information hiding, or Law of Demeter: [article 1 (same principle applied to methods)](https://understandlegacycode.com/blog/are-wrapper-functions-bad-design/), [article 2](https://dzone.com/articles/the-genius-of-the-law-of-demeter)

src/main/java/com/google/api/generator/gapic/composer/MethodSampleCodeHelperComposer.java
miraleung: PTAL at the information hiding concerns above.

src/main/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposer.java
miraleung: Wouldn't the call site of the ServiceClient's methods and class sample code be the same (i.e. ServiceClientClassComposer)?

It would be fine for PackageInfoComposer to call ServiceClientSampleCodeComposer directly, since it is really using that class's sample code.

src/test/java/com/google/api/generator/gapic/composer/MethodSampleCodeHelperComposerTest.java
miraleung: That line is probably fine, but I'll review this unit test once the other changes are addressed.

src/main/java/com/google/api/generator/gapic/composer/MethodSampleCodeHelperComposer.java
summer-ji-eng: Thanks for sharing your thought. SG

src/main/java/com/google/api/generator/gapic/composer/MethodSampleCodeHelperComposer.java
summer-ji-eng: SG

src/main/java/com/google/api/generator/gapic/composer/MethodSampleCodeHelperComposer.java
summer-ji-eng: SG

src/main/java/com/google/api/generator/gapic/composer/ServiceClientCommentComposer.java
summer-ji-eng: SG

src/test/java/com/google/api/generator/gapic/composer/BUILD.bazel
summer-ji-eng: `ConcreteReference.withClazz(com.google.api.resourcenames.ResourceName.class)`
This is used in unit tests. To create resource name type.

src/main/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposer.java
summer-ji-eng: SG

src/main/java/com/google/api/generator/gapic/composer/MethodSampleCodeHelperComposer.java
summer-ji-eng: Instead of touching up the `DefaultValueComposer`, I propose other process:
1. create list of method argument variable, because the index of this list is same as list of a method arguments,
2. Use for-loop on the index to get default value expr and its variable expr. If the argument is String-type has reference name, we create new variable expr for it. And it is only place we need to use.
Please see `createMethodArgumentsVariableExprs` for more details.

Looking forward your suggestion on this. Thanks

src/main/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposer.java
```diff
@@ -169,4 +179,169 @@ public static String composeClassHeaderEndpointSampleCode(
...
+                    .setMethodName("toString")
+                    .build();
+              }
+              return argVarExpr;
+            })
+        .collect(Collectors.toList());
+  }
+
+  // Return a list of AssignmentExpr for method argument with its default value.
+  private static List<Expr> assignMethodArgumentsWithDefaultValues(
```
miraleung, line 284: After addressing the next comment, let's rename this method to be more representative of its behavior. That is, it takes only a list of VarExprs and does not need to know anything about the `MethodArgument` type.

src/main/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposer.java
```diff
@@ -169,4 +179,169 @@ public static String composeClassHeaderEndpointSampleCode(
...
+      List<MethodArgument> arguments,
```
miraleung, line 285: We should avoid this kind of coupling (i.e. `arguments` and `argVarExprs`) where possible. Instead, process the list of `argVarExprs`  (making any conversions separately) before calling this method, then we won't need to pass in `arguments` at all.

src/main/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposer.java
```diff
@@ -169,4 +179,169 @@ public static String composeClassHeaderEndpointSampleCode(
...
+      assignmentExprs.add(
+          AssignmentExpr.builder()
+              .setVariableExpr(argVarExpr.toBuilder().setIsDecl(true).build())
+              .setValueExpr(defaultValueExpr)
+              .build());
+    }
+    return assignmentExprs;
+  }
+
+  private static Expr createAssignExprForVariableWithClientMethod(
```
miraleung, line 325: Can we inline this into the call site since there's only one usage? If we need to reuse it in a future PR, let's add this method only then, as that will help us pick a more generic name.

src/test/java/com/google/api/generator/gapic/composer/BUILD.bazel
miraleung: Thanks for clarifying.

src/main/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposer.java
```diff
@@ -169,4 +179,169 @@ public static String composeClassHeaderEndpointSampleCode(
...
+            MethodInvocationExpr.builder()
+                .setStaticReferenceType(clientVarExpr.variable().type())
+                .setReturnType(clientVarExpr.variable().type())
+                .setMethodName("create")
+                .build())
+        .build();
+  }
+
+  // Create a list of Expr for method argument variable expression.
+  private static List<Expr> createMethodArgVarExprs(
```
miraleung, line 258: Can we rename this method to more specifically represent its summarized behavior? E.g. createClientMethodArgVarExprs or createRpcMethodArgVarExprs.

miraleung (changes requested):

src/main/java/com/google/api/generator/gapic/composer/ServiceClientClassComposer.java
```diff
@@ -490,7 +498,11 @@ private static boolean hasLroMethods(Service service) {
   }
 
   private static List<MethodDefinition> createMethodVariants(
-      Method method, Map<String, Message> messageTypes, Map<String, TypeNode> types) {
+      Method method,
+      Map<String, Message> messageTypes,
+      Map<String, TypeNode> types,
+      String clientName,
```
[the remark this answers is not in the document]
summer-ji-eng: SG

src/main/java/com/google/api/generator/gapic/composer/ServiceClientClassComposer.java
```diff
@@ -471,11 +474,16 @@ private static boolean hasLroMethods(Service service) {
   }
 
   private static List<MethodDefinition> createServiceMethods(
-      Service service, Map<String, Message> messageTypes, Map<String, TypeNode> types) {
+      Service service,
+      Map<String, Message> messageTypes,
+      Map<String, TypeNode> types,
+      Map<String, ResourceName> resourceNames) {
     List<MethodDefinition> javaMethods = new ArrayList<>();
+    String clientName = getClientClassName(service);
```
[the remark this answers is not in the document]
summer-ji-eng: Sure.

src/main/java/com/google/api/generator/gapic/composer/MethodSampleCodeHelperComposer.java
miraleung: We'd like to avoid the index-based list coupling if we can. What is the motivation for going about it this way, rather than refactoring DefaultValueComposer?

src/main/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposer.java
```diff
@@ -169,4 +179,169 @@ public static String composeClassHeaderEndpointSampleCode(
...
+      TypeNode clientType,
+      Map<String, ResourceName> resourceNames) {
+    VariableExpr clientVarExpr =
+        VariableExpr.withVariable(
+            Variable.builder()
+                .setName(JavaStyle.toLowerCamelCase(clientType.reference().name()))
+                .setType(clientType)
+                .build());
+    // Assign each method arguments with its default value.
+    List<Expr> methodArgExprs = createMethodArgVarExprs(arguments, resourceNames);
```
miraleung, line 205: Here's another approach to doing this. We replace the index-based coupling between the list of MethodArguments and VariableExprs, and instead use a coupling that makes more semantic sense - those of the AssignmentExprs. We’ll also streamline the ordering, and rely less on circumstantial checks. Let's put aside the method wrapping and start by focusing on the core logic that needs to happen.

```java
List<Expr> rpcMethodArgVarExprs = arguments.stream()
        .map(arg -> 
             VariableExpr.withVariable(Variable.builder()
                          .setName(JavaStyle.toLowerCamelCase(arg.name()))
                          .setType(arg.type())
                          .build()))
List<Expr> rpcMethodArgValueExprs = arguments.stream()
    .map(arg ->
          isStringTypedResourceName(arg) 
          ? DefaultValueComposer.createDefaultValue(
                 resourceNames.get(arg.field().resourceReference().resourceTypeString()))
          : DefaultValueComposer.createDefaultValue(arg, resourceNames))

List<Expr> bodyExprs = new ArrayList<>();
Preconditions.checkState(rpcMethodArgVarExprs.size() == rpcMethodArgValueExprs.size(), “Error message here”);
bodyExprs.addAll(
    IntStream.range(0, rpcMethodArgVarExprs.size())
    .mapToObj(i -> AssignmentExpr.builder().setVariableExpr(rpcMethodArgVarExprs.get(i)).setValueExpr(rpcMethodArgValueExprs.get(i)).build())
    
```

miraleung (changes requested):

src/main/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposer.java
miraleung: Meant to also add this:

```java
private boolean isStringTypedResourceName(MethodArgument arg, 
    return arg.type().equals(TypeNode.STRING)
        && arg.field().hasResourceReference()
        && resourceNames.containsKey(
        arg.field().resourceReference().resourceTypeString());
}
```

summer-ji-eng:
Many thanks on discussion of default value expr for method arguments in the chat.

src/main/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposer.java
summer-ji-eng: removed this method.

src/main/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposer.java
summer-ji-eng: Many thanks on discussion of default value expr for method arguments in the chat.
SG.

xiaozhenliu-gg5 (approved):

src/test/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposerTest.java
```diff
@@ -0,0 +1,452 @@
...
+            .setMethodSignatures(signatures)
+            .setInputType(inputType)
+            .setOutputType(outputType)
+            .build();
+    String results =
+        SampleCodeWriter.write(
+            ServiceClientSampleCodeComposer.composeUnaryRpcMethodSampleCode(
+                unaryMethod, signatures.get(0), clientType, resourceNames));
+    String expected =
+        "try (EchoClient echoClient = EchoClient.create()) {\n"
```
miraleung, line 91: Instead of the newline chars, could we please use [`LineFormatter.lines()`](https://github.com/googleapis/gapic-generator-java/blob/master/src/test/java/com/google/api/generator/testutils/LineFormatter.java#L18)? Same elsewhere in this file; example usages are in other tests.

src/test/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposerTest.java
```diff
@@ -0,0 +1,452 @@
...
+            ServiceClientSampleCodeComposer.composeUnaryRpcMethodSampleCode(
+                unaryMethod, methodSignatures.get(0), clientType, resourceNames));
+    String expected =
+        "try (EchoClient echoClient = EchoClient.create()) {\n"
+            + "  String name = \"name3373707\";\n"
+            + "  echoClient.delete(name);\n"
+            + "}";
+    Assert.assertEquals(results, expected);
+  }
+}
```
miraleung, line 452: One more obscure edge case to test: The case where the method has no arguments. 

I recently learned this case actually exists. Please see [cloud_billing.proto](https://github.com/googleapis/googleapis/blob/master/google/cloud/billing/v1/cloud_billing.proto#L55) and the relevant client class in [java-billing](https://github.com/googleapis/java-billing).

test/integration/goldens/library/LibraryServiceClient.java
```diff
@@ -404,6 +479,16 @@ public final Shelf mergeShelves(String name, ShelfName otherShelfName) {
    * <p>Returns NOT_FOUND if either shelf does not exist. This call is a no-op if the specified
    * shelves are the same.
    *
+   * <p>Sample code:
+   *
+   * <pre>{@code
+   * try (LibraryServiceClient libraryServiceClient = LibraryServiceClient.create()) {
+   *   String name = "name3373707";
+   *   String otherShelfName = "other_shelf_name145746959";
```
miraleung, line 487: FYI this will be fixed in #585 (lower_snake_case -> lowerCamelCase).

miraleung (changes requested):

test/integration/goldens/library/LibraryServiceClient.java
summer-ji-eng: Nice. Approved #585

src/main/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposer.java
```diff
@@ -169,4 +180,140 @@ public static String composeClassHeaderEndpointSampleCode(
...
+                                    resourceNameList,
+                                    arg.field().name()))
+                            .setMethodName("toString")
+                            .setReturnType(TypeNode.STRING)
+                            .build())
+            .collect(Collectors.toList());
+
+    List<Expr> bodyExprs = new ArrayList<>();
+    Preconditions.checkState(
+        rpcMethodArgVarExprs.size() == rpcMethodArgDefaultValueExprs.size(), "Error message here");
```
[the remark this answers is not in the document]
summer-ji-eng: Updated.

src/main/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposer.java
```diff
@@ -169,4 +180,140 @@ public static String composeClassHeaderEndpointSampleCode(
...
+      TypeNode clientType,
+      Map<String, ResourceName> resourceNames) {
+    VariableExpr clientVarExpr =
+        VariableExpr.withVariable(
+            Variable.builder()
+                .setName(JavaStyle.toLowerCamelCase(clientType.reference().name()))
+                .setType(clientType)
+                .build());
+    // List of rpc method arguments' variable expressions.
+    List<Expr> rpcMethodArgVarExprs =
```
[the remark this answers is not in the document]
summer-ji-eng: The reason I use `List<VarialbeExpr>` here is `rpcMethodArgVarExprs` is used to pass in as `List<Expr>` for `MethodInvocationExpr.setArguments`  in other two places.  See [line262](https://github.com/googleapis/gapic-generator-java/pull/573/files#diff-7f479f381fc848c9556f6d57565536498b3f517034d769e0d36ad4d5136f13f3R262) and [line273](https://github.com/googleapis/gapic-generator-java/pull/573/files#diff-7f479f381fc848c9556f6d57565536498b3f517034d769e0d36ad4d5136f13f3R273).
And L246 is good place to cast as variable expr because it is the place need for-loop on each element. Otherwise, we need create another for-loop to cast expr to variable expr somewhere.

src/main/java/com/google/api/generator/gapic/model/GapicClass.java
```diff
@@ -19,6 +19,8 @@
 
 @AutoValue
 public abstract class GapicClass {
+  // TODO(miraleung): Clean up the hierarchy to avoid pass another parameter (resourceNames is
```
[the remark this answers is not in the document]
summer-ji-eng: Cool. Removed.

src/test/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposerTest.java
summer-ji-eng: Thanks for pointing out. Updated the code.

src/test/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposerTest.java
summer-ji-eng: Good to know. Added this edge case.

src/main/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposer.java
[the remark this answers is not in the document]
miraleung: Good point. In this case, let's use `List<VariableExpr>` and remove the cast, then in MethodArgument, use a stream to cast back up to Expr. The reason is that casting wider is safer - `(Expr) VariableExpr` will always be OK, but `(VariableExpr) Expr` is not, since the `Expr` could be anything else.

src/test/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposerTest.java
```diff
@@ -0,0 +1,518 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+public class ServiceClientSampleCodeComposerTest {
+  private static final String PACKAGE_NAME = "com.google.showcase.v1beta1";
+  private static final TypeNode clientType =
```
miraleung, line 43: Nit: Would prefer to have this local to each test. Code reuse for tests generally follows slightly different principles than for the core implementation. Since every test should be localized, certain "common" test-dependent variables (where the output depends on the variable's value) are better created inside the test for locality and readability.

miraleung (changes requested):
LGTM with comments addressed.

src/main/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposer.java
```diff
@@ -169,4 +181,142 @@ public static String composeClassHeaderEndpointSampleCode(
             ExprStatement.withExpr(initSettingsVarExpr),
             ExprStatement.withExpr(initClientVarExpr)));
   }
+
+  public static String composeRpcMethodHeaderSampleCode(
+      Method method,
```
miraleung, line 186: Nit: Consider adjusting the parameter order to be consistent with the chosen order in #576.

miraleung (changes requested):

src/main/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposer.java
```diff
@@ -169,4 +181,142 @@ public static String composeClassHeaderEndpointSampleCode(
...
+                                    arg.field().name()))
+                            .setMethodName("toString")
+                            .setReturnType(TypeNode.STRING)
+                            .build())
+            .collect(Collectors.toList());
+
+    List<Expr> bodyExprs = new ArrayList<>();
+    Preconditions.checkState(
+        rpcMethodArgVarExprs.size() == rpcMethodArgDefaultValueExprs.size(),
+        "The method arguments' the number of variable expressions should equal to the number of default value expressions.");
```
miraleung, line 242: Nit: Something like, "Expected the number of method arguments to match the number of default values"

miraleung (changes requested):

src/main/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposer.java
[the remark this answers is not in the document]
summer-ji-eng: Make sense.

src/test/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposerTest.java
summer-ji-eng: SG

src/main/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposer.java
summer-ji-eng: Sure.

src/main/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposer.java
summer-ji-eng: Thanks.

src/test/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposerTest.java
```diff
@@ -0,0 +1,548 @@
...
+import java.util.List;
+import java.util.Map;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+public class ServiceClientSampleCodeComposerTest {
+  private static final String PACKAGE_NAME = "com.google.showcase.v1beta1";
+
+  FileDescriptor echoFileDescriptor;
```
miraleung, line 44: Nit: Same comment as before on locality for these two variables, their decl can be moved inside the methods.

miraleung (approved):
LGTM with last comment addressed.

src/test/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposerTest.java
summer-ji-eng: Make sense to me.

PR #577, opened 2020-12-01
[DO NOT REVIEW][Merged In #573][CloseSoon]Use resource name class to construct String Default value

summer-ji-eng:
Default value on a string-type method argument of resource name is constructed by resources name class.
Use
```
   * try (ConfigClient configClient = ConfigClient.create()) {
   *   LogSinkName sinkName = LogSinkName.ofProjectSinkName("[PROJECT]", "[SINK]");
   *   configClient.deleteSink(sinkName.toString());
   * }
```

### DIFF
--- a/src/main/java/com/google/api/generator/gapic/composer/DefaultValueComposer.java
+++ b/src/main/java/com/google/api/generator/gapic/composer/DefaultValueComposer.java
@@ -50,7 +50,7 @@ public class DefaultValueComposer {
       TypeNode.withReference(ConcreteReference.withClazz(ByteString.class));
 
   static Expr createDefaultValue(
-      MethodArgument methodArg, Map<String, ResourceName> resourceNames) {
+      MethodArgument methodArg, Map<String, ResourceName> resourceNames, boolean useSampleCode) {
     if (methodArg.isResourceNameHelper()) {
       Preconditions.checkState(
           methodArg.field().hasResourceReference(),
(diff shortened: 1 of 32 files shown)


src/main/java/com/google/api/generator/gapic/composer/DefaultValueComposer.java
```diff
@@ -123,6 +134,13 @@ static Expr createDefaultValue(Field f, boolean useExplicitInitTypeInGenerics) {
     }
 
     if (f.type().equals(TypeNode.STRING)) {
+      if (useSampleCode
```
miraleung, line 137: Why not pass in the ResourceName-typed field instead if the field is a String? Then we don't need to change DefaultValueComposer at all.

src/main/java/com/google/api/generator/gapic/composer/MethodSampleCodeHelperComposer.java
```diff
@@ -118,10 +117,46 @@ private static AssignmentExpr assignClientVariableWithCreateMethodExpr(
       Map<String, ResourceName> resourceNames) {
     return arguments.stream()
         .map(
-            arg ->
-                createAssignmentExpr(
-                    argVarExprs.get(arg.name()),
-                    DefaultValueComposer.createDefaultValue(arg, resourceNames)))
+            arg -> {
+              Expr defaultValueExpr =
+                  DefaultValueComposer.createDefaultValue(arg, resourceNames, true);
```
miraleung, line 122: Here: If `arg` is String-typed and it has a matching resource name, pass that arg in instead.

src/main/java/com/google/api/generator/gapic/composer/DefaultValueComposer.java
```diff
@@ -71,18 +71,29 @@ static Expr createDefaultValue(
...
 
     return createDefaultValue(
-        Field.builder().setName(methodArg.name()).setType(methodArg.type()).build());
+        Field.builder().setName(methodArg.name()).setType(methodArg.type()).build(),
+        resourceNames,
+        useSampleCode);
+  }
+
+  static Expr createDefaultValue(
+      Field f, Map<String, ResourceName> resourceNames, Boolean useSampleCode) {
```
miraleung, line 84: What was the motivation for using the boxed type?

src/main/java/com/google/api/generator/gapic/composer/MethodSampleCodeHelperComposer.java
```diff
@@ -45,9 +46,7 @@ public static TryCatchStatement composeUnaryRpcMethodSampleCode(
     List<Expr> methodArgumentsAssignmentExpr =
         assignMethodArgumentsWithDefaultValues(arguments, methodArgVarExprMap, resourceNames);
     List<Expr> methodVarExprs =
-        arguments.stream()
-            .map(arg -> methodArgVarExprMap.get(arg.name()))
-            .collect(Collectors.toList());
+        createMethodArgVarExprs(arguments, methodArgVarExprMap, resourceNames);
```
miraleung, line 49: PTAL at my comments on splitting out the VarExprs in #573.

miraleung (changes requested):

summer-ji-eng:
Thanks for valuable comments. Make sense to me. Instead of pass boolean value into DefaultValueComposer, I will use your advice and merged in #573 .

PR #602, opened 2020-12-15
[ggj][engx] fix: remove unused package option from integration test rules

miraleung:
The Bazel rule will be updated once all clients are no longer providing a `package` option.


summer-ji-eng:
Can we get rebase on the PR #599?

xiaozhenliu-gg5 (approved):

PR #603, opened 2020-12-16
[ggj][engx] fix: refactor SettingsSampleCodeComposer to be more modular

miraleung:
This is more consistent with our existing `samplecode` structure and not piping sample code generation inside the comment composer.


summer-ji-eng (approved):
LGTM

PR #615, opened 2020-12-22
[ggj][engx] fix: move varexpr helper to local builder for better locality

miraleung:


xiaozhenliu-gg5 (approved):

src/main/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposer.java
```diff
@@ -27,12 +27,17 @@
...
 import com.google.api.generator.gapic.utils.JavaStyle;
 import java.util.Arrays;
+import java.util.function.BiFunction;
 
 public class ServiceClientSampleCodeComposer {
   // TODO(summerji): Add unit tests for ServiceClientSampleCodeComposer.
 
   public static String composeClassHeaderCredentialsSampleCode(
       TypeNode clientType, TypeNode settingsType) {
+    // Funcdtion instead of a helper for better config / locality reference.
```
summer-ji-eng, line 37: Nit: Function

src/main/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposer.java
```diff
@@ -27,12 +27,17 @@
...
+    BiFunction<String, TypeNode, VariableExpr> createVarExprFn =
+        (s, t) -> VariableExpr.withVariable(Variable.builder().setName(s).setType(t).build());
+
```
summer-ji-eng, line 40: In `samplecode-master`, we have discussed to unwrap the `createVariableExpr` method. [code](https://github.com/googleapis/gapic-generator-java/blob/samplecode-master/src/main/java/com/google/api/generator/gapic/composer/ServiceClientSampleCodeComposer.java#L60-L68)

For learning purpose, 
1. what's the motivation to use the BiFunction for create variable expr. One point of previous discussion about unwrap the helper function is to make reader easy to see the logic behind creating variable expr with set different attributes.
2. If we apply BiFunction in the `samplecode-master`, then each method in `ServiceClientSampleCodeComposer` needs to define this `BiFunction`.  Because it is used multiple times, should we define it as class-level variable instead of method-level variable.

miraleung: Since this is fixed in `samplecode-master`, I'll revert this PR. The method-local BiFunction is another way of making it easy to see the logic, since its function-only scoping ensures locality is kept.

miraleung: s/revert/fix this PR to match.

PR #630, opened 2020-12-31
[composer][engx] seperate default value related files and same to resource name

summer-ji-eng:
After change on #616, sample code generate depend on default value composer, it causes bazel error of a cycle in bazel dependency graph.

In this PR,  I create `defaultvalue` bazel package, `resourcename` bazel package and `constants` bazel package. Also do the same for their matched tests.

dependency graph:
defaultvalue -> resourcename -> constants

### DIFF
--- a/src/main/java/com/google/api/generator/gapic/composer/BUILD.bazel
+++ b/src/main/java/com/google/api/generator/gapic/composer/BUILD.bazel
@@ -18,6 +18,8 @@ java_library(
         "//src/main/java/com/google/api/generator/engine/writer",
         "//src/main/java/com/google/api/generator/gapic:status_java_proto",
         "//src/main/java/com/google/api/generator/gapic/composer/comment",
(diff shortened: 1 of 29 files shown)


summer-ji-eng:
> The resource name change makes sense, but I'm not sure the default value and constant classes need their own package. Could you please elaborate on the motivation for these changes?

Because `ServiceClientSampleCodeComposer` refer to default value, after `ServiceClientSampleCodeComposer` moved to `src/.../gapic/composer/samplecode`, it cause bazel build with cycle in dependency graph error. The solution on this is to move default value on its own package, so that `BUILD.bazel` add `dep` like "//src/main/java/com/google/api/generator/gapic/composer/defaultvalue",

and then `defaultvalue` refer to `resourcename`, `resourcename` refer to `constant`, they all have same error.

miraleung (approved):

PR #657, opened 2021-02-17
[ggj] feat(drift): Generate gapic_metadata.json, add integ test framework support

miraleung:


miraleung:
@vchudnov-g PTAL at the generated `gapic_metadata.json` files.

src/main/java/com/google/api/generator/gapic/composer/ServiceClientClassComposer.java
```diff
@@ -1667,4 +1704,49 @@ private static boolean isProtoEmptyType(TypeNode type) {
...
+    Collections.sort(sortedRpcNames);
+    for (String rpcName : sortedRpcNames) {
+      GapicMetadata.MethodList methodList =
+          GapicMetadata.MethodList.newBuilder()
+              .addAllMethods(grpcRpcToJavaMethodNames.get(rpcName))
+              .build();
+      serviceClientProtoBuilder.putRpcs(rpcName, methodList);
+    }
+
+    metadataBuilder =
```
vchudnov-g, line 1744: I'm curious: I see you use the pattern `fooBuilder = fooBuilder.doSomething().setSomething()`. Is the assignment redundant? I thought the builder methods were mutating, rather than functional methods returning a new copy of the object with the changes requested.

vchudnov-g (commented):
LGTM! Thanks for doing this!

src/main/java/com/google/api/generator/gapic/composer/ServiceClientClassComposer.java
miraleung: No, the original builder doesn't mutate -  AFAIK AutoValue classes are all immutable.

vchudnov-g: Oh, weird. In the monolith, builders were used as though their setters mutate. My mental model had been that AutoValue classes are immutable but AutoValue.Builder clases were mutable; I'll need to figure this out. Is there more than one AutoValue impl around?

PR #678, opened 2021-03-03
fix(mixins): enable RPC overrides to clobber mixed-in RPCs [ggj]

miraleung:


miraleung:
The new logic is covered in the integration tests.

noahdietz (approved):
LGTM thanks @miraleung

Issue #680, opened 2021-03-03
Request initialization logic is missing in in-code samples for *Callable() paginated methods


vam-google:
This seems to be happening only in a very limited scope of the paginated `<smth>Callable()` methods.

The incode-samples for paginated methods seem missing request creation logic. This is how they [look now](https://github.com/googleapis/java-dlp/blob/master/google-cloud-dlp/src/main/java/com/google/cloud/dlp/v2/DlpServiceClient.java#L2339)

```java
 // AUTO-GENERATED DOCUMENTATION AND METHOD.
  /**
   * Lists DeidentifyTemplates. See https://cloud.google.com/dlp/docs/creating-templates-deid to
   * learn more.
   *
   * <p>Sample code:
   *
   * <pre>{@code
   * try (DlpServiceClient dlpServiceClient = DlpServiceClient.create()) {
   *   while (true) {
   *     ListDeidentifyTemplatesResponse response =
   *         dlpServiceClient.listDeidentifyTemplatesCallable().call(request);
   *     for (DeidentifyTemplate element : response.getResponsesList()) {
   *       // doThingsWith(element);
   *     }
   *     String nextPageToken = response.getNextPageToken();
   *     if (!Strings.isNullOrEmpty(nextPageToken)) {
   *       request = request.toBuilder().setPageToken(nextPageToken).build();
   *     } else {
   *       break;
   *     }
   *   }
   * }
   * }</pre>
   */
  public final UnaryCallable<ListDeidentifyTemplatesRequest, ListDeidentifyTemplatesResponse>
      listDeidentifyTemplatesCallable() {
    return stub.listDeidentifyTemplatesCallable();
  }
```

Notice that the `request` **is not defined** when first used in the example above.
Also there is an **extra `}`** right at the end (before `</pre>` element).

This is how they used to be:

```java
 // AUTO-GENERATED DOCUMENTATION AND METHOD
  /**
   * Lists DeidentifyTemplates. See https://cloud.google.com/dlp/docs/creating-templates-deid to
   * learn more.
   *
   * <p>Sample code:
   *
   * <pre><code>
   * try (DlpServiceClient dlpServiceClient = DlpServiceClient.create()) {
   *   OrganizationLocationName parent = OrganizationLocationName.of("[ORGANIZATION]", "[LOCATION]");
   *   ListDeidentifyTemplatesRequest request = ListDeidentifyTemplatesRequest.newBuilder()
   *     .setParent(parent.toString())
[...]

PR #690, opened 2021-03-06
[samplecode]fix: add request with default value in regular paged callable method

summer-ji-eng:
fix issue #680

### DIFF
--- a/test/integration/goldens/logging/LoggingClient.java
+++ b/test/integration/goldens/logging/LoggingClient.java
@@ -613,6 +613,14 @@ public class LoggingClient implements BackgroundResource {
    *
    * <pre>{@code
    * try (LoggingClient loggingClient = LoggingClient.create()) {
+   *   ListLogEntriesRequest request =
+   *       ListLogEntriesRequest.newBuilder()
+   *           .addAllResourceNames(new ArrayList<String>())
+   *           .setFilter("filter-1274492040")
+   *           .setOrderBy("orderBy-1207110587")
+   *           .setPageSize(883849137)
+   *           .setPageToken("pageToken873572522")
+   *           .build();
    *   while (true) {
    *     ListLogEntriesResponse response = loggingClient.listLogEntriesCallable().call(request);
    *     for (LogEntry element : response.getResponsesList()) {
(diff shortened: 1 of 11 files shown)


miraleung (approved):

test/integration/goldens/logging/LoggingClient.java
```diff
@@ -613,6 +613,14 @@ public final ListLogEntriesPagedResponse listLogEntries(ListLogEntriesRequest re
    *
    * <pre>{@code
    * try (LoggingClient loggingClient = LoggingClient.create()) {
+   *   ListLogEntriesRequest request =
+   *       ListLogEntriesRequest.newBuilder()
+   *           .addAllResourceNames(new ArrayList<String>())
+   *           .setFilter("filter-1274492040")
+   *           .setOrderBy("orderBy-1207110587")
+   *           .setPageSize(883849137)
+   *           .setPageToken("pageToken873572522")
```
vam-google, line 622: Setting page token in the initial request seems harmful before you start paginated call there is no page token (i.e. page token makes sense only for pages 2+, but the first page does not have page token).

test/integration/goldens/logging/LoggingClient.java
```diff
@@ -613,6 +613,14 @@ public final ListLogEntriesPagedResponse listLogEntries(ListLogEntriesRequest re
    *
    * <pre>{@code
    * try (LoggingClient loggingClient = LoggingClient.create()) {
+   *   ListLogEntriesRequest request =
+   *       ListLogEntriesRequest.newBuilder()
+   *           .addAllResourceNames(new ArrayList<String>())
+   *           .setFilter("filter-1274492040")
+   *           .setOrderBy("orderBy-1207110587")
+   *           .setPageSize(883849137)
```
vam-google, line 621: Can we assing something reasonable (i.e smaller, like 10 or 100) to these kind of fields (random huge number for page size is confusing and not realistic)?

vam-google (commented):
LGTM, but I was wondering how difficult would it be to address those the two usability-related comments?

test/integration/goldens/logging/LoggingClient.java
summer-ji-eng: Currently, we use hashcode of field' name. [source code](https://github.com/googleapis/gapic-generator-java/blob/c5a32cc3fe74a62898228f0c39a455e7701d2f1c/src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java#L133-L139)

IMO, smaller number like 10 make more sense in reality. And PageSize is fixed filed to be set for default request variable. Looks like it is doable in code. @miraleung what is your opinion?

test/integration/goldens/logging/LoggingClient.java
summer-ji-eng: FYI @miraleung

PR #708, opened 2021-04-07
[ggj] feat(gapic): propagate protobuf 'deprecated' to classes/methods

### LINKED ISSUE #703: Generator should respect deprecated annotations
For gapic service rpcs:

`option deprecated = true`

Example: [proto](https://github.com/googleapis/googleapis/blob/d76f26f335b2b570b7e19d938f35f9894236e565/google/container/v1/cluster_service.proto#L175) [client](https://github.com/googleapis/java-container/blob/c9c5eed29645a849b4a4ddb604c1fbdc76534785/google-cloud-container/src/main/java/com/google/cloud/container/v1/ClusterManagerClient.java#L1234-L1238)

Expected:
* The method should be annotationed `@Deprecated`
* The javadoc should contain `@deprecated` with a note to replacement (if possible)

miraleung:
Closes #703. Note that the per-RPC methods in ServiceStub are not marked with `@deprecated` or `@Deprecated` since they are overrides, and the superclass's method has already been marked as such.

### DIFF
--- a/src/main/java/com/google/api/generator/gapic/composer/ServiceClientClassComposer.java
+++ b/src/main/java/com/google/api/generator/gapic/composer/ServiceClientClassComposer.java
@@ -138,7 +138,7 @@ public class ServiceClientClassComposer implements ClassComposer {
             .setHeaderCommentStatements(
                 createClassHeaderComments(service, typeStore, resourceNames, messageTypes))
             .setPackageString(pakkage)
-            .setAnnotations(createClassAnnotations(pakkage, typeStore))
+            .setAnnotations(createClassAnnotations(pakkage, service.isDeprecated(), typeStore))
             .setScope(ScopeNode.PUBLIC)
             .setName(className)
             .setImplementsTypes(createClassImplements(typeStore))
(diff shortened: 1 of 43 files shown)


test/integration/goldens/kms/KeyManagementServiceStubSettings.java
```diff
@@ -1301,6 +1290,17 @@ public Builder applyToAllUnaryMethods(
...
+    public UnaryCallSettings.Builder<SetIamPolicyRequest, Policy> setIamPolicySettings() {
+      return setIamPolicySettings;
+    }
+
+    /** Returns the builder for the settings used for calls to testIamPermissions. */
+    public UnaryCallSettings.Builder<TestIamPermissionsRequest, TestIamPermissionsResponse>
+        testIamPermissionsSettings() {
+      return testIamPermissionsSettings;
+    }
+
```
chingor13, line 1303: non-blocking comment: without debugging, I'm not seeing the cause for why these methods are moved in the generated code.

src/main/java/com/google/api/generator/gapic/composer/ServiceClientClassComposer.java
```diff
@@ -158,11 +158,15 @@ public GapicClass generate(GapicContext context, Service service) {
     return GapicClass.create(kind, classDef);
   }
 
-  private static List<AnnotationNode> createClassAnnotations(String pakkage, TypeStore typeStore) {
+  private static List<AnnotationNode> createClassAnnotations(
+      String pakkage, boolean isDeprecated, TypeStore typeStore) {
```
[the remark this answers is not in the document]
miraleung: Done.

chingor13 (approved):

test/integration/goldens/kms/KeyManagementServiceStubSettings.java
miraleung: To avoid arbitrary reordering, I'll sort everything in alphabetical order (except for mixin APIs). Adding this to a separate PR, so this diff should go away then. Thanks!

PR #718, opened 2021-04-22
chore: temporarily disable codecov [ggj]

miraleung:
Temporarily disabling Codecov due to https://about.codecov.io/security-update/.


miraleung:
Closing since this is using bazel coverage, not codecov.

PR #719, opened 2021-04-22
chore: Pre-DIREGAPIC Refactoring #1

vam-google:
This PR is a minimal subset of https://github.com/googleapis/gapic-generator-java/pull/717

 This is as minimal of a PR that I could make such that:
1) Only a single composer is refactored.
2) It compiles and tests run.
3) All the key changes are present (abstract/grpc class, comment composer, class names, golden files move)

I had to add a few quite ugly and artificial   constructs:
 - fake `ClassComposer` under `composer` package which simply extends `composer.common.Composer` to void changing the rest of the classes
- put back static methods in `ClassNames`,  `StubCommentComposer` and `TestProtoLoaderUtil` (for the same reasons as above). All of those chagnes are marked with `// TODO: remove after Pre-DIREGAPIC refactoring is fully merged` statement.


src/test/java/com/google/api/generator/test/framework/Differ.java
```diff
@@ -50,22 +50,28 @@
...
     List<String> unifiedDiff =
         UnifiedDiffUtils.generateUnifiedDiff("golden", "codegen", original, diff, 2);
     return unifiedDiff;
   }
 
   private static class GoldenFileReadException extends RuntimeException {
-    public GoldenFileReadException(String errorMessage) {
-      super(errorMessage);
+
+    private static final long serialVersionUID = 7423787084310530945L;
```
miraleung, line 62: Would this be used somewhere? Same below.

src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceSettingsClassComposer.java
```diff
@@ -75,22 +74,25 @@
...
+  private final Class<?> instantiatingChannelProviderClass;
+  private final String defaultTransportProviderBuilderName;
+
+  protected AbstractServiceSettingsClassComposer(
+      TypeStore fixedTypeStore,
+      Class<?> instantiatingChannelProviderClass,
+      String defaultTransportProviderBuilderName) {
+    this.fixedTypeStore = fixedTypeStore;
+    this.instantiatingChannelProviderClass = instantiatingChannelProviderClass;
+    this.defaultTransportProviderBuilderName = defaultTransportProviderBuilderName;
```
miraleung, line 95: Responding to [this comment](https://github.com/googleapis/gapic-generator-java/pull/717#discussion_r618675308):

To recap, we currently have three possible approaches:
- Instantiate any transport-specific values in the transport composer (e.g. GrpcServiceSettingsClassComposer sets `defaultTransportProviderBuilderName = "defaultGrpcTransportProviderBuilder"` (current approach in this PR).
  - Advantages: Maximizes per-composer locality.
  - Drawbacks: Splits per-transport state across many composers, so it may be harder to debug or change transport-specific things. When `ClassNames` is later involved, there will be another potential set of classes to look up.
- Pass the Transport enum into the Abstract class, which can then decide whether to use `"Grpc"` or `"HttpJson"` in the string (**most suboptimal**).
  - Advantages: Maximizes locality in the Abstract class's common logic (of which there's a significant amount).
  - Drawbacks: The abstract class should not know anything about per-transport state. 
- Have a per-transport context object that holds the respective values. 
  - Advantages: Maximizes per-transport locality, transport-specific debugging may be easier.
  - Drawbacks: Need to look up context in a transport-specific file.

My $0.02: The per-transport context object doesn't have to hold all the state, just anything that needed to be passed into the superclass. Motivation: Passing in differing string values in each per-transport composer class may possibly be less maintainable, especially since the values change within the same transport (e.g. `"gRPC"` vs `"Grpc"`).

We can also discuss more offline. :)

vam-google: I can do the contexts, no problem, two questions before doing it (we can discuss it via GVC, just mentioning it here to not forget it):

Please clarify what you mean by context maximizing per-transport locality?

Yes, the context will need to hold the stuff to be passed to the superclass, but there are multiple superlcasses (like 5+) and each of them needs a different set of things (with some intersection, but not much). This means we will either need 5+ different transport specific contexts, or one transport-specific context with stuff shared across different composers in, which will tightly couple the composers class hierarchy with the context and with each other.

src/test/java/com/google/api/generator/test/framework/Differ.java
vam-google: This is technically a best-practice requirement for all `Serializable` classes (which all `Exceptions` are), a bit dogmatic, agree.

src/main/java/com/google/api/generator/gapic/composer/utils/ClassNames.java
```diff
@@ -26,9 +26,21 @@
   private static final String SERVICE_SETTINGS_CLASS_NAME_PATTERN = "%sSettings";
   private static final String SERVICE_STUB_SETTINGS_CLASS_NAME_PATTERN = "%sStubSettings";
   private static final String SERVICE_STUB_CLASS_NAME_PATTERN = "%sStub";
+
+  // TODO: remove after Pre-DIREGAPIC refactoring is fully merged
   private static final String GRPC_SERVICE_STUB_CLASS_NAME_PATTERN = "Grpc%sStub";
+  // TODO: remove after Pre-DIREGAPIC refactoring is fully merged
```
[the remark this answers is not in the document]
vam-google: The problem is, as you mentioned in the other comment that it can be `gRPC` or `Grpc` and the enum itself is `GRPC`, which is ugly and the whole issue is about just generating a comment string. I guess technically the ClassNames and CommentComposer should repeat the fate of the other composers - have their own hierarchy to match the rest of the code architecturally, I just felt like that was an overkill for such a simple case, that is why went with a simple parameter. We can do something fancier, if you feel like it is worth it. Should we?

src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceSettingsClassComposer.java
```diff
@@ -110,7 +112,7 @@ public GapicClass generate(GapicContext ignored, Service service) {
             .setName(className)
             .setExtendsType(
                 TypeNode.withReference(
-                    FIXED_TYPESTORE
+                    getFixedTypeStore()
```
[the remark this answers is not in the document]
vam-google: Also  just basically a best practice thing for java - do not use protected fields, use protected setter/getter for them instead. This field can't be private, because it is accessed in the subclasses, so it must be either protected or have a protected getter, which is what is done here.

src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceSettingsClassComposer.java
```diff
@@ -75,22 +74,25 @@
...
-    return INSTANCE;
+  private final TypeStore fixedTypeStore;
+  private final Class<?> instantiatingChannelProviderClass;
+  private final String defaultTransportProviderBuilderName;
+
+  protected AbstractServiceSettingsClassComposer(
+      TypeStore fixedTypeStore,
+      Class<?> instantiatingChannelProviderClass,
+      String defaultTransportProviderBuilderName) {
+    this.fixedTypeStore = fixedTypeStore;
```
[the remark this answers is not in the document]
vam-google: Sorry I'm not sure I understand your comment instantiating the intersection of gRPC and REST types.

src/test/java/com/google/api/generator/test/framework/Differ.java
miraleung: SG, thanks.

src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceSettingsClassComposer.java
miraleung: Discussed offline.

src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceSettingsClassComposer.java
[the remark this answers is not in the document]
miraleung: Leaving open to revisit after the upcoming changes - if this becomes unneeded by the transport-specific composers, consider making it inaccessible to the subclasses.

vam-google:
@miraleung PTAL

src/main/java/com/google/api/generator/gapic/composer/utils/ClassNames.java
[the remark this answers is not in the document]
miraleung: Resolved with the latest refactoring.

src/main/java/com/google/api/generator/gapic/model/TransportContext.java
```diff
@@ -0,0 +1,103 @@
...
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+public abstract class TransportContext {
+  public enum Transport {
+    REST,
+    GRPC,
+    GRPC_REST;
+
+    public static Transport parse(String name) {
```
miraleung, line 28: Nit: Consider adding a comment with the expected input. As someone familiar with what we're doing, it looks like this will be parsing a Bazel param string like `"grpc+rest"`, so it'd be great to document this method's usage.

miraleung (approved):
LGTM. Given the scope of DIREGAPIC's refactorings, could we start a `diregapic` main branch for all the refactoring PRs, so we can merge that into master when it's all done?

src/main/java/com/google/api/generator/gapic/model/TransportContext.java
vam-google: Will in the PR which will introduce the actual parameters (here it is just hardcoded everywhere as grpc for now, because unill regapic prs are merged, grpc is the only actually supported transport by the generator.

PR #723, opened 2021-05-11
fix(types): Use fully-qualified message type names [ggj]

miraleung:
Addresses the use case where a proto message `foo.bar.Car` has an internal field with the same short name, such as `bar.foo.Car`.

### DIFF
--- a/src/main/java/com/google/api/generator/gapic/model/Message.java
+++ b/src/main/java/com/google/api/generator/gapic/model/Message.java
@@ -30,6 +30,9 @@ import javax.annotation.Nullable;
 public abstract class Message {
   public abstract String name();
 
+  // The fully-qualified proto name, which differs from the Java fully-qualified name.
+  public abstract String fullProtoName();
+
   // TODO(unsupported): oneof fields are parsed as separate ones because field flattening refers to
   // a specific field.
   public abstract ImmutableList<Field> fields();
@@ -87,6 +90,8 @@ public abstract class Message {
   @AutoValue.Builder
   public abstract static class Builder {
     public abstract Builder setName(String name);
+
+    public abstract Builder setFullProtoName(String fullProtoName);
 
     public abstract Builder setFields(List<Field> fields);
 
--- a/src/main/java/com/google/api/generator/gapic/model/Message.java
+++ b/src/main/java/com/google/api/generator/gapic/model/Message.java
@@ -31,6 +31,8 @@ public abstract class Message {
   public abstract String name();
 
   // The fully-qualified proto name, which differs from the Java fully-qualified name.
+  // For example, this would be google.showcase.v1beta1.EchoRequest for echo.proto (see testdata),
+  // whereas that message's Java fully-qualified name is com.google.showcase.v1beta1.EchoRequest.
   public abstract String fullProtoName();
 
   // TODO(unsupported): oneof fields are parsed as separate ones because field flattening refers to
--- a/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
+++ b/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
@@ -420,11 +420,13 @@ public class Parser {
     for (EnumDescriptor enumDescriptor : fileDescriptor.getEnumTypes()) {
       String name = enumDescriptor.getName();
       List<EnumValueDescriptor> valueDescriptors = enumDescriptor.getValues();
+      TypeNode enumType = TypeParser.parseType(enumDescriptor);
       messages.put(
-          name,
+          enumType.reference().fullName(),
           Message.builder()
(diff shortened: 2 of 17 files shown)


src/main/java/com/google/api/generator/gapic/model/Message.java
```diff
@@ -30,6 +30,9 @@
 public abstract class Message {
   public abstract String name();
 
+  // The fully-qualified proto name, which differs from the Java fully-qualified name.
+  public abstract String fullProtoName();
```
vam-google, line 34: Can you please specify how exactly they are different? Is it about the extra `com.` prefix for java names, or more than that?

src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -641,10 +640,51 @@ static LongrunningOperation parseLro(
...
         methodDescriptor.getOptions().getExtension(OperationsProto.operationInfo);
 
-    String responseTypeName = parseNestedProtoTypeName(lroInfo.getResponseType());
-    String metadataTypeName = parseNestedProtoTypeName(lroInfo.getMetadataType());
-    Message responseMessage = messageTypes.get(responseTypeName);
-    Message metadataMessage = messageTypes.get(metadataTypeName);
+    // These can be short names (e.g. FooMessage) or fully-qualified names with the *proto* package.
+    String responseTypeName = lroInfo.getResponseType();
+    String metadataTypeName = lroInfo.getMetadataType();
+    String[] responseTypeNameComponents = responseTypeName.split("\\.");
```
vam-google, line 646: Minor, but it seems like you don't need the full set of compontents, only the last element after dot, so it can be something like.
```
String responseTypeShortName = responseTypeName.substring(responseTypeName.lastIndexOf(".") + 1)
```
It is not about performance, mainly just to make it clear that the other compnents are not needed to the readers of the code.

src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -641,10 +640,51 @@ static LongrunningOperation parseLro(
...
+    boolean isResponseTypeNameShortOnly = responseTypeName.equals(responseTypeShortName);
+    boolean isMetadataTypeNameShortOnly = metadataTypeName.equals(metadataTypeShortName);
+
+    Message responseMessage = null;
+    Message metadataMessage = null;
+
+    // The messageTypes map keys to the Java fully-qualified name.
+    for (Map.Entry<String, Message> messageEntry : messageTypes.entrySet()) {
+      String[] packageComponents = messageEntry.getKey().split("\\.");
+      String messageShortName = packageComponents[packageComponents.length - 1];
```
vam-google, line 663: Similar situation as with `responseTypeShortName` and `metadataTypeShortName`

src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -641,10 +640,51 @@ static LongrunningOperation parseLro(
...
+      if (responseMessage == null) {
```
vam-google, line 664: The `responseMessage` logic and `metadataMessage` logic seem to be a copypaste of each other. Please consider moving that into a private or method combining this to a for loop with 2 elements in it (response and metadata messages).

vam-google (approved):
LGTM with few minor comments, please address those before submitting

src/main/java/com/google/api/generator/gapic/model/Message.java
miraleung: Yes, it's the proto package versus the `java_package`. Added an example.

src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
miraleung: Done.

src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
miraleung: Done.

src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
miraleung: I'd considered using these, which would result in the following tradeoffs:
- Nested loop: Need an extra map or data structure to associate `responseMessage`, `responseTypeName`, `isResponseTypeNameShortOnly`, which would be less readable, or recompute the latter in the loop (which is inefficient).
- Separate method: Need to take four parameters at best - the three above plus `messageEntry`, which would make the call site, helper signature, and method itself less readable.

Given that the logic is relatively small, copy-pasting does not seem to have significantly better or worse tradeoffs compared to the alternative approches.

PR #724, opened 2021-05-14
fix(types): Reorder nested types for legacy protos w/ java_outer_class in one file [ggj]

miraleung:
PR #723 addressed only modern protos with both `java_outer_classname` and `java_multiple_files` set to true. This PR handles cases like [Bigquery's model.proto](https://github.com/googleapis/googleapis/blob/master/google/cloud/bigquery/v2/model.proto), which does not set the latter.

This was tested by `bazel test`-ing every `java_gapic_test_suite` target in `googleapis`.


vam-google (approved):
LGTM

PR #730, opened 2021-05-21
chore: DIREGAPIC #2 refactoring of the rest Composer Classes

vam-google:
Each composer class refactoring is pushed as a separate commit.
This includes the following classes:

`AbstractServiceCallableFactoryComposer`
`AbstractClientTestClassComposer`
`AbstractServiceStubClassComposer`
`ServiceStubSettingsClassComposer`

Plus move of the non transport-specific composers under `com.google.api.generator.gapic.composer.common` package.

Changes from the original [DIREGAPIC #1](https://github.com/googleapis/gapic-generator-java/pull/719) refactoring PR:
- Moved `TransportContext` from `model` to `composer.common` package
- Moved `Transport` enum to the top-class level (used to be a public static internal class under `TransportContext`
- Moved replaced `TransportContext` in `GapicContext` with just the `Transport` enum.

The changes above were done because keeping `TransportContext` as part of `GapicContext` started looking not right, after `TransportContext` actually started being used by the classes. The original [DIREGAPIC #1](https://github.com/googleapis/gapic-generator-java/pull/719) PR was done on the simplest class `AbstractServiceSettingsClassComposer` and it was lacking some usage patterns of more sophisticated classes.


src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceClientTestClassComposer.java
```diff
@@ -96,15 +96,18 @@
...
+  protected static final AnnotationNode TEST_ANNOTATION =
+        AnnotationNode.withType(FIXED_TYPESTORE.get("Test"));
+
+  private final TransportContext transportContext;
+
+  protected AbstractServiceClientTestClassComposer(TransportContext transportContext) {
+    this.transportContext = transportContext;
+  }
+
+  public TransportContext getTransportContext() {
```
miraleung, line 109: Should this be part of ClassComposer?

src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceStubSettingsClassComposer.java
```diff
@@ -201,7 +196,30 @@ public GapicClass generate(GapicContext context, Service service) {
...
   }
 
-  private static List<AnnotationNode> createClassAnnotations(Service service) {
+  protected abstract MethodDefinition createDefaultTransportTransportProviderBuilderMethod();
+
+  protected abstract MethodDefinition createDefaultApiClientHeaderProviderBuilderMethod(
+      Service service, TypeStore typeStore);
+
+  public abstract MethodDefinition createDefaultTransportChannelProviderMethod();
+  //
```
miraleung, line 205: Delete?

miraleung (changes requested):
LGTM with nits addressed. Will stamp on the next review so we can do a "checkpoint" release that contains only this refactoring PR.

Thanks for splitting the refactorings into individual commits! That greatly helped make the changes clearer.

src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceClientTestClassComposer.java
vam-google: Yep, if you are fine with adding it to the general interface I'm more than  happy to do it!

src/main/java/com/google/api/generator/gapic/model/Transport.java
```diff
@@ -0,0 +1,28 @@
...
+package com.google.api.generator.gapic.model;
+
+public enum Transport {
+  REST,
+  GRPC,
+  // Never used in the context as is, must be split into two contexts (REST and GRPC respectively)
+  // instead.
+  GRPC_REST;
+
+  public static Transport parse(String name) {
```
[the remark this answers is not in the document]
vam-google: sure

src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceStubSettingsClassComposer.java
vam-google: Deleted

src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceClientTestClassComposer.java
vam-google: There are many transport-unaware composers, which will make this method not-needed for them. The possible solutions would be adding something `interface TransportAwareClassComposer extends ClassComposer` with the new method. WDYT? I'll leave it unchanged for this PR, and can make this change in a subsequent PR as it would be a trivial one.

vam-google:
@miraleung PTAL, also please check the last commit specifically

src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceClientTestClassComposer.java
miraleung: Subsequent PR SGTM. If it's not needed in most composers, feel free to skip.

miraleung (approved):

PR #733, opened 2021-05-24
fix: Support enums in proto HTTP annotations [ggj]

miraleung:


vam-google (approved):
LGTM.
Can we also add a test for it? Only if it can be done with little effort, please do not add the full set of new test just for this feature (really not worth it).

miraleung:
> LGTM.
> Can we also add a test for it? Only if it can be done with little effort, please do not add the full set of new test just for this feature (really not worth it).

Clicked merge too soon, will look into this as a followup.

PR #736, opened 2021-05-25
feat: add AST support for lambdas [ggj]

miraleung:
These lambdas will eventually replace anonymous classes once Java 7 goes into maintenance mode for all the client libraries.


src/main/java/com/google/api/generator/engine/writer/JavaWriterVisitor.java
```diff
@@ -509,6 +510,45 @@ public void visit(AssignmentOperationExpr assignmentOperationExpr) {
...
+      // Print just the variable.
+      lambdaExpr.arguments().get(0).variable().identifier().accept(this);
+    } else {
+      // Stylistic choice - print the types and variable names for clarity.
+      leftParen();
+      int numArguments = lambdaExpr.arguments().size();
+      for (int i = 0; i < numArguments; i++) {
+        lambdaExpr.arguments().get(i).accept(this);
+        if (i < numArguments - 1) {
+          buffer.append(COMMA);
```
vam-google, line 528: For natural constants (like `0`, `1`, or `','` in this case) it is basically a good practice to use them directly, no need to declare constants, otherwise it would be similar to having something like `for (i = ZERO ...)` in the for loop above. Or compare it to the line 532, where the lambda arrow operator is used directly (which is also absolutely fine).

src/main/java/com/google/api/generator/engine/writer/JavaWriterVisitor.java
```diff
@@ -509,6 +510,45 @@ public void visit(AssignmentOperationExpr assignmentOperationExpr) {
     assignmentOperationExpr.valueExpr().accept(this);
   }
 
+  @Override
+  public void visit(LambdaExpr lambdaExpr) {
+    if (lambdaExpr.arguments().isEmpty()) {
+      leftParen();
```
vam-google, line 516: Minor: I usually don't mind abbreviations (even though, they are against best practices in java), but for this one `Paren` may stand for `Parent` or `Parenthesis`, both of which are possible in a typical programming tasks. I know it was not part of this PR, but can you please rename it to `left/rightParenthesis` instead?

vam-google (approved):
LGTM with a few minor comments. Also can you please add in the PR description why you are doing it?

src/main/java/com/google/api/generator/engine/writer/JavaWriterVisitor.java
miraleung: Would prefer sticking with the small, string-literal constants in this case to eliminate potential typos, ease maintainability, and avoid depending on the compiler to optimize away unnecessary new object creation (as such optimizations may not necessarily find all applicable instances).

src/main/java/com/google/api/generator/engine/writer/JavaWriterVisitor.java
miraleung: Will do in a subsequent PR.

miraleung:
Updated the PR description.

PR #738, opened 2021-05-25
chore(AST): Prevent duplicate-named or modified/scoped method args [ggj]

miraleung:


src/main/java/com/google/api/generator/engine/ast/MethodDefinition.java
```diff
@@ -314,28 +317,66 @@ public MethodDefinition build() {
...
+      performArgumentChecks();
+      performThrownExceptionChecks();
 
       return method;
     }
 
-    void performNullChecks() {
+    private void performArgumentChecks() {
+      // Must be a declaration.
+      arguments().stream()
```
vam-google, line 328: This is strange - two subsequent iterations over the same `arguments()` to do validation. Why can't it be done in the same loop?

vam-google (approved):
LGTM.
Why do we need this? Please put in the description.

src/main/java/com/google/api/generator/engine/ast/MethodDefinition.java
miraleung: They're done separately to clarify the error messages - separates out `isDecl` and the modifier/scope issues. Alternatively, this could go in a for-loop, but since there are typically very few arguments (e.g. nearly always < 10), the n vs 2n time doesn't really matter in the end.

PR #743, opened 2021-06-01
fix(tests): Ensure deterministic field ordering in test classes [ggj]

miraleung:
This is needed for the Bazel integration test fix in #744.


vam-google (approved):
LGTM

PR #745, opened 2021-06-01
fix(bazel): ensure integration tests diff files recursively [ggj]

miraleung:
Take two on #744 so we can merge this into the main branch after #743.


miraleung:
Merging this is the same as #744.

PR #754, opened 2021-06-07
chore: Add comment context on null protobuf check [ggj]

miraleung:
Addresses #752.


src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceStubSettingsClassComposer.java
```diff
@@ -579,6 +579,13 @@ private static Expr createPagedListDescriptorAssignExpr(
                 String.format("get%sList", JavaStyle.toUpperCamelCase(repeatedFieldName)))
             .setReturnType(returnType)
             .build();
+
+    // While protobufs should not be null, this null-check is needed to protect against NPEs
+    // in paged iteration, as some clients actually return null instead of an empty list.
```
dpcollins-google, line 584: This comment does not provide accurate details. No client which deals in protobufs can possibly return null. Those that can use the legacy HTTP/JSON types. You can see that the referenced type from the issue and discussion extends GenericJson, not AbstractMessage. https://developers.google.com/resources/api-libraries/documentation/compute/v1/java/latest/com/google/api/services/compute/model/Image.html

miraleung: Updated.

vam-google (approved):
LGTM

PR #758, opened 2021-06-08
fix(bazel): Replace monolith deps with rules_gapic [ggj]

miraleung:
This completely removes the Java microgenerator's remaining dependency on the monolith, which takes us a step closer to removing the monolith from googleapis.


miraleung:
Done in #764.

PR #761, opened 2021-06-10
fix(resnames): Use anon resname classes when only wildcards are present

miraleung:
Use an anonymous ResourceName class as a default value when only wildcard resource name types are present. This primarily affects GCS and IAM. These anonymous classes will be used only for explicit field assignments, and not when the value is being set on a message. Specifically, note that the IAM integration goldens are unchanged.

The GCS use case is covered by the default value unit test. Integration tests may be added once the GCS protos are published.


noahdietz (approved):

vam-google (approved):
LGTM

PR #764, opened 2021-06-10
fix(bazel): Remove monolith rule deps from the Java µgen Bazel rules [ggj]

miraleung:
This PR is needed for the preliminary removal of the monolith rules from googleapis' `switched_rules_by_language`.


.github/workflows/ci.yaml
```diff
@@ -49,7 +49,9 @@ jobs:
         run: bazel --batch test $(bazel query "//src/test/..." | grep "Test$") --noshow_progress --test_output=errors
 
       - name: Integration Tests
-        run: bazel --batch test //test/integration:asset //test/integration:credentials //test/integration:iam //test/integration:kms //test/integration:logging //test/integration:pubsub //test/integration:redis //test/integration:library --noshow_progress
+        run: |
+          bazel clean --expunge --async
+          bazel --batch test //test/integration:asset //test/integration:credentials //test/integration:iam //test/integration:kms //test/integration:logging //test/integration:pubsub //test/integration:redis //test/integration:library --noshow_progress
```
vam-google, line 54: Why is each integration tests mentioned individually? Why not just `bazel test //...` for all tests? This would greatly simplify CI & workstation consistency between the tests and will not need any extra knowledge to maintain it (basically I have forgotten to add compute test in this list).

If there is a need to launch integration tests as a separate entity (to separate them from unit tests), please consider using [test_suite](https://docs.bazel.build/versions/main/be/general.html#test_suite) instead of mentioning each test individually.

We also need to add compute tests here, either as part of this PR or separately (I'll add it, if you prefer it to be done separately).

test/integration/BUILD.bazel
```diff
@@ -29,20 +31,20 @@ INTEGRATION_TEST_LIBRARIES = [
...
 
 # Keys must match the values in INTEGRATION_TEST_LIBRARIES above.
 API_GAPIC_TARGETS = {
-    "asset": "@com_google_googleapis//google/cloud/asset/v1:asset_java_gapic",
-    "credentials": "@com_google_googleapis//google/iam/credentials/v1:credentials_java_gapic",
+    "asset": ":asset_java_gapic",
+    "credentials": ":credentials_java_gapic",
     "iam": ":iam_java_gapic",  # Googleapis' LRO does not have a Java Gapic.
     "kms": ":kms_java_gapic",  # Local target because mixins are not rolled out yet.
     "pubsub": ":pubsub_java_gapic",
```
vam-google, line 43: Why are some tests coming from googleapis/googleapis-discovery and some are "native" to gapic-generator-java? (i.e. what is special about iam, kms and pubsub?)

test/integration/goldens/library/com/google/cloud/example/library/v1/stub/LibraryServiceStubSettings.java
```diff
@@ -408,28 +401,8 @@ protected LibraryServiceStubSettings(Builder settingsBuilder) throws IOException
...
-              .setInitialRetryDelay(Duration.ofMillis(100L))
-              .setRetryDelayMultiplier(1.3)
-              .setMaxRetryDelay(Duration.ofMillis(60000L))
-              .setInitialRpcTimeout(Duration.ofMillis(60000L))
-              .setRpcTimeoutMultiplier(1.0)
-              .setMaxRpcTimeout(Duration.ofMillis(60000L))
-              .setTotalTimeout(Duration.ofMillis(60000L))
-              .build();
-      definitions.put("retry_policy_0_params", settings);
+      settings = RetrySettings.newBuilder().setRpcTimeoutMultiplier(1.0).build();
```
vam-google, line 404: Why is this getting removed? Seems like something that would happen when switching from monolith to microgen, or loosing grpc_service_config somewhere

.github/workflows/ci.yaml
miraleung: Because `bazel test //...` runs all the update (i.e. non-test) targets as well. `test_suite` looks like a good option, will explore this separately.

Adding compute tests SGTM, done.

test/integration/BUILD.bazel
miraleung: I added some custom changes to IAM, KMS, and PubSub to exercise edge cases that the `googleapis` target does not cover.

test/integration/goldens/library/com/google/cloud/example/library/v1/stub/LibraryServiceStubSettings.java
miraleung: Fixed, thanks for catching that.

.github/workflows/ci.yaml
vam-google: bazel test //... will run only what is marked as `Test = true` in bazel [rule](https://docs.bazel.build/versions/4.1.0/skylark/lib/globals.html#parameters-30) definition. 
The integration tests are defined subject for `bazel test` https://github.com/googleapis/gapic-generator-java/blob/master/rules_bazel/java/integration_test.bzl#L87, `_update` targets are not https://github.com/googleapis/gapic-generator-java/blob/master/rules_bazel/java/integration_test.bzl#L164

vam-google: bazel test //... will run only what is marked as `Test = true` in bazel [rule](https://docs.bazel.build/versions/4.1.0/skylark/lib/globals.html#parameters-30) definition. 
The integration tests are defined subject for `bazel test` https://github.com/googleapis/gapic-generator-java/blob/master/rules_bazel/java/integration_test.bzl#L87, `_update` targets are not https://github.com/googleapis/gapic-generator-java/blob/master/rules_bazel/java/integration_test.bzl#L164

vam-google (commented):
LGTM, but please check the `bazel test //...` topic comment.

.github/workflows/ci.yaml
miraleung: Done.

vam-google (approved):

test/integration/goldens/library/com/google/cloud/example/library/v1/stub/LibraryServiceStubSettings.java
miraleung: On testing again - this will now fail because of the monolith rule removal, since it runs all the non-golden tests from the googleapis' targets. Will make this change on my next pass of removing the monolith's rules entirely from this codebase.

PR #765, opened 2021-06-11
feat: Add code request generator dumper and an ability to run generator from dumped file

vam-google:
For debugging purposes.

Usage example:

```bzl
load("@gapic_generator_java//rules_java_gapic:java_gapic.bzl", "java_generator_request_dump")
java_generator_request_dump(
    name = "compute_small_request_dump",
    srcs = [":compute_small_proto_with_info"],
    transport = "rest",
)
```


miraleung (changes requested):
I came across #767 first, so I left some comments on these files in there.

vam-google:
@miraleung The plan was to review just this one first, merge it and then I could easily merge that into the #767.

BUILD.bazel
```diff
@@ -45,6 +45,34 @@ java_binary(
     ],
 )
 
+java_binary(
```
vam-google, line 48: @miraleung (from https://github.com/googleapis/gapic-generator-java/pull/767): Nit: Can we add a summary comment above these two rules, which also indicate these are debugging tools?

src/main/java/com/google/api/generator/Main.java
```diff
@@ -20,15 +20,13 @@
...
-import com.google.protobuf.Descriptors.DescriptorValidationException;
 import com.google.protobuf.ExtensionRegistry;
 import com.google.protobuf.compiler.PluginProtos.CodeGeneratorRequest;
 import com.google.protobuf.compiler.PluginProtos.CodeGeneratorResponse;
 import java.io.IOException;
 
 public class Main {
-  public static void main(String[] args)
-      throws IOException, InterruptedException, DescriptorValidationException {
+  public static void main(String[] args) throws IOException {
```
vam-google, line 29: @miraleung (from https://github.com/googleapis/gapic-generator-java/pull/767): Could we move this into a separate file? Having the main classes all call something like `Registry.registerAllExtensions` has more consistent semantics than `Main.registerAllExtensions`.

Up to you whether to put this in a subpackage or keep it here.

vam-google: Ok

src/main/java/com/google/api/generator/debug/CodeGeneratorRequestDumper.java
```diff
@@ -0,0 +1,37 @@
...
+
+package com.google.api.generator.debug;
+
+import com.google.api.generator.Main;
+import com.google.protobuf.ExtensionRegistry;
+import com.google.protobuf.compiler.PluginProtos.CodeGeneratorRequest;
+import com.google.protobuf.compiler.PluginProtos.CodeGeneratorResponse;
+import java.io.IOException;
+
+public class CodeGeneratorRequestDumper {
```
vam-google, line 23: @miraleung (from https://github.com/googleapis/gapic-generator-java/pull/767): If we're using `Main` in the class names, could we append it here for consistency? Alternatively, remove `Main` from the other debugger's main class's name.

vam-google: Those are different things basically. The MainFromFile is literally another variation of the Main class. This, on the other hand, is a very different class, doing different thing. That is why they are named differently. And strictly speaking, this file is really very-very non-main =) Either way, if you have preferrence how to name these classes, please let me know and I'll rename them. For now, those were the best names I could come up with.

src/main/java/com/google/api/generator/debug/CodeGeneratorRequestDumper.java
```diff
@@ -0,0 +1,37 @@
...
+  public static void main(String[] args) throws IOException {
+    ExtensionRegistry registry = ExtensionRegistry.newInstance();
+    Main.registerAllExtensions(registry);
+    CodeGeneratorRequest request = CodeGeneratorRequest.parseFrom(System.in, registry);
+
+    CodeGeneratorResponse.Builder response = CodeGeneratorResponse.newBuilder();
+    response
+        .setSupportedFeatures(CodeGeneratorResponse.Feature.FEATURE_PROTO3_OPTIONAL_VALUE)
```
vam-google, line 31: @miraleung (from https://github.com/googleapis/gapic-generator-java/pull/767): Curious - is this still explicitly needed now that it's a public feature?

src/main/java/com/google/api/generator/debug/MainFromFile.java
```diff
@@ -0,0 +1,43 @@
...
+import com.google.protobuf.ExtensionRegistry;
+import com.google.protobuf.compiler.PluginProtos.CodeGeneratorRequest;
+import com.google.protobuf.compiler.PluginProtos.CodeGeneratorResponse;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+public class MainFromFile {
```
vam-google, line 28: @miraleung (from https://github.com/googleapis/gapic-generator-java/pull/767): Could we give this a more descriptive name, e.g. GapicToFileMain

vam-google: Ok, I'll rename. WDYT about naming the Main class to GapicMain also?

vam-google:
@miraleung I've migrated your comments from the other PR here. Please if anything like this happens, just let me know you prefer a PR to be rebased, so I can rebase it first and then we can continue with the review. Otherwise it generates a lot of unnecessary work for both of us (like migrating these comments).

BUILD.bazel
vam-google: These are not rules, these are java_binaries used by one of the rules only. I clarified the purpose of the binaries in the comments

src/main/java/com/google/api/generator/debug/CodeGeneratorRequestDumper.java
vam-google: I don't know, but really don't want to mess with it right now (who knows if somebody is still using this generator with older version of protoc). Unless you have strong opinion about it, I think keeping it as is is ok and safest option.

vam-google:
@miraleung PTAL

vam-google:
@miraleung PTAL

src/main/java/com/google/api/generator/debug/CodeGeneratorRequestDumper.java
miraleung: MainStandalone could be something like `CodeGeneratorRequestFileToGapicMain`?

BUILD.bazel
miraleung: Thanks for adding the comments. Could we copy-paste these into the Java src files as well? Up to you whether to replace these comments with a pointer to the Java source, or leave these as they are.

Could we also add a usage example? Up to you whether that goes here or in the Java source comments.

miraleung (approved):
LGTM with comments addressed.

src/main/java/com/google/api/generator/debug/CodeGeneratorRequestDumper.java
vam-google: Renamed

BUILD.bazel
vam-google: Done

PR #767, opened 2021-06-14
feat: Add DIREGAPIC-specific pagination

vam-google:


BUILD.bazel
```diff
@@ -45,6 +45,34 @@ java_binary(
     ],
 )
 
+java_binary(
```
miraleung, line 48: Nit: Can we add a summary comment above these two rules, which also indicate these are debugging tools?

src/main/java/com/google/api/generator/Main.java
```diff
@@ -37,7 +35,7 @@ public static void main(String[] args)
   }
 
   /** Register all extensions needed to process API protofiles. */
-  private static void registerAllExtensions(ExtensionRegistry extensionRegistry) {
+  public static void registerAllExtensions(ExtensionRegistry extensionRegistry) {
```
miraleung, line 38: Could we move this into a separate file? Having the main classes all call something like `Registry.registerAllExtensions` has more consistent semantics than `Main.registerAllExtensions`.

Up to you whether to put this in a subpackage or keep it here.

src/main/java/com/google/api/generator/debug/CodeGeneratorRequestDumper.java
```diff
@@ -0,0 +1,37 @@
...
+
+public class CodeGeneratorRequestDumper {
+  public static void main(String[] args) throws IOException {
+    ExtensionRegistry registry = ExtensionRegistry.newInstance();
+    Main.registerAllExtensions(registry);
+    CodeGeneratorRequest request = CodeGeneratorRequest.parseFrom(System.in, registry);
+
+    CodeGeneratorResponse.Builder response = CodeGeneratorResponse.newBuilder();
+    response
+        .setSupportedFeatures(CodeGeneratorResponse.Feature.FEATURE_PROTO3_OPTIONAL_VALUE)
```
miraleung, line 31: Curious - is this still explicitly needed now that it's a public feature?

src/main/java/com/google/api/generator/debug/MainFromFile.java
```diff
@@ -0,0 +1,43 @@
...
+import com.google.protobuf.ExtensionRegistry;
+import com.google.protobuf.compiler.PluginProtos.CodeGeneratorRequest;
+import com.google.protobuf.compiler.PluginProtos.CodeGeneratorResponse;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+public class MainFromFile {
```
miraleung, line 28: Could we give this a more descriptive name, e.g. GapicToFileMain?

src/main/java/com/google/api/generator/debug/CodeGeneratorRequestDumper.java
```diff
@@ -0,0 +1,37 @@
...
+
+package com.google.api.generator.debug;
+
+import com.google.api.generator.Main;
+import com.google.protobuf.ExtensionRegistry;
+import com.google.protobuf.compiler.PluginProtos.CodeGeneratorRequest;
+import com.google.protobuf.compiler.PluginProtos.CodeGeneratorResponse;
+import java.io.IOException;
+
+public class CodeGeneratorRequestDumper {
```
miraleung, line 23: If we're using `Main` in the class names, could we append it here for consistency? Alternatively, remove `Main` from the other debugger's main class's name.

src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceClientTestClassComposer.java
```diff
@@ -580,19 +590,46 @@ private MethodDefinition createRpcTestMethod(
...
-                      "get%sList", JavaStyle.toUpperCamelCase(repeatedPagedResultsField.name())))
-              .build();
-      expectedPagedResponseExpr =
-          MethodInvocationExpr.builder()
-              .setExprReferenceExpr(expectedPagedResponseExpr)
-              .setMethodName("get")
-              .setArguments(zeroExpr)
-              .build();
+
+      Expr expectedPagedResponseExpr;
```
miraleung, line 594: Consider adding a brief code snippet here (which we should have done more often throughout this codebase), similar to [this one](https://github.com/googleapis/gapic-generator-java/blob/master/src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java#L381).

src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
```diff
@@ -346,18 +347,39 @@ public static Expr createSimplePagedResponse(
...
-            .setArguments(
-                MethodInvocationExpr.builder()
-                    .setStaticReferenceType(
-                        TypeNode.withReference(ConcreteReference.withClazz(Arrays.class)))
-                    .setMethodName("asList")
-                    .setArguments(responseElementVarExpr)
-                    .build())
-            .build();
+    if (isMap) {
+      pagedResponseExpr =
```
miraleung, line 351: Ditto with the sample code, since this is longer and has some nested nodes.

src/main/java/com/google/api/generator/gapic/model/Message.java
```diff
@@ -71,7 +73,22 @@ public boolean hasResource() {
 
   /** Returns the first list repeated field in a message, unwrapped from its list type. */
   @Nullable
-  public Field findAndUnwrapFirstRepeatedField() {
+  public Field findAndUnwrapPaginatedRepeatedField() {
+    for (Field field : fields()) {
+      if (field.isMap()) {
+        List<Reference> repeatedGenericMapRefs = field.type().reference().generics();
+
+        TypeNode paginateType =
```
miraleung, line 81: Nit: Did you mean `paginatedType` or `paginationType`?

src/main/java/com/google/api/generator/gapic/model/Method.java
```diff
@@ -41,6 +41,9 @@
 
   public abstract boolean isPaged();
 
+  @Nullable
+  public abstract String pageSizeFieldName();
```
miraleung, line 45: Should we add a check on `build()` so that folks can't set this without `isPaged`?

src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceStubSettingsClassComposer.java
```diff
@@ -573,14 +572,56 @@ private static Expr createPagedListDescriptorAssignExpr(
...
                 .build());
-    Expr getResponsesListExpr =
-        MethodInvocationExpr.builder()
-            .setExprReferenceExpr(payloadVarExpr)
-            .setMethodName(
-                String.format("get%sList", JavaStyle.toUpperCamelCase(repeatedFieldName)))
-            .setReturnType(returnType)
-            .build();
 
+    Expr getResponsesExpr;
```
miraleung, line 576: Nit: Assign `null` for consistency (since that seems to be what we're doing below)?

src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -787,9 +802,22 @@ static boolean parseIsPaged(
...
-        && inputMessage.fieldMap().containsKey("page_token")
-        && outputMessage.fieldMap().containsKey("next_page_token");
+    String pagedFieldName = null;
+
+    if (inputMessage.fieldMap().containsKey("page_token")
+        && outputMessage.fieldMap().containsKey("next_page_token")) {
+      List<String> fieldNames = new ArrayList<>();
+      fieldNames.add("page_size");
+      if (transport == Transport.REST) {
+        fieldNames.add("max_results");
```
miraleung, line 812: Do we need to do something to check that either `max_results` xor `page_size` are present, and not both?

miraleung (changes requested):
Could we base this PR on #765 if possible?

src/main/java/com/google/api/generator/debug/CodeGeneratorRequestDumper.java
vam-google: Those are different things basically. The `MainFromFile` is literally another variation of the `Main` class. This, on the other hand, is a very different class, doing different thing. That is why they are named differently. And strictly speaking, this file is really very-very non-main =) Either way, if you have preferrence how to name these classes, please let me know and I'll rename them. For now, those were the best names I could come up with.

src/main/java/com/google/api/generator/debug/MainFromFile.java
vam-google: Ok, I'll rename. WDYT about naming the Main class to `GapicMain` also?

src/main/java/com/google/api/generator/Main.java
vam-google: Ok

vam-google:
@miraleung There was a merge commit with conflicts in between, so rebasing was painful. I just cleaned up the other pr form this one manually, since it was easier (just 6 files with straightforward changes).

src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceClientTestClassComposer.java
vam-google: Added per code snippet (not for the full method, since the whole method is big and  has branching, so it would be not possible to come up with full method snippet which covers all important branches.

src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
vam-google: Added two code snippets (one for maps, one for arrays)

src/main/java/com/google/api/generator/gapic/model/Message.java
vam-google: Yeah, a typo, I think I meant `paginatedType`, renamed.

src/main/java/com/google/api/generator/gapic/model/Method.java
vam-google: We can, but doubt it will give practical benefits, but will make the code more verbose and will add too  much manual boilerplate stuff in what is supposed to be boilerplate-autogenerated class (`@AutoValue`). Adding little checks like this moves us from simplistic semantics of `AutoValue` classes, where most of the stuff is generated  in a simplistic and standard way, to not that far away from relying on manually created builders (because this is what we will be doing in `autoValue()` method, which is supposed to be is rarely used feature).

src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceStubSettingsClassComposer.java
vam-google: Redundant `null` assignments like this are strictly discouraged and to some extent are a vestige from (old-standard) C/C++ times where compiler was not able to determine that the variable would be unequivocally assigned after going through if-else block(s). Being able to determine that a variable would get assigned under all possible execution branches and not requiring a default value assignment was a "cool" feature of Java compiler back then

The benefit of avoiding default null assignment in cases like this is, if somebody forgets to assign either `if` or `else` versions of this variable (both meaningful, not just `null`), we would get java compiler error. If we have default assignment and somebody forgets to assign the variable value at one of the potential execution branches (either initially or during future refactorings) there will be no compiler error and we may end up having `null` instead of actually forgotten true assigned value for the variable.

src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
vam-google: No. This logic explicitly prefers `page_size` over `max_results` if both are present, for safer backward compatibility (i.e. to not let DIREGAPIC-speficif logic mess with normal GAPICS, if somebody for whatever reason decides to add `max_results` field into a regular paginated gapic method.

src/main/java/com/google/api/generator/gapic/model/Method.java
miraleung: I'd prefer that we add this check, so that we can catch slip-ups as early as compile-time (which does occur in practice).

src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -787,9 +802,22 @@ static boolean parseIsPaged(
     // This should technically handle the absence of either of these fields (aip.dev/158), but we
     // gate on their collective presence to ensure the generated surface is backawrds-compatible
     // with monolith-gnerated libraries.
-    return inputMessage.fieldMap().containsKey("page_size")
-        && inputMessage.fieldMap().containsKey("page_token")
-        && outputMessage.fieldMap().containsKey("next_page_token");
+    String pagedFieldName = null;
```
miraleung, line 805: As per the comment thread above, can we remove this null-assignment?

src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
miraleung: SGTM, could we add a one-line comment to that effect, so that this behavior is clear to future readers who may be unfamiliar with this codebase?

miraleung (changes requested):

src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
vam-google: Added

src/main/java/com/google/api/generator/gapic/model/Method.java
vam-google: Added

src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
vam-google: This is a legit null assignment, as there is no else block. I.e. if there are not signs of field being paginated, `null` is a legit value, representing non-paginated case.

vam-google:
@miraleung PTAL

miraleung (approved):

PR #774, opened 2021-06-18
feat: Implement field presence support for DIREGAPIC

vam-google:


src/main/java/com/google/api/generator/gapic/model/HttpBindings.java
```diff
@@ -29,15 +29,31 @@
...
+    public abstract String name();
+
+    public abstract boolean isOptional();
+
+    public static HttpBinding create(String name, boolean isOptional) {
+      return new AutoValue_HttpBindings_HttpBinding(name, isOptional);
+    }
+
+    @Override
+    public int compareTo(HttpBinding o) {
```
miraleung, line 43: Do we need to account for other fields? If not, can we add an explanatory comment?

miraleung (approved):
LGTM with one comment on `HttpBindings.compare`.

src/main/java/com/google/api/generator/gapic/model/HttpBindings.java
vam-google: Good point! It does not matter for our use case (since protobuf guarantees that there can't be optional and non-optional fields in the same message with the same name), but by omitting `isOptional()` in `compareTo()` I made `compareTo()` inconsistent with `equals()`, which is plain wrong and may lead to nondeterministic behavior of sorted collections. Fixed.

PR #778, opened 2021-06-21
fix(bazel): Eradicate monolith deps from Java µgen repo

miraleung:
Any golden file updates are due to bumping the googleapis hash. This can't go in until googleapis_discovery has removed the monolith as well.


vam-google (approved):
LGTM

PR #781, opened 2021-06-25
fix(protoc): Mirror protoc's field name conflict resolution logic in client generation [ggj]

miraleung:
Mirrors protoc’s [field name conflict handling in Java](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/compiler/java/java_context.cc#L161). This is needed to enable GAPIC coverage for additional APIs.

This change will result in changes [like these](https://github.com/googleapis/gapic-generator-java/commit/82a3836e579890226bbf440592afbf2695ce4da7) in the generated libraries.

cc @alexander-fenster


src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -830,14 +830,47 @@ static String sanitizeDefaultHost(String rawDefaultHost) {
     // Sort by ascending field index order. This is important for paged responses, where the first
     // repeated type is taken.
     fields.sort((f1, f2) -> f1.getIndex() - f2.getIndex());
+
+    // Mirror protoc's name conflict resolution behavior for fields.
+    // If a singular field's name equals that of a repeated field with "Count" or "List" suffixed,
+    // append the protobuf's field number to both fields' names.
+    // See:
+    // https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/compiler/java/java_context.cc#L60
```
vam-google, line 838: Sooner or later java_context.cc will evolve and it is likely that on L60 there will be something else. Please consider putting a link to a specific commit/tag.

src/main/java/com/google/api/generator/gapic/model/Field.java
```diff
@@ -73,6 +85,7 @@ public boolean equals(Object o) {
   @Override
   public int hashCode() {
     return 17 * name().hashCode()
+        + 57 * originalName().hashCode()
```
vam-google, line 88: not like it is of any importance, but why not 31? =)

vam-google (approved):
LGTM, just a few minor comments

src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
miraleung: Done.

src/main/java/com/google/api/generator/gapic/model/Field.java
miraleung: Done.

PR #785, opened 2021-06-29
fix(resnames)!: Do not generate unused resname classes [ggj]

miraleung:
Prevents resource name helper classes from being generated for definition-only resource names and references. Previously, classes for any such definitions or references were generated in the monolith, resulting in unused code in the GAPIC client libraries. With the pending major version update for Java client libraries, we can stop generating these classes.

This branch will not be merged into main until all stakeholders are ready for the major version update.

Relevant PRs that added the backwards-compatibility logic:
  - [#595](https://github.com/googleapis/gapic-generator-java/pull/595)
  - [#597 - resolving name conflicts](https://github.com/googleapis/gapic-generator-java/pull/597)

cc @chingor13


Neenu1995 (approved):

miraleung:
For posterity - "usage" in this case pertains to "usage in the generated GAPIC library." That is, if there is no field used in a `method_signature` that has a `resource_reference`, then it is not considered "used." A message definition with a `resource_reference` on a field not used in a `method_signature` annotation will never be called within the client library, so the corresponding resource name helper class may likely be unused. Hence, we will generate helper classes only for the former case.

CC @BenWhitehead

PR #787, opened 2021-07-09
fix: (rest transport) Add `@BetaApi` to the generated `TransportServiceFactory` class and lro-specific method

vam-google:
This is to make implementing LRO post-GA safer. Note, `TransportServiceFactory` class is public for technical reasons, and is not supposed to be used by users directly.

### DIFF
--- a/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceCallableFactoryClassComposer.java
+++ b/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceCallableFactoryClassComposer.java
@@ -137,8 +137,7 @@ public abstract class AbstractServiceCallableFactoryClassComposer implements Cla
         /*callSettingsVariantName=*/ methodVariantName,
         /*callSettingsTemplateObjects=*/ methodTemplateNames.stream()
             .map(n -> (Object) n)
-            .collect(Collectors.toList()),
-        Collections.emptyList());
+            .collect(Collectors.toList()));
   }
 
   protected MethodDefinition createPagedCallableMethod(TypeStore typeStore) {
@@ -160,8 +159,7 @@ public abstract class AbstractServiceCallableFactoryClassComposer implements Cla
         /*callSettingsVariantName=*/ methodVariantName,
         /*callSettingsTemplateObjects=*/ methodTemplateNames.stream()
             .map(n -> (Object) n)
-            .collect(Collectors.toList()),
-        Collections.emptyList());
+            .collect(Collectors.toList()));
   }
 
   protected MethodDefinition createBatchingCallableMethod(TypeStore typeStore) {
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceCallableFactoryClassComposer.java
```diff
@@ -191,7 +195,8 @@ protected MethodDefinition createGenericCallableMethod(
       String methodVariantName,
       List<Object> transportCallSettingsTemplateObjects,
       String callSettingsVariantName,
-      List<Object> callSettingsTemplateObjects) {
+      List<Object> callSettingsTemplateObjects,
+      List<AnnotationNode> annotations) {
```
miraleung, line 199: What about having a thin wrapper around this that doesn't take `annotations`? Then the invocations that don't need `annotations` won't need to pass in an empty list/

src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceCallableFactoryClassComposer.java
```diff
@@ -44,6 +46,18 @@ public static HttpJsonServiceCallableFactoryClassComposer instance() {
     return INSTANCE;
   }
 
+  @Override
+  protected List<AnnotationNode> createClassAnnotations(Service service, TypeStore typeStore) {
+    List<AnnotationNode> annotations = super.createClassAnnotations(service, typeStore);
+    // Always add @BetaApi annotation to the generated CallableFactory for now. It is a public class
```
miraleung, line 52: Could we please add the criteria for removing the `BetaApi` annotation? Consider also adding that to the generated comments so end users understand what to expect.

src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceCallableFactoryClassComposer.java
```diff
@@ -63,6 +77,15 @@ protected MethodDefinition createOperationCallableMethod(TypeStore typeStore) {
...
     List<String> methodTemplateNames =
         Arrays.asList(requestTemplateName, responseTemplateName, "MetadataT");
+
+    // Always add @BetaApi annotation to the generated createOperationCallable()method for now,
+    // until LRO is fully implemented.
+    AnnotationNode betaAnnotation =
+        AnnotationNode.withTypeAndDescription(
+            typeStore.get("BetaApi"),
+            "The surface for long-running operations is not stable yet and may change in the"
+                + " future.");
```
miraleung, line 87: Ditto - can we use a concrete termination criteria instead of sometime in the future?

miraleung (changes requested):

src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceCallableFactoryClassComposer.java
vam-google: I can do that if you prefer it that way.

vam-google:
@miraleung PTAL

miraleung (approved):

PR #793, opened 2021-07-20
chore: Update googleapis hash (incl. mono removal) [ggj]

miraleung:
Golden file updates are due to the proto updates in googleapis.


vam-google (approved):
LGTM

PR #794, opened 2021-07-22
feat: enable self signed jwt for gapic clients

arithmetic1728:
For GAPIC clients, this PR enables self signed jwt feature by changing the implementation of `defaultCredentialsProviderBuilder`  from 
```
public static GoogleCredentialsProvider.Builder defaultCredentialsProviderBuilder() {
    return GoogleCredentialsProvider.newBuilder().setScopesToApply(DEFAULT_SERVICE_SCOPES);
}
```
to
```
    return GoogleCredentialsProvider.newBuilder().setScopesToApply(DEFAULT_SERVICE_SCOPES)
        .setUseJwtAccessWithScope(true);
```

There is no changes for DIREGAPIC clients.

This PR relies on https://github.com/googleapis/gax-java/pull/1420.

### DIFF
--- a/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceStubSettingsClassComposer.java
+++ b/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceStubSettingsClassComposer.java
@@ -137,10 +137,6 @@ public abstract class AbstractServiceStubSettingsClassComposer implements ClassC
 
   private static final String DOT = ".";
 
-  protected static final TypeStore FIXED_TYPESTORE = createStaticTypes();
-
-  protected static final VariableExpr DEFAULT_SERVICE_SCOPES_VAR_EXPR =
-      createDefaultServiceScopesVarExpr();
   private static final VariableExpr NESTED_UNARY_METHOD_SETTINGS_BUILDERS_VAR_EXPR =
       createNestedUnaryMethodSettingsBuildersVarExpr();
   private static final VariableExpr NESTED_RETRYABLE_CODE_DEFINITIONS_VAR_EXPR =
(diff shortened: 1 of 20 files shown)


src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceStubSettingsClassComposer.java
```diff
@@ -139,7 +139,7 @@
 
   protected static final TypeStore FIXED_TYPESTORE = createStaticTypes();
 
-  private static final VariableExpr DEFAULT_SERVICE_SCOPES_VAR_EXPR =
+  protected static final VariableExpr DEFAULT_SERVICE_SCOPES_VAR_EXPR =
```
miraleung, line 142: Could we please move this into another section, separate from the other `private static final`s?

src/main/java/com/google/api/generator/gapic/composer/grpc/ServiceStubSettingsClassComposer.java
```diff
@@ -94,6 +96,43 @@ protected MethodDefinition createDefaultTransportTransportProviderBuilderMethod(
         .build();
   }
 
+  @Override
```
miraleung, line 99: Could we please add a comment to indicate which part that is specific to gRPC GAPICs?

miraleung (approved):

src/main/java/com/google/api/generator/gapic/composer/grpc/ServiceStubSettingsClassComposer.java
arithmetic1728: added a comment inside this method

src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceStubSettingsClassComposer.java
arithmetic1728: moved both `protect static final` to the end of all `private static final`s

arithmetic1728:
added `do not merge` as we are waiting for https://github.com/googleapis/gax-java/pull/1420

PR #800, opened 2021-07-29
demo: Structure for HttpJsonServiceStubClassComposer DO NOT SUBMIT

miraleung:
Golden files not yet updated, and remaining OperationSnapshot missing type error, left as an exercise for the reader.


miraleung:
@GabrielGonzalezDiaz FYI

PR #801, opened 2021-07-30
feat(ast): Add support for Throwable causes [ggj]

miraleung:
This is needed for #804.

### DIFF
--- a/src/main/java/com/google/api/generator/engine/ast/ThrowExpr.java
+++ b/src/main/java/com/google/api/generator/engine/ast/ThrowExpr.java
@@ -28,6 +28,9 @@ public abstract class ThrowExpr implements Expr {
   @Nullable
   public abstract Expr messageExpr();
 
+  @Nullable
+  public abstract Expr causeExpr();
+
   @Override
   public void accept(AstNodeVisitor visitor) {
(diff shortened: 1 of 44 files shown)


src/main/java/com/google/api/generator/engine/ast/ThrowExpr.java
```diff
@@ -47,22 +50,37 @@ public Builder setMessageExpr(String message) {
...
+
       if (messageExpr() != null) {
         Preconditions.checkState(
             messageExpr().type().equals(TypeNode.STRING),
             String.format("Message expression type must be a string for exception %s", type()));
       }
+
+      if (causeExpr() != null) {
+        Preconditions.checkState(
+            ConcreteReference.withClazz(Throwable.class)
```
vam-google, line 77: Minor, but probably we don't have to create reference each time, and can have it as a static final constant in this class.

vam-google (approved):
LGTM

src/main/java/com/google/api/generator/engine/ast/ThrowExpr.java
miraleung: Done, added to TypeNode.

PR #760, opened 2021-06-09
fix(mocks): Use java.lang.Object if there are protos named 'Object' [ggj]

miraleung:

PR #802, opened 2021-07-30
chore: add baseline storage golden files [ggj]

miraleung:
Storage has several generation-time edge cases (e.g. messages named `Object`, which cause a naming collision in Java). This will also be used as a baseline diff for #803.

### DIFF
--- a/test/integration/BUILD.bazel
+++ b/test/integration/BUILD.bazel
@@ -30,6 +30,7 @@ INTEGRATION_TEST_LIBRARIES = [
     "pubsub",  # Special=case handling for "_deleted-topic_" resource name patterns.
     "logging",  # Java package remapping in gapic.yaml.
     "redis",  # Has a gapic.yaml.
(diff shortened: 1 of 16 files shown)


vam-google (approved):
LGTM, though I'm a bit concerned by the amount of baseline files (if there are too many it is hard to make sense of them and most likely peoploe will just start blindly updating them, insetad of carefully checking for differences in case of regression.

miraleung:
> LGTM, though I'm a bit concerned by the amount of baseline files (if there are too many it is hard to make sense of them and most likely peoploe will just start blindly updating them, insetad of carefully checking for differences in case of regression.

This is primarily because Storage is relatively newer to GAPICs (AFAIU), and these should have been added to catch #760 as well. But if we're eventually confident that the unit tests are sufficient, or we have a general way to catch such naming conflicts, feel free to remove them at a later point.

PR #804, opened 2021-07-30
fix: prevent hanging by call backgroundResources.close() on stub.close() [ggj]

### LINKED ISSUE #792: stub.close() should call close() on background resources
Currently generated stubs delegate close() to BackgroundResourceAggregation#shutdown(). Which will effectively end up calling channel.shutdown() followed immediately by executor.shutdown(). This can result in RPCs hanging indefinitely.

Steps leading to a hang:

future = stub.someCallable().futureCall(); // start rpc
// prevents new RPCs from starting, but existing RPCs can continue
// executor is shutdown
stub.close();
// grpc receives a response, but doesn't have executor to notify the future
// so this will hang indefinitely
future.get()

To avoid this situation, stub.close() should call close() on background resources, which wait for the channel to terminate before shutting down the executor.

See https://github.com/googleapis/java-bigtable/pull/916 for more details

miraleung:
Closes #792.


src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/GrpcPublisherStub.golden
```diff
@@ -430,7 +430,11 @@ public class GrpcPublisherStub extends PublisherStub {
 
   @Override
   public final void close() {
-    shutdown();
+    try {
```
vam-google, line 433: Looks like this (exception wrapping) exists solely because `backgroundResources.close()` is checked (`throws Exception`), but the `EchoStub.close()`is declared as non-checked, so we need do something like that for things to compile. Either way it will be a breaking behavioral change. To mitigate it, I think we can wrap only what we must (checked exceptions), but throw as is all the unchecked exceptions, so please consider generating somethign like this instead:

```java
   try {
      backgroundResources.close();
    } catch (RungimeException e) {
      thrown e;
    } catch (Exception e) {
      throw new IllegalStateException("Failed to close resource", e);
    }
  }
```
This will let us to avoid unnecessary exception wrapping, making this change differ less from the previous behavior.

vam-google (commented):
I was a bit hesitant originally, but giving it a closer look, seems like this PR really does the right thing: `close()` methods call `close()` underneath (instead of `shutdown()`, which was the old behavior). I mean `shutdown()` must call `shutdown()` for child resources, `close()` must call `close()` for child resources (instead of `shutdown()`), and this PR brings the generated code to that correct state. 

Please address one comment about exception wraping before mergin.

src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/GrpcPublisherStub.golden
miraleung: Done.

PR #810, opened 2021-08-02
feat(ast): support throwing all kinds of expressions [ggj]

### LINKED ISSUE #808: [ggj] Support throwing any kind of expression
This will enable expressions like `throw e`.

miraleung:
Closes #808. This is needed for #804, so we can rethrow exceptions.


vam-google (approved):
LGTM

PR #815, opened 2021-08-03
feat(java8): lambda-ize single-method anon classes [ggj]

miraleung:


Neenu1995 (approved):

vam-google (approved):
Very nice to have lambdas! Though this will mean I will have some painful marge conflicts, but I think it is worth it.

PR #821, opened 2021-08-06
fix: bring back unused resnames for Ads [ggj]

### LINKED ISSUE #819: v2.0.0's unused resname class removal deletes useful classes
The unused resname class removal in #814 causes the removal of ~4 *Name.java classes from googleads API. It's likely doing its job, as I think the resource names in question are indeed not referenced anywhere in a method_signature or a resource_reference. 

However, googleads has a generic query `search` method for fetching data, and some of the resource name classes in question are valid filter predicates in our search query language. Thus, generating the *Name.java classes for them is still a useful developer surface that can assist our users in constructing queries that filter on resource name values, even if they're not formally used elsewhere in the GAPIC surface. 

Can we either revert this or at least supply a flag value we can set in our Bazel java targets to turn off this behavior?

miraleung:
Closes #819. Undoes #785 for Ads under an allowlist. With this change, the following extra unused resource names are generated for Ads' v8 on comparing the one generated with release 2.0.0 (of the microgenerator) and one generated with this fix:

```sh
$ cd /path/to/output/orig
$ diff <(find googleads-java/ -name "*Name.java" | cut -d  '/' -f2-) <(find ../fixed/googleads-java/ -name "*Name.java" | cut -d  '/' -f4-)
> proto-googleads-java/src/main/java/com/google/ads/googleads/v8/resources/InvoiceName.java
43a45
> proto-googleads-java/src/main/java/com/google/ads/googleads/v8/resources/CallViewName.java
52a55
> proto-googleads-java/src/main/java/com/google/ads/googleads/v8/resources/PaymentsAccountName.java
87a91
> proto-googleads-java/src/main/java/com/google/ads/googleads/v8/resources/ChangeEventName.java
```

CC @aohren

### DIFF
--- a/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
+++ b/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
@@ -142,8 +142,11 @@ public class Parser {
     Map<String, ResourceName> resourceNames = parseResourceNames(request);
     messages = updateResourceNamesInMessages(messages, resourceNames.values());
 
-    // Contains only resource names that are actually used. That is, resource name definitions
-    // or references that are simply defined, but not used, will not have corresponding Java helper
+    // Contains only resource names that are actually used. Usage refers to the presence of a
+    // request message's field in an RPC's method_signature annotation. That is,  resource name
+    // definitions
+    // or references that are simply defined, but not used in such a manner, will not have
+    // corresponding Java helper
     // classes generated.
     Set<ResourceName> outputArgResourceNames = new HashSet<>();
     List<Service> mixinServices = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


miraleung:
Merging due to stakeholder timelines.

src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -160,6 +163,37 @@ public static GapicContext parse(CodeGeneratorRequest request) {
             transport);
 
     Preconditions.checkState(!services.isEmpty(), "No services found to generate");
+
+    // Temporary workaround for Ads, who still need these resource names.
+    if (services.get(0).protoPakkage().startsWith("google.ads.googleads.v")) {
```
vam-google, line 168: Can you please put some TODO here (I hope we do plan to get rid of google.ads.googleads.v specific hack, don't we)?

vam-google (commented):
LGTM

src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
miraleung: Will do. Up to you whether we should bring back all these "unused" resource name helper classes altogether, or just for Ads.

PR #823, opened 2021-08-12
chore: add hardcoded LRO implementation for rest transport

GabrielGonzalezDiaz:
# Adding hardcoded machinery to composer/rest

### HttpJsonServiceStubClassComposer

Added 2 new methods to generate `.setSnapshotFactory()` and `.setPollingRequestFactory()` using the AST model already established.

Problem:
This version is currently hard coded as it does not use the new annotations for LRO.

### HttpJsonServiceCallableFactoryClassComposer

Modified 1 method to generate `createOperationCallable()` using the AST model already established. The previous version generated the header of the method with an empty body and returned null. I took the already generated `MethodDefinition` to add the variables needed in the body for the return statement.

Problem:
Generic types in the generics for a new object are not currently supported, so to get around that I created a `VaporReference` with the same `pakkage` as the class being generated such that the import will be ignored. This is a temporary solution.

### DIFF
--- a/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
+++ b/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
@@ -164,6 +164,8 @@ public class Parser {
 
     Preconditions.checkState(!services.isEmpty(), "No services found to generate");
 
+    // TODO(vam-google): Figure out whether we should keep this allowlist or bring
+    // back the unused resource names for all APIs.
     // Temporary workaround for Ads, who still need these resource names.
     if (services.get(0).protoPakkage().startsWith("google.ads.googleads.v")) {
       Function<ResourceName, String> typeNameFn =


src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
```diff
@@ -351,12 +363,241 @@ protected Expr createTransportSettingsInitExpr(
...
+
+    // Generate check status expression
+    MethodInvocationExpr getStatusExpr = MethodInvocationExpr.builder()
+        .setExprReferenceExpr(responseVarExpr)
+        .setMethodName("getStatus")
+        .build();
+    TypeNode statusType = TypeNode.withReference(
+        VaporReference.builder()
+            .setName("Status")
+            .setPakkage("com.google.cloud.compute.v1")
```
vam-google, line 446: For every line of code, which has now compute hardcoded, add a corresponding TODO (that will be easy to search for , when we will be replacing the hardcoded parts).

vam-google (commented):
Please include the baseline and integration test files changes (you added changes in the generator, so now it generates new stuff, and that new stuff now  must be reflected in the expected baseline and integration test files).

PR #826, opened 2021-08-17
DIREGAPIC LRO implementation without annotations

GabrielGonzalezDiaz:
Added methods to HttpJsonServiceCallableFactoryClassComposer and HttpJsonServiceStubClassComposer to generate LRO machinery without annotations. Used the established AST format.

Methods generated:
- getOperationSnapshotFactory
- setPollingRequestFactory
- createOperationCallable

TODO:
Add annotations
Find alternative way of representing generic types. Currently using a vapor reference with the same package as the class being generated.


vam-google (commented):
Please make sure that the new LRO logic is  generated only for the methods which return Operation. Apart from that LGTM

vam-google (approved):
LGTM

PR #832, opened 2021-08-30
Feat: Implement DIREGAPIC LRO annotations

GabrielGonzalezDiaz:
Generated methods that were previously hard coded now use the new annotations.
Annotations implemented:
-operation_polling_service
-operation_service
-operation_field
-operation_request_field
-operation_response_field

The parts which appends fields for `opName` are in alphabetical order based on the field name and read from annotations
The `.get[FIELD]` methods are also read from annotations.
```java
.setOperationSnapshotFactory(
    (DeleteAddressRequest request, Operation response) -> {
      StringBuilder opName = new StringBuilder(response.getName());
      opName.append(":").append(request.getProject());
      opName.append(":").append(request.getRegion());
      return HttpJsonOperationSnapshot.newBuilder()
          .setName(opName.toString())
          .setMetadata(response)
          .setDone(response.getStatus().equals(Status.DONE))
          .setResponse(response)
          .setError(response.getHttpErrorStatusCode(), response.getHttpErrorMessage())
          .build();
```

The `.set[FIELD]` methods are read from annotations. The order is as follows: First go the fields that have the `operation_response_field` annotation, and they are listed alphabetically based on the value of the annotation. The fields that do not have the annotation are listed alphabetically based on the name of the field itself.
```java
.setPollingRequestFactory(
    compoundOperationId -> {
      List<String> idComponents = Arrays.asList(compoundOperationId.split(":"));
      return GetRegionOperationRequest.newBuilder()
          .setOperation(idComponents.get(0))
          .setProject(idComponents.get(1))
          .setRegion(idComponents.get(2))
          .build();
    })
```

Fixed logic to determine which methods require new LRO machinery using `operation_polling_method` and `operation_service`.
```java
if (protoMethod.isOperationPollingMethod() || protoMethod.operationService() != null) {
  expr =
      methodMaker
          .apply(
              "setOperationSnapshotFactory",
              setOperationSnapshotFactoryExpr(protoMethod, messageTypes))
          .apply(expr);
}
if (protoMethod.isOperationPollingMethod()) {
  expr =
      methodMaker
          .apply(
              "setPollingRequestFactory",
              setPollingRequestFactoryExpr(protoMethod, messageTypes))
          .apply(expr);
}
```

Data model classes `Message.java` and `Method.java` were changed to hold the new data from the annotations


src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceCallableFactoryClassComposer.java
```diff
@@ -121,11 +120,21 @@ protected MethodDefinition createOperationCallableMethod(TypeStore typeStore) {
 
     List<Statement> createOperationCallableBody = new ArrayList<Statement>(2);
 
-    List<VariableExpr> arguments = method.arguments();
+    List<VariableExpr> arguments = new ArrayList<>(method.arguments());
+    // arguments.set(
```
vam-google, line 124: Please remove all comments like this (debugging artifacts)

src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
```diff
@@ -380,9 +384,74 @@ protected Expr createTransportSettingsInitExpr(
     return Collections.singletonList(expr);
   }
 
+  // Generates expr.methodName();
+  private MethodInvocationExpr invokeMethodfromVar(Expr expr, String methodName) {
```
vam-google, line 388: YOu probably don't need this method and use the MethodInvocationExpr construction code directly where you need it It will make it a bit more verbose, but oddly enough it will make it more idiomatic for gapic-generator-java. The same applies to the staticBuilder() method below

src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
```diff
@@ -598,36 +623,45 @@ protected Expr createTransportSettingsInitExpr(
...
+    TypeNode getOperationRequestType = TypeNode.withReference(protoMethod.inputType().reference());
+    MethodInvocationExpr newBuilderExpr = staticBuilder(getOperationRequestType);
+    Map<String, String> responseFields = inputOperationMessage.operationResponseFields();
+    // TODO: revise variable name
+    List<String> responseFieldAnnotationNames = new ArrayList<String>(responseFields.keySet());
+    Collections.sort(responseFieldAnnotationNames);
+    ImmutableSet<String> fieldNames = inputOperationMessage.fieldMap().keySet();
+    ArrayList<String> nonResponseAnnotationFields = new ArrayList<String>();
+    for (String fieldName : fieldNames) {
+      if (!responseFields.containsValue(fieldName)) {
```
vam-google, line 635: This is not correct usage of maps. `map.containsValue()` will give you the linear time (do not confuse with containsKey(), which will give amortized O(1)).

This is not critical for this application, since there will never be many elements in the map, but basically if you need both keys and values to be indexed, you may consider using two-way index (direct and reverse, assuming that values are also unique).

src/main/java/com/google/api/generator/gapic/model/OperationResponse.java
```diff
@@ -0,0 +1,62 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.generator.gapic.model;
+
+import com.google.auto.value.AutoValue;
+import javax.annotation.Nullable;
+
+// In composer:
```
vam-google, line 21: Please remove this, and all similar artifacts in comments.

src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -663,6 +697,14 @@ private static boolean isMapType(Descriptor messageDescriptor) {
                       /* protoPakkage */ protoMethod.getFile().getPackage(),
                       serviceDescriptor.getName(),
                       protoMethod.getName());
+      boolean operationPollingMethod = false;
+      if (protoMethod.getOptions().hasExtension(ExtendedOperationsProto.operationPollingMethod))
```
vam-google, line 701: Please never use if statements without `{}`. If you have ever used them like that elswere, please add curly brackets there too.

vam-google (commented):
Overall looks good! A few initial comments. I will do a more thorough review later today.

src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
GabrielGonzalezDiaz: image I created a collection used regular contains. I tried to find the runtime for this and couldn't find it, but I believe it is O(1)

repositories.bzl
```diff
@@ -76,6 +76,15 @@ def gapic_generator_java_repositories():
         ],
     )
 
+    #    local(
```
vam-google, line 79: Please remove

src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceCallableFactoryClassComposer.java
vam-google: Still here...

src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
```diff
@@ -417,57 +482,26 @@ protected Expr createTransportSettingsInitExpr(
...
-            .setMethodName("getStatus")
-            .build();
+        invokeMethodfromVar(
+            responseVarExpr, getMethodFormat(operationResponse.getStatusFieldName()));
     TypeNode statusType =
         TypeNode.withReference(
             VaporReference.builder()
                 .setName("Status")
-                .setPakkage("com.google.cloud.compute.v1")
+                .setPakkage("com.google.cloud.compute.v1.Operation")
```
vam-google, line 497: All explicit mentions of "compute" must be gone. Why is this still needed?
Please check for all other mentions of compute in your codebase. There should be 0 mentions of compute in gapic-generator-java codebase.

src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
vam-google: `Collection` is just an interface it can have many implementations. Here you did the same thing but in a more convoluded synax: `responseFieldName = map.values()` - which is still a non-indexed "array" of values. In other words, the proposed solution is exactly the same as it was, but a more verbose. 

In IntelliJ you can go to the SDK library classes implementations the same way as you go to your code implementation (ctrl + lef_mouse_click in most basic case). Reading the code is usually the best way to asses how it works and which time properties it has.

Basically only key operations are O(1) in maps. 

Side note: Please consider posting code as text instead of posting as a picture (it makes it mucheasier to work with

There is a way to post it higlighted:
```java
public static void main(String[] args) {}
```

test/integration/goldens/compute/com/google/cloud/compute/v1/stub/HttpJsonRegionOperationsStub.java
```diff
@@ -85,9 +85,7 @@
                       .build())
               .setOperationSnapshotFactory(
                   (GetRegionOperationRequest request, Operation response) -> {
-                    StringBuilder opName = new StringBuilder(response.getId());
-                    opName.append(":").append(request.getProject());
-                    opName.append(":").append(request.getRegion());
+                    StringBuilder opName = new StringBuilder(response.getName());
```
vam-google, line 88: Sorry, I missed it during the previous review, can you  generate the line 92 a bit differently. Instead of 
`response.getStatus().equals(Status.DONE)`
do it in reverse instead?
`Status.DONE.equals(response.getStatus())`

This should give the same result but is safer, because the second variant will not produce null pointer exception if `response.getStatus()` is null (it is a general well known trick of doing comparisons in reverse)

src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -663,6 +697,16 @@ private static boolean isMapType(Descriptor messageDescriptor) {
                       /* protoPakkage */ protoMethod.getFile().getPackage(),
                       serviceDescriptor.getName(),
                       protoMethod.getName());
+      boolean operationPollingMethod = false;
+      if (protoMethod.getOptions().hasExtension(ExtendedOperationsProto.operationPollingMethod)) {
```
vam-google, line 701: consider using ternary operator (`?`) for stuff like this (this is basically its mos common usage pattern).

vam-google (commented):
For the polling stubs (`HttpJsonRegionOprationStub`, i.e. the ones marked with 
`operation_polling_method = true` these parts should be generated too:

```java
this.longRunningClient =
        new HttpJsonLongRunningClient<GetRegionOperationRequest, Operation>(
            getCallable,
            getMethodDescriptor.getOperationSnapshotFactory(),
            getMethodDescriptor.getPollingRequestFactory());  
```

```java            
@Override
  public LongRunningClient longRunningClient() {
    return longRunningClient;
  }
```

```java
private final LongRunningClient longRunningClient;
```

Apart from the stuff above and some other relatively minor comments, looks good!

src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
GabrielGonzalezDiaz: So status is generated somewhere else in compute, and is used in this class. I am not using compute to generate anything. This is a Vapor reference and that is just the package that will be imported in the generated class. 

See: https://github.com/vam-google/google-cloud-compute-small-v1-lro/blob/7ff5244a579476088a222c50a7155e47e107a277/gapic-google-cloud-compute-small-v1-java/src/main/java/com/google/cloud/compute/v1/stub/HttpJsonAddressesStub.java#L46

vam-google (approved):
LGTM, assuming you will add those few missing pieces which we discussed offline

PR #837, opened 2021-09-08
Fix development doc

chanseokoh:
@meltsufin

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -15,13 +15,13 @@
 ## Running the Plugin
 
 1.  Clone [googleapis](https://github.com/googleapis/googleapis) and
-    [gapic-showcase](https://github.com/googleapis/gapic-showcase/) and install
-    protoc.
+    [gapic-showcase](https://github.com/googleapis/gapic-showcase/).
 
 2.  Copy the protos from Showcase into googleapis/google/showcase.
 
     ```sh
-    cp gapic-showcase/schema/google/showcase/v1beta1 googleapis/google/showcase/v1beta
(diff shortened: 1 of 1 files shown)


miraleung (approved):

DEVELOPMENT.md
```diff
@@ -15,13 +15,13 @@
...
 
 2.  Copy the protos from Showcase into googleapis/google/showcase.
 
     ```sh
-    cp gapic-showcase/schema/google/showcase/v1beta1 googleapis/google/showcase/v1beta
+    mkdir googleapis/google/showcase
+    cp -r gapic-showcase/schema/google/showcase/v1beta1 googleapis/google/showcase/v1beta1
     ```
 
 3.  Add the new microgenerator rules to the protobuf directory's `BUILD.bazel`
```
meltsufin, line 27: I think it's better to clarify which file needs to be modified.
```suggestion
3.  Add the new microgenerator rules to `googleapis/google/showcase/v1beta/BUILD.bazel`
```

meltsufin (commented):
For Step 4, I would also clarify that building the target should be from the `googleapis` project/directory.

DEVELOPMENT.md
chanseokoh: Good idea. Note it's `v1beta1`, not `v1beta`, which is what I'm fixing in this PR too.

meltsufin: Right, sorry, fixed.

chanseokoh:
@miraleung could you approve test workflows?

PR #840, opened 2021-09-10
Remove warnings (unused vars, dead code, missing annotations, missing generic reference, etc)

chanseokoh:
Doesn't change the code behavior. Purely removing warnings and code smells.

- unused local and private variables, and private methods
- missing `@Override` annotations
- missing generic reference
- invalid static method reference
- etc.


miraleung (approved):
Thanks!

PR #845, opened 2021-09-20
Fix preserving sorted order of a collection

chanseokoh:
`toSet()` makes `sorted()` irrelevant. Looking at other code locations, `toList()` is really intended.

Other changes are only minor simplifications.

### DIFF
--- a/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
+++ b/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
@@ -128,9 +128,7 @@ public class Parser {
     Optional<String> serviceYamlConfigPathOpt =
         PluginArgumentParser.parseServiceYamlConfigPath(request);
     Optional<com.google.api.Service> serviceYamlProtoOpt =
-        serviceYamlConfigPathOpt.isPresent()
-            ? ServiceYamlParser.parse(serviceYamlConfigPathOpt.get())
-            : Optional.empty();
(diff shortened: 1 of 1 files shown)


meltsufin (approved):
I guess it's more about just preserving order, not necessarily "sorted order".

chanseokoh:
> I guess it's more about just preserving order, not necessarily "sorted order".

Just so you know, the collection is sorted with `.sorted()` before `.collection()`. So, preserving the sorted order in that sense.

```java
            .sorted((s1, s2) -> s2.name().compareTo(s1.name()))
```

meltsufin:
> > I guess it's more about just preserving order, not necessarily "sorted order".
> 
> Just so you know, the collection is sorted with `.sorted()` before `.collect()`. So, preserving the sorted order in that sense.
> 
> ```java
>             .sorted((s1, s2) -> s2.name().compareTo(s1.name()))
>             .collect(Collectors.toList()));
> ```

I saw that, but I was curious if that's actually needed since your PR description said that it's irrelevant.

chanseokoh:
Ah, yeah, the comment on that part says sorting is intended and needed.
```java
    // Use a list to ensure ordering for deterministic tests.
    outputMixinServices.addAll(
        outputMixinServiceSet.stream()
            .sorted((s1, s2) -> s2.name().compareTo(s1.name()))
```

chanseokoh:
Umm... actually whether sorting is necessary is unclear if it's only for deterministic order, but I think it's still a good idea to sort. Eliminates any concern about which implementations of a set is being used.

PR #850, opened 2021-09-29
feat: enable self signed JWT for http

arithmetic1728:


arithmetic1728:
The codecov seems to have some calculation problem. Only 2 files are affected, one coverage stays 100% and the other increases by 0.02%.

chanseokoh (approved):
I wouldn't care about Codecov.

PR #863, opened 2021-10-21
fix: gax batching regression

### LINKED ISSUE #853: Generator does not support batching configuration
#### Environment details

  - Programming language: Java
  - OS: Any
  - Language runtime version: Java JDK 8
  - Package version: ≥2.1.4

#### Steps to reproduce

  1. Write logs using java-logging library by calling `writeLogEntries()` interface multiple time
  2. Confirm that the number of `entries.write` API calls match the number of calls to the interface

Relates to googleapis/java-logging#632

chanseokoh:
Fixes #853.

### DIFF
--- a/src/main/java/com/google/api/generator/gapic/composer/Composer.java
+++ b/src/main/java/com/google/api/generator/gapic/composer/Composer.java
@@ -153,8 +153,10 @@ public class Composer {
         s -> {
           if (context.transport() == Transport.REST) {
             // REST transport tests do not use mock services.
-          } else if (context.transport() == Transport.GRPC
-              || context.transport() == Transport.GRPC_REST) {
+          } else if (context.transport() == Transport.GRPC) {
+            clazzes.add(MockServiceClassComposer.instance().generate(context, s));
+            clazzes.add(MockServiceImplClassComposer.instance().generate(context, s));
+          } else if (context.transport() == Transport.GRPC_REST) {
             clazzes.add(MockServiceClassComposer.instance().generate(context, s));
             clazzes.add(MockServiceImplClassComposer.instance().generate(context, s));
           }
--- a/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
+++ b/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
@@ -202,7 +202,8 @@ public abstract class AbstractTransportServiceStubClassComposer implements Class
             .setAnnotations(createClassAnnotations(service))
             .setScope(ScopeNode.PUBLIC)
             .setName(className)
-            .setExtendsType(typeStore.get(ClassNames.getServiceStubClassName(service)))
+            .setExtendsType(
+                typeStore.get(getTransportContext().classNames().getServiceStubClassName(service)))
             .setStatements(classStatements)
             .setMethods(
                 createClassMethods(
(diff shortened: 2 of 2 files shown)


src/main/java/com/google/api/generator/gapic/composer/Composer.java
```diff
@@ -152,11 +152,9 @@ public static GapicPackageInfo composePackageInfo(GapicContext context) {
     services.forEach(
         s -> {
           if (context.transport() == Transport.REST) {
-            // REST transport tests donot not use mock services.
-          } else if (context.transport() == Transport.GRPC) {
-            clazzes.add(MockServiceClassComposer.instance().generate(context, s));
-            clazzes.add(MockServiceImplClassComposer.instance().generate(context, s));
-          } else if (context.transport() == Transport.GRPC_REST) {
+            // REST transport tests do not use mock services.
+          } else if (context.transport() == Transport.GRPC
```
vam-google, line 156: Please consider still keeping these separate, even if both if statemetns are the same - for consistency reasons and cleaner structure (each transport type has its own section).

src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
```diff
@@ -200,11 +202,11 @@ public GapicClass generate(GapicContext context, Service service) {
             .setAnnotations(createClassAnnotations(service))
             .setScope(ScopeNode.PUBLIC)
             .setName(className)
-            .setExtendsType(
-                typeStore.get(getTransportContext().classNames().getServiceStubClassName(service)))
+            .setExtendsType(typeStore.get(ClassNames.getServiceStubClassName(service)))
```
vam-google, line 205: This does not look correct. The expectation is to move the other direction - convert static usages of class names to the dynamic ones.

src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
```diff
@@ -473,8 +477,7 @@ protected abstract Expr createTransportSettingsInitExpr(
         argList ->
             NewObjectExpr.builder().setType(creatorMethodReturnType).setArguments(argList).build();
 
-    TypeNode stubSettingsType =
-        typeStore.get(getTransportContext().classNames().getServiceStubSettingsClassName(service));
+    TypeNode stubSettingsType = typeStore.get(ClassNames.getServiceStubSettingsClassName(service));
```
vam-google, line 480: Why is this getting replaced? Is it solely because a static method is accessed via instance reference? 

Note, since we already have `ClassNames` as part of transport context, all the static usages of ClassNames are potentially error-prone (a change in context one will not affect places, which are potentially the subject to change, because they are called statically), so basiclaly we need to convert it the other way around to reach consistency -> convert all Static ClassNames calls to dynamic instance-level ones (and making the method non-static).

The reason why those static methods are still static is historical.

src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
```diff
@@ -825,26 +836,37 @@ private Expr createCallableInitExpr(
...
-        streamName = "ServerStreaming";
-      } else if (typeName.startsWith("Bidi")) {
-        streamName = "BidiStreaming";
-      } else if (typeName.startsWith("Operation")) {
-        streamName = "Operation";
-      }
+      return Optional.of("createPagedCallable");
+    }
+
+    String typeName = callableVarExprType.reference().name();
```
vam-google, line 856: Minor: why is this style preferable? The older one (if-else) is technically shorter, so seems nicer.

vam-google (commented):
LGTM (but please roll back the the ClassNames method access change).

Please consider putting general refactoring chagnes in separate PRs (this one is Ok to keep as is, but for any future changes like this), as most of this PR changes are not related to its description (fixing batch regression) and it makes it hard to reason about and review the actual chagnes.

PR #874, opened 2021-11-30
feat: update generated samples to executable format

eaball35:
b/207550139, b/207550144

go/java-sample-gen
go/java-sample-gen-bugs


amanda-tarafa (approved):
LGTM on how the resulting samples are looking, this is great.

meltsufin (commented):
@amanda-tarafa @eaball35 
I see that this only updates inline samples, and the single-file samples as described in the design are not part of this PR.
I'm a little surprised to see essentially the whole executable sample file with imports and a wrapping application class inline. Was that the intent of the original design all along? Do you feel like this might be adding too much boilerplate repetitive code to the javadoc samples?

amanda-tarafa:
@meltsufin I'm starting a thread internally to discuss more in-depth, but basically, these aspects are up for decision for each language, as they know their community better.

src/main/java/com/google/api/generator/gapic/composer/samplecode/SampleUtil.java
```diff
@@ -0,0 +1,49 @@
...
+
+  public static MethodInvocationExpr systemOutPrint(String content) {
+    return composeSystemOutPrint(ValueExpr.withValue(StringObjectValue.withValue(content)));
+  }
+
+  public static MethodInvocationExpr systemOutPrint(VariableExpr variableExpr) {
+    return composeSystemOutPrint(variableExpr.toBuilder().setIsDecl(false).build());
+  }
+
+  static MethodInvocationExpr composeSystemOutPrint(Expr content) {
```
chanseokoh, line 36: These method are not called by non-test code. Let's remove them.

src/main/java/com/google/api/generator/gapic/composer/samplecode/ServiceClientSampleCodeComposer.java
```diff
@@ -14,6 +14,9 @@
 
 package com.google.api.generator.gapic.composer.samplecode;
 
+import static com.google.api.generator.gapic.composer.samplecode.ExecutableSampleComposer.createExecutableSample;
+import static com.google.api.generator.gapic.composer.samplecode.SampleUtil.composeSampleMethodName;
```
chanseokoh, line 18:
```suggestion
import static com.google.api.generator.gapic.composer.samplecode.ExecutableSampleComposer;
import static com.google.api.generator.gapic.composer.samplecode.SampleUtil;
```

chanseokoh (commented):
Just went over the top part of the PR. Will take another look later.

src/main/java/com/google/api/generator/gapic/composer/samplecode/SampleUtil.java
chanseokoh: That's actually what I suspected. Let's add them in the future when they are needed.

eaball35:
Forking as new branch as we decided to not make inline snippets "executable" format but instead link to full example if users wants it

src/main/java/com/google/api/generator/gapic/composer/samplecode/SampleCodeWriter.java
```diff
@@ -34,4 +36,18 @@ public static String write(List<Statement> statements) {
     // Escape character "@" in the markdown code block <pre>{@code...} tags.
     return formattedSampleCode.replaceAll("@", "{@literal @}");
   }
+
+  public static String write(ClassDefinition classDefinition) {
+    JavaWriterVisitor visitor = new JavaWriterVisitor();
+    classDefinition.accept(visitor);
+    // Escape character "@" in the markdown code block <pre>{@code...} tags.
+    return visitor.write().replaceAll("@", "{@literal @}");
```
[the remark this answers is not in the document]
eaball35: These methods are separate from each other. A class definition has statements but writing a class definition doesn't use the write statements method so it's not double-escaping. We are either writing the sample as statements (inline samples) or writing as a class definition (individual files). I actually think I'm missing the formatting here and it should be in both places.

PR #875, opened 2021-12-01
deps: upgrade gax to 2.7.1, protobuf to 3.19.1, grpc to 1.42.1, and truth to 1.1.2, upgrade Bazel in GitHub Actions, and fix build error

chanseokoh:
The versions are aligned with [those in gax-java](https://github.com/googleapis/gax-java/blob/b629488ffc7d68158158d9197695158f97229c7b/dependencies.properties#L26-L28).

Also, builds were failing for [the same reason](https://github.com/googleapis/gax-java/pull/1571) we observed in gax-java. Applying the same fix to use `PROTOBUF_MAVEN_ARTIFACTS`.


chanseokoh:
Oops, something wrong. Will fix it.

chanseokoh:
Turns out the protobuf dependency is conditionally loaded via the custom `_maybe` method, only when it has not been loaded previously.
```bazel
def _maybe(repo_rule, name, strip_repo_prefix = "", **kwargs):
    if not name.startswith(strip_repo_prefix):
        return
    repo_name = name[len(strip_repo_prefix):]
    if repo_name in native.existing_rules():
        return
    repo_rule(name = repo_name, **kwargs)
```
All in all, the build will start building successfully once we make a new release of gax-java and update `WORKSPACE` in this repo to use that version. Will leave this open until then.

chanseokoh:
@meltsufin fixed all issues. Ready for review.

chanseokoh:
@vam-google I'm syncing the versions with those in gax-java. Any concern?

meltsufin (approved):

.github/workflows/ci.yaml
```diff
@@ -7,11 +7,6 @@ name: ci
 jobs:
   build:
     runs-on: ubuntu-latest
-    container: gcr.io/gapic-images/googleapis-bazel:20210105
```
vam-google, line 10: Why was this removed?

vam-google (approved):
LGTM

.github/workflows/ci.yaml
chanseokoh: Sorry, forgot to explain in the description. Bazel version there is old, so builds fail with these changes. We encountered the same error on gax-java, and the J-Team had a discussion around this. `ubuntu-latest` comes with a recent Bazel, and we didn't find a good reason to use a custom image. We came to the agreement that we would just use `ubuntu-latest`.

PR #876, opened 2021-12-02
Upgrade Gradle build system in generated library repo and fix setup issue

chanseokoh:
Gradle 4.10.3 is ancient. We cannot delay upgrading Gradle forever.

Also, the following logic was incomplete in that it included directories (basically every non-hidden directory) that were not supposed to be included.
```gradle
include rootDir.listFiles().findAll {
    it.isDirectory() && !(it =~ ".*/\\..*") && !(it =~ "^\\..*")
}.collect {
    it.getName()
}.<String> toArray(new String[0])
```
For example, a generated library always contains a directory called `gradle/`, and it was being included. (It's a minor point, but the regex `".*/\\..*"` doesn't make sense to me either.)

Tested the changes by running `bazel build google/cloud/compute/v1:google-cloud-compute-v1-java`, unzipping `google-cloud-compute-v1-java.tar.gz`, and running `./gradlew check publishToMavenLocal`.


meltsufin (approved):

vam-google:
@chanseokoh @meltsufin This change, since now the generated templates depend on gradle 7 broke the self-service workflow, because `./gradlew clean install` does not work anymore (and there may be more issues).

chanseokoh:
#884 will fix https://github.com/googleapis/gapic-generator-java/pull/876#issuecomment-1002803437

PR #882, opened 2021-12-15
feat: ExecutableSampleComposer

eaball35:
b/207550139, b/207550144

go/java-sample-gen
go/java-sample-gen-bugs


vam-google (commented):
Thanks for doing it! Looks like a big milestone in gapic-generator's life.

Looking at the code it seems like in this specific PR the `ExecutableSampleComposer` is capable of generating a basic class skeleton (with `main()` method and stuff), but is not doing real client runnable examples, calling client library yet. Is it correct? If yes, can you please clarify in PR descritpion:

1) what kind of functionality this specific PR (in its current state) ads to gapic generator. 
2) how is it triggered (now) during generation of gapic clients
3) what would be the next step/PR?
4) do you anticipate samples differ depending on the transport used (grpc or REST)? note, client instantiation differs depending on if it is grpc or rest client.

Thanks

Issue #869, opened 2021-11-03
Add support for explicit routing headers and the google.api.routing annotation


meltsufin:
See: https://github.com/googleapis/api-common-protos/blob/master/google/api/routing.proto
go/actools-dynamic-routing-proposal

PR #887, opened 2022-01-06
feat: Support explicit dynamic routing header

blakeli0:
For issue: #869.
This PR is to support explicit dynamic routing header feature proposed in go/actools-dynamic-routing-proposal. 
A few notes:

- All examples provided in `routing.proto` are covered by golden files unit tests.
- Added validation that the field has to be of String type as suggested(Compare to implicit routing headers support primitive types as well).
-  For nested fields, when getting field values, added null check for each level to prevent null pointer exception during runtime. Something like [this](https://github.com/googleapis/gapic-generator-java/blob/165c324df1bbeada73ff99d280a7a0c814235042/src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/GrpcRoutingHeadersStub.golden#L421).
- Will create another PR to generate some unit tests for the client code.


src/main/java/com/google/api/generator/gapic/model/Message.java
```diff
@@ -80,6 +89,32 @@ public boolean hasResource() {
     return resource() != null;
   }
 
+  public void validateField(String fieldName, Map<String, Message> messageTypes) {
+    String[] subFields = fieldName.split("\\.");
```
chanseokoh, line 93: IMO, `String.split` is fine in simple and obvious cases, but make sure you understand [the pitfall](
https://errorprone.info/bugpattern/StringSplitter).

src/main/java/com/google/api/generator/gapic/model/Message.java
```diff
@@ -80,6 +89,32 @@ public boolean hasResource() {
...
+        Field field = nestedMessage.fieldMap().get(subFieldName);
+        nestedMessage = messageTypes.get(field.type().reference().fullName());
+        Preconditions.checkNotNull(
+            nestedMessage,
+            String.format(
+                MESSAGE_NOT_FOUND_ERROR_MESSAGE,
+                field.name(),
+                field.type().reference().simpleName()));
+      } else {
+        Preconditions.checkState(
```
chanseokoh, line 107: Why do we check this only for the last `subFieldName`? Should we do this for all?

src/main/java/com/google/api/generator/gapic/model/Message.java
```diff
@@ -80,6 +89,32 @@ public boolean hasResource() {
...
+            !Strings.isNullOrEmpty(subFieldName),
+            String.format(EMPTY_FIELD_ERROR_MESSAGE, nestedMessage.name()));
+        Preconditions.checkState(
+            nestedMessage.fieldMap().containsKey(subFieldName),
+            String.format(FIELD_DOES_NOT_EXIST_ERROR_MESSAGE, nestedMessage.name(), subFieldName));
```
chanseokoh, line 112: I think it's better to restructure the loop to call `Field field = nestedMessage.fieldMap().get(subFieldName)` here right after this check, since the check goes hand in hand.

Actually, I wonder if it's better to remove the branching in the loop by only looping from 0 to subFields.length - 2 and then check the last element outside the loop. But I haven't tried.

src/main/java/com/google/api/generator/gapic/protoparser/RoutingRuleParser.java
```diff
@@ -0,0 +1,88 @@
...
+  private static ImmutableSortedSet.Builder<String> getPatternBindings(String pattern) {
+    ImmutableSortedSet.Builder<String> bindings = ImmutableSortedSet.naturalOrder();
+    if (pattern.isEmpty()) {
+      return bindings;
+    }
+
+    PathTemplate template = PathTemplate.create(pattern);
+    // Filter out any unbound variable like "$0, $1, etc.
+    bindings.addAll(
+        template.vars().stream().filter(s -> !s.contains("$")).collect(Collectors.toSet()));
```
chanseokoh, line 85: Probably simply
`template.vars().stream().filter(s -> !s.contains("$")).forEach(bindings:add)`?

src/main/java/com/google/api/generator/gapic/protoparser/RoutingRuleParser.java
```diff
@@ -0,0 +1,88 @@
...
+
+      RoutingHeader routingHeader = RoutingHeader.create(field, name, pathTemplate);
+      builder.addRoutingHeader(routingHeader);
+    }
+
+    return builder.build();
+  }
+
+  // TODO: duplicate of HttpRuleParser.getPatternBindings, move to a helper class
+  private static ImmutableSortedSet.Builder<String> getPatternBindings(String pattern) {
```
chanseokoh, line 76: I think we should just return a set, not a builder. I don't see a good reason for returning a builder.

Also, since we assert `size() == 1`, I don't think we need a sorted set for our purposes.

But I admit this will be moot after all, if this is a shared helper method.

src/main/java/com/google/api/generator/gapic/composer/grpc/GrpcServiceStubClassComposer.java
```diff
@@ -335,4 +447,120 @@ private LambdaExpr createRequestParamsExtractorClassInstance(Method method) {
...
+      bodyExprs.add(addParamsMethodExpr);
+    }
+  }
+
+  private void addRequestParamsForHttpBindings(
+      Method method,
+      VariableExpr paramsVarExpr,
+      List<Expr> bodyExprs,
+      VariableExpr requestVarExpr) {
+    // TODO: Are these params used for anything else other than implicit dynamic routing?
```
chanseokoh, line 516: Hmm... can we answer this? I think it's possible that they are used for other purposes? What we have been doing is really constructing the request parameters? It's just that we are directly formulating the implicit routing header out of the request params?

src/main/java/com/google/api/generator/gapic/composer/grpc/GrpcServiceStubClassComposer.java
```diff
@@ -335,4 +447,120 @@ private LambdaExpr createRequestParamsExtractorClassInstance(Method method) {
...
+    // Can we skip this method if routing headers are configured?
+    if (!method.routingHeaders().routingHeadersList().isEmpty()) {
+      return;
+    }
+    for (HttpBinding httpBindingFieldBinding : method.httpBindings().pathParameters()) {
+      // Handle foo.bar cases by descending into the subfields.
```
chanseokoh, line 522:
```suggestion
```

src/main/java/com/google/api/generator/gapic/composer/grpc/GrpcServiceStubClassComposer.java
```diff
@@ -335,4 +447,120 @@ private LambdaExpr createRequestParamsExtractorClassInstance(Method method) {
...
         .build();
   }
+
+  private void addRequestParamsForRoutingHeders(
+      Method method,
+      List<Statement> classStatements,
+      VariableExpr paramsVarExpr,
+      List<Expr> bodyExprs,
+      VariableExpr requestVarExpr) {
+    ImmutableList<RoutingHeader> routingHeaders = method.routingHeaders().routingHeadersList();
```
chanseokoh, line 457: I think `routingHeaders()` is nullable?

src/main/java/com/google/api/generator/gapic/composer/grpc/GrpcServiceStubClassComposer.java
```diff
@@ -335,4 +447,120 @@ private LambdaExpr createRequestParamsExtractorClassInstance(Method method) {
...
+                      .setReturnType(pathTemplateType)
+                      .build())
+              .build();
+      Statement pathTemplateClassVar = ExprStatement.withExpr(pathTemplateExpr);
+      classStatements.add(pathTemplateClassVar);
+      MethodInvocationExpr addParamsMethodExpr =
+          MethodInvocationExpr.builder()
+              .setStaticReferenceType(
+                  TypeNode.withReference(ConcreteReference.withClazz(RoutingHeaderHelper.class)))
+              .setMethodName("addParams")
```
chanseokoh, line 499: Should this be singular? (https://github.com/googleapis/gax-java/pull/1598#discussion_r781574600)

chanseokoh (commented):
Initial pass. I'll take a deeper look later.

src/main/java/com/google/api/generator/gapic/model/Message.java
blakeli0: Good to know, I can switch to Guava split.

src/main/java/com/google/api/generator/gapic/composer/grpc/GrpcServiceStubClassComposer.java
blakeli0: The builder of RoutingHeaders will always create an empty map on initialization.

src/main/java/com/google/api/generator/gapic/model/Message.java
blakeli0: This validation logic is extracted from the existing code since the current implementation for implicit dynamic routing headers also need this validation, https://github.com/googleapis/gapic-generator-java/blob/b711799d7f631afd68f4a8f48b80b8ea5589da99/src/main/java/com/google/api/generator/gapic/protoparser/HttpRuleParser.java#L125. We can probably improve it if we want though.

src/main/java/com/google/api/generator/gapic/composer/grpc/GrpcServiceStubClassComposer.java
blakeli0: Yes, I agree. I'll change it to singular.

src/main/java/com/google/api/generator/gapic/model/RoutingHeaders.java
```diff
@@ -0,0 +1,56 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.api.generator.gapic.model;
+
+import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableList;
+
+@AutoValue
+public abstract class RoutingHeaders {
```
meltsufin, line 21: Please add a javadoc.

meltsufin (commented):
This is a massive amount of work and it's looking pretty good. Let's finalize the smaller PR in gax-java and then update this one. After that, I think we'll be ready to merge this in. We can do additional tests and edge case handling in follow-up PRs.

src/test/java/com/google/api/generator/gapic/protoparser/BUILD.bazel
```diff
@@ -45,6 +46,7 @@ filegroup(
         "//src/test/java/com/google/api/generator/gapic/testdata:showcase_java_proto",
         "//src/test/java/com/google/api/generator/gapic/testdata:testgapic_java_proto",
         "@com_google_api_api_common//jar",
+        "@com_google_api_grpc_proto_google_common_protos",
```
vam-google, line 49: You should not be needing this, as this dependency must be pulled transitively (many stuff in protoparser package already depends on this artifact). What fails if you remove it?

src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/GrpcTestingStub.golden
```diff
@@ -142,6 +144,11 @@ public class GrpcTestingStub extends TestingStub {
   private final GrpcOperationsStub operationsStub;
   private final GrpcStubCallableFactory callableFactory;
 
+  private static final PathTemplate PATH_TEMPLATE_GETTEST_0 =
```
vam-google, line 147: WDYT about making `GETTEST` suffix snake case (`GET_TEST`) (generator has lots of utility 
 methods to do camel-snake case conversion) and instead of adding indexes add field name in templates for easier matching of stuff and making it a bit  more readable. Also maybe make `PATH_TEMPLATE `a suffix instead of prefix.

i.e. `PATH_TEMPLATE_GETTEST_0` to become `PATH_TEMPLATE_GET_TEST_RENAME`

src/main/java/com/google/api/generator/gapic/composer/grpc/GrpcServiceStubClassComposer.java
```diff
@@ -335,4 +447,120 @@ private LambdaExpr createRequestParamsExtractorClassInstance(Method method) {
...
+                  requestFieldGetterExpr,
+                  routingHeaderKeyExpr,
+                  routingHeaderPatternExpr)
+              .build();
+
+      bodyExprs.add(addParamsMethodExpr);
+    }
+  }
+
+  private void addRequestParamsForHttpBindings(
```
vam-google, line 511: If i remember correctly, this logic is used heavily for rest transport and it does not have proper tests in this codebase yet (i.e. in other words it is very easy to break it). Please be very careful with this change. Ideally generate compute client and make sure it still works. (at least  unit tests pass).

vam-google (commented):
LGTM with a few comments.

src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/GrpcTestingStub.golden
blakeli0: Yes I have a TODO to make it proper snake style. However, I added the index on purpose because field name alone or even the combination of field name and header key is not enough to make it unique, there could be multiple patterns configured for the same field name/header key.

src/main/java/com/google/api/generator/gapic/model/Message.java
```diff
@@ -80,6 +89,32 @@ public boolean hasResource() {
     return resource() != null;
   }
 
+  public void validateField(String fieldName, Map<String, Message> messageTypes) {
```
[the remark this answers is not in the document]
blakeli0: Will make it clearer. Yes there is similar logic in HttpBindingParser to validate the existence of a field, I plan to do some refactoring to reuse it later. Even if we think refactoring is too risky, I think validate the existence of a field in a Message is a common logic that could be reused in the future.

src/main/java/com/google/api/generator/gapic/composer/grpc/GrpcServiceStubClassComposer.java
```diff
@@ -229,6 +239,141 @@ protected Expr createTransportSettingsInitExpr(
         .build();
   }
 
+  // TODO: remove this method once we finalized the approach, keep it as a reference for now.
+  private MethodDefinition createAddParamsMethod() {
```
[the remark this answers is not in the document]
blakeli0: Thanks for the advice, it did took me a few hours but it was a good practice for me to get familiar with the code base as well, so didn't lose much lol.

src/test/java/com/google/api/generator/gapic/protoparser/BUILD.bazel
blakeli0: I just tried and nothing broke, I'll remove it.

src/main/java/com/google/api/generator/gapic/composer/grpc/GrpcServiceStubClassComposer.java
blakeli0: Answered that the params are not used for any other purposes.

src/main/java/com/google/api/generator/gapic/composer/grpc/GrpcServiceStubClassComposer.java
blakeli0: Removed

src/main/java/com/google/api/generator/gapic/model/Message.java
```diff
@@ -32,6 +34,13 @@
 
 @AutoValue
 public abstract class Message {
+
+  static final String EMPTY_FIELD_ERROR_MESSAGE = "Null or empty field name found for message %s";
+  static final String FIELD_DOES_NOT_EXIST_ERROR_MESSAGE =
+      "Expected message %s to contain field %s but none found";
+  static final String MESSAGE_NOT_FOUND_ERROR_MESSAGE =
+      "No containing message found for field %s with type %s";
```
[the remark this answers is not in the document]
blakeli0: Changed to inline formatting

src/main/java/com/google/api/generator/gapic/composer/grpc/GrpcServiceStubClassComposer.java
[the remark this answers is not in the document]
blakeli0: Removed.

src/main/java/com/google/api/generator/gapic/model/RoutingHeaders.java
```diff
@@ -0,0 +1,56 @@
...
+
+  public abstract ImmutableList<RoutingHeader> routingHeadersList();
```
meltsufin, line 23: Don't use Guava API here. `List<>`.

src/main/java/com/google/api/generator/gapic/composer/grpc/GrpcServiceStubClassComposer.java
```diff
@@ -245,10 +255,171 @@ protected String getProtoRpcFullMethodName(Service protoService, Method protoMet
...
+        .setReturnType(returnType)
+        .build();
+  }
+
+  private Expr fieldValuesNotNullConditionExpr(
+      VariableExpr requestVarExpr, List<String> fieldNames) {
+    MethodInvocationExpr.Builder requestFieldGetterExprBuilder =
+        MethodInvocationExpr.builder().setExprReferenceExpr(requestVarExpr);
+    Expr fieldValuesNotNullExpr = null;
+    for (int i = 0; i < fieldNames.size(); i++) {
```
vam-google, line 391: it does not seem like there are any other usages for `i` except `get(i)`, can we just switch to foreach loop then?

src/main/java/com/google/api/generator/gapic/composer/grpc/GrpcServiceStubClassComposer.java
```diff
@@ -245,10 +255,171 @@ protected String getProtoRpcFullMethodName(Service protoService, Method protoMet
...
+      String currFieldName = fieldNames.get(i);
+      String bindingFieldMethodName =
+          String.format("get%s", JavaStyle.toUpperCamelCase(currFieldName));
+      requestFieldGetterExprBuilder =
+          requestFieldGetterExprBuilder.setMethodName(bindingFieldMethodName);
+      // set return type of each method invocation to String just to pass the validation for
+      // RelationalOperationExpr that both side of relational operation needs to be a valid equality
```
vam-google, line 398: Can  you please elaborate on this? RelationalOperationExpr seems to be used to check for equality with a `null`, which does not have any type, so it seems like it should work with any type.

vam-google (commented):
Overall looks good. Please consider adding a corresponding section in generated unit tests to cover this new feature. A lightweight version of it (like adding extra assert in existing tests) should be sufficient.

src/main/java/com/google/api/generator/gapic/composer/grpc/GrpcServiceStubClassComposer.java
```diff
@@ -245,10 +255,171 @@ protected String getProtoRpcFullMethodName(Service protoService, Method protoMet
...
+    // Overrides extract().
+    // https://github.com/googleapis/gax-java/blob/8d45d186e36ae97b789a6f89d80ae5213a773b65/gax/src/main/java/com/google/api/gax/rpc/RequestParamsExtractor.java#L55
+    return LambdaExpr.builder()
+        .setArguments(requestVarExpr.toBuilder().setIsDecl(true).build())
+        .setBody(bodyStatements)
+        .setReturnExpr(returnExpr)
+        .build();
+  }
+
+  private Expr addRequestParamsForRoutingHeaders(
```
chanseokoh, line 291: Just realized this method returns two values: (1) a return expression for a lambda; and (2) a body for that lambda. I don't like that one is returned as a return value while the other as a mutable argument. Also, `bodyStatements` is in the middle of the argument list, and it's not distinguishable from other read-only arguments.

src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -709,6 +710,9 @@ private static boolean isMapType(Descriptor messageDescriptor) {
                   .getOptions()
                   .getExtension(ExtendedOperationsProto.operationPollingMethod)
               : false;
+      // may not need to pass in messageTypes?
```
chanseokoh, line 713: I don't understand what this is saying and why. Remove or resolve it. Or add an unambiguous TODO.

src/main/java/com/google/api/generator/gapic/protoparser/RoutingRuleParser.java
```diff
@@ -0,0 +1,86 @@
...
+      String field = routingParameter.getField();
+      // validate if field exist in Message or nested Messages
+      inputMessage.validateField(field, messageTypes);
+      String name;
+      // if specified, the pattern must contain one and only one named segment
+      if (Strings.isNullOrEmpty(pathTemplate)) {
+        name = field;
+        pathTemplate = String.format("{%s=**}", name);
+      } else {
+        Set<String> params = getPatternBindings(pathTemplate).build();
```
chanseokoh, line 57: `params` too general and difficult to comprehend.
```suggestion
        Set<String> bindings = getPatternBindings(pathTemplate).build();
```

src/test/java/com/google/api/generator/gapic/composer/common/TestProtoLoader.java
```diff
@@ -218,6 +219,31 @@ public GapicContext parseShowcaseTesting() {
...
   }
 
+  public GapicContext parseExplicitDynamicRoutingHeaderTesting() {
+    FileDescriptor testingFileDescriptor = ExplicitDynamicRoutingHeaderTestingOuterClass.getDescriptor();
+    ServiceDescriptor testingService = testingFileDescriptor.getServices().get(0);
+    assertEquals(testingService.getName(), "ExplicitDynamicRoutingHeaderTesting");
+
+    Map<String, Message> messageTypes = Parser.parseMessages(testingFileDescriptor);
+    Map<String, ResourceName> resourceNames = Parser.parseResourceNames(testingFileDescriptor);
+    Set<ResourceName> outputResourceNames = new HashSet<>();
```
chanseokoh, line 229: Just to confirm the order doesn't matter. If it matters, consider using `LinkedHashSet`.

src/main/java/com/google/api/generator/gapic/composer/grpc/GrpcServiceStubClassComposer.java
```diff
@@ -245,10 +255,171 @@ protected String getProtoRpcFullMethodName(Service protoService, Method protoMet
...
+            .setVariable(
+                Variable.builder().setName("builder").setType(routingHeadersBuilderType).build())
+            .build();
+    for (int i = 0; i < routingHeaders.size(); i++) {
+      RoutingHeader routingHeader = routingHeaders.get(i);
+      MethodInvocationExpr requestFieldGetterExpr =
+          createRequestFieldGetterExpr(requestVarExpr, routingHeader.fieldName());
+      Expr routingHeaderKeyExpr =
+          ValueExpr.withValue(StringObjectValue.withValue(routingHeader.name()));
+      // TODO: Create proper snake style upper case name
```
chanseokoh, line 325: You can use `JavaStyle.toUpperSnakeCase()`?

src/test/java/com/google/api/generator/gapic/model/MessageTest.java
```diff
@@ -0,0 +1,110 @@
...
+import com.google.api.generator.engine.ast.TypeNode;
+import com.google.api.generator.engine.ast.VaporReference;
+import com.google.common.collect.ImmutableMap;
+import java.util.Map;
+import org.junit.Test;
+
+public class MessageTest {
+
+  public static final String MESSAGE_NAME = "TestMessage";
+  public static final Message.Builder TEST_MESSAGE_BUILDER =
```
chanseokoh, line 29: A builder is probably a mutable instance. Generally don't make it look like a constant for non-constants.
```suggestion
  public static final Message.Builder testMessageBuilder =
```
I wouldn't mind dropping `static` as well.

src/test/java/com/google/api/generator/gapic/model/MessageTest.java
```diff
@@ -0,0 +1,110 @@
...
+          .setFullProtoName("com.google.test.TestMessage")
+          .setType(TypeNode.OBJECT);
+
+  @Test
+  public void validateField_shouldThrowExceptionIfFieldNameIsEmpty() {
+    Message message = TEST_MESSAGE_BUILDER.build();
+    IllegalStateException illegalStateException =
+        assertThrows(
+            IllegalStateException.class, () -> message.validateField("", ImmutableMap.of()));
+    assertThat(illegalStateException.getMessage())
```
chanseokoh, line 41:
```suggestion
    assertThat(illegalStateException).hasMessageThat()
```
I think `illegalStateException` is too verbose in this context. I prefer `assertThat(exception)`.

src/test/java/com/google/api/generator/gapic/model/RoutingHeadersTest.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.List;
+import org.junit.Test;
+
+public class RoutingHeadersTest {
+
+  @Test
+  public void getDescendantFieldNames_shouldSplitFieldNameByDot() {
+    RoutingHeader routingHeader = RoutingHeader.create("table.name", "name", "/abc/dec");
+    List<String> descendantFieldNames = routingHeader.getDescendantFieldNames();
+    assertThat(descendantFieldNames).containsExactly("table", "name");
```
chanseokoh, line 29: Testing order is important?

```suggestion
    assertThat(descendantFieldNames).containsExactly("table", "name").inOrder();
```

src/test/java/com/google/api/generator/gapic/protoparser/RoutingRuleParserTest.java
```diff
@@ -0,0 +1,109 @@
...
+import com.google.protobuf.Descriptors.MethodDescriptor;
+import com.google.protobuf.Descriptors.ServiceDescriptor;
+import java.util.Map;
+import org.junit.Test;
+
+public class RoutingRuleParserTest {
+
+  private static final FileDescriptor TESTING_FILE_DESCRIPTOR =
+      RoutingRuleParserTestingOuterClass.getDescriptor();
+  private static final Map<String, Message> MESSAGES =
```
chanseokoh, line 34: `Map` is mutable.

src/test/java/com/google/api/generator/gapic/protoparser/RoutingRuleParserTest.java
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  @Test
+  public void parse_shouldParseRoutingRulesWithMultipleParameter() {
+    RoutingHeaders actual = getRoutingHeaders(5);
+    RoutingHeader expectedHeader1 =
+        RoutingHeader.create("name", "rename", "/v1beta1/{rename=tests/*}");
+    RoutingHeader expectedHeader2 =
+        RoutingHeader.create("routing_id", "id", "/v1beta1/{id=projects/*}/tables/*");
+    assertThat(actual.routingHeadersList()).containsExactly(expectedHeader1, expectedHeader2);
```
chanseokoh, line 88:
```suggestion
    assertThat(actual.routingHeadersList()).containsExactly(expectedHeader1, expectedHeader2).inOrder();
```

src/test/java/com/google/api/generator/gapic/composer/common/TestProtoLoader.java
blakeli0: No, it does not matter

src/main/java/com/google/api/generator/gapic/composer/grpc/GrpcServiceStubClassComposer.java
blakeli0: Yep, thanks for the suggestion, it work perfectly!

src/test/java/com/google/api/generator/gapic/model/MessageTest.java
blakeli0: In my opinion, as long as the field is final(not reassignable), then it's a constant, but I do see that Java style guide says differently go/java-style#s5.2.4-constant-names. I can change it to follow our style guide.

src/test/java/com/google/api/generator/gapic/protoparser/RoutingRuleParserTest.java
blakeli0: Thought `containsExactly` is already checking the order, tested and it does not, TIL, thanks for the suggestion.

src/test/java/com/google/api/generator/gapic/model/RoutingHeadersTest.java
blakeli0: Added.

src/test/java/com/google/api/generator/gapic/protoparser/RoutingRuleParserTest.java
blakeli0: Refactored `MESSAGES` not to be a static final field.

src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
blakeli0: Removed

src/main/java/com/google/api/generator/gapic/protoparser/RoutingRuleParser.java
blakeli0: Renamed

src/main/java/com/google/api/generator/gapic/composer/grpc/GrpcServiceStubClassComposer.java
blakeli0: if we want to check `if (request.getName() != null)`, we have to set the `returnType` of `request.getName` to a concrete type(which we don't have that info at this moment), otherwise it would fail the validation since it's expecting the the type of left hand side and right hand side of `RelationalOperationExpr` should be comparable(which makes sense, to prevent something like `if (5 != null)`). In our case, we know we are either checking an Object or a String, so setting the `returnType` to String just to pass validation.

src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/GrpcTestingStub.golden
```diff
@@ -220,9 +227,18 @@ public class GrpcTestingStub extends TestingStub {
             .setMethodDescriptor(getTestMethodDescriptor)
             .setParamsExtractor(
                 request -> {
-                  ImmutableMap.Builder<String, String> params = ImmutableMap.builder();
-                  params.put("name", String.valueOf(request.getName()));
-                  return params.build();
+                  RoutingHeaderParamsBuilder builder = new RoutingHeaderParamsBuilder();
+                  if (request.getName() != null) {
+                    builder.add(
+                        String.valueOf(request.getName()), "rename", PATH_TEMPLATE_GETTEST_0);
```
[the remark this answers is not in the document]
blakeli0: I thought the same thing too, but we will have to change the method signature to `Object` and cast it to String in the builder, which is not something I prefer either. That being said, I'm considering adding the type check for String so that we don't have to do this kind of casting, I'll start a new thread and discuss with you guys.

src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/GrpcTestingStub.golden
```diff
@@ -220,9 +227,18 @@ public class GrpcTestingStub extends TestingStub {
             .setMethodDescriptor(getTestMethodDescriptor)
             .setParamsExtractor(
                 request -> {
-                  ImmutableMap.Builder<String, String> params = ImmutableMap.builder();
-                  params.put("name", String.valueOf(request.getName()));
-                  return params.build();
+                  RoutingHeaderParamsBuilder builder = new RoutingHeaderParamsBuilder();
```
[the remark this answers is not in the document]
blakeli0: Makes sense, will change it.

src/main/java/com/google/api/generator/gapic/composer/grpc/GrpcServiceStubClassComposer.java
blakeli0: I totally understand where you are coming from and I gave it a lot of thought to make it better. I don't really like the way it is right now as well because the exactly same reason you mentioned, but it's kind of a compromise between reuse as much code as possible and modularize the code. 
That being said, there is still some ways we can improve, we can either pass an ExprBuilder to the method and return void(my preferred approach) or wrap everything we manipulated to a class and return it. I'll update the code and let you know.

src/main/java/com/google/api/generator/gapic/model/RoutingHeaders.java
blakeli0: Guava API is not OK in GAX surface but OK in this repo.

src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/GrpcTestingStub.golden
chanseokoh: I think this is resolved? https://github.com/googleapis/gapic-generator-java/pull/887#discussion_r787976169

src/main/java/com/google/api/generator/gapic/composer/grpc/GrpcServiceStubClassComposer.java
chanseokoh: > ither pass an ExprBuilder to the method and return void(my preferred approach) or wrap everything we manipulated to a class and return it

Yeah, I think either one is better than the current code. Consider renaming the method too.

src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
chanseokoh: Make sure you push your changes before making this as resolved.

src/main/java/com/google/api/generator/gapic/model/Message.java
blakeli0: Restructured the validation logic. The pre-existing validation logic actually has a bug, it assumes the first field always exist, if it does not, there will be a null pointer exception.

src/main/java/com/google/api/generator/gapic/protoparser/RoutingRuleParser.java
blakeli0: Updated as suggested.

src/main/java/com/google/api/generator/gapic/protoparser/RoutingRuleParser.java
blakeli0: Created a new [helper class](https://github.com/googleapis/gapic-generator-java/blob/d1fe9e138a4958735cffcb398797e5af764b15bc/src/main/java/com/google/api/generator/gapic/protoparser/PatternParser.java#L23) to handle this logic, also returning Set instead of builder now.

src/main/java/com/google/api/generator/gapic/model/RoutingHeaders.java
blakeli0: Added.

src/main/java/com/google/api/generator/gapic/model/Message.java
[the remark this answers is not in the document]
blakeli0: Added Javadoc to explain the usage of the method.

src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/GrpcTestingStub.golden
blakeli0: Yes, updated to proper snake style name.

src/main/java/com/google/api/generator/gapic/composer/grpc/GrpcServiceStubClassComposer.java
blakeli0: This class is in Grpc package which I assume it's only used for grpc requests not rest? I don't see any usage of this class for generating Http client either.

src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/GrpcTestingStub.golden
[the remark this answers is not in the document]
blakeli0: Updated as suggested.

src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/GrpcTestingStub.golden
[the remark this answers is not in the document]
blakeli0: Removed `String.valueOf()` since the validation of String type for field is added.

src/main/java/com/google/api/generator/gapic/composer/grpc/GrpcServiceStubClassComposer.java
blakeli0: Updated the logic to not adding null check for the last field, so we need the index now to skip the last one.

src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
blakeli0: Changes pushed.

src/test/java/com/google/api/generator/gapic/model/MessageTest.java
blakeli0: Updated.

src/main/java/com/google/api/generator/gapic/composer/grpc/GrpcServiceStubClassComposer.java
blakeli0: Passing the builder now, also renamed the methods.

src/main/java/com/google/api/generator/gapic/model/RoutingHeaders.java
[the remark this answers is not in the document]
vam-google: It is always a trade-off between more flexibility vs simplicity of the code and having it less verbose. It is fine to keep it as is if you want.

vam-google (approved):
LGTM

meltsufin (approved):

Issue #888, opened 2022-01-06
Compiling generated libraries with Java 9+ fails due to javax.annotation.Generated


chanseokoh:
For example, compiling a "self-service" generated library with `./gradlew check` fails:
```
> Task :proto-google-cloud-retail-v2alpha-java:compileJava
/tmp/google-cloud-retail-v2alpha-java/proto-google-cloud-retail-v2alpha-java/src/main/java/com/google/cloud/retail/v2alpha/LocationName.java:27: error: cannot find symbol
import javax.annotation.Generated;
```

The `javax.annotation` package space is really a mess from the perspective of Java 9+ JPMS, and it has caused a lot of trouble to many people with Java 9+.

The right solution for Java 9+ is to drop `javax.annotation.Generated` and migrate to `javax.annotation.processing.Generated`. However, doing so will cause Java 8 to fail with the same "cannot find symbol" error. An option would be to completely drop using the annotation. However, the gRPC proto compiler generated files with the annotation, so they are out of our control:
```
@javax.annotation.processing.Generated(
    value = "by gRPC proto compiler",
    comments = "Source: google/cloud/retail/v2alpha/product_service.proto")
```

A workaround for the users hitting this issue is to declare the following dependency, but this isn't really the right solution.
```gradle
implementation 'javax.annotation:javax.annotation-api:1.2'
```

Refs:
https://github.com/google/dagger/issues/880
https://github.com/grpc/grpc-java/issues/3633
https://github.com/vavr-io/vavr/issues/2154

chanseokoh:
#889 added the dependency as a workaround. Eventually when we drop Java 8, we should migrate our code to use `javax.annotation.processing.Generated`.

vam-google:
@chanseokoh Is this bug still relevant? Keeping this bug open for the Java8 support drop time is probably not practical, as we are guaranteed to not drop it till 2030 at minimum (current end of Java8 Extended Suppport by Oracle https://www.oracle.com/java/technologies/java-se-support-roadmap.html)

chanseokoh:
We need way to track this issue, so that we don't forget to fix it when the time comes.

Are we not allowed to leave a bug open for some set period of time? If so, I'll have no choice but to close it.

suztomo:
Why it's not happening in Java 11 and Java 17 CIs?

[...]

PR #889, opened 2022-01-06
feat: support Java 9+ compilation of generated libraries

chanseokoh:
Stopgap support for generated libraries. See #888.

Tested this with java-compute and java-retail using Java 11.


meltsufin (approved):
What a mess with this annotation, but this seems like the best compromise when having to support Java 8 and 9+.

Issue #886, opened 2022-01-05
Run tests for "self-service" libraries in CI


chanseokoh:
Doing so will help validating changes to the "self-service" library generation workflow (e.g., as in #876).

Theoretically, we should be able to generate some libraries against actual API service protos like the following (may need to change a few things in `WORKSPACE` or some `BUILD` files).
```
bazel build @com_google_googleapis//google/cloud/compute/v1:google-cloud-compute-v1-java
```
Then unpack `bazel-bin/external/com_google_googleapis/google/cloud/compute/v1/google-cloud-compute-v1-java.tar.gz` and run  `./gradlew clean check publishToMavenLocal` at the root of the generated library.

`compute` is REST and all others are gRPC at the moment, so `compute` and another library are good candidates.

PR #892, opened 2022-01-07
chore: sync googleapis and disco-to-proto3-converter repos

chanseokoh:


chanseokoh:
Updated the golden files accordingly. PTAL.

meltsufin (commented):
It seems fine, but the size of the change is a little alarming. Can you give a bit more context to this upgrade? how big is the version jump? It's not clear looking just at the commit hashes.

chanseokoh:
googleapis `ba30d8097582039ac4cc4e21b4e4baa426423075` is the state of Aug 23, 2021. I'm moving it to the latest commit.
```
commit ba30d8097582039ac4cc4e21b4e4baa426423075
Author: Google APIs <runner@fv-az154-630.kzav01uq0i3u1f1hch3ggm0mbb.bx.internal.cloudapp.net>
Date:   Mon Aug 23 23:09:34 2021 +0000

    chore: regenerate API index
```
I only changed the commits. All the integration golden file changes are auto-generated, which would match the current state of googleapis.

I think advancing commits in itself is better than making them stall forever, as our generator is tested against more recent APIs. However, the reason I'm trying to do it now is to be able to see Compute API so that I can fix #886. [There's no Compute](  https://github.com/googleapis/googleapis/tree/ba30d8097582039ac4cc4e21b4e4baa426423075/google/cloud) in the googleapis commit of Aug 23.

For the disco-to-proto3-converter, I noticed a build failure without advancing both commits together.

meltsufin (approved):

PR #896, opened 2022-01-10
chore: add CI test for Gradle builds on generated libraries

### LINKED ISSUE #886: Run tests for "self-service" libraries in CI

chanseokoh:
Fixes #886.

Storage is already generated and pre-packaged as part of the integration test, so we can just unpack it.

Compute isn't the case, and I opted to generate against the "small" API.


.github/workflows/ci.yaml
```diff
@@ -46,6 +46,23 @@ jobs:
       - name: Integration Tests
         run: bazel --batch test //test/integration/...
 
+      - name: Gradle Build Tests on Generated Libraries
```
meltsufin, line 49: Can we just split it into two steps? (1) Gradle Build Generated Storage Client Library; (2) Generate and Gradle Build Compute Client Library?
Otherwise, it implies like all client libraries are being tested.

meltsufin (approved):

PR #903, opened 2022-01-21
feat: make generated test values comply with url path template

vam-google:
This includes nested messages creation when there are url paths with subfields mentioned (like `/v1/{field.name=projects/*/databases/*/collectionGroups/*/fields/*}`)

This is needed for rest transports tests because, unlike grpc, request fields must match path templates for rest logic pass the tests.

Main changes are in `DefaultValueComposer` and `HttpRuleParser` classes.

The generated pattern-matching samples are in the following format:
  given the pattern pattern:
  `/v1/{field.name=projects/*/databases/*/collectionGroups/*/fields/*}`
  the value will be:
  `field.name=projects/project-1234/databases/database-1234/collectionGroups/collectionGroup-1234/fields/field-1234`


src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceClientTestClassComposer.java
```diff
@@ -472,7 +476,13 @@ private MethodDefinition createRpcTestMethod(
             VariableExpr.withVariable(
                 Variable.builder().setType(methodArg.type()).setName(methodArgName).build());
         argExprs.add(varExpr);
-        Expr valExpr = createDefaultValue(methodArg, resourceNames);
+        Map<String, String> valuePatterns = Collections.emptyMap();
+        if (getTransportContext().useValuePatterns() && method.hasHttpBindings()) {
+          valuePatterns = method.httpBindings().getPathParametersValuePatterns();
```
chanseokoh, line 481: Same operation for all iterations. Move out of the `for` loop to signify that (and avoid unnecessary repetition).

src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceClientTestClassComposer.java
```diff
@@ -771,7 +783,14 @@ protected abstract Expr createDefaultValue(
             VariableExpr.withVariable(
                 Variable.builder().setType(methodArg.type()).setName(methodArgName).build());
         argVarExprs.add(varExpr);
-        Expr valExpr = createDefaultValue(methodArg, resourceNames);
+        Map<String, String> valuePatterns = Collections.emptyMap();
```
chanseokoh, line 786: ditto

src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
```diff
@@ -94,78 +98,100 @@ public static Expr createDefaultValue(
...
     }
 
-    if (f.isMessage()) {
+    if (field.isMessage()) {
+      String nestedFieldName = field.name();
+      Map<String, String> nestedValuePatterns = new HashMap<>();
+      for (Map.Entry<String, String> entry : valuePatterns.entrySet()) {
+        if (entry.getKey().startsWith(nestedFieldName + '.')) {
+          nestedValuePatterns.put(
+              entry.getKey().substring(nestedFieldName.length() + 1), entry.getValue());
```
chanseokoh, line 150: We don't consider two or more levels of nesting?

src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
```diff
@@ -557,4 +592,24 @@ static AnonymousClassExpr createAnonymousResourceNameClass(String fieldOrMessage
...
+
+  private static String constructValueMatchingPattern(String fieldName, String pattern) {
+    if (pattern == null || pattern.isEmpty()) {
+      return fieldName + fieldName.hashCode();
+    }
+
+    final String suffix = "-" + (Math.abs((fieldName + pattern).hashCode()) % 10000);
+
+    String value = pattern;
+    value = value.replace("**", "*");
```
chanseokoh, line 604:
```suggestion
    String value = pattern.replace("**", "*");
```

src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
```diff
@@ -557,4 +592,24 @@ static AnonymousClassExpr createAnonymousResourceNameClass(String fieldOrMessage
...
+
+    String prevTempl = null;
+    while (!value.equals(prevTempl)) {
+      prevTempl = value;
+      value = REPLACER_PATTERN.matcher(value).replaceFirst("$1$2$1" + suffix);
+    }
+
+    value = value.replace("*", fieldName + suffix);
+    return value;
```
chanseokoh, line 613:
```suggestion
    return value.replace("*", fieldName + suffix);
```

src/main/java/com/google/api/generator/gapic/protoparser/HttpRuleParser.java
```diff
@@ -200,4 +212,18 @@ private static void checkHttpFieldIsValid(String binding, Message inputMessage,
         fieldCondition,
         String.format(messageFormat, field.name(), inputMessage.name(), field.type()));
   }
+
+  private static Map<String, String> constructPathValuePatterns(String pattern) {
+    Map<String, String> varPattern = new HashMap<>();
```
chanseokoh, line 217: The same question.

src/test/java/com/google/api/generator/gapic/composer/rest/goldens/ComplianceClientTest.golden
```diff
@@ -219,7 +220,30 @@ public class ComplianceClientTest {
     RepeatRequest request =
         RepeatRequest.newBuilder()
             .setName("name3373707")
-            .setInfo(ComplianceData.newBuilder().build())
```
meltsufin, line 222: What has caused this statement to get expanded like that?

src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceClientTestClassComposer.java
vam-google: moved

src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceClientTestClassComposer.java
vam-google: moved

src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
```diff
@@ -94,78 +98,100 @@ public static Expr createDefaultValue(
...
                   PrimitiveValue.builder().setType(TypeNode.INT).setValue("0").build()))
-          .setReturnType(f.type())
+          .setReturnType(field.type())
           .build();
     }
 
-    if (f.isMessage()) {
+    if (field.isMessage()) {
+      String nestedFieldName = field.name();
+      Map<String, String> nestedValuePatterns = new HashMap<>();
```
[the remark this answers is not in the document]
vam-google: No, it should not matter, as the map keys are field names, and on each level of object nesting the keys are guaranteed to be unique and the order in which they go does not matter, as it (the order in which setters are called in generated code) is determined by other parts of the logic, the valuePatterns map is used only to pull the specific pattern for a field with a specific name when we are finally ready to generate the specific value.

src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
vam-google: we do. createValue and createSimpleMessageBuilderValue call each other recursively

src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
vam-google: Thanks, I'm surprised that Intellij does not highlight stuff like this =)

src/main/java/com/google/api/generator/gapic/model/HttpBindings.java
```diff
@@ -81,6 +90,14 @@ public String patternLowerCamel() {
     return lowerCamelPattern;
   }
 
+  public Map<String, String> getPathParametersValuePatterns() {
+    Map<String, String> valuePatterns = new HashMap<>();
```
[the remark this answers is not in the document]
vam-google: So the order here does not matter, as this map is to map field to their patterns and it is never used (and can't reasonably be sued) to determine order of things. This map is being iterated once in `DefaultValue` composer to create a nested fields map (i.e. filter out all fields, which are outside of the nested field scope) but that is once again - to get another non-ordered map, which eventually will be used for only one thing - get the field pattern by field name, and order does not matter in that case.

src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
vam-google: changed

src/main/java/com/google/api/generator/gapic/protoparser/HttpRuleParser.java
vam-google: Order does not matter (see other response for details)

chanseokoh:
I'll LGTM once the changes are in.

chanseokoh (approved):

PR #911, opened 2022-01-26
chore: re-work generating unit goldens

chanseokoh:
Continuing in the same spirit to #904.


chanseokoh:
Oh, wait. Maybe I can do this better.

PR #912, opened 2022-01-26
chore: re-work generating unit golden files

chanseokoh:
Now we can completely remove the `rules_bazel` folder.

To generate a golden file, all we have to do is to run a unit test, because those unit tests are written to generate and save a golden. For example,

```java
  @Test
  public void generateGrpcServiceStubClass_createBatchingCallable() {
    ...
    // Generate a golden while running this test.
    Utils.saveCodegenToFile(this.getClass(), "GrpcLoggingStub.golden", visitor.write());
    ...
  }
```

I think this is better than my previous attempt: #911


rules_bazel/java/java_diff_test.bzl
```diff
@@ -1,84 +0,0 @@
-def _overwrite_golden_impl(ctx):
-    test_class = ctx.attr.test_class
-    inputs = ctx.files.srcs
-    goldens_output_zip = ctx.outputs.goldens_output_zip
-    test_runner = ctx.executable.test_runner
-
-    # Generate the goldens from tests.
-    generate_goldens_script = """
-    mkdir local_tmp
-    TEST_OUTPUT_HOME="$(pwd)/local_tmp" {test_runner_path} $@
```
chanseokoh, line 10: FTR: this is the line running `SingleJUnitTestRunner`.

{test_runner_path} $@ is roughly `java -cp ... com.google.api.generator.test.framework.SingleJUnitTestRunner <SomeUnitTestClass>`, as shown [here](https://github.com/googleapis/gapic-generator-java/pull/911/files#diff-4e75eece56df81431426a0c3063d4173a34cd1cc2786460ba2032dc46abb1637R9).

PR #913, opened 2022-01-27
chore: rework build infrastructure

chanseokoh:
Removes `repository_rules.bzl`  and `dependencies.properties`. Creates `PROPERTIES.bzl` instead (which was dynamically generated by the rule above previously) to make the whole process straightforward.

The `dependencies.properties`-pattern is used in other repos (including `gax-java`). Its purpose is to avoid dependency versions going out of sync when compiling source between Bazel and Gradle. However, we don't have Gradle, and we won't enable building with two different build systems, so YAGNI.


meltsufin (approved):
I like the simplification. I wish we had some way of automatically updating the versions using a bot like we do in other repos.

chanseokoh:
Looks like googleapis/googleapis is referencing `repository_rules.bzl` that is removed in this PR.

I'll have to look into that to resolve the issue.

chanseokoh:
Turns out it's a trivial fix to remove unused code from googleapis/googleapis. However, there was a chicken-and-egg-like situation, so I ended up upgrading gax-java and gapic-generator-java in googleapis and removing the code after releasing gapic-generator-java. (Internal change: 424868012)

PR #918, opened 2022-02-02
feat: Use gapic-generator-java jar in the client library generation process

chanseokoh:
**Note(Updated on 12/29/2022)**: The description is now almost re-written since a lot of the implementation details changed. Thanks to @chanseokoh who created this PR originally and enlightened me the new approach.

The main goal of this PR is to use the gapic-generator-java jar in the client library generation process, with both released jars and local SNAPSHOT jars. Hence, Bazel will not be used for loading dependencies and building the generator from source going forward. Instead, Maven is now used for building a gapic-generator-jar first, and the jar is passed into the Bazel plugin to generate the client libraries. This PR also removed as much Bazel files/configs as possible.

Nothing changes from the "UI" or "interface" side where how currently outsiders (including googleapis/googleapis) generate GAPIC libraries. The way gapic-generator-java is built is an implementation detail (that changed from Bazel to Maven) and is completely hidden from outsiders.

**Local development** is almost the same like before, two notable changes:
1. Unit tests can only be done in Maven now.
2. Before running integration tests, you would need to build a local gapic-generator-java jar.

**Next Steps[Must haves]:**

1. After this PR is merged and a new version of gapic-generator-java is released, we need to update the googleapis repo to make changes accordingly.
2. The gradle files generated by the self-service client libraries are still using the `dependencies.properties` file from `gax-java` repo, which may include different versions from the pom.xml that is used during generation runtime. It is to be determined to find the best way to manage the versions for the generated Gradle files.

**Next Steps[Nice to haves, not an exhaustive list]:**

1. Integration Tests still rely on Bazel to run, this is expected as the Bazel interface didn't change, and integration tests are triggering the actual library generation process which is still based on Bazel. However, we could wrap it as a Maven plugin so that the whole development flow will not dependent on Bazel at all.
2. This repo still exposes Java Bazel rules in `rules_java_gapic` folder, which is referenced by googleapis in the generation process, hence googleapis will need to dependent on both the jar and the source of gapic-generator-java. We could move these Bazel rules to googleapis so that gapic-generator-java is completely self-contained.


chanseokoh:
More files removed that are noteworthy: `PROPERTIES.bzl` and `repositories.bzl`.

Note, these don't affect "self-service" library generation.

chanseokoh:
> Unit tests are not wokring. `mvn test` is basically functioning, but tests are currently failing in this PoC. (`mvn package` without `-DskipTests` will thus fail too.) Updating unit golden files is not yet implemented either.

`mvn test` is now working.

.github/workflows/ci.yaml
```diff
@@ -42,7 +42,9 @@ jobs:
           echo "The old one will disappear after 7 days."
 
       - name: Integration Tests
-        run: bazelisk --batch test //test/integration/...
+        run: |
+          mvn -B -ntp install -DskipTests -Dcheckstyle.skip -Dfmt.skip
+          bazel --batch test //test/integration/...
```
suztomo, line 47: Use bazelisk

blakeli0: Done

burkedavison:
Sorry for the unintentional pushes to this branch.

meltsufin (approved):
LGTM. Deferring to @ddixit14 on the CI config changes.

blakeli0 (approved):

PR #927, opened 2022-02-10
chore: update DEVELOPEMENT.md

chanseokoh:
This doesn't work in Bazel.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -139,9 +139,3 @@
     ```sh
     bazel run //test/integration:update_redis
     ```
-
--   Update all integration test golden files.
-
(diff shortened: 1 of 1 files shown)


blakeli0 (approved):

meltsufin (approved):
OK, but is it possible to create a target that would update all?

chanseokoh:
> OK, but is it possible to create a target that would update all?

I am not aware of such a thing to run all binary targets. What you can do is
```
bazel query //test/integration/... | grep :update_ | xargs --max-args=1 bazel run
```
or
```
for t in $(bazel query //test/integration/... | grep :update_); do bazel run $t; done
```

PR #928, opened 2022-02-11
fix: Routing annotation was not recognized on parsing protos

blakeli0:
Routing annotation was not recognized on parsing protos, this is due to RoutingProto's extensions are not registered. This also indicate that we are missing integration tests for the new routing annotation, hence added bigtable to integration tests as well.


meltsufin (approved):

vam-google (approved):
LGTM

PR #937, opened 2022-02-17
chore: set up fully working Maven pom.xml

chanseokoh:
**This is on top of #936. Please review that first.**

 Adds a fully working `pom.xml` that compiles the generator and runs JUnit tests. Type `mvn verify`, and everything will just work.

Doesn't touch Bazel, so there's no side-effect.


chanseokoh:
What's incredible is that the Maven unit tests on CI took ~45 seconds even though nothing was cached for the first time. (OTOH, Bazel currently takes ~10 minutes.)

meltsufin (approved):
This is great! Thank you so much @chanseokoh!

pom.xml
```diff
@@ -0,0 +1,328 @@
...
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+
+  <build>
+    <extensions>
+      <extension>
+        <groupId>kr.motd.maven</groupId>
+        <artifactId>os-maven-plugin</artifactId>
+        <version>1.7.0</version>
```
blakeli0, line 30: Do we usually extract versions to properties?

chanseokoh: I think it's a matter of preference. Usually we do it if there's a merit and reason. For example, https://github.com/googleapis/java-bigquery/blob/main/pom.xml

PR #938, opened 2022-02-17
chore: clean up Bazel build

chanseokoh:
`com.google.googlejavaformat:google-java-format:jar:all-deps:1.7` has been declared very early (almost at the top) in `WORKSPACE`, so there is no point to do "maybe do jvm_maven_import_external if it has not been declared" through `repositories.bzl`, which is imported later.

Also, `jvm_maven_import_external()` is an old rule, so replacing that with more modern `maven_install()`.


chanseokoh:
Oh, wait, something's broken.

chanseokoh:
Not sure what's really going on, but I guess mixing `jvm_maven_import_external()` and `maven_install()` doesn't work well. I think everything should be migrated once.

PR #951, opened 2022-02-23
fix: Update copyright to 2022 for newly generated files.

blakeli0:


src/main/java/com/google/api/generator/gapic/composer/comment/CommentComposer.java
```diff
@@ -20,7 +20,7 @@
 
 public class CommentComposer {
   private static final String APACHE_LICENSE_STRING =
-      "Copyright 2021 Google LLC\n\n"
+      "Copyright 2022 Google LLC\n\n"
```
meltsufin, line 23: Can we just put in the current year, so we don't have to update it every year?

BUILD.bazel
```diff
@@ -209,7 +209,6 @@ GOLDEN_UPDATING_UNIT_TESTS = [
     args = [test_class],
     runtime_deps = [
         ":gapic_generator_java",
-        ":gapic_generator_java-test",
```
meltsufin, line 212: why this change?

meltsufin (changes requested):

BUILD.bazel
blakeli0: `gapic_generator_java-test` does not exist, `bazel test //...` would error out without this change.

src/main/java/com/google/api/generator/gapic/composer/comment/CommentComposer.java
blakeli0: I thought about it, the concern is that our golden tests would start failing all of sudden in the beginning of next year, which is not a good behavior in my opinion. I can update it to use current year if this behavior is OK.

meltsufin (approved):

PR #953, opened 2022-02-24
chore: Add jacoco plugin to support code coverage

meltsufin:


pom.xml
```diff
@@ -210,6 +210,27 @@
         </configuration>
       </plugin>
 
+      <!-- Code coverage for SonarCloud -->
+      <plugin>
+        <groupId>org.jacoco</groupId>
+        <artifactId>jacoco-maven-plugin</artifactId>
+        <version>0.8.7</version>
```
danielgazineu, line 217: Might be good to manage these plugins somewhere as well (shared deps?). Upper layers like handwritten libraries and boms should probably use the same setup.

danielgazineu (approved):
LGTM

pom.xml
meltsufin: parent pom is usually a good place

PR #960, opened 2022-03-02
fix: setups for manually runs single JUnit test

zhumin8:
This seems to be added setup in #936, but later accidentally deleted?

Verified test runs with command `bazel run //:update_com_google_api_generator_engine_JavaCodeGeneratorTest`

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -209,6 +209,7 @@ GOLDEN_UPDATING_UNIT_TESTS = [
     args = [test_class],
     runtime_deps = [
         ":gapic_generator_java",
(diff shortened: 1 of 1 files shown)


meltsufin:
@blakeli0 I think you removed it.

blakeli0:
> This seems to be added setup in #936, but later accidentally deleted?
> 
> Verified test runs with command `bazel run //:update_com_google_api_generator_engine_JavaCodeGeneratorTest`

hmm, I removed it on purpose because it was failing other commands, but I just tested and they are working fine, maybe my branch was not up to date.

blakeli0 (approved):

PR #970, opened 2022-03-17
feat: inline disclaimer comment, use Sample/RegionTag, collect GapicClass samples (pt 2)

eaball35:
go/java-sample-gen


src/main/java/com/google/api/generator/gapic/composer/samplecode/SampleComposerUtil.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  static boolean isStringTypedResourceName(
+      MethodArgument arg, Map<String, ResourceName> resourceNames) {
+    return arg.type().equals(TypeNode.STRING)
+        && arg.field().hasResourceReference()
+        && resourceNames.containsKey(arg.field().resourceReference().resourceTypeString());
+  }
+
+  static boolean isProtoEmptyType(TypeNode type) {
```
blakeli0, line 50: This helper method can be moved to TypeNode since it's checking internal attributes of itself.

src/main/java/com/google/api/generator/gapic/model/GapicClass.java
```diff
@@ -45,6 +62,58 @@ static Builder builder() {
...
+
+    abstract List<Sample> samples();
+
+    public final GapicClass build() {
+      setSamples(handleDuplicateSamples(samples()));
+      return autoBuild();
+    }
+  }
+
+  private static List<Sample> handleDuplicateSamples(List<Sample> samples) {
```
blakeli0, line 77: What is the scenario for "duplicate samples"? Is it for RPCs that have multiple samples? If it is, they are not really duplicate samples, they are overloaded samples for the same RPC. This name kind of implies we are removing duplicates, which could be misleading.

I see a few issues with the logic in this method:
1. The logic is very complex and not easy to understand, please consider simplify it or break it down.
2. This complex logic is being called in a build() method. Builder methods(Or any initialization methods like constructors) should only have simple logics like input validation or build a field from other fields at most, a complex logic like this should probably be handled before passing values to build().
3. If there is no way, please consider add unit tests to cover all the cases. I know it may already be covered by some golden tests, but golden tests are not really unit tests, they are more like a component tests that test multiple classes, which is something we can really improve for this repo in my opinion. Unit test composers/writers may be difficult and don't make sense in some cases, for logic in model classes though, we should be able to unit test them without much difficulties.

eaball35: I haven't actually seen these cases come up in the integration tests and the few other googleapis libs I tried, but it does occur in some of the unit tests and could potentially occur. If duplicates are not at least considered, we run into errors with duplicate files. 

We can't disambiguate on every possible dimension or names get too long/complex so at some point the samples are the "same" sample with different content. See golden units for [examples](https://github.com/googleapis/gapic-generator-java/tree/b5800afd3b5a34be1a3b9b25f7e90b096d84602d/src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/samples/echoclient). 

Overload samples will have unique names, but consider this example using different variable names:
Sample 1: 
```
String parent = "parent";
rpcCall(parent);
```

Sample 2:
```
String child = "child";
rpcCall(child)
```

Another possible example is with slight difference in response:
Sample 1: 
```
String resource = "resource";
rpcCall(parent);
```

Sample 2:
```
String resource = "resource";
rpcCall(resource).get()
```

I'm not sure if there's a much better place to put this logic. We should consider duplicates within a gapic class. If I disambiguate duplicates as I collect the samples, I'm just doing the same work over and over again. If I try to disambiguate duplicate samples as I create them, I'm not sure there's a guarantee it will always be a duplicate until they're collected.  If I wait to do this until I write the samples, it's just pushing the work down the road.

To separate this from build, I will 1.) move this to a helper method that I call before passing them in to create gapic class.  and I will 2.) also write some additional unit tests for this logic. The goal for phase 1 snippet gen is the canonical samples, I can still pivot this overload strategy if it doesn't make sense as we start generating sample files.

src/main/java/com/google/api/generator/gapic/composer/samplecode/SampleComposerUtil.java
eaball35: I didn't actually write this code. This originally existed in the original ServiceClientSampleComposer and just got moved here.  I didn't catch this in that refactor PR I pushed, I forgot to rename this from ServiceClientSampleComposerUtil to just SampleComposerUtil.

That being said, you're right. I will move this + add tests.

src/main/java/com/google/api/generator/gapic/model/GapicClass.java
eaball35: I've broken down all the samples and added additional details at in go/java-sample-gen under 'Current Generated Sample Types'. 

I cleaned up the sample names one final time and the main duplicate I'm seeing is the first example above with the overload with different variable names. I've only seen it in the unit tests, haven't seen it in the integration tests. I moved this out of GapicClass and handle before I create GapicClass. As this isn't part of phase 1 and a small edge case, I'm ok leaving as if for now. I added some additional comments in the method to better explain what's happening.

src/test/java/com/google/api/generator/gapic/composer/samplecode/ServiceClientHeaderSampleComposerTest.java
```diff
@@ -27,24 +24,319 @@
...
+            VaporReference.builder()
+                .setName("EchoClient")
+                .setPakkage(SHOWCASE_PACKAGE_NAME)
+                .build());
+    String results =
+        writeStatements(
+            ServiceClientHeaderSampleComposer.composeClassHeaderSample(
+                service, clientType, resourceNames, messageTypes));
+    String expected =
+        LineFormatter.lines(
```
blakeli0, line 208: Thanks for writing these comprehensive unit tests, just want to share my experience with testing composers. 
Another way(my preference) is to have a testing proto that includes all test cases, load that proto as the input, and compare the output to golden files, see `explicit_dynamic_routing_header_testing.proto` and the usage of it for example. I know it's not TRUE unit tests for composers since it's dependent on parser/writer logics(which is pretty stable), but it's usually easier to write and modify especially if you have a lot of test cases, it's more straightforward for others to understand as well.
That being said, there is nothing wrong with your way, and this way is preferred in most of cases since you are only testing one method. So nothing need to be changed here, just want to share my 2 cents.

src/main/java/com/google/api/generator/gapic/composer/samplecode/SampleComposerUtil.java
```diff
@@ -0,0 +1,113 @@
...
+        .map(
+            arg ->
+                JavaStyle.toUpperCamelCase(
+                    arg.variable().type().reference() == null
+                        ? arg.variable().type().typeKind().name().toLowerCase()
+                        : arg.variable().type().reference().name().toLowerCase()))
+        .collect(Collectors.joining());
+  }
+
+  public static List<Sample> handleDuplicateSamples(List<Sample> samples) {
```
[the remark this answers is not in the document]
eaball35: Just added SampleComposerUtil testing

src/test/java/com/google/api/generator/gapic/composer/samplecode/ServiceClientHeaderSampleComposerTest.java
eaball35: Thanks for sharing. Makes sense, I will consider this in future refactors + testing

PR #980, opened 2022-04-05
feat: write samples to file (pt3)

eaball35:
go/java-sample-gen - see 'Post gapic-generator-java' section

Can easily include/not include generated samples - plan to roll out to a few before rolling out everywhere. This change won't automatically start generating samples in repos. To do so:
- googleapis libs BUILD.bazels 'java_gapic_assembly_gradle_pkg' needs to include ":lib_java_gapic_samples" in deps
- need to update repos OwlBot.yaml to copy 'samples/snippets/generated'


src/main/java/com/google/api/generator/gapic/protowriter/Writer.java
```diff
@@ -43,52 +45,104 @@ public static CodeGeneratorResponse write(
...
+  private static void writeSamples(
+      GapicClass gapicClazz, String pakkage, String clazzPath, JarOutputStream jos) {
+    for (Sample sample : gapicClazz.samples()) {
+      JarEntry jarEntry =
+          new JarEntry(
+              String.format(
+                  "samples/snippets/generated/%s/%s/%s/%s.java",
+                  clazzPath,
+                  sample.regionTag().serviceName().toLowerCase(),
+                  sample.regionTag().rpcName().toLowerCase(),
```
blakeli0, line 110: Can you please check Vadym's recent PR regarding empty services? With his code change and your code change here, we are going to create an empty Sample file under the path /empty, which I think may not be a good idea going forward. I think we need to either filter out empty services here or when creating Samples, unless you think empty sample is a valid case.

src/main/java/com/google/api/generator/gapic/composer/Composer.java
```diff
@@ -186,6 +188,40 @@ public static List<GapicClass> generateTestClasses(GapicContext context) {
...
+              List<Sample> samples =
+                  gapicClass.samples().stream()
+                      .map(sample -> updateSample(sample, apiShortName, apiVersion))
+                      .collect(Collectors.toList());
+              return gapicClass.withSamples(samples);
+            })
+        .collect(Collectors.toList());
+  }
+
+  private static Sample updateSample(Sample sample, String apiShortName, String apiVersion) {
```
blakeli0, line 218: The method could be renamed to something like `addRegionTagAndHeaderToSample` to make it more specific, `updateSample` is too generic and does not indicate what exactly it's doing.

src/main/java/com/google/api/generator/gapic/protowriter/Writer.java
```diff
@@ -43,52 +45,104 @@ public static CodeGeneratorResponse write(
...
+    codeWriter.clear();
+
+    String path = getPath(clazz.packageString(), clazz.classIdentifier().name());
+    String className = clazz.classIdentifier().name();
+    JarEntry jarEntry = new JarEntry(String.format("%s/%s.java", path, className));
+    try {
+      jos.putNextEntry(jarEntry);
+      jos.write(code.getBytes(StandardCharsets.UTF_8));
+    } catch (IOException e) {
+      throw new GapicWriterException(
```
blakeli0, line 93: Can we include the original exception as the cause of `GapicWriterException`? This is usually how we handle and re-throw exceptions because you can keep the stack trace. You would need an overloaded constructor for `GapicWriterException` but that's totally worth it.

src/main/java/com/google/api/generator/gapic/protowriter/Writer.java
eaball35: I replied to the PR. I don't think empty should be passed in. There's a reason a region tag can't be built without rpcName.

src/main/java/com/google/api/generator/gapic/composer/Composer.java
```diff
@@ -186,6 +188,40 @@ public static List<GapicClass> generateTestClasses(GapicContext context) {
     return clazzes;
   }
 
+  @VisibleForTesting
+  static List<GapicClass> composeSamples(List<GapicClass> clazzes, String protoPackage) {
```
blakeli0, line 192: This method is basically adding region tags and headers to all samples in all classes, which should probably be renamed as well. But before you do that, I have a basic question regarding why we want to do it here after samples are already mostly composed? The package info is already in the context which is passes into `generateServiceClasses`, so potentially we should be able to generate a full Sample right?

src/main/java/com/google/api/generator/gapic/protowriter/Writer.java
eaball35: done

src/main/java/com/google/api/generator/gapic/composer/Composer.java
eaball35: done

src/main/java/com/google/api/generator/gapic/composer/Composer.java
eaball35: Before this step, we can generate an inline sample but not an "executable" one because inline format doesn't have a proper region tag. I could parse shortname/version out before and pass into composeServiceClasses in [Generator](https://github.com/googleapis/gapic-generator-java/blob/main/src/main/java/com/google/api/generator/gapic/Generator.java) and keep passing these details around when creating Samples or I can add these details before writing the files. I already have to iterate over the samples to add the license header here.

rules_java_gapic/java_gapic_pkg.bzl
```diff
@@ -83,13 +94,14 @@ def _gapic_pkg_tar_impl(ctx):
...
         package_dir = paths.package_dir,
+        samples = " ".join(["'%s'" % s.path for s in samples]),
         pkg = ctx.outputs.pkg.path,
         tar_cd_suffix = paths.tar_cd_suffix,
         tar_prefix = paths.tar_prefix,
     )
 
     ctx.actions.run_shell(
-        inputs = deps,
+        inputs = deps + samples,
```
blakeli0, line 104: What does this `run_shell` do? Do we need it? I see that pretty much all other changes in bazel files are net new, but this is modifying existing scripts so a little alarming to me.

eaball35: Yes, it runs the script above it for the _gapic_pkg_tar_impl. 'samples' is being used as input {samples}. This is where they get unpacked into the correct location in the tar.

src/main/java/com/google/api/generator/gapic/composer/Composer.java
blakeli0: Yeah, it's unfortunate these info are not easily accessible and you have to pass them around. I have some refactoring ideas but should definitely not be the scope of this PR. For now, can you please rename this method to make more readable/precise?

blakeli0:
Overall looks really good. Just have one additional comment that is not directly related to this PR, can you please add more comments/Javadoc to the new model classes you added in pt1 and pt2? Especially [Sample.java](https://github.com/googleapis/gapic-generator-java/blob/main/src/main/java/com/google/api/generator/gapic/model/Sample.java) and [RegionTag.java](https://github.com/googleapis/gapic-generator-java/blob/main/src/main/java/com/google/api/generator/gapic/model/RegionTag.java)?  I know not every model class have Javadocs currently and I should've done a better job in reviewing those earlier as well, but they would benefit new people in our team greatly.

src/main/java/com/google/api/generator/gapic/composer/Composer.java
eaball35: Makes sense and I agree it's not ideal, assuming the behaviors won't change I can test this out overall and refactor in parallel.

done, I renamed to make more readable

eaball35:
> Overall looks really good. Just have one additional comment that is not directly related to this PR, can you please add more comments/Javadoc to the new model classes you added in pt1 and pt2? Especially [Sample.java](https://github.com/googleapis/gapic-generator-java/blob/main/src/main/java/com/google/api/generator/gapic/model/Sample.java) and [RegionTag.java](https://github.com/googleapis/gapic-generator-java/blob/main/src/main/java/com/google/api/generator/gapic/model/RegionTag.java)? I know not every model class have Javadocs currently and I should've done a better job in reviewing those earlier as well, but they would benefit new people in our team greatly.

blakeli0 (approved):

PR #989, opened 2022-05-10
fix: [REGAPIC] Fix repeated fields handling for query parameters

vam-google:
if a field is declared as repeated in proto then its getter's name is `get<FieldName>List` not `get<FieldName>` instead.

### DIFF
--- a/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
+++ b/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
@@ -907,7 +907,9 @@ public class HttpJsonServiceStubClassComposer extends AbstractTransportServiceSt
       for (int i = 0; i < descendantFields.length; i++) {
         String currFieldName = descendantFields[i];
         String bindingFieldMethodName =
-            String.format("get%s", JavaStyle.toUpperCamelCase(currFieldName));
+            (i < descendantFields.length - 1 || !httpBindingFieldName.isRepeated())
+                ? String.format("get%s", JavaStyle.toUpperCamelCase(currFieldName))
+                : String.format("get%sList", JavaStyle.toUpperCamelCase(currFieldName));
         requestFieldGetterExprBuilder =
             requestFieldGetterExprBuilder.setMethodName(bindingFieldMethodName);
 
--- a/src/main/java/com/google/api/generator/gapic/model/HttpBindings.java
+++ b/src/main/java/com/google/api/generator/gapic/model/HttpBindings.java
@@ -40,12 +40,15 @@ public abstract class HttpBindings {
 
     public abstract boolean isOptional();
 
+    public abstract boolean isRepeated();
+
     @Nullable
     public abstract String valuePattern();
 
-    public static HttpBinding create(String name, boolean isOptional, String valuePattern) {
+    public static HttpBinding create(
+        String name, boolean isOptional, boolean isRepeated, String valuePattern) {
       return new AutoValue_HttpBindings_HttpBinding(
-          name, JavaStyle.toLowerCamelCase(name), isOptional, valuePattern);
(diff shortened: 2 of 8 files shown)


blakeli0:
> if a field is declared as repeated in proto then its getter's name is `get<FieldName>List` not `get<FieldName>` instead.

Why do we want to change this? Looks cosmetic to me. If it is indeed required change, do we want to port this over to GRPC client as well, like [here](https://github.com/googleapis/gapic-generator-java/blob/d22b966703644ea150b9ceee6df39cde17cf8df4/src/main/java/com/google/api/generator/gapic/composer/grpc/GrpcServiceStubClassComposer.java#L485)? Also in the same class, there is another [similar logic](https://github.com/googleapis/gapic-generator-java/blob/d22b966703644ea150b9ceee6df39cde17cf8df4/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java#L802) as well.

src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
```diff
@@ -907,7 +907,9 @@ private Expr createFieldsExtractorClassInstance(
       for (int i = 0; i < descendantFields.length; i++) {
         String currFieldName = descendantFields[i];
         String bindingFieldMethodName =
-            String.format("get%s", JavaStyle.toUpperCamelCase(currFieldName));
+            (i < descendantFields.length - 1 || !httpBindingFieldName.isRepeated())
```
blakeli0, line 910: Is it easier to read if we change it to `httpBindingFieldName.isRepeated() && i == descendantFields.length - 1`? Also is it a valid case that a repeated field appear in positions other than the last one? Or you are just trying to be defensive here?
If it is indeed a valid case, can you please add a test case for it as well?

vam-google:
@blakeli0 This is not cosmetic, this is required for the generated code to compile, as `get<FieldName>List()` getter is generted by protobuf for repeated fields and `get<FieldName>()` for scalar fields. 

As for porting it to grpc code - maybe, but it is not in scope of this PR as I don't want to risk breaking grpc code while working on regapic part (it is going to be a big rollout and I would like keep regapic changes separate to simplify debugging and minimize risk of breaking what is already working).

eaball35 (approved):

src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
vam-google: I just keep `i < descendantFields.length - 1` construct in the if statement to be consistent with the rest of similar statements in the same method (there are several of them in this method). 

No, repeated field can't have "nested" fields because repeated field is a List. But most importantly `httpBindingFieldName.isRepeated()` always represents the characteristic of the last field in the `.` chain only. I.e. if we have `field_one.field_two.field_three` there will be only one binding object with `name = "field_one.field_two.field_three"` and `isRepeated = false` if `field_three` is a regular field and `isRepeated = true` if it is repeated. `field_one` and `field_two` are guaranteed to be non-primitive, non-repeated message fields (when the type of `field_one` has a field named `field_two`), otherwise it would be a broken API definition.

blakeli0: If it's not a valid case, can we simplify it to `if (!httpBindingFieldName.isRepeated())`?

blakeli0:
> As for porting it to grpc code - maybe, but it is not in scope of this PR as I don't want to risk breaking grpc code while working on regapic part (it is going to be a big rollout and I would like keep regapic changes separate to simplify debugging and minimize risk of breaking what is already working).

That makes sense and I think it's OK not porting over too. Because on gRPC side, the request extractor is only used for routing headers and the field can not be repeated AFAIK. 

However, this is another example that we have long similar logics in multiple places which we could benefit a lot if extracted to common utility methods.

PR #990, opened 2022-05-12
fix: [REGAPIC] Fix snake_case nested resource names in generated unit tests

vam-google:
Prior to this fix we were comparing snake_case and camelCase of the same resource. Most often field names consist of one word (so both cases are the same), so this was not causing troubles.


eaball35 (approved):

eaball35:
lgtm. Is there an issue for this?

vam-google:
I guess buganizer automatically started creating issues for "fix" PRs. But no issue in github. I thought it was too small of a problem to have an issue for it. Do you want me to create one?

PR #993, opened 2022-05-13
feat: Add support for additional_bindings

vam-google:
This PR depends on the corresponding one in gax: https://github.com/googleapis/gax-java/pull/1680

### DIFF
--- a/src/main/java/com/google/api/generator/gapic/model/HttpBindings.java
+++ b/src/main/java/com/google/api/generator/gapic/model/HttpBindings.java
@@ -18,8 +18,10 @@ import com.google.api.generator.gapic.utils.JavaStyle;
 import com.google.auto.value.AutoValue;
 import com.google.common.collect.ImmutableSet;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 
 @AutoValue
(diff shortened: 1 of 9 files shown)


src/main/java/com/google/api/generator/gapic/model/HttpBindings.java
```diff
@@ -63,6 +65,8 @@ public int compareTo(HttpBinding o) {
 
   public abstract String pattern();
 
+  public abstract List<String> additionalPatterns();
```
blakeli0, line 68: Do we need to use `List` here? Can we use `Set`? Does order matter? Do duplicates matter?

src/main/java/com/google/api/generator/gapic/protoparser/PatternParser.java
```diff
@@ -22,7 +22,7 @@ public class PatternParser {
 
   // This method tries to parse all named segments from pattern and sort in natual order
   // e.g. /v1beta1/{table_name=tests/*}/{routing_id=instances/*}/** -> (routing_id, table_name)
-  public static Set<String> getPattenBindings(String pattern) {
```
blakeli0, line 25: Good catch! Thanks for correcting my mistake!

src/main/java/com/google/api/generator/gapic/model/HttpBindings.java
vam-google: I think technically duplicates are allowed even though they will be useless. Bindings come in list from proto file descriptor, so keeping them as list here.

src/main/java/com/google/api/generator/gapic/protoparser/PatternParser.java
vam-google: Oh, I actually thought it was my typo, only now realized that it was the class you created =).

blakeli0 (approved):

PR #994, opened 2022-05-18
chore: avoid using implicitly declared samples target in BUILD.bazel

eaball35:
Now to include samples in the package we have to pass `include_samples = True` in the build file such as:

```
java_gapic_assembly_gradle_pkg(
    name = "google-cloud-accessapproval-v1-java",
    deps = [
        ":accessapproval_java_gapic",
        ":accessapproval_java_grpc",
        ":accessapproval_java_proto",
        ":accessapproval_proto",
    ],
    include_samples = True,
)
```

Not including (as currently is) will result in no samples being included allowing us to roll out slower. This isn't whether samples will be generated but whether they would be included in the assembly package, that would require a more complicated change.

Follow up to: cl/446288272

### DIFF
--- a/rules_java_gapic/java_gapic_pkg.bzl
+++ b/rules_java_gapic/java_gapic_pkg.bzl
@@ -317,6 +317,7 @@ java_gapic_srcs_pkg = rule(
 def java_gapic_assembly_gradle_pkg(
         name,
         deps,
+        includeSamples = False,
         assembly_name = None,
         transport = None,
         **kwargs):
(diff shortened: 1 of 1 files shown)


blakeli0 (approved):

rules_java_gapic/java_gapic_pkg.bzl
```diff
@@ -317,6 +317,7 @@ java_gapic_srcs_pkg = rule(
 def java_gapic_assembly_gradle_pkg(
         name,
         deps,
+        includeSamples = False,
```
vam-google, line 320: please use snake_case

vam-google (approved):
LGTM, but please change the name of the new param to include_samples to match bazel naming convention

rules_java_gapic/java_gapic_pkg.bzl
eaball35: done

PR #997, opened 2022-05-25
fix: Multiple REST transport related fixes

vam-google:
1) Stop failing for grpc+rest and rest transports in case if there are bidi- or client- streaming methods (generated a method throwing "Unimplemented" exception if called)
2) Make sure that the resource name suggested for wildcard resource names actually matches url pattern (most of the changes in goldens are caused by this change)
3) Fix the order of "actual" and "expected" in diff_gen_and_golden to match
4) Fix multipattern `of%Name` method resoluiton to make sure that the chosen pattern matches url pattern (the changes in `MessagingClient.golden` and related files are caused by this change).
5) Fix anonymous resource name class construction (add proper toString() method matching pattern)
6) Gracefully treat methods without http bindings as not supported in rest transport (instead of failing on generation)

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -147,6 +147,7 @@ java_binary(
     name = "code_generator_request_file_to_gapic_main",
     main_class = "com.google.api.generator.debug.CodeGeneratorRequestFileToGapicMain",
     runtime_deps = [":gapic_generator_java"] + MAIN_DEPS + MAIN_DEPS_DEBUG_RUNTIME_ONLY,
+    data = [":google/pubsub/v1/pubsub_v1.yaml"],
 )
 
 # another test resource
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -147,7 +147,6 @@ java_binary(
     name = "code_generator_request_file_to_gapic_main",
     main_class = "com.google.api.generator.debug.CodeGeneratorRequestFileToGapicMain",
     runtime_deps = [":gapic_generator_java"] + MAIN_DEPS + MAIN_DEPS_DEBUG_RUNTIME_ONLY,
-    data = [":google/pubsub/v1/pubsub_v1.yaml"],
 )
 
 # another test resource
--- a/src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
+++ b/src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
@@ -32,6 +32,7 @@ import com.google.api.generator.engine.ast.Variable;
 import com.google.api.generator.engine.ast.VariableExpr;
 import com.google.api.generator.gapic.composer.resourcename.ResourceNameTokenizer;
 import com.google.api.generator.gapic.model.Field;
+import com.google.api.generator.gapic.model.HttpBindings;
 import com.google.api.generator.gapic.model.Message;
 import com.google.api.generator.gapic.model.MethodArgument;
 import com.google.api.generator.gapic.model.ResourceName;
@@ -66,7 +67,7 @@ public class DefaultValueComposer {
       Map<String, ResourceName> resourceNames,
       Map<String, Message> messageTypes,
       Map<String, String> valuePatterns,
-      String bindingPattern) {
+      HttpBindings bindings) {
     if (methodArg.isResourceNameHelper()) {
       Preconditions.checkState(
           methodArg.field().hasResourceReference(),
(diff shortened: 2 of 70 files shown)


src/main/java/com/google/api/generator/gapic/model/Method.java
```diff
@@ -62,6 +62,11 @@ public boolean isPaged() {
   @Nullable
   public abstract HttpBindings httpBindings();
 
+  @Nullable
```
blakeli0, line 65: Can we add some unit tests for the model classes? The composers are hard to test but these helper methods in model classes are very easy to be unit tested, I know it might already be covered by some golden tests, but having unit tests can help us catch the issue earlier and also easier to troubleshoot if it fails.

src/main/java/com/google/api/generator/gapic/model/ResourceName.java
```diff
@@ -55,6 +57,31 @@ public abstract class ResourceName {
 
   public abstract boolean isOnlyWildcard();
 
+  @Nullable
+  public String getMatchingPattern(String binding) {
```
blakeli0, line 61: Same comment as last method. The logic here also looks much more complicated, so unit tests would be really helpful.

BUILD.bazel
```diff
@@ -147,6 +147,7 @@ java_binary(
     name = "code_generator_request_file_to_gapic_main",
     main_class = "com.google.api.generator.debug.CodeGeneratorRequestFileToGapicMain",
     runtime_deps = [":gapic_generator_java"] + MAIN_DEPS + MAIN_DEPS_DEBUG_RUNTIME_ONLY,
+    data = [":google/pubsub/v1/pubsub_v1.yaml"],
```
blakeli0, line 150: Is this change intentional?

src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
```diff
@@ -249,8 +254,10 @@ static Expr createResourceHelperValue(
       List<ResourceName> unexaminedResnames = new ArrayList<>(resnames);
       for (ResourceName resname : resnames) {
         unexaminedResnames.remove(resname);
-        if (!resname.isOnlyWildcard()) {
-          return createResourceHelperValue(resname, false, unexaminedResnames, fieldOrMessageName);
+        if (!resname.isOnlyWildcard()
```
blakeli0, line 257: I see a couple of new logics in this method, and I understand it's in composers so unit testing it might get complicated, can we add some test cases in the protos if it's not too hard?

src/main/java/com/google/api/generator/gapic/model/Method.java
vam-google: This method got deleted actually in the final version

src/main/java/com/google/api/generator/gapic/model/ResourceName.java
vam-google: We don't have tests for Resourcename class directly, but it all gets tested as part of bigger test picture.

BUILD.bazel
vam-google: No, it is not. Good catch! removed

src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
vam-google: My general feeling about this codebase is that it is beyond comprehension unfortunately and this affects the testing and developing strategy.
This new logic is being tested by the current set of tests already.

src/main/java/com/google/api/generator/gapic/model/ResourceName.java
blakeli0: Can we please add a test class for it? This method is pretty complicated and having unit tests will not only help coverage and help us to understand it as well(easier for us to maintain in the future). It may be already tested somewhere but unit test will always fail fast, so that it's much easier to troubleshoot.

src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
blakeli0: I understand the sentiment and I totally feel you when I was developing dynamic routing headers, but do we want to make it worse?  e.g. I added this [proto](https://github.com/googleapis/gapic-generator-java/blob/main/src/test/proto/explicit_dynamic_routing_header_testing.proto) to test different combinations for dynamic routing headers, it may still not cover everything but I think it helps.

vam-google: I really believe that those extra tests are not making it better, but making it worse. Basically we add more unmaintainable code. This codebase has 10+ times more code than it should for the task it performs.

vam-google:
@blakeli0 PTAL

blakeli0 (approved):

PR #1003, opened 2022-06-08
fix: More REST transport fixes

vam-google:
1) Generate an empty unit test  method even if the method is not supported in a particular transport. Bazel fails test execution, if the test class has no `@Test` methods. In case if there is a service with only unsupported methods, and we do not generate unit tests for those, we would get an empty unit test, which will fail execution in our automatic bazel build pipeline.

2) For fields of `google.protobuf.Value` type set an actual value for it (to avoid non-set vs null value differences after json serialization/deserializaiton roudtrip). This looks like an issue in protobuf support of `google.protobuf.Value` type in JSON and not specific to GAPIC.

### DIFF
--- a/src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
+++ b/src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
@@ -168,6 +168,20 @@ public class DefaultValueComposer {
               .setStaticReferenceType(field.type())
               .setMethodName("newBuilder")
               .build();
+      if (field.type().equals(TypeNode.VALUE)) {
+        newBuilderExpr =
+            MethodInvocationExpr.builder()
+                .setExprReferenceExpr(newBuilderExpr)
+                .setMethodName("setBoolValue")
+                .setArguments(
+                    ValueExpr.withValue(
+                        PrimitiveValue.builder()
+                            .setType(TypeNode.BOOLEAN)
+                            .setValue("true")
+                            .build()))
+                .build();
+      }
+
       return MethodInvocationExpr.builder()
           .setExprReferenceExpr(newBuilderExpr)
           .setMethodName("build")
--- a/src/test/java/com/google/api/generator/gapic/composer/grpcrest/goldens/EchoClientHttpJsonTest.golden
+++ b/src/test/java/com/google/api/generator/gapic/composer/grpcrest/goldens/EchoClientHttpJsonTest.golden
@@ -18,6 +18,7 @@ import com.google.longrunning.Operation;
 import com.google.protobuf.Any;
 import com.google.protobuf.Duration;
 import com.google.protobuf.Timestamp;
+import com.google.protobuf.Value;
 import com.google.rpc.Status;
 import com.google.showcase.grpcrest.v1beta1.stub.HttpJsonEchoStub;
 import java.io.IOException;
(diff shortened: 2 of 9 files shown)


src/test/java/com/google/api/generator/gapic/composer/grpcrest/goldens/EchoClientHttpJsonTest.golden
```diff
@@ -805,4 +814,16 @@ public class EchoClientHttpJsonTest {
       // Expected exception.
     }
   }
+
+  @Test
```
blakeli0, line 818: Is it possible to not generate this whole test class if this case happens? I know it would be a rare case, so if it takes a lot of effort, I think this is fine.

src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
```diff
@@ -168,6 +168,20 @@ public static Expr createValue(
               .setStaticReferenceType(field.type())
               .setMethodName("newBuilder")
               .build();
+      if (field.type().equals(TypeNode.VALUE)) {
```
blakeli0, line 171: Not sure if I understand why we need to set the default value for response, can you please provide a little more explanation? Also looks like this method is only called by test composers?

src/test/java/com/google/api/generator/gapic/composer/grpcrest/goldens/EchoClientHttpJsonTest.golden
vam-google: It is much-much simpler to generate empty tests, but I believe it is a better approach to have empty tests instead of not having a class (eve if it was harder to do)

src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
vam-google: Honestly, I don't know myself what happens. Looks like a but in json serialization/deserialization for Value class. It is a special class (like `Object` in java or `void *` in C). Due to this when we instantiate `Value.newBuilder.build()` it is recognized as an empty value, but when serialized->deserialized to/from json, then it gets recognized as special `null` value, and as a result the `Assert.equals()` throws.

vam-google:
@blakeli0 PTAL, this is the last PR I need to submit before rolling our regapic

src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
blakeli0: Maybe I'm missing something, what I don't understand is that this code only adds a default boolean value "true" to the expected response in the generated tests, where is this value coming from in the actual response?

blakeli0 (approved):

src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
vam-google: In each generated test there is `MockHttpService` instance used. Before calling a tested method, the expected response is added to the mockService. The response returned from the tested methos is the expectedResult object which went through serialization/deserialization cycle.

PR #1004, opened 2022-06-14
fix: Remove HttpMethod compile time dependency from genrated HttpJson stubs

vam-google:


Neenu1995 (approved):

src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
```diff
@@ -14,7 +14,6 @@
 
 package com.google.api.generator.gapic.composer.rest;
 
-import com.google.api.client.http.HttpMethods;
```
blakeli0, line 17: Just wondering what's the benefit of removing the dependency of this class? I think we still need the http client library so we can not remove the dependency entirely in gax?

Neenu1995: The REGAPIC files were adding an additional dependency on google-http-client in the downstream GAPIC libraries. This fix avoids that.
http-client is a legacy library we don't want in our GAPIC dependency tree.

PR #1011, opened 2022-07-06
feat(ast): add support for LambdaExpr to infer type from return expr type

zhumin8:
This will enable it to be assigned to an appropriate variable

**Before this change:**  cannot assign a lambda expression to a variable because its type is always `VOID`. 
**After this change:** lambda expression builder will infer type from `returnExpr`’s expression type. Allowing it to be assigned to an appropriate variable. Because `returnExpr` is a required field, its type should always be accessible.
**Example:** a naive sample usage is provided in [JavaWriterVisitorTest.java](https://github.com/googleapis/gapic-generator-java/pull/1011/files#diff-60163f71e845b4cc97f9ccd7101646aa243fa5317a77cd326a7f61143618f62f) as test. Another example is what I am trying to generate for spring autoconfig code: `this.projectIdProvider = () -> clientProperties.getProjectId();`. Ideally through method reference, but lambda assignment does the same thing and it’s an easier feature addition.

### DIFF
--- a/src/main/java/com/google/api/generator/engine/ast/LambdaExpr.java
+++ b/src/main/java/com/google/api/generator/engine/ast/LambdaExpr.java
@@ -25,8 +25,6 @@ import java.util.stream.Collectors;
 
 @AutoValue
 public abstract class LambdaExpr implements Expr {
-  @Override
-  public abstract TypeNode type();
 
   public abstract ImmutableList<VariableExpr> arguments();
 
--- a/src/main/java/com/google/api/generator/engine/ast/LambdaExpr.java
+++ b/src/main/java/com/google/api/generator/engine/ast/LambdaExpr.java
@@ -26,7 +26,9 @@ import java.util.stream.Collectors;
 @AutoValue
 public abstract class LambdaExpr implements Expr {
   @Override
-  public abstract TypeNode type();
+  public TypeNode type() {
+    return returnExpr().expr().type();
+  }
 
   public abstract ImmutableList<VariableExpr> arguments();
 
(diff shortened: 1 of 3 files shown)


zhumin8:
Yes, the improvement is exactly as you described. 
I've updated the description above to include a bit more explanation.

src/main/java/com/google/api/generator/engine/ast/LambdaExpr.java
```diff
@@ -26,11 +26,7 @@
 @AutoValue
 public abstract class LambdaExpr implements Expr {
   @Override
-  public TypeNode type() {
-    // TODO(v2): Support set of FunctionalInterface  parameterized on the args and return type,
-    // which would enable assignment to an appropriate variable.
-    return TypeNode.VOID;
-  }
+  public abstract TypeNode type();
```
zhumin8, line 29: keeping this despite sonar complains about already defined in `Expr` interface. It's seems the recommend way for @AutoValue as explained in [this doc](https://github.com/google/auto/blob/master/value/userguide/howto.md#-have-autovalue-implement-a-concrete-or-default-method). Also keeps consistent with other expression classes like [`ValueExpr`](https://github.com/googleapis/gapic-generator-java/blob/main/src/main/java/com/google/api/generator/engine/ast/ValueExpr.java#L21), [`CastExpr`](https://github.com/googleapis/gapic-generator-java/blob/main/src/main/java/com/google/api/generator/engine/ast/CastExpr.java#L23)

blakeli0: If you want, we can still expose the `setType()` method and set a default value in the `builder()` method, but like you said `ReturnExpr().expr()` should always have `type` associated so we should never call `setType()` explicitly.

blakeli0 (approved):
LGTM

PR #1012, opened 2022-07-07
feat(ast): extend support for annotation named parameters

zhumin8:
[updated] 
This change is to extend annotation arguments support.
**Before**: Only one argument that consist of a single `String` can be added to annotation description.
**After**:  For single unnamed parameter/description, add support for `ValueExpr` and `VariableExpr` in addition to `String`. Also allow adding multiple named parameters of type `VariableExpr`. So that annotations like below can be generated:

```
@ConditionalOnClass(VisionServiceClient.class)
@ConditionalOnProperty(value = "spring.cloud.gcp.vision.enabled", matchIfMissing = true)
```
this improvement is an addition and does not affect existing usage of `setDescription(String description)`.
Added tests in [JavaWriterVisitorTest.java](https://github.com/googleapis/gapic-generator-java/pull/1012/files#diff-60163f71e845b4cc97f9ccd7101646aa243fa5317a77cd326a7f61143618f62f) also shows the new feature usages.


src/main/java/com/google/api/generator/engine/ast/ClazzValue.java
```diff
@@ -0,0 +1,45 @@
...
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.api.generator.engine.ast;
+
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+public abstract class ClazzValue implements ObjectValue {
```
blakeli0, line 19: We already have a way to achieve the same thing, see the code [here](https://github.com/googleapis/gapic-generator-java/blob/6e02c70bf1b2fdd4ea382427fe8271a00744d915/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceStubSettingsClassComposer.java#L331) and the golden test file [here](https://github.com/googleapis/gapic-generator-java/blob/6e02c70bf1b2fdd4ea382427fe8271a00744d915/src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/LoggingServiceV2StubSettings.golden#L463). I think it's better not to have different ways to achieve the same thing, it would create more maintenance burden in the future. If you want, you could extract the common logic to a util method/class, something like `createClazzValueExpr` or `createClazzVariableExpr`, which only takes type as an argument.

src/main/java/com/google/api/generator/engine/ast/AnnotationNode.java
```diff
@@ -62,12 +66,16 @@ public abstract static class Builder {
     public abstract Builder setType(TypeNode type);
 
     public Builder setDescription(String description) {
-      return setDescriptionExpr(ValueExpr.withValue(StringObjectValue.withValue(description)));
+      return setDescriptionExprs(
+          Arrays.asList(ValueExpr.withValue(StringObjectValue.withValue(description))));
+    }
+
+    public Builder setDescriptions(List<Expr> exprList) {
```
blakeli0, line 73: I like this improvement in general, but we should be careful exposing a setter that can set descriptions to any `Expr`, you can technically pass a `NewObjectExpr` or a `ReturnExpr` and everything will work until we try to compile the generated code, which could be too late to find an issue. 
Correct me if I'm wrong, I think we only allow `ValueExpr`, `AssignmentExpr`, or `VariableExpr`(for clazz if you remove `CLazzValue`). So we could either add some type checks in the `build()` method, or only expose methods to add descriptions for the above types, like `addDescription(ValueExpr expr)`, `addDescription(AssignmentExpr expr)` and `addDescription(VariableExpr expr)`.
Basically anything that could restrict people from passing any types of `Expr` would be great.

src/main/java/com/google/api/generator/engine/ast/ClazzValue.java
zhumin8: Ah, thanks for the tip. I couldn't wrap my head around needing a `Value` instead of `Variable`.

src/main/java/com/google/api/generator/engine/ast/AnnotationNode.java
zhumin8: Good point, let me refactor this part and expose methods for each expr type.

zhumin8:
Adding a small enhancement to this PR since it's related.
In addition to previous changes, I added an optional field to VariableExpr allowing Annotations. 
Prior to this, only annotations on ClassDefinition and MethodDefinition are supported. With this addition, I will be able to generate code like:
```
  @NestedConfigurationProperty
  private final Credentials credentials = new Credentials();

  // or 
  @Autowired
  private LanguageServiceClient autoClient;
```
Field annotations are pretty common in Spring syntax, this enhancement is needed to generate Spring autoconfig code as I planned. 
Note: Added a todo note in `VariableExpr` only as a nice to have feature for future, it's not blocking any use cases for now.   But it seems reasonable to have target info in `AnnotationNode` and apply checks on it when adding annotation nodes.

src/main/java/com/google/api/generator/engine/ast/VariableExpr.java
```diff
@@ -133,7 +143,20 @@ public VariableExpr build() {
                   })
               .collect(Collectors.toList()));
 
+      // Remove duplicates while maintaining insertion order.
+      ImmutableList<AnnotationNode> processedAnnotations = annotations();
+      setAnnotations(
+          new LinkedHashSet<>(processedAnnotations).stream().collect(Collectors.toList()));
```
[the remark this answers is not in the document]
zhumin8: The `ImmutableList` can have duplicates. For example if you attempt to add 2 `@Deprecated` to a variable, will result in code that cannot compile if not de-duped.
That said, I don't feel too strongly about this check itself, but this duplicate check at least does no harm. 
I am really trying to follow similar logic in `MethodDefinition` [here](https://github.com/googleapis/gapic-generator-java/blob/66782d1109fae176b0f3266a3dacb4b55cf3b677/src/main/java/com/google/api/generator/engine/ast/MethodDefinition.java#L256-L258). Adding annotation to `VariableExpr` should be parallel to adding annotation to `MethodDefinition` (except `@Override` is only applicable to method, so `isOverride()` is irrelevant here). Thus I would rather avoid having different logics in builder method here vs. `MethodDefinition`.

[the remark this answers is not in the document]
blakeli0: I see, I think it was a poor decision to use a List and de dup it later, but since we already have a pattern, let's follow it for now and refactor it later.

blakeli0 (approved):

Issue #843, opened 2021-09-14
LambdaExpr doesn't support void-typed return expressions


chanseokoh:
https://github.com/googleapis/gapic-generator-java/blob/92f7c1cf7b343947a64943fd7ee7ffb4d67a9d5a/src/main/java/com/google/api/generator/engine/ast/LambdaExpr.java#L72-L74

`LambdaExpr` requires a `ReturnExpr` to be set. And `LambdaExpr.build()` explicitly forbids void-typed return expressions. So it's not possible to create a lambda like this:
```java
    Runnable r1 = () -> System.out.println();
    Runnable r2 = () -> {
      System.out.println();
    };
    Runnable r3 = System.out::println;
```

PR #1019, opened 2022-07-26
fix(ast): removed lambda void return type check

diego92sigma6:
Fix for Issue #843. This pull request modifies allows the creation of lambda expressions that have _void_ as return type. The main change was made in `ast/LambdaExpr.java`


blakeli0 (approved):

blakeli0:
LGTM. Please link the fixed issue in the description.

PR #1020, opened 2022-07-26
feat: Add support for rest numeric enums.

blakeli0:
See b/232457244 for exact requirements. In short, this PR added four things:
1. Add a flag to the Java microgenerator Bazel plugin to determine whether we need to generate a client library that supports rest numeric enums. This flag is to make sure we only generate client libraries with numeric enums for the services that have already upgraded to the latest ESF. See go/gapic-numericenum-release for details.
2. Serialize enum object to int if the above flag is set to true
3. Add a query parameter `$alt=json;enum-encoding=int` to every request if the above flag is set to true so that the backend will return response with numeric enums.
4. Upgrade gax-java to the latest since it's dependent on https://github.com/googleapis/gax-java/pull/1743.

### DIFF
--- a/rules_java_gapic/java_gapic.bzl
+++ b/rules_java_gapic/java_gapic.bzl
@@ -189,6 +189,7 @@ def _java_gapic_srcjar(
         service_yaml,
         # possible values are: "grpc", "rest", "grpc+rest"
         transport,
+        rest_numeric_enums,
         # Can be used to provide a java_library with a customized generator,
         # like the one which dumps descriptor to a file for future debugging.
         java_generator_name = "java_gapic",
@@ -212,6 +213,9 @@ def _java_gapic_srcjar(
 
     if transport:
         opt_args.append("transport=%s" % transport)
+
+    if rest_numeric_enums:
+        opt_args.append("rest-numeric-enums")
 
     # Produces the GAPIC metadata file if this flag is set. to any value.
     # Protoc invocation: --java_gapic_opt=metadata
@@ -240,6 +244,7 @@ def java_gapic_library(
         test_deps = [],
         # possible values are: "grpc", "rest", "grpc+rest"
         transport = None,
+        rest_numeric_enums = False,
         **kwargs):
     srcjar_name = name + "_srcjar"
     raw_srcjar_name = srcjar_name + "_raw"
@@ -251,6 +256,7 @@ def java_gapic_library(
         gapic_yaml = gapic_yaml,
         service_yaml = service_yaml,
         transport = transport,
+        rest_numeric_enums = rest_numeric_enums,
         java_generator_name = "java_gapic",
         **kwargs
     )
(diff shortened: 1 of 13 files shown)


meltsufin (approved):

src/main/java/com/google/api/generator/gapic/protoparser/PluginArgumentParser.java
```diff
@@ -89,8 +94,8 @@ private static Optional<String> parseConfigArgument(String pluginProtocArgument,
   }
 
   @VisibleForTesting
-  static boolean hasMetadataFlag(String pluginProtocArgument) {
-    return Arrays.stream(pluginProtocArgument.split(COMMA)).anyMatch(s -> s.equals(KEY_METADATA));
+  static boolean hasFlag(String pluginProtocArgument, String flagKey) {
+    return Arrays.asList(pluginProtocArgument.split(COMMA)).contains(flagKey);
```
vam-google, line 98: I believe stream should be preferred way, as it does not create actual list and things happen on the fly (like one pass and no intermediate list, unlike two passes (one to create list and another one to do contains()).

src/test/java/com/google/api/generator/gapic/composer/grpcrest/goldens/HttpJsonEchoStub.golden
```diff
@@ -78,7 +78,7 @@ public class HttpJsonEchoStub extends EchoStub {
                       })
                   .setRequestBodyExtractor(
                       request ->
-                          ProtoRestSerializer.create().toBody("*", request.toBuilder().build()))
+                          ProtoRestSerializer.create().toBody(request.toBuilder().build(), false))
```
vam-google, line 81: I agree it is a bit subjective, but can you please keep the field name thing, as it looks clearer and matches the proto config more directly (config does have the names so it is easier to search and match things between generated code and proto file). Also it is basically an implementaiton detail that the field name is not actually used now, it may be used later.

src/main/java/com/google/api/generator/gapic/protoparser/PluginArgumentParser.java
blakeli0: Reverted to stream.

src/test/java/com/google/api/generator/gapic/composer/grpcrest/goldens/HttpJsonEchoStub.golden
blakeli0: Added the fieldName back.

rules_java_gapic/java_gapic.bzl
```diff
@@ -213,6 +214,9 @@ def _java_gapic_srcjar(
     if transport:
         opt_args.append("transport=%s" % transport)
 
+    if rest_numeric_enums:
+        opt_args.append("rest-numeric-enums")
```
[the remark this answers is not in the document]
blakeli0: Yes, this is a standardized name and type for all languages.

PR #1023, opened 2022-08-02
[WIP] feat(composer): Primitive query params

### LINKED ISSUE #1016: Non-primitive values should not be put into query parameters
In some cases, the generated client library would put complex object to query parameters, like [here](https://github.com/googleapis/gapic-generator-java/blob/889f316d7dfcfdaf826cb8024166cc8eabeaef0d/src/test/java/com/google/api/generator/gapic/composer/rest/goldens/HttpJsonComplianceStub.golden#L176) , the proto definition for this method is [here](https://github.com/googleapis/gapic-generator-java/blob/889f316d7dfcfdaf826cb8024166cc8eabeaef0d/src/test/proto/compliance.proto#L64). The generated client library will error out during runtime if non-primitive values are put into query parameters.

Even though a rpc method definition like this should never happen, but in case it happens, we should produce errors during generation time instead of runtime.

TODO: Create a separate issue for path parameters.

diegomarquezp:
Fixes #1016

### DIFF
--- a/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
+++ b/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
@@ -912,7 +912,6 @@ public class HttpJsonServiceStubClassComposer extends AbstractTransportServiceSt
     VariableExpr requestVarExpr =
         VariableExpr.withVariable(
             Variable.builder().setType(method.inputType()).setName("request").build());
-
     for (HttpBinding httpBindingFieldName : httpBindingFieldNames) {
       // Handle foo.bar cases by descending into the subfields.
       MethodInvocationExpr.Builder requestFieldGetterExprBuilder =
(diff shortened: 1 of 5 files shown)


blakeli0 (approved):

src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
```diff
@@ -52,12 +52,9 @@
 import com.google.api.generator.engine.ast.VariableExpr;
 import com.google.api.generator.gapic.composer.common.AbstractTransportServiceStubClassComposer;
 import com.google.api.generator.gapic.composer.store.TypeStore;
+import com.google.api.generator.gapic.model.*;
```
blakeli0, line 55: Wildcard import should not be used based on go/java-style#s3.3.1-wildcard-imports

vam-google (changes requested):
The original got redefined, thus the PR fixing it is expected to be modified to match that. Please see the updated bug description.

blakeli0:
The original issue is not valid anymore.

PR #1040, opened 2022-09-15
fix: update sample region tag to parse host instead of proto package

alicejli:
Per SnippetGen design doc- the ApiShortName in the RegionTag should be generated from the host of the proto. 

Prior to this change, the Java generator parses the proto package name for the ApiShortName, leading to occasional discrepancies.

### DIFF
--- a/src/main/java/com/google/api/generator/gapic/composer/Composer.java
+++ b/src/main/java/com/google/api/generator/gapic/composer/Composer.java
@@ -37,10 +37,11 @@ import com.google.api.generator.gapic.model.Sample;
 import com.google.api.generator.gapic.model.Service;
 import com.google.api.generator.gapic.model.Transport;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Splitter;
+import com.google.common.collect.Iterables;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
-import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 public class Composer {
@@ -223,25 +224,14 @@ public class Composer {
   // Will keep backup name parsed from proto package for now.
   @VisibleForTesting
   protected static String parseDefaultHost(String defaultHost) {
-    String apiShortName;
-    String apiShortNameHelper;
     // If the defaultHost is of the format "**.googleapis.com", take the name before the first
     // period.
-    Pattern standardApiHost = Pattern.compile("\\.");
+    String apiShortName = Iterables.getFirst(Splitter.on(".").split(defaultHost), defaultHost);
     // If the defaultHost is of the format "**-**-**.googleapis.com", take the section before the
     // first period and after the last dash to follow CSharp's implementation here:
     // https://github.com/googleapis/gapic-generator-csharp/blob/main/Google.Api.Generator/Generation/ServiceDetails.cs#L70
-    Pattern regionalApiHost = Pattern.compile("-");
-    apiShortNameHelper = standardApiHost.split(defaultHost)[0];
-    apiShortName = defaultHost;
-    if (standardApiHost.split(defaultHost).length > 1
-        && regionalApiHost.split(defaultHost).length > 1) {
-      apiShortName =
-          regionalApiHost
-              .split(apiShortNameHelper)[regionalApiHost.split(apiShortNameHelper).length - 1];
-    } else {
-      apiShortName = apiShortNameHelper;
-    }
+    apiShortName = Iterables.getLast(Splitter.on("-").split(apiShortName), defaultHost);
+    // TODO: alicejli will need to handle special case for `iam-meta-api` service
     return apiShortName;
   }
 
--- a/src/main/java/com/google/api/generator/gapic/composer/Composer.java
+++ b/src/main/java/com/google/api/generator/gapic/composer/Composer.java
@@ -220,7 +220,7 @@ public class Composer {
     return clazzesWithSamples;
   }
 
-  // Parse defaultHost for apiShortName. Need to account for regional default endpoints like
+  // Parse defaultHost for apiShortName for the RegionTag. Need to account for regional default endpoints like
   // "us-east1-pubsub.googleapis.com".
   @VisibleForTesting
   protected static String parseDefaultHost(String defaultHost) {
--- a/src/main/java/com/google/api/generator/gapic/composer/Composer.java
+++ b/src/main/java/com/google/api/generator/gapic/composer/Composer.java
@@ -15,6 +15,7 @@
 package com.google.api.generator.gapic.composer;
 
 import com.google.api.generator.engine.ast.ClassDefinition;
+import com.google.api.generator.engine.ast.CommentStatement;
 import com.google.api.generator.gapic.composer.comment.CommentComposer;
 import com.google.api.generator.gapic.composer.grpc.GrpcServiceCallableFactoryClassComposer;
 import com.google.api.generator.gapic.composer.grpc.GrpcServiceStubClassComposer;
@@ -29,16 +30,12 @@ import com.google.api.generator.gapic.composer.grpcrest.HttpJsonServiceClientTes
 import com.google.api.generator.gapic.composer.resourcename.ResourceNameHelperClassComposer;
 import com.google.api.generator.gapic.composer.rest.HttpJsonServiceCallableFactoryClassComposer;
 import com.google.api.generator.gapic.composer.rest.HttpJsonServiceStubClassComposer;
-import com.google.api.generator.gapic.model.GapicClass;
-import com.google.api.generator.gapic.model.GapicContext;
-import com.google.api.generator.gapic.model.GapicPackageInfo;
-import com.google.api.generator.gapic.model.Sample;
-import com.google.api.generator.gapic.model.Service;
-import com.google.api.generator.gapic.model.Transport;
+import com.google.api.generator.gapic.model.*;
 import com.google.common.annotations.VisibleForTesting;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 public class Composer {
@@ -191,17 +188,14 @@ public class Composer {
 
   @VisibleForTesting
   static List<GapicClass> prepareExecutableSamples(List<GapicClass> clazzes, String protoPackage) {
-    //  parse protoPackage for apiVersion and apiShortName
+    //  parse protoPackage for apiVersion
     String[] pakkage = protoPackage.split("\\.");
     String apiVersion;
-    String apiShortName;
     //  e.g. v1, v2, v1beta1
     if (pakkage[pakkage.length - 1].matches("v[0-9].*")) {
       apiVersion = pakkage[pakkage.length - 1];
-      apiShortName = pakkage[pakkage.length - 2];
     } else {
       apiVersion = "";
-      apiShortName = pakkage[pakkage.length - 1];
     }
     //  Include license header, apiShortName, and apiVersion
     return clazzes.stream()
@@ -210,17 +204,48 @@ public class Composer {
               List<Sample> samples =
                   gapicClass.samples().stream()
                       .map(
-                          sample -> addRegionTagAndHeaderToSample(sample, apiShortName, apiVersion))
+                          sample ->
+                              addRegionTagAndHeaderToSample(
+                                  sample, parseDefaultHost(gapicClass.defaultHost()), apiVersion))
                       .collect(Collectors.toList());
               return gapicClass.withSamples(samples);
             })
         .collect(Collectors.toList());
   }
 
-  private static Sample addRegionTagAndHeaderToSample(
+  // Parse defaultHost for apiShortName. Need to account for regional default endpoints like
+  // "us-east1-pubsub.googleapis.com".
+  // Will keep backup name parsed from proto package for now.
+  @VisibleForTesting
+  protected static String parseDefaultHost(String defaultHost) {
+    String apiShortName;
+    String apiShortNameHelper;
+    // If the defaultHost is of the format "**.googleapis.com", take the name before the first
+    // period.
+    Pattern standardApiHost = Pattern.compile("\\.");
+    // If the defaultHost is of the format "**-**-**.googleapis.com", take the section before the
+    // first period and after the last dash to follow CSharp's implementation here:
+    // https://github.com/googleapis/gapic-generator-csharp/blob/main/Google.Api.Generator/Generation/ServiceDetails.cs#L70
+    Pattern regionalApiHost = Pattern.compile("-");
+    apiShortNameHelper = standardApiHost.split(defaultHost)[0];
+    apiShortName = defaultHost;
+    if (standardApiHost.split(defaultHost).length > 1
+        && regionalApiHost.split(defaultHost).length > 1) {
+      apiShortName =
+          regionalApiHost
+              .split(apiShortNameHelper)[regionalApiHost.split(apiShortNameHelper).length - 1];
+    } else {
+      apiShortName = apiShortNameHelper;
+    }
+    return apiShortName;
+  }
+
+  @VisibleForTesting
+  protected static Sample addRegionTagAndHeaderToSample(
       Sample sample, String apiShortName, String apiVersion) {
+    final List<CommentStatement> header = Arrays.asList(CommentComposer.APACHE_LICENSE_COMMENT);
     return sample
-        .withHeader(Arrays.asList(CommentComposer.APACHE_LICENSE_COMMENT))
+        .withHeader(header)
         .withRegionTag(
             sample.regionTag().withApiVersion(apiVersion).withApiShortName(apiShortName));
   }
--- a/src/main/java/com/google/api/generator/gapic/composer/Composer.java
+++ b/src/main/java/com/google/api/generator/gapic/composer/Composer.java
@@ -231,7 +231,10 @@ public class Composer {
     // first period and after the last dash to follow CSharp's implementation here:
     // https://github.com/googleapis/gapic-generator-csharp/blob/main/Google.Api.Generator/Generation/ServiceDetails.cs#L70
     apiShortName = Iterables.getLast(Splitter.on("-").split(apiShortName), defaultHost);
-    // TODO: alicejli will need to handle special case for `iam-meta-api` service
+    // `iam-meta-api` service is an exceptional case and is handled as a one-off
+      if(defaultHost.contains("iam-meta-api")){
+          apiShortName="iam";
+      }
     return apiShortName;
   }
 
--- a/src/main/java/com/google/api/generator/gapic/composer/Composer.java
+++ b/src/main/java/com/google/api/generator/gapic/composer/Composer.java
@@ -220,7 +220,8 @@ public class Composer {
     return clazzesWithSamples;
   }
 
-  // Parse defaultHost for apiShortName for the RegionTag. Need to account for regional default endpoints like
+  // Parse defaultHost for apiShortName for the RegionTag. Need to account for regional default
+  // endpoints like
   // "us-east1-pubsub.googleapis.com".
   @VisibleForTesting
   protected static String parseDefaultHost(String defaultHost) {
--- a/src/main/java/com/google/api/generator/gapic/composer/Composer.java
+++ b/src/main/java/com/google/api/generator/gapic/composer/Composer.java
@@ -203,20 +203,21 @@ public class Composer {
     } else {
       apiVersion = "";
     }
-    //  Include license header, apiShortName, and apiVersion
-    return clazzes.stream()
-        .map(
-            gapicClass -> {
-              List<Sample> samples =
-                  gapicClass.samples().stream()
-                      .map(
-                          sample ->
-                              addRegionTagAndHeaderToSample(
-                                  sample, parseDefaultHost(gapicClass.defaultHost()), apiVersion))
-                      .collect(Collectors.toList());
-              return gapicClass.withSamples(samples);
-            })
-        .collect(Collectors.toList());
+    // Include license header, apiShortName, and apiVersion
+    List<GapicClass> clazzesWithSamples = new ArrayList<>();
+    clazzes.forEach(
+        gapicClass -> {
+          List<Sample> samples = new ArrayList<>();
+          gapicClass
+              .samples()
+              .forEach(
+                  sample ->
+                      samples.add(
+                          addRegionTagAndHeaderToSample(
+                              sample, parseDefaultHost(gapicClass.defaultHost()), apiVersion)));
+          clazzesWithSamples.add(gapicClass.withSamples(samples));
+        });
+    return clazzesWithSamples;
   }
 
   // Parse defaultHost for apiShortName. Need to account for regional default endpoints like
--- a/src/main/java/com/google/api/generator/gapic/composer/Composer.java
+++ b/src/main/java/com/google/api/generator/gapic/composer/Composer.java
@@ -232,9 +232,9 @@ public class Composer {
     // https://github.com/googleapis/gapic-generator-csharp/blob/main/Google.Api.Generator/Generation/ServiceDetails.cs#L70
     apiShortName = Iterables.getLast(Splitter.on("-").split(apiShortName), defaultHost);
     // `iam-meta-api` service is an exceptional case and is handled as a one-off
-      if(defaultHost.contains("iam-meta-api")){
-          apiShortName="iam";
-      }
+    if (defaultHost.contains("iam-meta-api")) {
+      apiShortName = "iam";
+    }
     return apiShortName;
   }
 
--- a/src/main/java/com/google/api/generator/gapic/composer/Composer.java
+++ b/src/main/java/com/google/api/generator/gapic/composer/Composer.java
@@ -222,7 +222,6 @@ public class Composer {
 
   // Parse defaultHost for apiShortName. Need to account for regional default endpoints like
   // "us-east1-pubsub.googleapis.com".
-  // Will keep backup name parsed from proto package for now.
   @VisibleForTesting
   protected static String parseDefaultHost(String defaultHost) {
     // If the defaultHost is of the format "**.googleapis.com", take the name before the first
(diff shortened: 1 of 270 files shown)


src/test/java/com/google/api/generator/gapic/composer/ComposerTest.java
```diff
@@ -42,8 +37,13 @@ public class ComposerTest {
...
       Arrays.asList(
           GrpcServiceCallableFactoryClassComposer.instance().generate(context, echoProtoService));
-  private final String protoPackage = context.gapicMetadata().getProtoPackage();
-  private final List<Sample> samples = clazzes.get(0).samples();
+  private final Sample sample =
+      Sample.builder()
+          .setRegionTag(
+              RegionTag.builder().setServiceName("servicename").setRpcName("rpcname").build())
+          .build();
+  private List<Sample> ListofSamples = Arrays.asList(new Sample[] {sample});
```
diego92sigma6, line 45: this non-constant class field should use lowerCamelCase

src/main/java/com/google/api/generator/gapic/composer/Composer.java
```diff
@@ -29,16 +30,12 @@
 import com.google.api.generator.gapic.composer.resourcename.ResourceNameHelperClassComposer;
 import com.google.api.generator.gapic.composer.rest.HttpJsonServiceCallableFactoryClassComposer;
 import com.google.api.generator.gapic.composer.rest.HttpJsonServiceStubClassComposer;
-import com.google.api.generator.gapic.model.GapicClass;
-import com.google.api.generator.gapic.model.GapicContext;
-import com.google.api.generator.gapic.model.GapicPackageInfo;
-import com.google.api.generator.gapic.model.Sample;
-import com.google.api.generator.gapic.model.Service;
-import com.google.api.generator.gapic.model.Transport;
+import com.google.api.generator.gapic.model.*;
```
diego92sigma6, line 33: Seems like the IDE defaulted to use a star for a bunch of changed files

alicejli:
Adding a `do not merge` label as I confirmed with Amanda that the `iam` service ApiShortName should be fixed (it currently reads as `api` because the hostname is `iam-meta-api.googleapis.com`. Will wait until a decision is made on how to handle this exception before fixing and merging.

src/test/java/com/google/api/generator/gapic/composer/ComposerTest.java
alicejli: thanks! just updated.

src/main/java/com/google/api/generator/gapic/composer/Composer.java
alicejli: Thanks for catching! I reverted.

src/main/java/com/google/api/generator/gapic/composer/Composer.java
```diff
@@ -210,17 +209,48 @@ static List<GapicClass> prepareExecutableSamples(List<GapicClass> clazzes, Strin
...
+    String apiShortName;
+    String apiShortNameHelper;
+    // If the defaultHost is of the format "**.googleapis.com", take the name before the first
+    // period.
+    Pattern standardApiHost = Pattern.compile("\\.");
+    // If the defaultHost is of the format "**-**-**.googleapis.com", take the section before the
+    // first period and after the last dash to follow CSharp's implementation here:
+    // https://github.com/googleapis/gapic-generator-csharp/blob/main/Google.Api.Generator/Generation/ServiceDetails.cs#L70
+    Pattern regionalApiHost = Pattern.compile("-");
+    apiShortNameHelper = standardApiHost.split(defaultHost)[0];
```
blakeli0, line 235: Try to use `StringSplitter` from Guava instead of the default String split, see https://errorprone.info/bugpattern/StringSplitter

src/main/java/com/google/api/generator/gapic/composer/Composer.java
```diff
@@ -210,17 +209,48 @@ static List<GapicClass> prepareExecutableSamples(List<GapicClass> clazzes, Strin
...
+  @VisibleForTesting
+  protected static String parseDefaultHost(String defaultHost) {
+    String apiShortName;
+    String apiShortNameHelper;
+    // If the defaultHost is of the format "**.googleapis.com", take the name before the first
+    // period.
+    Pattern standardApiHost = Pattern.compile("\\.");
+    // If the defaultHost is of the format "**-**-**.googleapis.com", take the section before the
+    // first period and after the last dash to follow CSharp's implementation here:
+    // https://github.com/googleapis/gapic-generator-csharp/blob/main/Google.Api.Generator/Generation/ServiceDetails.cs#L70
```
blakeli0, line 233: I think the solution in C# is much cleaner, we can do similar things i n Java as well. We don't have to declare each pattern as a separate variable, and we don't have to have a `apiShortNameHelper` as well. Something like
```
String apiShortName = split using Splitter with . and get the first one
apiShortName = split using Splitter with - and get the last one
```
Usually it's a good practice to check the length before we get it to prevent runtime exception, but we don't have to check it here since a well formatted defaultHost should not error out.

src/main/java/com/google/api/generator/gapic/composer/samplecode/SettingsSampleComposer.java
```diff
@@ -145,6 +145,7 @@ public static Optional<Sample> composeSettingsSample(
             .map(e -> ExprStatement.withExpr(e))
             .collect(Collectors.toList());
 
+    // TODO: alicejli edit RegionTag to match other languages
```
blakeli0, line 148: Do we still need this TODO?

src/main/java/com/google/api/generator/gapic/model/GapicClass.java
```diff
@@ -44,8 +46,28 @@ public static GapicClass create(
     return builder().setKind(kind).setClassDefinition(classDefinition).setSamples(samples).build();
   }
 
+  public static GapicClass create(Kind kind, ClassDefinition classDefinition, String defaultHost) {
```
blakeli0, line 49: It's better to have less overloaded factory methods since it loses the benefit of having builders. See [here](go/java-practices/builders#when) for details. I would remove these two methods and use builders to add `defaultHost`. We should also examine all the places that created `GapicClass` and see if we really need to add `defaultHost` every time we create `GapicClass`.

src/main/java/com/google/api/generator/gapic/composer/Composer.java
```diff
@@ -191,17 +193,14 @@ public static List<GapicClass> generateTestClasses(GapicContext context) {
...
     //  e.g. v1, v2, v1beta1
     if (pakkage[pakkage.length - 1].matches("v[0-9].*")) {
       apiVersion = pakkage[pakkage.length - 1];
-      apiShortName = pakkage[pakkage.length - 2];
     } else {
       apiVersion = "";
-      apiShortName = pakkage[pakkage.length - 1];
     }
     //  Include license header, apiShortName, and apiVersion
     return clazzes.stream()
```
blakeli0, line 206: Nice to have: I know this is pre-existing code, it would be nice to refactor this part to  be more readable/understandable

- `map` is usually used to map an object to another, here we are not really mapping, we are adding something to the existing object, so we can use `foreach` or just traditional for loop. 
- Nested mappings and lambda expressions make it really hard to read, there are six levels of indentation here, it's usually a code smell that you have more than two levels of indentation.

src/main/java/com/google/api/generator/gapic/composer/samplecode/SettingsSampleComposer.java
alicejli: Yes - this comment in particular is in reference to https://buganizer.corp.google.com/issues/239821325 which I decided to keep as a separate PR from this b/c just parsing the host got more complex than I initially realized.

src/main/java/com/google/api/generator/gapic/composer/Composer.java
alicejli: Thanks for the tip! Makes sense; updated.

src/main/java/com/google/api/generator/gapic/composer/Composer.java
alicejli: Thank you - the Guava splitter makes it much easier to mimic C#. Updated!

src/main/java/com/google/api/generator/gapic/model/GapicClass.java
alicejli: Thank you; that makes sense. I've gone through and updated to use the Builder as well as removed it where we don't need to add the `defaultHost`.

src/test/java/com/google/api/generator/gapic/composer/ComposerTest.java
```diff
@@ -65,58 +65,90 @@ public void gapicClass_addApacheLicense() {
...
+    testClass = clazzes.get(0).withSamples(ListofSamples).withDefaultHost(defaultHost);
+    testClassList = Arrays.asList(new GapicClass[] {testClass});
+    composedSamples = Composer.prepareExecutableSamples(testClassList, protoPack).get(0).samples();
+    // If samples is empty, the test automatically passes without checking.
+    assertFalse(composedSamples.isEmpty());
+
     for (Sample sample : composedSamples) {
-      assertEquals("ApiShortName should be vision", sample.regionTag().apiShortName(), "vision");
-      assertEquals("ApiVersion should be v1p1beta1", sample.regionTag().apiVersion(), "v1p1beta1");
+      assertEquals("ApiShortName should be Vision", sample.regionTag().apiShortName(), "Vision");
```
amanda-tarafa, line 141: Hmmm, we don't capitalize the first letter of the API short name. Is this just for output or is it literally being used like this on the region tags?

src/test/java/com/google/api/generator/gapic/composer/ComposerTest.java
```diff
@@ -14,22 +14,17 @@
 
 package com.google.api.generator.gapic.composer;
 
-import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.*;
```
blakeli0, line 17: Can you please update Intellij to prevent it from importing *?

src/test/java/com/google/api/generator/gapic/composer/ComposerTest.java
```diff
@@ -65,58 +65,90 @@ public void gapicClass_addApacheLicense() {
...
+          "File header should be APACHE",
+          Arrays.asList(CommentComposer.APACHE_LICENSE_COMMENT),
+          sample.fileHeader());
+      assertEquals("ApiShortName should be empty", "", sample.regionTag().apiShortName());
+      assertEquals("ApiVersion should be V1beta1", "V1Beta1", sample.regionTag().apiVersion());
     }
   }
 
+  @Test
+  public void parseDefaultHostTest() {
```
blakeli0, line 86: We usually follow this [pattern](go/java-style#s5.2.3-method-names) for unit test names

src/main/java/com/google/api/generator/gapic/composer/Composer.java
```diff
@@ -210,17 +210,37 @@ static List<GapicClass> prepareExecutableSamples(List<GapicClass> clazzes, Strin
...
+  @VisibleForTesting
+  protected static String parseDefaultHost(String defaultHost) {
+    // If the defaultHost is of the format "**.googleapis.com", take the name before the first
+    // period.
+    String apiShortName = Iterables.getFirst(Splitter.on(".").split(defaultHost), defaultHost);
+    // If the defaultHost is of the format "**-**-**.googleapis.com", take the section before the
+    // first period and after the last dash to follow CSharp's implementation here:
+    // https://github.com/googleapis/gapic-generator-csharp/blob/main/Google.Api.Generator/Generation/ServiceDetails.cs#L70
+    apiShortName = Iterables.getLast(Splitter.on("-").split(apiShortName), defaultHost);
+    // TODO: alicejli will need to handle special case for `iam-meta-api` service
```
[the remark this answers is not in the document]
alicejli: Thanks! Made an exception for the iam service.

src/main/java/com/google/api/generator/gapic/composer/Composer.java
alicejli: Thanks for the advice - makes sense. I've refactored to use `forEach`.

test/integration/goldens/asset/samples/snippets/generated/main/java/com/google/cloud/asset/v1/assetserviceclient/analyzeiampolicy/AsyncAnalyzeIamPolicy.java
```diff
@@ -16,7 +16,7 @@
 
 package com.google.cloud.asset.v1.samples;
 
-// [START asset_v1_generated_assetserviceclient_analyzeiampolicy_async]
+// [START cloudasset_v1_generated_assetserviceclient_analyzeiampolicy_async]
```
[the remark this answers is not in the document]
alicejli: Thanks! That makes sense to me; I dug into this a bit- the region tag builder is sanitizing the attributes to use `JavaStyle.toUpperCamelCase()` which is then getting converted to all lowercase in the actual sample, so changing the current setup is a little more involved than I thought. 

I will make a TODO to refactor this in a different PR.

src/test/java/com/google/api/generator/gapic/composer/ComposerTest.java
alicejli: Thanks - I've updated my IntelliJ settings so this shouldn't happen anymore.

src/test/java/com/google/api/generator/gapic/composer/ComposerTest.java
alicejli: Thanks; updated.

test/integration/goldens/kms/samples/snippets/generated/main/java/com/google/cloud/kms/v1/keymanagementserviceclient/updatecryptokeyversion/SyncUpdateCryptoKeyVersion.java
```diff
@@ -16,7 +16,7 @@
 
 package com.google.cloud.kms.v1.samples;
 
-// [START kms_v1_generated_keymanagementserviceclient_updatecryptokeyversion_sync]
+// [START cloudkms_v1_generated_keymanagementserviceclient_updatecryptokeyversion_sync]
```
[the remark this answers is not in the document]
alicejli: Thanks Amanda! Yep I'm planning on fixing that here (still in draft): https://github.com/googleapis/gapic-generator-java/pull/1047 - can you confirm if the capitalization of the service / RPC name is important? Or is it okay to leave it as all lowercase?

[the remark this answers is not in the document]
alicejli: Confirmed that capitalization is important so will plan to fix that in the other PR as well.

src/test/java/com/google/api/generator/gapic/composer/ComposerTest.java
alicejli: Answered below :)

alicejli:
https://github.com/googleapis/gapic-generator-java/issues/1051

test/integration/goldens/kms/samples/snippets/generated/main/java/com/google/cloud/kms/v1/keymanagementservicesettings/getkeyring/SyncGetKeyRing.java
```diff
@@ -16,7 +16,7 @@
 
 package com.google.cloud.kms.v1.samples;
 
-// [START kms_v1_generated_keymanagementservicesettings_getkeyring_sync]
+// [START cloudkms_v1_generated_keymanagementservicesettings_getkeyring_sync]
```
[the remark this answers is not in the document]
alicejli: Ah thanks for clarifying - the generator was already set up to generate these so I'll plan to keep for now, but good to know that these are optional.

amanda-tarafa (approved):
LGTM from the perspective of SnippetGen output and counting on #1047 to come after.
Thanks!

src/test/java/com/google/api/generator/gapic/composer/ComposerTest.java
```diff
@@ -65,58 +71,90 @@ public void gapicClass_addApacheLicense() {
...
+          "File header should be APACHE",
+          Arrays.asList(CommentComposer.APACHE_LICENSE_COMMENT),
+          sample.fileHeader());
+      assertEquals("ApiShortName should be empty", "", sample.regionTag().apiShortName());
+      assertEquals("ApiVersion should be V1beta1", "V1Beta1", sample.regionTag().apiVersion());
     }
   }
 
+  @Test
+  public void parseDefaultHost_returnsApiShortName() {
```
blakeli0, line 92: It's usually better to create three separate test cases for this scenario, in case one of them failed, it's easier for us to identify the issue. The test name can be something like `parseDefaultHost_shouldReturnApiShortNameIfHostContainsDash`, also can you please add a test case for the hardcode `iam`?

src/test/java/com/google/api/generator/gapic/composer/ComposerTest.java
```diff
@@ -65,58 +71,90 @@ public void gapicClass_addApacheLicense() {
...
+    String apiShortName = Composer.parseDefaultHost(defaultHost);
+    org.junit.Assert.assertEquals("pubsub", apiShortName);
+
+    defaultHost = "logging.googleapis.com";
+    apiShortName = Composer.parseDefaultHost(defaultHost);
+    org.junit.Assert.assertEquals("logging", apiShortName);
+
+    defaultHost = "localhost:7469";
+    apiShortName = Composer.parseDefaultHost(defaultHost);
+    org.junit.Assert.assertEquals("localhost:7469", apiShortName);
```
blakeli0, line 103: Can we just use the static imported `assertEquals`?

src/main/java/com/google/api/generator/gapic/composer/grpc/MockServiceClassComposer.java
```diff
@@ -69,7 +69,7 @@ public GapicClass generate(GapicContext ignored, Service service) {
             .setStatements(createFieldDeclarations(service, typeStore))
             .setMethods(createClassMethods(service, typeStore))
             .build();
-    return GapicClass.create(kind, classDef);
+    return GapicClass.create(kind, classDef).withDefaultHost(service.defaultHost());
```
blakeli0, line 72: I feel like we might be adding `defaulHost` to too many composers, I'm pretty sure we don't need it here and `MockServiceImplClassComposer`, probably not in `ResourceNameHelperClassComposer.java` as well, can you try to run all the tests without adding defaultHost to these composers? It will make the tests much easier

src/test/java/com/google/api/generator/gapic/composer/grpc/GrpcServiceCallableFactoryClassComposerTest.java
```diff
@@ -30,6 +30,7 @@ public void generateServiceClasses() {
 
     Assert.assertGoldenClass(this.getClass(), clazz, "GrpcEchoCallableFactory.golden");
     Assert.assertEmptySamples(clazz.samples());
+    Assert.assertCodeEquals(clazz.defaultHost(), "");
```
blakeli0, line 33: I don't think we need the assertions in this tests, because we didn't touch `AbstractServiceCallableFactoryClassComposer` or `GrpcServiceCallableFactoryClassComposer`.

src/test/java/com/google/api/generator/gapic/composer/grpc/GrpcServiceStubClassComposerTest.java
```diff
@@ -29,6 +29,7 @@ public void generateGrpcServiceStubClass_simple() {
 
     Assert.assertGoldenClass(this.getClass(), clazz, "GrpcEchoStub.golden");
     Assert.assertEmptySamples(clazz.samples());
+    Assert.assertCodeEquals(clazz.defaultHost(), "");
```
blakeli0, line 32: I don't think we need the assertions in this tests, because we didn't touch `GrpcServiceStubClassComposer` or `AbstractTransportServiceStubClassComposer`.

src/test/java/com/google/api/generator/gapic/composer/grpc/MockServiceImplClassComposerTest.java
```diff
@@ -30,8 +30,12 @@ public class MockServiceImplClassComposerTest {
   public static Collection<Object[]> data() {
     return Arrays.asList(
         new Object[][] {
-          {"MockEchoImpl", GrpcTestProtoLoader.instance().parseShowcaseEcho()},
-          {"MockDeprecatedServiceImpl", GrpcTestProtoLoader.instance().parseDeprecatedService()}
+          {"MockEchoImpl", GrpcTestProtoLoader.instance().parseShowcaseEcho(), "localhost:7469"},
```
blakeli0, line 33: We don't need these tests if we can remove the `withDefaultHost` part from `MockServiceImplClassComposer`.

src/test/java/com/google/api/generator/gapic/composer/grpc/ServiceClientTestClassComposerTest.java
```diff
@@ -30,15 +30,16 @@ public class ServiceClientTestClassComposerTest {
   public static Collection<Object[]> data() {
     return Arrays.asList(
         new Object[][] {
-          {"EchoClientTest", GrpcTestProtoLoader.instance().parseShowcaseEcho(), 0},
+          {"EchoClientTest", GrpcTestProtoLoader.instance().parseShowcaseEcho(), 0, ""},
```
blakeli0, line 33: We don't need to touch this test since we didn't touch `ServiceClientTestClassComposer` or `AbstractServiceClientTestClassComposer`

src/test/java/com/google/api/generator/gapic/composer/resourcename/ResourceNameHelperClassComposerTest.java
```diff
@@ -110,6 +110,7 @@ public void generateResourceNameClass_echoFoobarMultiplePatterns() {
     Utils.saveCodegenToFile(this.getClass(), "FoobarName.golden", visitor.write());
     Path goldenFilePath = Paths.get(Utils.getGoldenDir(this.getClass()), "FoobarName.golden");
     Assert.assertCodeEquals(goldenFilePath, visitor.write());
+    Assert.assertCodeEquals(clazz.defaultHost(), "localhost:7469");
```
blakeli0, line 113: We can remove these new assertions if we found `withDefaultHost` is not needed for `ResourceNameHelperClassComposerTest`

src/test/java/com/google/api/generator/gapic/composer/ComposerTest.java
alicejli: Makes sense - updated!

src/test/java/com/google/api/generator/gapic/composer/ComposerTest.java
alicejli: Yes good catch; updated

src/test/java/com/google/api/generator/gapic/composer/grpc/GrpcServiceCallableFactoryClassComposerTest.java
alicejli: Thanks that makes sense - I forgot to update after I refactored. Updated!

src/main/java/com/google/api/generator/gapic/composer/grpc/MockServiceClassComposer.java
alicejli: That makes sense. Confirmed that removing `withDefaultHost` from the builders of `MockServiceClassComposer`, `MockServiceImplClassComposer`, and `ResourceNameHelperClassComposer` is fine as those composers are not used in generating samples. Have removed from builders as well as the unit tests.

src/test/java/com/google/api/generator/gapic/composer/grpc/MockServiceImplClassComposerTest.java
alicejli: Confirmed; removed

src/test/java/com/google/api/generator/gapic/composer/resourcename/ResourceNameHelperClassComposerTest.java
alicejli: Confirmed, removed

src/test/java/com/google/api/generator/gapic/composer/grpc/ServiceClientTestClassComposerTest.java
alicejli: Removed

src/main/java/com/google/api/generator/gapic/model/GapicClass.java
```diff
@@ -62,6 +72,8 @@ abstract static class Builder {
 
     abstract Builder setSamples(List<Sample> samples);
 
+    abstract Builder setDefaultHost(String defaultHost);
```
blakeli0, line 75: Looking at the PR again, I noticed that it's not the default host we are setting here, we are actually the apiShortName that are parsed from default host, do you mind renaming this field to `apiShortName`?

src/main/java/com/google/api/generator/gapic/model/HttpBindings.java
```diff
@@ -36,21 +36,54 @@ public enum HttpVerb {
...
 
     abstract String lowerCamelName();
 
-    public abstract boolean isOptional();
+    // An object that contains all info of the leaf level field
+    @Nullable
+    public abstract Field field();
 
-    public abstract boolean isRepeated();
+    public boolean isOptional() {
```
blakeli0, line 49: These code looks like from my another PR, you may need to merge from main again.

src/main/java/com/google/api/generator/gapic/model/GapicClass.java
blakeli0: You are right, I was confused with what's parsed and what's being set in GapicClass. SpringCodeGen may want to reuse the parsing logic so I was thinking we might want to move the parsing logic earlier in the process, but that can be done in a different PR as well.

src/main/java/com/google/api/generator/gapic/model/HttpBindings.java
alicejli: Good catch - updated!

blakeli0 (approved):

PR #1042, opened 2022-09-22
fix: Get numeric value for Enum fields if it is configured as query param or path param

blakeli0:
Get numeric value for Enum fields if it is configured as query param or path param. Refactor HttpBinding to include Field info and use builder.
Found an existing issue regarding how query params are populated. Add comments to explain the current behavior and expected behavior for populating query params. Created issue https://github.com/googleapis/gapic-generator-java/issues/1041 to track the future work.

### DIFF
--- a/src/test/java/com/google/api/generator/gapic/model/HttpBindingsTest.java
+++ b/src/test/java/com/google/api/generator/gapic/model/HttpBindingsTest.java
@@ -38,7 +38,7 @@ public class HttpBindingsTest {
   }
 
   @Test
-  public void isOptional_shouldReturnFalseIfFieldExistAndIsOptionalIsFalse() {
+  public void isOptional_shouldReturnFalseIfFieldExistsAndIsOptionalIsFalse() {
     HttpBinding httpBinding =
         httpBindingBuilder.setField(fieldBuilder.setIsProto3Optional(false).build()).build();
 
@@ -46,7 +46,7 @@ public class HttpBindingsTest {
   }
 
   @Test
-  public void isOptional_shouldReturnTrueIfFieldExistAndIsOptionalIsTue() {
+  public void isOptional_shouldReturnTrueIfFieldExistsAndIsOptionalIsTue() {
     HttpBinding httpBinding =
         httpBindingBuilder.setField(fieldBuilder.setIsProto3Optional(true).build()).build();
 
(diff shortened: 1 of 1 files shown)


diego92sigma6 (approved):

src/test/java/com/google/api/generator/gapic/model/HttpBindingsTest.java
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  @Test
+  public void isOptional_shouldReturnFalseIfFieldIsNull() {
+    HttpBinding httpBinding = httpBindingBuilder.build();
+    Truth.assertThat(httpBinding.isOptional()).isFalse();
+  }
+
+  @Test
+  public void isOptional_shouldReturnFalseIfFieldExistAndIsOptionalIsFalse() {
```
alicejli, line 41: nit: I think it should read "IfFieldExists" to keep present tense (and elsewhere in tests)

src/test/proto/compliance.proto
```diff
@@ -80,6 +80,44 @@ service Compliance {
       get: "/v1beta1/repeat/{info.f_string=first/*}/{info.f_child.f_string=second/**}:pathtrailingresource"
     };
   }
+
+  // This method requests an enum value from the server. Depending on the contents of EnumRequest, the enum value returned will be a known enum declared in the
```
alicejli, line 84: general question - is there a reason this proto doesn't match https://github.com/googleapis/gapic-showcase/blob/main/schema/google/showcase/v1beta1/compliance.proto exactly?

alicejli (commented):
Some small nits and a general question on the `compliance.proto`. Otherwise it looks good; thank you for the thorough comments and your hard work on this feature!

src/main/java/com/google/api/generator/gapic/model/HttpBindings.java
```diff
@@ -36,21 +36,54 @@ public enum HttpVerb {
 
   @AutoValue
   public abstract static class HttpBinding implements Comparable<HttpBinding> {
+
+    // The full qualified name of the field. e.g. request.complex_object.another_object.name
```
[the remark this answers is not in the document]
blakeli0: Updated

src/test/java/com/google/api/generator/gapic/model/HttpBindingsTest.java
blakeli0: Updated

src/test/proto/compliance.proto
blakeli0: This was way before I joined, but I guess what happened was that the showcase protos have a lot of edge cases that we want to use, so instead of creating protos from scratch for our test cases, we just copied over these showcase protos and use them as they are. However, protos in showcase evolve over time and we didn't update them regularly.

PR #1044, opened 2022-09-26
fix(Spring CodeGen): switch to Spring boot 2.7 style in registering auto-configurations

zhumin8:
Spring Boot 2.7 has introduced [changes to how auto-configurations](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.7-Release-Notes#changes-to-auto-configuration) are registered. Both changes are backward compatible, but the older `spring.factories` style [will be removed in Spring Boot 3.0](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0.0-M5-Release-Notes#auto-configuration-registration). 
Also consider that Spring Boot 2.6.x goes out of support by 11-24-2022, switching this code to 2.7's new style.


src/main/java/com/google/api/generator/spring/SpringWriter.java
```diff
@@ -139,13 +139,14 @@ private static void writeMetadataFile(GapicContext context, String path, JarOutp
   }
 
   private static void writeSpringFactories(GapicContext context, JarOutputStream jos) {
```
emmileaf, line 141: Would this mean we need to rebrand this method and the exception message as well? (To something like `writeAutoConfigRegistration`)

emmileaf (approved):

src/main/java/com/google/api/generator/spring/SpringWriter.java
zhumin8: ah-ha! Good catch.

PR #1045, opened 2022-09-27
fix(Spring CodeGen): Update missing and placeholder annotations

emmileaf:
**In this PR:**
- Following #1012, 
  * updates annotation parameter placeholder strings in `SpringAutoConfigClassComposer` for parameters of type class
  * updates annotation parameter placeholder strings in `SpringAutoConfigClassComposer` for multiple parameters
  * adds `@Qualifier` annotations on variables
  * adds `@NestedConfigurationProperty` annotation on credentials field in `SpringPropertiesClassComposer`

@zhumin8 ~~Do you have any preferences on whether I use this this branch/PR to also work on the other annotation TODOs in the code, or have the changes be separate for now?~~ (Edit: will use this PR since they are pretty closely related) I also wasn't too sure what naming conventions to use for the variables, so would appreciate any feedback here as well.

**Potential conflicts:**
    * With #1046 (Switching Spring dependencies from dynamic to static types)
    * With #1044 (`@Configuration(proxyBeanMethods = false)` is no longer used in Spring Boot 2.7)


src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java
```diff
@@ -401,37 +403,84 @@ private static List<AnnotationNode> createClassAnnotations(
...
+    TypeNode propertiesClazzType =
+        TypeNode.withReference(
+            VaporReference.builder()
+                .setName(types.get(service.name() + "Properties").reference().name())
+                .setPakkage(service.pakkage())
+                .build());
+    VariableExpr propertiesClassVariableExpr =
+        VariableExpr.builder()
+            .setVariable(Variable.builder().setType(TypeNode.CLASS_OBJECT).setName("class").build())
+            .setStaticReferenceType(propertiesClazzType)
```
zhumin8, line 476: I have types are defined in 2 ways at the bottom of class:
- [dynamic types](https://github.com/googleapis/gapic-generator-java/blob/62cd6384e2afa58ee7b2f8d6c452f15e0ac05223/src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java#L984) - these are types that needs to use vapor references.
- [static types](https://github.com/googleapis/gapic-generator-java/blob/62cd6384e2afa58ee7b2f8d6c452f15e0ac05223/src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java#L966) - these dependencies are brought in to gapic-generator so the concrete types can be used.  One caveat is that I am also working on #1046 to bring in Spring dependencies so those can be static instead of dynamic.

In this case, the properties class is already defined in [here](https://github.com/googleapis/gapic-generator-java/blob/62cd6384e2afa58ee7b2f8d6c452f15e0ac05223/src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java#L1002-L1007), so you can use `types.get(service.name() + "Properties")` here and get rid of `propertiesClazzType`.

Sorry about the confusion that `types.get(service.name() + "Properties").reference().name()` brings, I was trying to remind myself this type is avalaible.

src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java
```diff
@@ -401,37 +403,84 @@ private static List<AnnotationNode> createClassAnnotations(
...
             .setType(types.get("ConditionalOnProperty"))
-            .setDescription(
-                "value = \""
-                    + Utils.springPropertyPrefix(libName, service.name())
-                    + ".enabled\", matchIfMissing = false")
+            .addDescription(valueStringAssignmentExpr)
+            .addDescription(matchIfMissingAssignmentExpr)
+            .build();
+
+    TypeNode clazzType =
```
zhumin8, line 432: I believe this TypeNode is also defined in [dynamic types](https://github.com/googleapis/gapic-generator-java/blob/62cd6384e2afa58ee7b2f8d6c452f15e0ac05223/src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java#L87). see details in comment below.

zhumin8 (changes requested):
LGTM! I just have two comments about reusing defined types. 
- About variable naming conventions, I think your naming looks good. I also try to spell-out the actual content as well as the ast type e.g. `AssignmentExpr`, and realize sometimes it can get quite long. In my draft code, you might still see some naming that's not comprehensive enough, feel free to change them. I will also try to do another cleanup later on.
- On pr naming, @blakeli0  I've tried to follow "fix():...", "feat():..." but wonder if we have any guidelines on naming? Also, I'd like to add a "spring" label to these PRs if that does not interfere with any of the automated tasks?

src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java
emmileaf: Thanks for the explanation here - I hadn't noticed the typeMaps until recently, will go ahead and make this update! 

Will keep an eye on #1046 and maybe rebase this branch if you wanted to merge that first though, since these two PRs will probably have conflicting changes.

zhumin8 (approved):
Looks good!

src/test/java/com/google/api/generator/spring/SpringAutoConfigClassComposerTest.java
```diff
@@ -189,8 +190,9 @@ public void generatePropertiesTest() {
           + "  @Bean\n"
           + "  @ConditionalOnMissingBean\n"
           + "  public EchoClient echoClient(\n"
-          + "      CredentialsProvider credentialsProvider,\n"
-          + "      TransportChannelProvider defaultTransportChannelProvider)\n"
+          + "      @Qualifier(\"googleCredentials\") CredentialsProvider credentialsProvider,\n"
```
zhumin8, line 193: Totally missed this: We want to add this `@Qualifier` so that when multiple services are present, there won't be conflict in which `CredentialsProvider` bean it takes. So need to rename `googleCredentials()` to `[serviceName]Credentials()` and match the name here. (same as `defaultEchoTransportChannelProvider()`)

emmileaf: Thanks for catching this - will address in a separate PR!

PR #1047, opened 2022-09-28
fix: update regionTag to use service name

alicejli:
Region tags for Java generated samples do not match other languages for Cloud Workflows.

Java current uses tags like `workflows_v1_generated_workflowsclient_listworkflows_sync` which includes `workflowsclient.` This doesn't match other languages that use tags like `workflows_v1_generated_Workflows_ListWorkflows_sync`. This prevents grouping of samples on the devsite.

This updates:
1) The RegionTag to use the service name instead of the client name. 
2) The capitalization of the service and RPC name in the Region Tag


alicejli:
https://github.com/googleapis/gapic-generator-java/issues/1051

src/main/java/com/google/api/generator/gapic/composer/samplecode/ServiceClientHeaderSampleComposer.java
```diff
@@ -81,7 +87,8 @@ public static Sample composeShowcaseMethodSample(
       TypeNode clientType,
       List<MethodArgument> arguments,
       Map<String, ResourceName> resourceNames,
-      Map<String, Message> messageTypes) {
+      Map<String, Message> messageTypes,
+      Service service) {
```
blakeli0, line 91: Do we still need to pass in Service if we are parsing Strings to get the service name?

src/main/java/com/google/api/generator/gapic/composer/utils/ClassNames.java
```diff
@@ -120,4 +122,9 @@ private static String monolithBackwardsCompatibleName(String rawServiceName) {
         ? rawServiceName.replace("IAM", "Iam")
         : rawServiceName;
   }
+
+  // For region tags, the service name should not include `Client` in the name
+  public static String getPureServiceName(String serviceClassName) {
+    return Iterables.getFirst(Splitter.on("Client").split(serviceClassName), serviceClassName);
```
blakeli0, line 128: If you decide to get service name by splitting the String, I would suggest extract the `Client` here and [here](https://github.com/googleapis/gapic-generator-java/blob/main/src/main/java/com/google/api/generator/gapic/composer/utils/ClassNames.java#L27) to a constant, in case it changes in the future.

alicejli:
Got stuck in rebasing hell - will generate a new PR for this change and link back here.

src/main/java/com/google/api/generator/gapic/composer/samplecode/ServiceClientHeaderSampleComposer.java
alicejli: Thanks for catching - when I switched approaches to instead parse the service name, I didn't do a thorough job cleaning up. Updated!

src/main/java/com/google/api/generator/gapic/composer/utils/ClassNames.java
alicejli: Thanks - updated.

src/main/java/com/google/api/generator/gapic/model/RegionTag.java
```diff
@@ -138,7 +138,8 @@ public String generate() {
     if (!apiVersion().isEmpty()) {
       rt = rt + apiVersion() + "_";
     }
-    rt = rt + "generated_" + serviceName() + "_" + rpcName();
+    rt = rt.toLowerCase() + "generated_" + serviceName() + "_" + rpcName();
+    // TODO: @alicejli overloadDisambiguation will need to be updated to match other languages
```
blakeli0, line 142: Is this a must have for Phase 1?

src/main/java/com/google/api/generator/gapic/composer/utils/ClassNames.java
```diff
@@ -120,4 +123,11 @@ private static String monolithBackwardsCompatibleName(String rawServiceName) {
         ? rawServiceName.replace("IAM", "Iam")
         : rawServiceName;
   }
+
+  // For region tags, the service name should not include `Client` in the name
+  public static String getPureServiceName(String serviceClassName) {
```
blakeli0, line 128: What I meant was that we can extract the word `Client` to a constant(e.g. `CLIENT_SUFFIX = "Client"`), and make `SERVICE_CLIENT_CLASS_NAME_PATTERN = "%s" + CLIENT_SUFFIX`. So that in case the suffix changes, we don't have to update multiple places.

src/main/java/com/google/api/generator/gapic/model/RegionTag.java
alicejli: No - to be fair I think the overloadDisambiguation hasn't been formalized across languages in the design; I just happened to notice Java's doesn't look the same as C# (which I'm taking as canonical). I'm going to keep as a TODO to check on later, but right now I think it's okay that Java is different.

src/main/java/com/google/api/generator/gapic/composer/utils/ClassNames.java
alicejli: Ah sorry that makes sense. Let me know if the update matches how you were thinking of it -

blakeli0 (approved):

PR #1052, opened 2022-10-03
test(Spring CodeGen): Set up golden file tests for spring class composers

emmileaf:
**In this PR:**
- Separates `SpringPropertiesClassComposerTest` from `SpringAutoConfigClassComposerTest`
- Replaces string comparisons with golden file testing
 
**Potential conflicts:**
- #1045 should be merged first; resolve conflicts and update golden files in the PR once that happens
- Update golden files to include logger statements as well if #1053 is merged first


emmileaf:
@zhumin8 Two open questions for this PR:

1.  Right now the golden files for `SpringComposerTest` vs. the individual class composer tests only differ in the apache license. Would we want to keep these as separate files (vs. combine and reuse), given the additional testing for `SpringComposer` steps that we anticipate adding in the future?

2. I’ve split out `SpringPropertiesClassComposerTest` from `SpringAutoConfigClassComposerTest`, but noticed that the same chunk of [setup](https://github.com/googleapis/gapic-generator-java/blob/3cd3419c1a07519321f1ed301eb0e157f6b88f08/src/test/java/com/google/api/generator/spring/SpringAutoConfigClassComposerTest.java#L53) code is shared and repeated. This setup seems to be replaceable by calling `TestProtoLoader`:
```
    this.context = TestProtoLoader.instance().parseShowcaseEcho();
    this.echoProtoService = this.context.services().get(0);

``` 
Was there a consideration for keeping the explicit setup steps here (perhaps for things that’d need customizing or flexibility in the future?) If not, I'll make the switch over in this PR.

BUILD.bazel
```diff
@@ -208,6 +208,9 @@ GOLDEN_UPDATING_UNIT_TESTS = [
     "com.google.api.generator.gapic.composer.rest.ServiceClientTestClassComposerTest",
     "com.google.api.generator.gapic.composer.rest.ServiceSettingsClassComposerTest",
     "com.google.api.generator.gapic.composer.rest.ServiceStubSettingsClassComposerTest",
+    "com.google.api.generator.spring.SpringAutoConfigClassComposerTest",
```
zhumin8, line 211: note: we will need to carry this update logic over if we separate from gapic-generator-java in future.

zhumin8 (commented):
Great work! this will make our work a lot easier when making changes to the Composer classes. 
I only got one comment: 
The setups for the three existing tests are exactly the same. Perhaps it makes sense to have an Abstract test class with the setups and each extend from it. This way if we ever want to test on more edge-cases in future, we are able to add the scenario to all test in sync.

emmileaf:
@zhumin8 Thank you for the review! A quick summary of follow-up changes to the PR:
- Replaced the tests’ setup code (as discussed above)  
- Moved tests and goldens into a `composer` subdirectory to align with package structure of the source code
- Tried changing the `SpringComposer` tests to reuse golden files from the individual class composer tests, but the extra setup felt more complicated than expected. Upon second thought, having a test use but not update the golden files (or having two different tests update the same ones) could get confusing, so I reverted back to separate files and just gave them better names. We can revisit this if needed later.

zhumin8 (approved):
Changes looks good.  Agreed on your reasoning on keeping separate golden files.

BUILD.bazel
```diff
@@ -208,6 +208,9 @@ GOLDEN_UPDATING_UNIT_TESTS = [
     "com.google.api.generator.gapic.composer.rest.ServiceClientTestClassComposerTest",
     "com.google.api.generator.gapic.composer.rest.ServiceSettingsClassComposerTest",
     "com.google.api.generator.gapic.composer.rest.ServiceStubSettingsClassComposerTest",
+    "com.google.api.generator.spring.composer.SpringAutoConfigClassComposerTest",
```
blakeli0, line 211: Can you try to use Maven command like this `mvn test -DupdateUnitGoldens -Dtest=JavaCodeGeneratorTest` to update these tests?  [DEVELOPMENT.md](https://github.com/googleapis/gapic-generator-java/blob/main/DEVELOPMENT.md) has more details about it. Ideally we don't want to touch anything in Bazel.

emmileaf: Ah I hadn't realized this alternative - thank you! Just made the switch to not modify this file.

blakeli0: Yeah, the Maven build in gapic-generator-java is not completely done, so some of the functionality may not be completely working and we never publicized it. Did you confirm that you can use the maven command to update golden test files?

emmileaf: Yup, I checked locally that the golden files were able to be generated using the maven command instead.

blakeli0: Perfect, thanks!

blakeli0 (approved):

PR #1056, opened 2022-10-05
fix: extract and expose simple util methods from RetrySettingsComposer

zhumin8:
Extract and expose some simple util methods from `RetrySettingsComposer` for use in Spring code.

### DIFF
--- a/src/main/java/com/google/api/generator/gapic/composer/utils/RetrySettingsUtils.java
+++ b/src/main/java/com/google/api/generator/gapic/composer/utils/RetrySettingsUtils.java
@@ -57,6 +57,7 @@ public class RetrySettingsUtils {
         .setStaticReferenceType(FIXED_TYPESTORE.get("Duration"))
         .setMethodName("ofMillis")
         .setArguments(valExpr)
+        .setReturnType(FIXED_TYPESTORE.get("Duration"))
         .build();
   }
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/api/generator/gapic/composer/utils/RetrySettingsUtils.java
```diff
@@ -0,0 +1,68 @@
...
+import com.google.api.generator.engine.ast.TypeNode;
+import com.google.api.generator.engine.ast.ValueExpr;
+import com.google.api.generator.gapic.composer.store.TypeStore;
+import com.google.protobuf.Duration;
+import com.google.protobuf.util.Durations;
+import java.util.Arrays;
+import java.util.List;
+
+/** Common utility methods to expose */
+public class RetrySettingsUtils {
```
blakeli0, line 28: I know these methods are only used by retry related logics, but they can be used by other features as well, can we have a more generic name for this class? 
I'm bad at naming as well, something like `CommonUtils` seems too generic but something like `DurationUtils` seems inaccurate. Maybe break it further into two utils? One is `ValueExprUtils`, another one is  `MethodInvocationExprUtils`?

zhumin8: How about `ConvertToExprUtils` as updated below? I tried to keep it generic in case of future additions. Is it too generic? I also don't want split into two classes as the `createDurationOfMillisExpr()` is clearly close related to the `toValExpr` ones.

blakeli0:
Cancelling this PR as we are revamping the RetrySettings design in spring-cloud-gcp

PR #1059, opened 2022-10-12
feat(Spring CodeGen): Add customized spring composer for package info

emmileaf:
**In this PR:**
- Add `SpringPackageInfoComposer.java` to generate `package-info.java`, similar to [ClientLibraryPackageInfoComposer.java](https://github.com/googleapis/gapic-generator-java/blob/main/src/main/java/com/google/api/generator/gapic/composer/ClientLibraryPackageInfoComposer.java) but customized for spring.
- In `SpringWriter.java`, remove generation of `gapic_metadata.json` for spring libraries.

**Todos:**
- [x] Also update `SpringComposerTest` to include this file
- [x] Libname and (spring) package name logic should be updated to call `Utils` after #1061 is merged


src/main/java/com/google/api/generator/spring/SpringWriter.java
```diff
@@ -202,8 +187,4 @@ private static String getPath(String pakkage, String className) {
     }
     return path;
   }
-
```
zhumin8, line 205: Oops, I missed this one. Thanks.

src/main/java/com/google/api/generator/spring/composer/SpringPackageInfoComposer.java
```diff
@@ -0,0 +1,67 @@
...
+import com.google.api.generator.engine.ast.TypeNode;
+import com.google.api.generator.gapic.model.GapicContext;
+import com.google.api.generator.gapic.model.GapicPackageInfo;
+import com.google.common.base.Preconditions;
+import com.google.common.base.Strings;
+import javax.annotation.Generated;
+
+public class SpringPackageInfoComposer {
+  private static final String PACKAGE_INFO_TITLE_PATTERN =
+      "Auto-configuration for %s integration with Spring.";
```
zhumin8, line 31:
```suggestion
      "Spring Boot Auto-configurations for %s.";
```

zhumin8 (commented):
A few comments below, and just a thought:
You probably started out from `ClientLibraryPackageInfoComposer.java` for `SpringPackageInfoComposer.java` and cleaned out irrelevant contents. 
What if we keep the descriptions for each service like in [here](https://github.com/googleapis/java-vision/blob/main/google-cloud-vision/src/main/java/com/google/cloud/vision/v1/package-info.java#L22-L26), remove the samples, and at the bottom, add a sample to use along the lines of:
"You can inject auto-configured service bean to your Spring Boot application with annotation based configuration `@Autowired [serviceBean]` or Constructor Injection "

src/main/java/com/google/api/generator/spring/composer/SpringPackageInfoComposer.java
emmileaf: Updated

emmileaf:
> A few comments below, and just a thought (this could totally be a separate pr if we want it): You probably started out from `ClientLibraryPackageInfoComposer.java` for `SpringPackageInfoComposer.java` and cleaned out irrelevant contents. What if we keep the descriptions for each service like in [here](https://github.com/googleapis/java-vision/blob/main/google-cloud-vision/src/main/java/com/google/cloud/vision/v1/package-info.java#L22-L26), remove the samples, and at the bottom, add a sample to use along the lines of: "You can inject auto-configured service bean to your Spring Boot application with annotation based configuration `@Autowired [serviceBean]` or Constructor Injection "

@zhumin8 Ah I think that would be helpful! One concern here might be that `ClientLibraryPackageInfoComposer` has [logic for formatting service descriptions](https://github.com/googleapis/gapic-generator-java/blob/01c14d41d4195963c0ed46f310360f39eec6d420/src/main/java/com/google/api/generator/gapic/composer/ClientLibraryPackageInfoComposer.java#L84-L113) that is not currently exposed, so copying it could run us into inconsistency if that part of gapic generator code gets updated in the future (though those lines haven’t been modified for a while).

A couple of possible approaches here come to mind, in order of estimated effort. Let me know if you have any preferences or suggestions?
1. We could skip the service description and just list each [service client name](https://github.com/googleapis/java-vision/blob/b62fb05a8b55e76de6752eeb798bf9ce901c8f8f/google-cloud-vision/src/main/java/com/google/cloud/vision/v1/package-info.java#L42), followed by the Spring-specific snippet
2. We could copy over [service description logic](https://github.com/googleapis/java-vision/blob/b62fb05a8b55e76de6752eeb798bf9ce901c8f8f/google-cloud-vision/src/main/java/com/google/cloud/vision/v1/package-info.java#L44-L58) from `ClientPackageInfoComposer`, and keep `SpringPackageInfoComposer` updated if this ever changes (any discrepancy would also only impact generated javadoc and not code)
3. We could possibly refactor `ClientPackageInfoComposer` to expose service description logic for reuse? (This option would definitely warrant a separate PR)

zhumin8:
I am inclined to option 3. 
But either way, you are right, it could be done in a separate PR. Let's get this merged in and talk about it separately.

emmileaf:
> I am inclined to option 3. But either way, you are right, it could be done in a separate PR. Let's get this merged in and talk about it separately.

Sounds good!

zhumin8 (approved):
LGTM

PR #1061, opened 2022-10-13
fix(Spring CodeGen): Update and move naming-related logic into utils

emmileaf:
This PR updates and moves three pieces of naming-related logic in spring codegen into util methods:

- getSpringPropertyPrefix (e.g. `com.google.cloud.vision.v1.spring.auto.image-annotator`)
  * Replaced earlier use of libname due to ambiguity (see [comment](https://github.com/googleapis/gapic-generator-java/pull/1030#discussion_r983837987))
  
- getLibName (e.g. `vision`)
  * Intended for descriptive purposes
  * Includes three options considered by design doc - currently this PR implements option 2 (parse from package name) with the intention to switch to option 3 (parse from default host) in alignment with SampleGen 

- getSpringPackageName (e.g. `com.google.cloud.vision.v1.spring`)
  * Since this was also used in a couple of places, extracted to util in case of future need to update

**Dependencies:**
- #1059 should be adjusted to use these util methods (merge order should be after this PR)
- Would we prefer to have this PR wait on the switch to  `parseDefaultHost`, or make the change later in a separate PR?

### DIFF
--- a/src/main/java/com/google/api/generator/spring/SpringWriter.java
+++ b/src/main/java/com/google/api/generator/spring/SpringWriter.java
@@ -163,7 +163,7 @@ public class SpringWriter {
     String path = "src/main/resources/META-INF";
     JarEntry jarEntry =
         new JarEntry(String.format("%s/additional-spring-configuration-metadata.json", path));
-    String libName = Utils.getLibName(context);
+    String packageName = context.services().get(0).pakkage();
     try {
       jos.putNextEntry(jarEntry);
       StringJoiner sb = new StringJoiner(",\n", "\n{\n    \"properties\": [\n", "\n    ]\n" + "}");
@@ -179,8 +179,8 @@ public class SpringWriter {
                               + "            \"description\": \"Auto-configure Google Cloud %s components.\",\n"
                               + "            \"defaultValue\": true\n"
                               + "        }",
-                          Utils.springPropertyPrefix(libName, service.name()),
-                          libName + "/" + service.name())));
+                          Utils.getSpringPropertyPrefix(packageName, service.name()),
+                          Utils.getLibName(context))));
 
       jos.write(sb.toString().getBytes(StandardCharsets.UTF_8));
     } catch (IOException e) {
--- a/src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java
+++ b/src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java
@@ -79,12 +79,11 @@ public class SpringAutoConfigClassComposer implements ClassComposer {
 
   @Override
   public GapicClass generate(GapicContext context, Service service) {
-    String packageName = service.pakkage() + ".spring";
+    String packageName = Utils.getSpringPackageName(service.pakkage());
     Map<String, TypeNode> types = createDynamicTypes(service, packageName);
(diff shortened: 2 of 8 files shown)


emmileaf:
Two questions for perhaps an initial review related to the code changes here:

1. Would we prefer to have this PR wait on the switch to `parseDefaultHost` in `getLibName`, or make the change later in a separate PR?
2. I saw that the work in #1057 also needs things like the client library name (descriptive title vs. short name), group, and version. Would it make sense for these parsing logic (or at least the tentative version) to also be added under `Utils`?

In parsing version, for example, perhaps we could also align with SampleGen’s approach:https://github.com/googleapis/gapic-generator-java/blob/d222af10cd6345e0a3660353d7d667e917be67ec/src/main/java/com/google/api/generator/gapic/composer/Composer.java#L194-L205

zhumin8:
1. I think it makes sense to make the change later in a separate PR, consider `parseDefaultHost` exposure may need to wait a bit.
2. As for #1057, I would suggest you not to worry about it in this pr. I will align later in that pr separately when possible. The client library coordinates likely needs to be acquired outside of gapic-gen from sources like this [metadata](https://github.com/googleapis/java-vision/blob/main/.repo-metadata.json#L11), and the library name in Spring Starter to produced can align with `parseDefaultHost` in `getLibName`.

emmileaf:
@zhumin8 Sounds good! In that case I’ll keep both items as out of scope for this PR and mark it ready for review. 

(Side note for Sonar: the missing code coverage for these utils should improve after #1059 adds package info composer tests)

src/main/java/com/google/api/generator/spring/SpringWriter.java
```diff
@@ -179,8 +179,8 @@ private static void writeSpringAdditionalMetadataJson(GapicContext context, JarO
                               + "            \"description\": \"Auto-configure Google Cloud %s components.\",\n"
                               + "            \"defaultValue\": true\n"
                               + "        }",
-                          Utils.springPropertyPrefix(libName, service.name()),
-                          libName + "/" + service.name())));
+                          Utils.getSpringPropertyPrefix(packageName, service.name()),
+                          Utils.getLibName(context))));
```
zhumin8, line 183: I think we still want `service.name()` some where in the description, as this block is generated per service. So for client lib with multiple services like `vision`, description for enabling each service is different.

src/main/java/com/google/api/generator/spring/SpringWriter.java
```diff
@@ -163,7 +163,7 @@ private static void writeSpringAdditionalMetadataJson(GapicContext context, JarO
     String path = "src/main/resources/META-INF";
     JarEntry jarEntry =
         new JarEntry(String.format("%s/additional-spring-configuration-metadata.json", path));
-    String libName = Utils.getLibName(context);
+    String packageName = context.services().get(0).pakkage();
```
zhumin8, line 166: nit: Do you mind also extract this to `Utils`? It's only one line, but just in case of future changes.

zhumin8 (commented):
Just minor comments.

src/main/java/com/google/api/generator/spring/SpringWriter.java
emmileaf: Oops - thanks for catching this!

PR #1062, opened 2022-10-14
feat(SpringGen): adding extra annotations at the `SpringComposer` step for all Spring classes generated

zhumin8:
this pr:
- adds a `@BetaApi` annotation for each class composed in `SpringComposer` 
- also moves the `@Generated("by gapic-generator-java")` annotation to `SpringComposer` for consistency. (this annotation was missing for the properties class composer.)

Updating annotations in `SpringComposer`  level for the ones that applies to all classes makes it easier for future changes when needed.


emmileaf (approved):
LGTM! A couple of curiosities on the `BetaApi` annotation:
- Would it be helpful to provide a context string, like `@BetaApi(“Autogenerated Spring autoconfiguration is not yet stable”)`?
- Assuming this blanket BetaApi annotation will eventually be removed here: would we only intend to support Spring autoconfig for stable API client libraries, or could there be cases where this annotation is needed to distinguish such things from upstream?

zhumin8:
- Yes, that's a good idea. I'll add a context string to say this is experimental.
- This can be open for discussions, and can be done both ways. but I think it be reasonable to start out with stable ones.

diegomarquezp (approved):
LGTM!

PR #1064, opened 2022-10-21
fix(Spring CodeGen): Add user-agent header to generated client for metrics

zhumin8:
Adding user-agent header to generated client for metrics. Will mark this ready when header is setup and tested.

Also included some cleanups on un-used code in class (second committish - 04814ef).


elefeint (changes requested):
Make sure to test that the underlying header provider is also still used (the one that sends Gax/client library versions).

zhumin8:
I am still waiting for test metric data to show-up, but in theory, our change should not affect the Gax/client library header provider.
Client library header is set through [`setInternalHeaderProvider()`](https://github.com/googleapis/java-vision/blob/main/google-cloud-vision/src/main/java/com/google/cloud/vision/v1/stub/ImageAnnotatorStubSettings.java#L373) and defined [here](https://github.com/googleapis/java-vision/blob/main/google-cloud-vision/src/main/java/com/google/cloud/vision/v1/stub/ImageAnnotatorStubSettings.java#L220-L226).  However, this `setInternalHeaderProvider()` method is protected in gax, and we are using [`setHeaderProvider()`](https://github.com/googleapis/gax-java/blob/main/gax/src/main/java/com/google/api/gax/rpc/ClientSettings.java#L202) for for custom static headers instead.

zhumin8:
Test results with new tool_name: `spring-autogen` aligns with expected.
```
[{
 "service_name": "vision",
 "client_library_type": "gapic",
 "tool_name": "spring-autogen",
 "is_internal": null,
 "obscured_client_project_number": xxxx,
 "f0_": "4"
}, {
 "service_name": "language",
 "client_library_type": "gapic",
 "tool_name": "spring-autogen",
 "is_internal": null,
 "obscured_client_project_number":  xxxx,
 "f0_": "24"
}]
```
Will resolve merging conflicts.

elefeint (approved):

PR #1066, opened 2022-10-24
chore: expose methods for parsing api short name and version

emmileaf:
This PR exposes `parseApiShortName` (renamed from parseDefaultHost in #1040) and `parseApiVersion` (refactored from parsing proto package in prepareExecutableSamples) as public static methods in `Composer.java`. This change will enable Spring Codegen (when eventually split out from this repo) to reuse this parsing logic in descriptive comments and metadata. 

An alternative refactoring to expose these two values as field accessors in the `Service` class can be found on the [`refactor-parse-defaulthost-2`](https://github.com/googleapis/gapic-generator-java/compare/refactor-parse-defaulthost-2) branch. It moves the parsing logic to earlier in the process, since the source fields (`defaultHost` and `protoPakkage`) are both defined per-service. It’s a more convoluted change compared to this PR, but if the approach is preferred or adds any value from the sample generation perspective, happy to switch and open a PR from that branch instead.

### DIFF
--- a/src/main/java/com/google/api/generator/gapic/composer/Composer.java
+++ b/src/main/java/com/google/api/generator/gapic/composer/Composer.java
@@ -192,39 +192,10 @@ public class Composer {
     return clazzes;
   }
 
-  @VisibleForTesting
-  static List<GapicClass> prepareExecutableSamples(List<GapicClass> clazzes, String protoPackage) {
-    //  parse protoPackage for apiVersion
-    String[] pakkage = protoPackage.split("\\.");
-    String apiVersion;
-    //  e.g. v1, v2, v1beta1
-    if (pakkage[pakkage.length - 1].matches("v[0-9].*")) {
-      apiVersion = pakkage[pakkage.length - 1];
-    } else {
-      apiVersion = "";
-    }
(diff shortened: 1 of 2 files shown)


blakeli0:
Thanks Emily! Both approaches look pretty good! I would prefer the other one, even though it's more code but it opens up more opportunities in the future, if we need `apiShortName` or `apiVersion` to generate other stuff in the composers. I also think the other approach is cleaner instead of convoluted since it has a clear separation of concerns, the parsing logic is very specific to `Service`, it's not really Composer's concern.
If you really like this approach, I would at least extract the logic out to a util to make it cleaner. It would be weird if we call `Composer.parseApiShortName()` from another composer just for parsing.

emmileaf:
Thanks @blakeli0! I was mainly concerned that the other approach is bringing too much change to the code for just one feature request from the spring side, but your rationale makes sense and it’s helpful to know this refactoring could benefit other development in the future. In this case, I’ll close this PR and open up another one from [refactor-parse-defaulthost-2](https://github.com/googleapis/gapic-generator-java/compare/refactor-parse-defaulthost-2).

PR #1070, opened 2022-10-25
feat(ast): array expressions

diegomarquezp:
Added `ArrayExpr` to allow anonymous arrays (e.g. `{"value1", "value2"}`). Same type validation is enforced (e.g. `{"string1", 1.2f}` is not allowed).
`AnnotationNode` now allows an `ArrayExpr` as a single description. For multiple descriptions, a wrapping `AssignmentExpr` is still needed.

This PR is to help generate annotations with an anonymous array like this `@EnableConfigurationProperties({LanguageProperties.class, SharedProperties.class})`

It also enables assignments to arrays in the form `String[] arr = {"str1", "str2"}` - both `arr` and the anon array should be of type `TypeNode.arrayOf(TypeNode.STRING)` for this example (`arrayOf()` was also introduced in this PR)

### DIFF
--- a/src/main/java/com/google/api/generator/engine/ast/AnnotationNode.java
+++ b/src/main/java/com/google/api/generator/engine/ast/AnnotationNode.java
@@ -121,11 +121,11 @@ public abstract class AnnotationNode implements AstNode {
     /**
      * To add an AnonymousArrayAnnotationExpr as parameter. Can be used repeatedly to add multiple parameters.
      *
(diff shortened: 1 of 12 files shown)


diegomarquezp:
Force pushed an alternate approach using `ArrayValue`

blakeli0:
Can you delete this PR and open another PR against main branch? As this would be a good feature added to the generator itself

PR #1076, opened 2022-10-31
fix(ast): add import generation for annotations on VariableExpr

emmileaf:
This PR is a follow-up patch to #1012 which added support for annotations on `VariableExpr`. It updates `ImportWriterVisitor `so that import generation also covers these annotations. 

Note: the Spring CodeGen branch ([autoconfig-gen-draft2](https://github.com/googleapis/gapic-generator-java/tree/autoconfig-gen-draft2)) will need a rebase after this is merged, and have its golden test files updated to reflect this change.

### DIFF
--- a/src/test/java/com/google/api/generator/engine/writer/ImportWriterVisitorTest.java
+++ b/src/test/java/com/google/api/generator/engine/writer/ImportWriterVisitorTest.java
@@ -37,6 +37,7 @@ import com.google.api.generator.engine.ast.MethodDefinition;
 import com.google.api.generator.engine.ast.MethodInvocationExpr;
 import com.google.api.generator.engine.ast.NewObjectExpr;
 import com.google.api.generator.engine.ast.PackageInfoDefinition;
+import com.google.api.generator.engine.ast.PrimitiveValue;
 import com.google.api.generator.engine.ast.Reference;
 import com.google.api.generator.engine.ast.ReferenceConstructorExpr;
 import com.google.api.generator.engine.ast.RelationalOperationExpr;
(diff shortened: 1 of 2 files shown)


src/test/java/com/google/api/generator/engine/writer/ImportWriterVisitorTest.java
```diff
@@ -363,6 +364,41 @@ public void writeVariableExprImports_nestedReference() {
...
+            .setDescription(
+                ValueExpr.withValue(
+                    PrimitiveValue.builder().setValue("1").setType(TypeNode.INT).build()))
+            .build();
+
+    VariableExpr variableExpr =
+        VariableExpr.builder()
+            .setVariable(variable)
+            .setIsDecl(true)
+            .setAnnotations(Arrays.asList(annotation))
```
zhumin8, line 391: Nit: extremely small point, might save just a few lines using the concrete `Generated` as test.
```suggestion=
            .setAnnotations(Arrays.asList(
                AnnotationNode.withType(
                    TypeNode.withReference(ConcreteReference.withClazz(Generated.class)))))
```

src/test/java/com/google/api/generator/engine/writer/ImportWriterVisitorTest.java
```diff
@@ -363,6 +364,41 @@ public void writeVariableExprImports_nestedReference() {
...
+            .build();
+
+    TypeNode fakeAnnotationType =
+        TypeNode.withReference(
+            VaporReference.builder().setName("FakeAnnotation").setPakkage("com.foo.bar").build());
+
+    AnnotationNode annotation =
+        AnnotationNode.builder()
+            .setType(fakeAnnotationType)
+            .setDescription(
```
blakeli0, line 382: Do we have to set descriptions? If not, removing it makes the test more concise.

zhumin8 (commented):
One line fixes are the best!

src/test/java/com/google/api/generator/engine/writer/ImportWriterVisitorTest.java
emmileaf: Ah true, not for the functionality this test intends to cover - replaced with Min's suggestion below!

src/test/java/com/google/api/generator/engine/writer/ImportWriterVisitorTest.java
emmileaf: Ohh I hadn't realized this - thank you for the suggestion and updated!

zhumin8 (approved):

PR #1078, opened 2022-11-01
fix(Spring CodeGen): To only include options to set to rest transport when supported

zhumin8:
As per earlier discussions, `defaultHttpJsonTransportProvider` is only applicable when Transport.REST_GRPC (e.g. in case of [java_vision](https://github.com/googleapis/googleapis/blob/master/google/cloud/vision/v1/BUILD.bazel#L86)). In this change:
- if Transport.GRPC: do not provide property `useRest` to override TransportProvider, use default `defaultTransportChannelProvider()`.
- if Transport.REST_GRPC: by default, set `useRest = false` and use default settings same as above. But override to `defaultHttpJsonTransportProvider` when property `useRest` is set to `true`.
- In both cases, keep `TransportChannelProvider` bean. This bean defaults to `defaultTransportChannelProvider()`, but allows advanced users to override in user application (e.g. to [set a proxy](https://github.com/googleapis/google-cloud-java#configuring-a-proxy))


diegomarquezp (approved):

src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java
```diff
@@ -653,50 +661,51 @@ private static MethodDefinition createClientBeanMethod(
...
+              .setExprReferenceExpr(thisClientPropertiesVarExpr)
+              .build();
+
+      // LanguageServiceSettings.defaultHttpJsonTransportProviderBuilder().build()
+      Expr defaultTransportProviderBuider =
+          MethodInvocationExpr.builder()
+              .setStaticReferenceType(types.get("ServiceSettings"))
+              .setMethodName("defaultHttpJsonTransportProviderBuilder")
+              .build();
+      defaultTransportProviderBuider =
```
emmileaf, line 683: Nit: consider naming this part something else instead of re-assigning to `defaultTransportProviderBuider` here? (Might help make parsing through the chained invocation logic a bit easier)

src/test/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposerTest.java
```diff
@@ -33,10 +34,21 @@ public void setUp() {
   }
 
   @Test
-  public void generateAutoConfigClazzTest() {
+  public void generateAutoConfigClazzGrpcTest() {
     GapicClass clazz =
         SpringAutoConfigClassComposer.instance().generate(this.context, this.echoProtoService);
-    String fileName = clazz.classDefinition().classIdentifier() + ".golden";
+    String fileName = clazz.classDefinition().classIdentifier() + "Grcp.golden";
```
emmileaf, line 40: Small fileName typo here, and in the generated file
```suggestion
    String fileName = clazz.classDefinition().classIdentifier() + "Grpc.golden";
```

emmileaf (commented):
Two minor comments, otherwise LGTM!

src/test/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposerTest.java
zhumin8: Ah, good catch. thanks!

emmileaf (approved):

PR #1080, opened 2022-11-02
feat(ast): array expressions

diegomarquezp:
_Copy of #1070 but to be merged in `main`_

Added `ArrayExpr` to allow anonymous arrays (e.g. `{"value1", "value2"}`). Same type validation is enforced (e.g. `{"string1", 1.2f}` is not allowed).
`AnnotationNode` now allows an `ArrayExpr` as a single description. For multiple descriptions, a wrapping `AssignmentExpr` is still needed.

This PR is to help generate annotations with an anonymous array like this `@EnableConfigurationProperties({LanguageProperties.class, SharedProperties.class})`

It also enables assignments to arrays in the form `String[] arr = {"str1", "str2"}` - both `arr` and the anon array should be of type `TypeNode.arrayOf(TypeNode.STRING)` for this example (`arrayOf()` was also introduced in this PR)

### DIFF
--- a/src/main/java/com/google/api/generator/engine/ast/ArrayExpr.java
+++ b/src/main/java/com/google/api/generator/engine/ast/ArrayExpr.java
@@ -68,7 +68,6 @@ public abstract class ArrayExpr implements Expr {
       return addExpr(ValueExpr.withValue(StringObjectValue.withValue(expr)));
     }
 
-    // this method is private, and called only by addExpr(Expr expr)
     public ArrayExpr.Builder addExpr(Expr expr) {
       return setExprs((new ImmutableList.Builder<Expr>().addAll(exprs()).add(expr).build()));
     }
--- a/src/main/java/com/google/api/generator/engine/ast/ArrayExpr.java
+++ b/src/main/java/com/google/api/generator/engine/ast/ArrayExpr.java
@@ -72,12 +72,13 @@ public abstract class ArrayExpr implements Expr {
       return addExprToList(expr);
     }
 
-      /**
-       * To set many string expressions as the content
-       *
-       * @param expr
-       * @return Builder
-       */
+    /**
+     * To add a string expression
+     * same-type validation is performed
+     *
+     * @param expr
+     * @return Builder
+     */
     public ArrayExpr.Builder addExpr(String expr) {
       return addExprToList(ValueExpr.withValue(StringObjectValue.withValue(expr)));
     }
--- a/src/main/java/com/google/api/generator/engine/ast/ArrayExpr.java
+++ b/src/main/java/com/google/api/generator/engine/ast/ArrayExpr.java
@@ -5,7 +5,6 @@ import com.google.common.base.Preconditions;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
-import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 
 @AutoValue
@@ -25,18 +24,14 @@ public abstract class ArrayExpr implements Expr {
     return new AutoValue_ArrayExpr.Builder();
   }
 
-  public static ArrayExpr withStrings(String ...stringValues) {
+  public static ArrayExpr withStrings(String... stringValues) {
     ArrayExpr.Builder builder = ArrayExpr.builder();
-    Arrays.asList(stringValues)
-        .stream()
-        .forEach(s -> builder.addExpr(s));
+    Arrays.asList(stringValues).stream().forEach(s -> builder.addExpr(s));
     return builder.build();
   }
 
-  public static ArrayExpr withExprs(Expr ...exprs) {
-    return ArrayExpr.builder()
-        .setExprsList(Arrays.asList(exprs))
-        .build();
+  public static ArrayExpr withExprs(Expr... exprs) {
+    return ArrayExpr.builder().setExprsList(Arrays.asList(exprs)).build();
   }
 
   @AutoValue.Builder
@@ -73,8 +68,7 @@ public abstract class ArrayExpr implements Expr {
     }
 
     /**
-     * To add a string expression
-     * same-type validation is performed
+     * To add a string expression same-type validation is performed
      *
      * @param expr
      * @return Builder
@@ -84,8 +78,8 @@ public abstract class ArrayExpr implements Expr {
     }
 
     /**
-     * To set the list of expressions for the anonymous array
-     * Validates that every expression is of the same type
+     * To set the list of expressions for the anonymous array Validates that every expression is of
+     * the same type
      *
      * @param exprs
      * @return Builder
@@ -99,7 +93,7 @@ public abstract class ArrayExpr implements Expr {
         Preconditions.checkState(currentType.equals(baseType), SAME_TYPE_EXPRS_MESSAGE);
         baseType = currentType;
       }
-      setType(baseType);
+      setType(TypeNode.arrayOf(baseType));
       return setExprs(exprs);
     }
 
--- a/src/main/java/com/google/api/generator/engine/ast/ArrayExpr.java
+++ b/src/main/java/com/google/api/generator/engine/ast/ArrayExpr.java
@@ -9,7 +9,7 @@ import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 
 @AutoValue
-public abstract class AnonymousArrayAnnotationExpr implements Expr {
+public abstract class ArrayExpr implements Expr {
 
   @Nullable
   public abstract List<Expr> exprs();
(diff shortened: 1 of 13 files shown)


src/main/java/com/google/api/generator/engine/ast/ArrayExpr.java
```diff
@@ -0,0 +1,139 @@
...
+     * @return Builder
+     */
+    private ArrayExpr.Builder setExprsList(List<Expr> exprs) {
+      Preconditions.checkState(exprs.size() > 0, EMPTY_EXPRS_MESSAGE);
+      // validate types
+      TypeNode baseType = exprs.get(0).type();
+      for (int i = 1; i < exprs.size(); i++) {
+        TypeNode currentType = exprs.get(i).type();
+        Preconditions.checkState(currentType.equals(baseType), SAME_TYPE_EXPRS_MESSAGE);
+        baseType = currentType;
```
blakeli0, line 108: I don't think we need this line to reassign `baseType`, also we can achieve the same by using streams like `exprs.stream().allMatch(expr -> expr.type().equals(baseType))`.

src/main/java/com/google/api/generator/engine/ast/ArrayExpr.java
```diff
@@ -0,0 +1,139 @@
...
+import com.google.common.base.Preconditions;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import javax.annotation.Nullable;
+
+@AutoValue
+public abstract class ArrayExpr implements Expr {
+
+  @Nullable
```
blakeli0, line 27: We can remove the `@Nullable` annotation here if we are always validating that it should not be null.

src/main/java/com/google/api/generator/engine/ast/ArrayExpr.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    // this setter is private, and called only by setExprsList() to ensured sanitized contents
+    abstract ArrayExpr.Builder setExprs(List<Expr> descriptionExprs);
+
+    abstract ArrayExpr autoBuild();
+
+    public ArrayExpr build() {
+      ArrayExpr anonymousArrayExpr = autoBuild();
+      Preconditions.checkState(
```
blakeli0, line 131: We can remove the null check here if we remove the `@Nullable` annotation from the field. Also I think empty Array a valid case, e.g. `String[] strArray = {};` is a totally valid expression.

src/main/java/com/google/api/generator/engine/ast/ArrayExpr.java
```diff
@@ -0,0 +1,139 @@
...
+          anonymousArrayExpr.exprs() != null && anonymousArrayExpr.exprs().size() > 0,
+          EMPTY_EXPRS_MESSAGE);
+      Reference ref = anonymousArrayExpr.exprs().get(0).type().reference();
+      Preconditions.checkNotNull(ref, "Annotations must be an Object type");
```
blakeli0, line 135: I think we should get the type info from `TypeNode` field and use it as the source of truth for all checks.

src/main/java/com/google/api/generator/engine/ast/ArrayExpr.java
```diff
@@ -0,0 +1,139 @@
...
+    private ArrayExpr.Builder setExprsList(List<Expr> exprs) {
+      Preconditions.checkState(exprs.size() > 0, EMPTY_EXPRS_MESSAGE);
+      // validate types
+      TypeNode baseType = exprs.get(0).type();
+      for (int i = 1; i < exprs.size(); i++) {
+        TypeNode currentType = exprs.get(i).type();
+        Preconditions.checkState(currentType.equals(baseType), SAME_TYPE_EXPRS_MESSAGE);
+        baseType = currentType;
+      }
+      setType(TypeNode.arrayOf(baseType));
```
blakeli0, line 110: I think we should do it the other way around, set `TypeNode` explicitly and use it as source of truth for all checks.

src/main/java/com/google/api/generator/engine/ast/ArrayExpr.java
```diff
@@ -0,0 +1,139 @@
...
+      return setExprs(exprs);
+    }
+
+    // this method is private, and called only by addExpr(Expr expr)
+    private ArrayExpr.Builder addExprToList(Expr expr) {
+      List<Expr> exprList = exprs();
+      if (exprList == null) {
+        exprList = new ArrayList<>();
+      }
+      exprList.add(expr);
```
blakeli0, line 120: We can add a type check here as well every time we try to add an `Expr`

src/main/java/com/google/api/generator/engine/ast/TypeNode.java
```diff
@@ -101,6 +101,14 @@ public enum TypeKind {
 
   public abstract boolean isArray();
 
+  public static TypeNode arrayOf(TypeNode type) {
```
blakeli0, line 104: The name is a little confusing, sounds like it's creating a new array, I'm not good at naming as well, maybe something like `createArrayTypeOf()` or `ofArrayType()`?

diegomarquezp: `createArrayTypeOf` seems more precise - I'll use that!

src/main/java/com/google/api/generator/engine/ast/ArrayExpr.java
diegomarquezp: i.e. Set type explicitly before adding elements

src/main/java/com/google/api/generator/engine/ast/ArrayExpr.java
```diff
@@ -0,0 +1,139 @@
...
+        baseType = currentType;
+      }
+      setType(TypeNode.arrayOf(baseType));
+      return setExprs(exprs);
+    }
+
+    // this method is private, and called only by addExpr(Expr expr)
+    private ArrayExpr.Builder addExprToList(Expr expr) {
+      List<Expr> exprList = exprs();
+      if (exprList == null) {
```
diegomarquezp, line 117: Do not check null (builder initializes array). Use `ImmutableList`.

src/main/java/com/google/api/generator/engine/ast/ArrayExpr.java
```diff
@@ -0,0 +1,139 @@
...
+        exprList = new ArrayList<>();
+      }
+      exprList.add(expr);
+      return setExprsList(exprList);
+    }
+
+    // this setter is private, and called only by setExprsList() to ensured sanitized contents
+    abstract ArrayExpr.Builder setExprs(List<Expr> descriptionExprs);
```
diegomarquezp, line 125: Consider moving check logic to a public overriden version of this function

src/main/java/com/google/api/generator/engine/ast/ArrayExpr.java
diegomarquezp: Decided to do check in whole array via public overriden `setExprs`

src/main/java/com/google/api/generator/engine/writer/ImportWriterVisitor.java
```diff
@@ -167,6 +168,11 @@ public void visit(AnnotationNode annotation) {
     annotation.type().accept(this);
   }
 
+  @Override
+  public void visit(ArrayExpr expr) {
+    expr.exprs().get(0).type().accept(this);
```
diegomarquezp, line 173: Add test to confirm this

src/main/java/com/google/api/generator/engine/ast/ArrayExpr.java
diegomarquezp: Using `ArrayExpr.type()` as source of truth

src/main/java/com/google/api/generator/engine/ast/ArrayExpr.java
diegomarquezp: Removed

src/main/java/com/google/api/generator/engine/ast/ArrayExpr.java
diegomarquezp: Now allowing empty arrays.

src/main/java/com/google/api/generator/engine/ast/ArrayExpr.java
diegomarquezp: Kept abstract methods - moved validation to `build()`

src/main/java/com/google/api/generator/engine/writer/ImportWriterVisitor.java
diegomarquezp: Modified and tested.

src/main/java/com/google/api/generator/engine/ast/ArrayExpr.java
```diff
@@ -0,0 +1,93 @@
...
+     * To add a string expression same-type validation is performed
+     *
+     * @param expr
+     * @return Builder
+     */
+    public ArrayExpr.Builder addExpr(String expr) {
+      return addExpr(ValueExpr.withValue(StringObjectValue.withValue(expr)));
+    }
+
+    // this method is private, and called only by addExpr(Expr expr)
```
blakeli0, line 71: We can remove this comment now.

diegomarquezp: Fixed!

blakeli0 (approved):

PR #1084, opened 2022-11-04
fix(spring): use parsed api shortname as libname

emmileaf:
Follow-up PR to switch `getLibName()` to use apiShortName parsed from default host.
- ~Added an extra conversion for showcase (which we unit test with), which uses `localhost:7469` and does not parse to a pretty libname. (This means the current spring golden tests don't really cover other cases of `getLibName()`, but the logic itself also lives upstream)~

### DIFF
--- a/src/main/java/com/google/api/generator/spring/utils/Utils.java
+++ b/src/main/java/com/google/api/generator/spring/utils/Utils.java
@@ -64,12 +64,7 @@ public class Utils {
 
     // Option 3: Use parsed apiShortName from service proto's default host
     // (e.g. vision.googleapis.com => vision)
-    String apiShortName = context.services().get(0).apiShortName();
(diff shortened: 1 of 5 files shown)


zhumin8:
I wonder if we should just stick to `localhost:7469` for `showcase` and update our expected test resutls instead? Judging from this [test](https://github.com/googleapis/gapic-generator-java/blob/main/src/test/java/com/google/api/generator/gapic/model/ServiceTest.java#L55-L60 ) in #1075, `localhost:7469` is the expected api-shortname. We could alternatively alter out test setup, but as you said, that would be unnecessary since it's testing upstream logic.

emmileaf:
@zhumin8 Hmm yeah, keeping it that way and updating our goldens also makes sense given this is expected upstream logic. Will update!

zhumin8 (approved):

PR #1089, opened 2022-11-07
refactor: extract and move comment formatting logic to util

emmileaf:
Both `ClientLibraryPackageInfoComposer` and `ServiceClientCommentComposer` currently have additional parsing logic to convert protobuf comments to a Javadoc-compatible format. For example, this handles [bulleted list items with asterisks in secret manager service proto](https://github.com/googleapis/googleapis/blob/db1fc8a59aeec7aabfb3dfed4ce9944a312ed55e/google/cloud/secretmanager/v1/service.proto#L44-L45). 

This PR extracts common logic to an util class, and exposes it for reuse by Spring Codegen.

### DIFF
--- a/src/main/java/com/google/api/generator/gapic/composer/utils/CommentFormatter.java
+++ b/src/main/java/com/google/api/generator/gapic/composer/utils/CommentFormatter.java
@@ -1,0 +1,63 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.api.generator.gapic.composer.utils;
+
+import com.google.api.generator.engine.ast.JavaDocComment;
+import com.google.common.base.Strings;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class CommentFormatter {
+
+  private CommentFormatter() {};
+
+  // Additional formatting logic for converting protobuf comment to Javadoc
+  public static JavaDocComment.Builder formatCommentForJavaDoc(
+      String comment, JavaDocComment.Builder originalCommentBuilder, String firstPattern) {
+
+    JavaDocComment.Builder javaDocCommentBuilder = originalCommentBuilder;
+
+    String[] descriptionParagraphs = comment.split("\\n\\n");
+    for (int i = 0; i < descriptionParagraphs.length; i++) {
+      boolean startsWithItemizedList = descriptionParagraphs[i].startsWith(" * ");
+      // Split by listed items, then join newlines.
+      List<String> listItems =
+          Stream.of(descriptionParagraphs[i].split("\\n \\*"))
+              .map(s -> s.replace("\n", ""))
+              .collect(Collectors.toList());
+      if (startsWithItemizedList) {
+        // Remove the first asterisk.
+        listItems.set(0, listItems.get(0).substring(2));
+      }
+
+      if (!startsWithItemizedList) {
+        if (i == 0 && !Strings.isNullOrEmpty(firstPattern)) {
+          javaDocCommentBuilder =
+              javaDocCommentBuilder.addParagraph(String.format(firstPattern, listItems.get(0)));
+        } else {
+          javaDocCommentBuilder = javaDocCommentBuilder.addParagraph(listItems.get(0));
+        }
+      }
+      if (listItems.size() > 1 || startsWithItemizedList) {
+        javaDocCommentBuilder =
+            javaDocCommentBuilder.addUnorderedList(
+                listItems.subList(startsWithItemizedList ? 0 : 1, listItems.size()));
+      }
+    }
+
+    return javaDocCommentBuilder;
+  }
+}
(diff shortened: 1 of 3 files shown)


diegomarquezp (approved):
LGTM - nice cleanup!

src/main/java/com/google/api/generator/gapic/composer/utils/CommentFormatter.java
```diff
@@ -0,0 +1,62 @@
...
+
+import com.google.api.generator.engine.ast.JavaDocComment;
+import com.google.common.base.Strings;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class CommentFormatter {
+
+  private CommentFormatter() {}
```
blakeli0, line 25: This empty constructor is not needed.

src/main/java/com/google/api/generator/gapic/composer/utils/CommentFormatter.java
```diff
@@ -0,0 +1,62 @@
...
+
+  // Additional formatting logic for converting protobuf comment to Javadoc
+  public static String formatAndAddToJavaDocComment(String comment, String prefixPattern) {
```
blakeli0, line 28: The method name may not reflect the latest changes, usually if you have an `and` in a method name, it's a code smell that either the method is doing too much or you can have a better name, maybe something like `fromProtobufCommentToJavaDocComment`?(I'm not good at naming as well, just throwing out an idea here)

src/main/java/com/google/api/generator/gapic/composer/utils/CommentFormatter.java
```diff
@@ -0,0 +1,62 @@
...
+import com.google.common.base.Strings;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class CommentFormatter {
+
+  private CommentFormatter() {}
+
+  // Additional formatting logic for converting protobuf comment to Javadoc
```
blakeli0, line 27: Can you provide a little more example about what exact this method is doing?

blakeli0 (changes requested):
Please resolve the comments

src/main/java/com/google/api/generator/gapic/composer/utils/CommentFormatter.java
zhumin8: I think this is from sonar recommendation to use private constructor to hide the implicit public one.
https://rules.sonarsource.com/java/RSPEC-1118

src/main/java/com/google/api/generator/gapic/composer/utils/CommentFormatter.java
emmileaf: Thanks for catching this, I missed updating the method name when switching to returning a string instead of adding to a builder. Renamed to `formatAsJavaDocComment`.

src/main/java/com/google/api/generator/gapic/composer/utils/CommentFormatter.java
emmileaf: Updated the unit tests to a couple of different parsing cases - hopefully this illustrates the method's behaviour a bit better!

emmileaf:
Regarding the code coverage:

The use of this logic/util method in the two composers is covered by the integration tests (e.g. [redis golden file](https://github.com/googleapis/gapic-generator-java/blob/main/test/integration/goldens/redis/src/com/google/cloud/redis/v1beta1/package-info.java)). Sonar is unhappy with coverage because the existing composer unit tests (e.g. through `ServiceClientClassComposerTest`) do not hit any of this logic -  it only applies when services or methods have a `.description()` field, which showcase services do not.
 
Todo: will need to look more into how `Parser` gets descriptions for services and methods to see if unit test resources can be updated to cover comment parsing in existing Composer tests, or if this behavior needs to be separately mocked.
- [service.description()](https://github.com/googleapis/gapic-generator-java/blob/ee5d7a5985d298e6d96dc700805870f3e362b1de/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java#L457-L464)
- [method.description()](https://github.com/googleapis/gapic-generator-java/blob/ee5d7a5985d298e6d96dc700805870f3e362b1de/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java#L686-L693)

src/main/java/com/google/api/generator/gapic/composer/utils/CommentFormatter.java
blakeli0: I see, this is one of the controversial rules though, see this [stackoverflow](https://stackoverflow.com/questions/14398747/how-to-disable-warning-on-sonar-hide-utility-class-constructor) discussion. IMO, it does not provide much value and add unnecessary code, maybe we should turn off this rule? WDYT @meltsufin ?

emmileaf:
Some additional test-related changes from addressing code coverage:
- Added description for echo service explicitly in `TestProtoLoader` since it is not parsed from source code location as an end-to-end call would
  - This allows the existing `ServiceClientClassTest` to cover class header comments that are also formatted with the util’s logic
- Added `ClientLibraryPackageInfoComposerTest` as a golden-updating test

src/main/java/com/google/api/generator/gapic/composer/utils/CommentFormatter.java
meltsufin: I kind of agree with Sonar here. A private constructor is helpful to avoid mistakes around instantiating the class when it's intended to be used as a collection of static utility methods.

blakeli0: I think it makes sense if this is a util class in `gax-java` or a backend of a web app, but maybe not for `gapic-generator-java`, because pretty much everything in the generator is called through `public static` methods in parsers and composers, the whole generator is basically a util. 

There may be unnecessary garbage collection if we accidentally instantiated a util class, which could cause some harm in a server if the util method is call multiple times, but there is no harm for the generator because the generator is run once per jvm.

blakeli0: Mike and I discussed offline and we are going to remove this rule from gapic-generator-java, but it's not critically enough to block this PR.

blakeli0 (approved):

PR #1090, opened 2022-11-10
ci: Add showcase build to CI test steps

burkedavison:
Introduces showcase client generation during `bazel test //test/integration:showcase`.


PROPERTIES.bzl
```diff
@@ -3,6 +3,7 @@ PROPERTIES = {
     # Version of google-java-format is downgraded from 1.8 to 1.7, because 1.8 supports java 11 minimum, while our JRE is java 8.
     "version.google_java_format": "1.7",
     "version.io_grpc_java": "1.42.1",
+    "commit.gapic_showcase": "90d73532a0cab753a85f45c158394f24fc21d91a",
```
blakeli0, line 6: We plan to remove this properties file soon and use maven to load dependencies, can we inline the showcase hash?

test/showcase/BUILD.bazel
```diff
@@ -0,0 +1,57 @@
...
+    deps = [
+        ":showcase_java_proto",
+        "@com_google_googleapis//google/api:api_java_proto",
+    ],
+)
+
+# Open Source Packages
+java_gapic_assembly_gradle_pkg(
+    name = "google-cloud-showcase-v1beta1-java",
+    include_samples = True,
```
blakeli0, line 49: We don't need to generate samples for showcase.

test/showcase/BUILD.bazel
```diff
@@ -0,0 +1,57 @@
...
+    srcs = ["@com_google_gapic_showcase//schema/google/showcase/v1beta1:showcase_proto"],
+    deps = [":showcase_java_proto"],
+)
+
+java_gapic_library(
+    name = "showcase_java_gapic",
+    srcs = [":showcase_proto_with_info"],
+    gapic_yaml = None,
+    grpc_service_config = "@com_google_gapic_showcase//schema/google/showcase/v1beta1:showcase_grpc_service_config.json",
+    service_yaml = "@com_google_gapic_showcase//schema/google/showcase/v1beta1:showcase_v1beta1.yaml",
```
blakeli0, line 35: Can we add `rest_numeric_enums = True` here? Similar to [here](https://github.com/googleapis/googleapis/blob/9140e5546470d945fc741f27707aa68f562088f0/google/cloud/aiplatform/v1/BUILD.bazel#L176), we can use numeric enums tests as our first integration tests.

.github/workflows/ci.yaml
```diff
@@ -47,6 +47,9 @@ jobs:
       - name: Integration Tests
         run: bazel --batch test //test/integration/...
 
+      - name: Showcase Tests
```
blakeli0, line 50: is this only generating the showcase client libraries at this moment?

burkedavison: Correct - and only once merged. More will be needed, but the intent was to get it executed by CI asap.

test/showcase/BUILD.bazel
burkedavison: Done

test/showcase/BUILD.bazel
burkedavison: Done

PROPERTIES.bzl
burkedavison: Done

blakeli0 (changes requested):
Looks pretty good. Two suggestions:
1. Please add some golden tests for the test class composers.
2. Consider splitting the two bug fixes out to a separate `fix` PR, and this PR can stay as a `ci`.

WORKSPACE
```diff
@@ -2,6 +2,7 @@ workspace(name = "gapic_generator_java")
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load("@bazel_tools//tools/build_defs/repo:jvm.bzl", "jvm_maven_import_external")
+load("//:PROPERTIES.bzl", "PROPERTIES")
```
blakeli0, line 5: This line can be removed now.

burkedavison: Done

test/integration/goldens/showcase/samples/snippets/generated/main/java/com/google/showcase/v1beta1/compliance/create/SyncCreateSetCredentialsProvider.java
```diff
@@ -0,0 +1,44 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.showcase.v1beta1.samples;
+
+// [START localhost7469_v1beta1_generated_Compliance_Create_SetCredentialsProvider_sync]
+import com.google.api.gax.core.FixedCredentialsProvider;
```
blakeli0, line 20: We don't need these auto generated samples, they shouldn't be generated anymore without `include_samples=True`, the whole `samples` folder can be deleted.

test/showcase/BUILD.bazel
```diff
@@ -0,0 +1,72 @@
...
+        ":showcase_java_grpc",
+    ],
+    transport = "grpc+rest",
+    deps = [
+        ":showcase_java_proto",
+        "@com_google_googleapis//google/api:api_java_proto",
+    ],
+)
+
+java_gapic_test(
```
blakeli0, line 49: Are these tests run with `bazel --batch build //test/showcase:all`?

test/integration/BUILD.bazel
```diff
@@ -76,6 +78,20 @@ java_gapic_test(
     runtime_deps = ["@com_google_googleapis//google/cloud/asset/v1:asset_java_gapic_test"],
 )
 
+# Showcase API.
+java_gapic_test(
```
blakeli0, line 82: Do we need this bazel target?

test/showcase/BUILD.bazel
burkedavison: They are invoked with `bazel test //test/showcase:all`, not `build`.

test/integration/goldens/showcase/samples/snippets/generated/main/java/com/google/showcase/v1beta1/compliance/create/SyncCreateSetCredentialsProvider.java
blakeli0: Actually they are here because how `include_samples` are implemented, it's implemented in `java_gapic_assembly_gradle_pkg` not `java_gapic_library`, so the samples are always generated but not included in the package. We may want to change this behavior though in the future.

test/integration/BUILD.bazel
burkedavison: This depends on how we want to invoke the showcase tests. Either they are to be integrated with the existing `//test/integration` stages, or we create a separate workflow to test them. I had assumed the `integrated with the existing` route due to how similar the syntax is (so far). Once we start working with the server, the tests and syntax will quickly diverge.

How would you like to handle this?

test/showcase/BUILD.bazel
blakeli0: I don't see `bazel test //test/showcase:all` in this PR, that's why I'm wondering if this is triggered or not at this point. There is an identical `java_gapic_test` in the Bazel build file in integration folder, which I think will be triggered by `bazel test //...` automatically, but is this one invoked or not?
The bottom line is I think we only want to have these target declared and invoked in one place to prevent confusion.

test/integration/BUILD.bazel
blakeli0: I think either way makes sense, we are not moving away from the golden integration tests yet soon, so we can continue to test the golden files and the generated unit tests in integration folder, while all the future handwritten showcase tests will be in the showcase folder. Or we move all showcase related stuff to showcase folder.

test/showcase/BUILD.bazel
burkedavison: Yes - makes sense and I agree. I've removed the test suite from /test/showcase so those can focus on the actual integration tests w/ the server.

test/integration/BUILD.bazel
burkedavison: Kept here, and removed from /test/showcase.

blakeli0 (approved):
LGTM. You may want to update the description of the PR before merging, now that we break the original PR to two.

PR #1093, opened 2022-11-14
fix(spring): update setters for properties of type Duration

emmileaf:
**This PR:** updates setters for Spring properties of type `org.threeten.bp.Duration` to accept argument of type `java.time.Duration` instead, and parse to `org.threeten.bp.Duration` in the setter body. 
- Converting between the two types: `parse()` and `toString()` methods for both will accept or return ISO-8601 seconds based representation, such as `PT8H6M12.345S`.

This addresses the issue that properties of type `org.threeten.bp.Duration` cannot be set from strings (e.g. specified in Spring property files) because the implicit conversion is unsupported. Spring boot has [dedicated support](https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.external-config.typesafe-configuration-properties.conversion.durations) for properties of type `java.time.Duration`.

**Note for review:** would appreciate any feedback or thoughts on the conversion approach here! Some alternative approaches explored were:
- The property type itself can be changed to `java.time.Duration`, but have getters modified to convert to `org.threeten.bp.Duration` ([poc](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/df71cb64ebc9e3355a4f688f137d49c1761ed6c8/generated/language/src/main/java/com/google/cloud/language/v1/spring/LanguageServiceSpringProperties.java#L138-L149))
- Adding custom converter for String to `org.threeten.bp.Duration` through `@ConfigurationPropertiesBinding` ([poc](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/df71cb64ebc9e3355a4f688f137d49c1761ed6c8/generated/language/src/main/java/com/google/cloud/language/v1/spring/StringToBackportDurationConverter.java))

### DIFF
--- a/src/main/java/com/google/api/generator/spring/composer/SpringPropertiesClassComposer.java
+++ b/src/main/java/com/google/api/generator/spring/composer/SpringPropertiesClassComposer.java
@@ -296,16 +296,16 @@ public class SpringPropertiesClassComposer implements ClassComposer {
     // Common building blocks
     Variable propertyVar = Variable.builder().setName(propertyName).setType(returnType).build();
     Expr thisExpr = ValueExpr.withValue(ThisObjectValue.withType(thisClassType));
-    TypeNode oldDurationType = staticTypes.get("org.threeten.bp.Duration");
-    TypeNode newDurationType = staticTypes.get("java.time.Duration");
+    TypeNode threetenBpDurationType = staticTypes.get("org.threeten.bp.Duration");
+    TypeNode javaTimeDurationType = staticTypes.get("java.time.Duration");
 
     // Default building blocks - may be updated in Duration condition below
     Variable argumentVar = propertyVar;
     Expr propertyValueExpr = VariableExpr.withVariable(argumentVar);
 
     // Setter logic for Duration accepts different type and handles conversion
-    if (returnType.equals(oldDurationType)) {
-      argumentVar = Variable.builder().setName(propertyName).setType(newDurationType).build();
+    if (returnType.equals(threetenBpDurationType)) {
+      argumentVar = Variable.builder().setName(propertyName).setType(javaTimeDurationType).build();
 
       MethodInvocationExpr durationToStringExpr =
           MethodInvocationExpr.builder()
@@ -316,10 +316,10 @@ public class SpringPropertiesClassComposer implements ClassComposer {
 
       propertyValueExpr =
           MethodInvocationExpr.builder()
-              .setStaticReferenceType(oldDurationType)
+              .setStaticReferenceType(threetenBpDurationType)
               .setMethodName("parse")
               .setArguments(durationToStringExpr)
-              .setReturnType(oldDurationType)
+              .setReturnType(threetenBpDurationType)
               .build();
     }
 
--- a/src/main/java/com/google/api/generator/spring/composer/SpringPropertiesClassComposer.java
+++ b/src/main/java/com/google/api/generator/spring/composer/SpringPropertiesClassComposer.java
@@ -25,6 +25,7 @@ import com.google.api.generator.engine.ast.ConcreteReference;
 import com.google.api.generator.engine.ast.Expr;
 import com.google.api.generator.engine.ast.ExprStatement;
 import com.google.api.generator.engine.ast.MethodDefinition;
+import com.google.api.generator.engine.ast.MethodInvocationExpr;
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/api/generator/spring/composer/SpringPropertiesClassComposer.java
```diff
@@ -256,7 +257,13 @@ private static List<MethodDefinition> createGetterSetters(
               String propertyName = Joiner.on("").join(methodAndPropertyName);
               getterAndSetter.add(
                   createGetterMethod(thisClassType, propertyName, propertyType, null));
-              getterAndSetter.add(createSetterMethod(thisClassType, propertyName, propertyType));
+              if (propertyType.equals(staticTypes.get("org.threeten.bp.Duration"))) {
+                // Use different setter logic for Duration handling conversion
+                getterAndSetter.add(
+                    createDurationSetterMethod(thisClassType, propertyName, propertyType));
+              } else {
+                getterAndSetter.add(createSetterMethod(thisClassType, propertyName, propertyType));
```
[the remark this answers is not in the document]
emmileaf: Ah, that’s a good point! Currently the only call pattern is through `createGetterSetters()` since these are not exposed as public methods, but it will probably benefit future maintenance. I’ve moved the logic to all live inside `createSetterMethod()`.

src/main/java/com/google/api/generator/spring/composer/SpringPropertiesClassComposer.java
```diff
@@ -315,6 +322,52 @@ private static MethodDefinition createSetterMethod(
...
+        AssignmentExpr.builder()
+            .setVariableExpr(
+                VariableExpr.withVariable(propertyVar)
+                    .toBuilder()
+                    .setExprReferenceExpr(thisExpr)
+                    .build())
+            .setValueExpr(parsedDurationExpr)
+            .build();
+
+    String methodName = "set" + CaseFormat.LOWER_CAMEL.to(CaseFormat.UPPER_CAMEL, propertyName);
```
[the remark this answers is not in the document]
emmileaf: Hmm good question - looks like under the hood `JavaStyle.ToUpperCamelCase()` uses `CaseFormat` to [convert from lower snake to upper camel case](https://github.com/googleapis/gapic-generator-java/blob/ee5d7a5985d298e6d96dc700805870f3e362b1de/src/main/java/com/google/api/generator/gapic/utils/JavaStyle.java#L53), whereas we are converting from lower camel in this line.

I'm inclined to keep the current implementation but practically they do yield the same result, since here we just need the first letter of `propertyName` to change from lower to upper case.

src/main/java/com/google/api/generator/spring/composer/SpringPropertiesClassComposer.java
```diff
@@ -365,14 +394,8 @@ private static Map<String, TypeNode> createDynamicTypes(Service service, String
                 .setPakkage("org.springframework.boot.context.properties")
                 .build());
 
-    // import org.threeten.bp.Duration;
```
zhumin8, line 368: Thanks for catching this, it should be concrete type instead.

src/main/java/com/google/api/generator/spring/composer/SpringPropertiesClassComposer.java
```diff
@@ -291,8 +292,36 @@ private static MethodDefinition createGetterMethod(
 
   private static MethodDefinition createSetterMethod(
       TypeNode thisClassType, String propertyName, TypeNode returnType) {
+
+    // Common building blocks
     Variable propertyVar = Variable.builder().setName(propertyName).setType(returnType).build();
     Expr thisExpr = ValueExpr.withValue(ThisObjectValue.withType(thisClassType));
+    TypeNode oldDurationType = staticTypes.get("org.threeten.bp.Duration");
+    TypeNode newDurationType = staticTypes.get("java.time.Duration");
```
zhumin8, line 300: nit: perhaps naming as `threetenDurationType` and `JavaDurationType`?

zhumin8 (approved):
LGTM!

src/main/java/com/google/api/generator/spring/composer/SpringPropertiesClassComposer.java
emmileaf: Yup - updated to `threetenBpDurationType` and `javaTimeDurationType`.

PR #1094, opened 2022-11-14
fix: Support testing nested argument method signatures and 'double' field assertions

burkedavison:
* **Handle [method signatures with nested arguments](https://github.com/googleapis/gapic-showcase/blob/main/schema/google/showcase/v1beta1/identity.proto#L44-L46)**.

  Before:
  ```java
  Assert.assertEquals(displayName, actualRequest.getDisplayName());
  ```
  After:
  ```java
  Assert.assertEquals(displayName, actualRequest.getUser().getDisplayName());
  ```

* **Provide a consistent tolerance value of `0.0001` for `assertEquals` on `double` and `float` values.**

  Before:
  ```java
  Assert.assertEquals(expected, actual);
  ```
  > bazel error: [DoNotCall] This method always throws java.lang.AssertionError. Use assertEquals(expected, actual, delta) to compare floating-point numbers

  After:
  ```java
  Assert.assertEquals(expected, actual, 0.0001);
  ```

### DIFF
--- a/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceClientTestClassComposer.java
+++ b/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceClientTestClassComposer.java
@@ -82,6 +82,7 @@ import org.junit.BeforeClass;
 import org.junit.Test;
 
 public abstract class AbstractServiceClientTestClassComposer implements ClassComposer {
+
   protected static final Statement EMPTY_LINE_STATEMENT = EmptyLineStatement.create();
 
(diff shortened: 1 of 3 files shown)


blakeli0 (approved):
Ideally for every composer change, we should have a golden unit test updated, like [this one](https://github.com/googleapis/gapic-generator-java/blob/main/src/test/java/com/google/api/generator/gapic/composer/rest/goldens/ComplianceClientTest.golden). But since we know it will be tested as part of the showcase integration tests, and we are moving away from generated unit tests, this PR is good to merge.

PR #1095, opened 2022-11-14
fix: eliminate raw-types and remove unnecessary cast

burkedavison:


emmileaf (approved):
LGTM - thank you for these fixes!

(Nit on PR naming: perhaps prefix with something like `fix(spring)` or `fix(Spring CodeGen)`, just to help differentiate with PRs merging into the main branch?)

src/main/java/com/google/api/generator/spring/utils/Utils.java
```diff
@@ -90,14 +91,14 @@ public static String getSpringPropertyPrefix(String packageName, String serviceN
...
-  public static List<? extends AstNode> processRetrySettings(
+  public static <T extends AstNode> List<T> processRetrySettings(
       Service service,
       GapicServiceConfig gapicServiceConfig,
       TypeNode thisClassType,
-      Function<String, List<? extends AstNode>> perMethodFuncBeforeSettings,
-      BiFunction<List<String>, Expr, List<? extends AstNode>> processFunc,
-      Function<String, List<? extends AstNode>> perMethodFuncAfterSettings) {
-    List resultList = new ArrayList<>();
+      Function<String, List<? extends T>> perMethodFuncBeforeSettings,
```
blakeli0, line 98: Can we change `<? extends T>` to `T` as well? Looks like what gets returned are always the same type?

burkedavison: `T` is valid for the current source; but as a `Util` function, the limit of the implementation is `<? extends T>` which defines how it can be used in the future as well.

zhumin8 (commented):
Thanks for fixing this!

src/main/java/com/google/api/generator/spring/utils/Utils.java
burkedavison: For example, without the `? extends T`, it wouldn't be possible to combine type-related-but-not-the-same helper functions:
```java
abstract List<CommentStatement> getHeaderComments(String methodName);
abstract List<Statement> getStatements(String methodName);
private static <T> List<T> NONE(String unused) { return Collections.emptyList(); }

private void example() {
  List<Statement> statements = Util.processRetrySettings(
    server,
    config,
    clazzType,
    this::getHeaderComments,
    this::getStatements,
    this::NONE
}
```

blakeli0: I agree with the general idea. But for this util method, it's a very specific use case(Create retry settings with some hardcoded method names) that is only used by the two composers, so I doubt it will be used by anything else. I could also argue that this method should not be a public Util, maybe it should be in something like a `AbstractSpringComposer` with default scope. 
That being said, I don't think this is critical enough to block the PR, so either way works.

blakeli0 (approved):

PR #1096, opened 2022-11-16
fix(spring): add parent and remove maven-jar-plugin from pom string.

zhumin8:
Two changes:
- add parent to generated module: `generated-parent`. This parent pom is manually created in [`spring-cloud-gcp/generated`](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/generated/pom.xml).
    -  parent pom coordinate is not expected to change. #1086 can be an improvement on top of this hardcoded parent.
- remove unnecessary `maven-jar-plugin` as module will inherit from spring parent.


emmileaf (approved):

diegomarquezp (approved):
LGTM
Following @emmileaf 's line, a good patter would be that of [rules_gapic](https://github.com/googleapis/rules_gapic/blob/main/bazel/src/main/java/com/google/api/codegen/bazel/BazelBuildFileTemplate.java) if we want to use templates for the pom.

diegomarquezp:
> parent pom coordinate is not expected to change. https://github.com/googleapis/gapic-generator-java/pull/1086 can be an improvement on top of this hardcoded parent.

Maybe we can just close #1086

zhumin8:
On the template note, haven't looked closely at @diegomarquezp's suggestion yet, but can we do something similar to [client_grpcrest.gradle.tmpl](https://github.com/googleapis/gapic-generator-java/blob/main/rules_java_gapic/resources/gradle/client_grpcrest.gradle.tmpl)?

PR #1098, opened 2022-11-18
ci: add showcase testing

burkedavison:


pom.xml
```diff
@@ -8,6 +10,8 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <!-- Should be in sync with repositories.bzl -->
     <googleapis.commit>44d6bef0ca6db8bba3fb324c8186e694bcc4829c</googleapis.commit>
+    <!-- Should be in sync with WORKSPACE -->
+    <showcase.commit>90d73532a0cab753a85f45c158394f24fc21d91a</showcase.commit>
```
blakeli0, line 14: Is this used?

src/test/java/com/google/showcase/v1beta1/NumericEnumsIT.java
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  @After
+  public void destroyClient() {
+    client.close();
+  }
+
+  // See
+  // https://github.com/googleapis/gapic-showcase/blob/v0.25.0/util/genrest/resttools/systemparam.go#L37-L46
+  @Test(expected = InvalidArgumentException.class)
```
blakeli0, line 55: Usually we use `assertThrows` instead because it's clearer where exactly is the exception thrown. But I know this test case is temporary as well because the showcase server issue.

src/test/java/com/google/showcase/v1beta1/BUILD.bazel
```diff
@@ -0,0 +1,48 @@
+load(
```
blakeli0, line 1: Can we move all showcase related code to a separate folder like `test/showcase`? We plan to have no bazel stuff in the `src` folder, not sure if we can achieve the same thing with Maven. I also think it's fine to leave all the golden integration tests and showcase integration tests with Bazel, because the main benefit of migrating the generator to Maven is to have a hermetic code generation process, there is not much benefit of migration the testing aspect to Maven. 

Also, a typical dev workflow would be `make changes -> unit tests with testing protos -> integration tests with showcase/real protos`, this change will require the unit tests to dependent on showcase client, which feels backwards. Now the dev workflow would be `make changes -> build showcase client -> run both unit tests and integration tests`.

scripts/showcase/README.md
```diff
@@ -0,0 +1,29 @@
+# Showcase Integration Testing
+
+## Requirements
+
+* Install `go` in your `PATH`
+
+## Scripts
+
+* `install.sh <version>`
```
blakeli0, line 9: This could be subjective, shell scripts are useful for encapsulating complex logics, but maybe these scripts are simple enough so that we don't have to wrap them up. We may also want to expose more functionality of `gapic-showcase` command since sometimes we may want to test directly, wrapping it up will kind of hide it from developers.

pom.xml
burkedavison: It was, and then I dropped it when refactoring. Thanks! Removed.

src/test/java/com/google/showcase/v1beta1/NumericEnumsIT.java
burkedavison: GTK. I've made some adjustments to use the recommended `assertThrows`

scripts/showcase/README.md
burkedavison: Understood. For now, I'm going to leave them in just for my own ease of iterating -- but we can remove them or rework them as we get closer to the merge.

burkedavison: Resolved by moving scripts into `DEVELOPMENT.md`

PR #1099, opened 2022-11-21
fix(spring): fix write pom placeholders.

zhumin8:
- `{{starter-version}}` are not needed anymore as we inherit version from parent pom.
- add parent version placeholder. Changes on spring-cloud-gcp side needed.
- remove redundant groupid.
- remove versions for `spring-cloud-gcp-core` and `spring-boot-starter`, specified in `genereated-parent` pom

### DIFF
--- a/src/main/java/com/google/api/generator/spring/SpringWriter.java
+++ b/src/main/java/com/google/api/generator/spring/SpringWriter.java
@@ -212,6 +212,7 @@ public class SpringWriter {
 
     String springStarterArtifactId = clientLibraryName + "-spring-starter";
     String springStarterName = "Spring Boot Starter - " + clientLibraryShortName;
+    String springParentVersion = "{{parent-version}}";
 
     StringJoiner sb = new StringJoiner(",\\\n");
     sb.add(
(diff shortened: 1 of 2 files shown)


emmileaf:
@zhumin8 Ah, I was also running into the issue that parent version needs to specified in the generated pom. Since we are aligning versions with `spring-cloud-gcp`, would the [`${revision}`](https://maven.apache.org/maven-ci-friendly.html) property in maven be something we could use as an alternative to the `{{parent-version}}` placeholder? 

Together with spring-cloud-gcp side edits, this would look like https://github.com/GoogleCloudPlatform/spring-cloud-gcp/commit/79e0864a1d7503fbba95e7649086093e98714a2b as a POC.

zhumin8:
@emmileaf Yeah,  `${revision}` is better here! I'll make the changes.

zhumin8:
reverting the `${revision}` for now to un-block generation process setup.

diegomarquezp (approved):

emmileaf (approved):

PR #1103, opened 2022-11-29
fix(spring): change @Generated description to spring specific.

zhumin8:
change `@Generated` description to spring specific and separate from "gapic-generator-java" for client libraries.
Also moved relevant annotation from `SpringPackageInfoComposer` to `SpringComposer`.


emmileaf:
Couldn't figure out how to add suggestions to unchanged files, but this description is also set in `SpringPackageInfoComposer` (and `SpringPackageInfo.golden`). Could you help make the quick change in this PR? Thanks!

src/main/java/com/google/api/generator/spring/composer/SpringComposer.java
```diff
@@ -81,7 +81,7 @@ protected static List<GapicClass> addExtraClassAnnotations(List<GapicClass> gapi
     AnnotationNode generatedAnnotation =
         AnnotationNode.builder()
             .setType(FIXED_TYPESTORE.get("Generated"))
-            .setDescription("by gapic-generator-java")
+            .setDescription("by spring-generator-java")
```
blakeli0, line 84: What do we want to call the spring auto config generator once we move the code to `spring-cloud-gcp`? Maybe we should come up with a long term name and stick to it, `spring-generator-java` seems a little odd to me. Maybe `spring-auto-config-generator` or just a generic `spring-generator`?

zhumin8: Thanks for the suggestions. `spring-generator` sounds better to me, to keep it open to expanding beyond auto-config in future.

meltsufin: Should we add "google" somewhere in there? For instance, `google-cloud-spring-generator`?

emmileaf (approved):

blakeli0 (approved):

PR #1104, opened 2022-11-29
feat: set logging level to trace

diegomarquezp:
As suggested after discussing https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1364


src/test/java/com/google/api/generator/spring/composer/goldens/EchoSpringAutoConfigurationFull.golden
```diff
@@ -132,162 +132,163 @@ public class EchoSpringAutoConfiguration {
...
           this.clientProperties.getCollectInitialRpcTimeout());
-      LOGGER.info(
+      LOGGER.trace(
           "CollectInitialRpcTimeout set to " + this.clientProperties.getCollectInitialRpcTimeout());
     }
     if (this.clientProperties.getCollectRpcTimeoutMultiplier() != null) {
       collectRetrySettingBuilder.setRpcTimeoutMultiplier(
           this.clientProperties.getCollectRpcTimeoutMultiplier());
-      LOGGER.info(
+      LOGGER.trace(
```
meltsufin, line 229: I think adding `isTraceEnabled()` would be useful here, since we're not only doing string concatenation, but also calling a method.

meltsufin (approved):

zhumin8 (approved):

PR #1105, opened 2022-11-29
chore: relocate gapic-generator-java pom.xml to subfolder

burkedavison:
@blakeli0 : This is the minimum number of edits needed to make the project work with a relocated `pom.xml`.

Let's discuss whether you'd prefer to have this stay in its own branch for now.


blakeli0:
> @blakeli0 : This is the minimum number of edits needed to make the project work with a relocated `pom.xml`.
> 
> Let's discuss whether you'd prefer to have this stay in its own branch for now.

The changes look good to me. The bazel build is failing though, if all the checks are passing, we could merge it in.

src/test/java/com/google/api/generator/engine/JavaCodeGeneratorTest.java
```diff
@@ -63,8 +63,9 @@
 import org.junit.Test;
 
 public class JavaCodeGeneratorTest {
+
   private static final String GOLDENFILES_DIRECTORY =
-      "src/test/java/com/google/api/generator/engine/goldens/";
+      "../src/test/java/com/google/api/generator/engine/goldens/";
```
meltsufin, line 68: Would it be possible to change the working directory for surefire or failsafe instead?

burkedavison: Good idea. I'll give it a try.

burkedavison: This suggestion gets bonus points for both simplifying the PR *and* fixing the bazel build.

meltsufin (approved):
Glad the working directory worked!

blakeli0 (approved):

PR #1111, opened 2022-12-05
feat: Add Maven release

blakeli0:
- Add `.kokoro` folder for releasing gapic-generaotr-java as a jar.
  - It is mostly copied from `.kokoro` folder in [gax-java](https://github.com/googleapis/gax-java/tree/main/.kokoro), and only keeping the necessary.
- Change release type from `simple` to `java-yoshi`
- Update the pom version to match the current version.


version.txt
```diff
@@ -1 +1 @@
-2.11.0
```
lqiu96, line 1: nit: Can you add 
```
# Format:
# module:released-version:current-version
```

like it is in: https://github.com/googleapis/gax-java/blob/main/versions.txt

blakeli0: Done

.kokoro/release/common.cfg
```diff
@@ -0,0 +1,49 @@
...
+# Download trampoline resources.
+gfile_resources: "/bigstore/cloud-devrel-kokoro-resources/trampoline"
+
+# Use the trampoline script to run in docker.
+build_file: "gapic-generator-java/.kokoro/trampoline.sh"
+
+# Configure the docker image for kokoro-trampoline.
+env_vars: {
+  key: "TRAMPOLINE_IMAGE"
+  value: "gcr.io/cloud-devrel-kokoro-resources/java8"
```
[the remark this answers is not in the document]
blakeli0: Unlike the other repos where we can compile with Java 11 and make the source compatible with Java 8, we have to compile gapic-generator-java in Java 8. It's not a dependency to client libraries, it has to load classes from Java 8 instead of Java 11 during compilation time.

gapic-generator-java/pom.xml
```diff
@@ -365,6 +384,11 @@
       <artifactId>junit</artifactId>
       <version>4.13.2</version>
     </dependency>
+    <dependency>
```
blakeli0, line 387: This is just to pass enforcer plugin for now. We should not import shared-dependency-pom and should declare each dependency explicitly.

burkedavison (approved):

gapic-generator-java/pom.xml
```diff
@@ -15,6 +15,12 @@
     <sonar.working.directory>../target/sonar</sonar.working.directory>
   </properties>
 
+  <parent>
+    <groupId>com.google.api</groupId>
+    <artifactId>gapic-generator-java-parent</artifactId>
```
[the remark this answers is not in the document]
suztomo: <parent>
    <groupId>com.google.cloud</groupId>
    <artifactId>google-cloud-shared-config</artifactId>
    <version>1.5.5</version>
  </parent>

pom.xml
```diff
@@ -38,4 +46,27 @@
       </plugin>
     </plugins>
   </build>
-</project>
\ No newline at end of file
+  
+  <profiles>
```
suztomo, line 50: Follow google-cloud-java's root pom.xml https://github.com/googleapis/google-cloud-java/blob/main/pom.xml#L176
It seems following it already.

.kokoro/release/stage.sh
```diff
@@ -0,0 +1,52 @@
...
+# environment variable. Fusion also lets us to specify this variable.
+#if [[ -n "${AUTORELEASE_PR}" ]]
+#then
+#  mvn nexus-staging:release -B \
+#    -P release-staging-repository \
+#    -DperformRelease=true \
+#    --settings=${MAVEN_SETTINGS_FILE}
+#else
+#  echo "AUTORELEASE_PR is not set. Not releasing."
+#fi
```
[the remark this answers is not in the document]
blakeli0: Done

pom.xml
blakeli0: Made it identical with google-cloud-java's root pom.xml

gapic-generator-java/pom.xml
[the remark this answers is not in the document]
blakeli0: Changed the parent to shared-config

PR #1114, opened 2022-12-08
ci: add showcase testing

burkedavison:


.github/workflows/ci-maven.yaml
```diff
@@ -3,14 +3,18 @@ on:
     branches:
       - main
   pull_request:
+
 name: ci-maven
+
+env:
+  SHOWCASE_VERSION: 0.25.0
```
blakeli0, line 10: Do we expect to manually change this version? A nice to have would be to set up renovate bot to automatically change it.

.github/workflows/ci.yaml
```diff
@@ -41,10 +54,13 @@ jobs:
...
           echo "The old one will disappear after 7 days."
 
+      - name: Java Linter
+        run: bazel --batch run //:google_java_format_verification
+
       - name: Unit Tests
         run: bazel --batch test //:units --noshow_progress --test_output=errors
 
-      - name: Integration Tests
+      - name: Golden Tests
```
blakeli0, line 63: Can we still reference `Integration` somewhere in the name? Some unit tests are golden tests too. Maybe something like `Golden Integration Tests`?

DEVELOPMENT.md
```diff
@@ -2,168 +2,255 @@
...
+the Showcase API which can communicate with a local Showcase server to perform integration tests.
+
+### Requirements
+
+* Install [Go](https://go.dev) in your `PATH`.
+
+### Installing the Server
+
+```shell
+$ GAPIC_SHOWCASE_VERSION=0.25.0
```
blakeli0, line 133: Can we mention that the showcase version should be the latest from `gapic-showcase` repo?

.github/workflows/ci-maven.yaml
burkedavison: Yes - this version will need to be manually changed (right now), and is coupled-but-distinct from [the selected showcase proto version](https://github.com/googleapis/gapic-generator-java/pull/1114/files#diff-5493ff8e9397811510e780de47c57abb70137f1afe85d1519130dc3679d60ce5R126) used to generate the client.

Agreed this server version is a good candidate for renovate-bot, as we'll always be pointing to a release for its associated binary artifacts.

The client version would be a good candidate if there were consistent releases, but there aren't today - and we're currently needing to point to specific commits. I don't think we want renovate bot to be updating the client with each change to head.

- [ ] Use renovate bot to update the showcase server version.

.github/workflows/ci.yaml
burkedavison: If you want to keep the 'Integration' word, we can. I'm trying to move away from its use due to it not having a consistent definition across projects. 
* Client libraries use 'integration test' to refer to large tests with broad scope. 
* These golden tests are small tests with medium scope. 
* The showcase tests are medium tests with medium scope.

Yet all are "integration". At some point it loses its meaning.

DEVELOPMENT.md
burkedavison: - [ ] Update automatically with renovate bot
- [ ] Add documentation to use the latest server version

showcase/google-cloud-jar-parent/pom.xml
```diff
@@ -0,0 +1,287 @@
...
+  <name>Google Cloud JAR Parent</name>
+  <description>
+    The parent for all client library JAR modules in the repository.
+  </description>
+
+  <parent>
+    <groupId>com.google.cloud</groupId>
+    <artifactId>google-cloud-pom-parent</artifactId>
+    <version>0.0.1-SNAPSHOT</version>
+    <relativePath>../google-cloud-pom-parent/pom.xml</relativePath>
```
blakeli0, line 19: What's the benefit of having this pom and the parent pom? Is it just to mimicking the same behavior of [google-cloud-java](https://github.com/googleapis/google-cloud-java/blob/main/google-cloud-jar-parent/pom.xml)?

rules_java_gapic/java_gapic.bzl
```diff
@@ -372,10 +372,11 @@ def java_gapic_library(
         **kwargs
     )
 
-def java_gapic_test(name, runtime_deps, test_classes, **kwargs):
+def java_gapic_test(name, runtime_deps, test_classes, size = "small", **kwargs):
```
blakeli0, line 375: Is this change used anywhere?

.github/workflows/ci-maven.yaml
burkedavison: Based on our in-person discussion, I'm resolving this as a "nice to have but not required" feature.

DEVELOPMENT.md
burkedavison: Based on our in-person discussion, I'm resolving this as a "nice to have but not required" feature.

rules_java_gapic/java_gapic.bzl
burkedavison: It prevents warning messages from appearing in the bazel builds. Otherwise, no.

showcase/google-cloud-jar-parent/pom.xml
burkedavison: Correct. I considered refactoring to reduce the complexity, but thought the similarity to google-cloud-java modules would be preferred.

blakeli0: Personally I feel we don't need these two sub-sub modules as they are introducing more complexity without much benefit, but I would wait for @suztomo's decision. If we do decide to have these two modules, we need at least change the artifact ids and remove most of the dependencies below.

suztomo: I don't see the benefit of google-cloud-jar-parent or google-cloud-pom-parent here. In google-cloud-java repository, it helps to apply the same configuration to many modules.

In this repository, if you need some configuration in pom.xml apply directly to the module.

burkedavison: - [ ] Condense showcase pom structure

PR #1115, opened 2022-12-09
deps: Explicitly declare dependencies instead of using shared-dependencies-bom

blakeli0:


meltsufin (approved):
This is fine, but why not move the dependency management to the root `pom.xml`?

blakeli0:
> This is fine, but why not move the dependency management to the root `pom.xml`?

I thought about it yesterday as well, but the main goal of this PR  is to make sure the Maven build is using correct dependencies so I didn't do it. How to organize dependencies may change. E.g. GAX will become a sibling module so we may not want to declare it in the dependency management section of the root pom

PR #1135, opened 2022-12-13
chore: Skip clirr check

blakeli0:


blakeli0:
Skipping clirr check is required for https://github.com/googleapis/gapic-generator-java/pull/1126, now that we have a jar in maven central

lqiu96 (approved):

lqiu96:
I've seen the `clirr-maven-plugin:2.8:check failed: Invalid byte tag in constant pool: 19` error before, but I don't recall how it was resolved. iirc, last time we also skipped the clirr check -- I'll try and retry the other PR and see if that resolves it.

blakeli0:
> I've seen the `clirr-maven-plugin:2.8:check failed: Invalid byte tag in constant pool: 19` error before, but I don't recall how it was resolved. iirc, last time we also skipped the clirr check -- I'll try and retry the other PR and see if that resolves it.

Thanks, it may help the multi module set up in the future and don't spend a lot of time on it though, since gapic-generator-java is not a library that customer would dependent on directly. Hence we probably don't need clirr check at all for it.

PR #1138, opened 2022-12-14
feat: prototype of Snippetgen Phase 2 samples

alicejli:
WIP

Prototype to generate golden file in cl/479130206
Hardcoded `snippet_config_language.proto` and sample input from cl/459090407

More details in go/snippetgen-java-phase2

Prototype assumes input .json file lives next to protos and has a `java_gapic.bzl` rule. The final input directory structure is still TBD but it is unlikely to be the same as what the prototype shows. I chose to do so for the prototype to demonstrate that it works end to end.


alicejli:
Closing in favor of https://github.com/googleapis/gapic-generator-java/pull/1190 which was the patch made post migration to multi-module generator.

PR #1143, opened 2022-12-16
fix(spring): refactor retry settings to service and method-level nested properties

emmileaf:
This PR goes together with https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1392 and builds on top of commits from https://github.com/googleapis/gapic-generator-java/pull/1140. It:

- Refactors the current setup (method-level and individual settings) to use a nested Retry object provided through properties at both service-level and method-level ([poc](https://github.com/zhumin8/language_poc1/pull/6)).

    * Previously, configuring granular settings per-method :
    ```
    <prefix>.language-service.annotate-text-retry-delay-multiplier=2
    <prefix>.language-service.annotate-text-max-retry-delay=PT0.9S
    ```

    * With this PR, retry settings will be configured through nested properties with the following precedence order: 
        - method-level properties > service-level properties > client library defaults 
        
    ```
    // settings provided at service-level will be applied to all rpc methods

    <prefix>.language-service.retry.retry-delay-multiplier=2
    <prefix>.language-service.retry.max-retry-delay=PT0.9S
    
    // settings provided at method-level will take precedence over service-level ones configured
    // e.g. annotateText's retry delay multiplier will be set to 3, while max retry delay remains at service-level configuration of 0.9s

    <prefix>.language-service.annotate-text-retry.retry-delay-multiplier=3
    ```


src/test/java/com/google/api/generator/spring/composer/goldens/EchoSpringPropertiesFull.golden
```diff
@@ -36,55 +36,17 @@ public class EchoSpringProperties implements CredentialsSupplier {
...
-  private Duration waitTotalTimeout;
-  private Duration blockInitialRpcTimeout;
-  private Double blockRpcTimeoutMultiplier;
-  private Duration blockMaxRpcTimeout;
-  private Duration blockTotalTimeout;
-  private Duration collideNameInitialRpcTimeout;
-  private Double collideNameRpcTimeoutMultiplier;
-  private Duration collideNameMaxRpcTimeout;
-  private Duration collideNameTotalTimeout;
+  private Retry retrySettings;
```
blakeli0, line 39: Do we need `NestedConfigurationProperty` on each of them?

src/test/java/com/google/api/generator/spring/composer/goldens/EchoSpringPropertiesGrpc.golden
```diff
@@ -87,398 +49,11 @@ public class EchoSpringProperties implements CredentialsSupplier {
     this.executorThreadCount = executorThreadCount;
   }
 
-  public Duration getEchoInitialRetryDelay() {
-    return this.echoInitialRetryDelay;
+  public Retry getRetrySettings() {
```
blakeli0, line 52: Since we are set on the name `Retry`, can we rename the field and getter/setters as well to prevent confusion?

src/test/java/com/google/api/generator/spring/composer/goldens/EchoSpringPropertiesFull.golden
emmileaf: That's a good point! I missed this since it didn't directly affect the properties binding, but will add to each field.

src/test/java/com/google/api/generator/spring/composer/goldens/EchoSpringPropertiesGrpc.golden
emmileaf: Do you mean naming the field/getter/setter to something like `retry`/`getRetry`/`setRetry`?

Though the helper class name is `Retry`, I think the rationale behind naming the field `retrySettings` is because it determines the configuration property name, i.e. the following mapping: 
```
private Retry retrySettings;
<prefix>.language-service.retry-settings.total-timeout

private Retry retry;
<prefix>.language-service.retry.total-timeout
```

So having `retry-settings` as part of the property name feels a bit more informative, but I also don't have a strong opinion on this. WDYT?

meltsufin: I think we agreed on `retry-settings` which requires us to name the field `retrySettings`, but I'm flexible to go with just `retry`. @zhumin8 WDYT?

src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java
```diff
@@ -689,77 +620,107 @@ private static MethodDefinition createSettingsBeanMethod(
...
-              return statements;
-            },
-            (String methodName) -> {
-              List<Statement> statements = new ArrayList<>();
-              statements.add(setRetrySettingsForMethod(methodName, settingBuilderVariable));
-              return statements;
-            });
-    bodyStatements.addAll(retrySettings);
+
+    // If service-level properties configured, update retry settings for each method
```
zhumin8, line 624: Perhaps add description of logic between setting service-level vs method-level retries to the javadoc comment of this `EchoSettings` bean?

zhumin8 (commented):
LGTM!
Dropped a small comment and 2 reminder notes below.

src/test/java/com/google/api/generator/spring/composer/goldens/EchoSpringPropertiesGrpc.golden
```diff
@@ -16,55 +16,17 @@ public class EchoSpringProperties implements CredentialsSupplier {
...
-  private Duration waitTotalTimeout;
-  private Duration blockInitialRpcTimeout;
-  private Double blockRpcTimeoutMultiplier;
-  private Duration blockMaxRpcTimeout;
-  private Duration blockTotalTimeout;
-  private Duration collideNameInitialRpcTimeout;
-  private Double collideNameRpcTimeoutMultiplier;
-  private Duration collideNameMaxRpcTimeout;
-  private Duration collideNameTotalTimeout;
+  @NestedConfigurationProperty private Retry retrySettings;
```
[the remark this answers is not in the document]
emmileaf: Yes - planning to revamp that javadoc PR along with this change. Thanks for the reminder!

src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java
emmileaf: Makes sense, updated! 

(FYI @diegomarquezp I only changed the part relevant to retry settings in https://github.com/googleapis/gapic-generator-java/pull/1143/commits/4d55a7a640248eeebf61cbde3c1300bf54d0d9c9 , hopefully it won't conflict too much with your updates for `ServiceClient` => `ServiceSettings` bean)

src/test/java/com/google/api/generator/spring/composer/goldens/EchoSpringPropertiesGrpc.golden
blakeli0: I see, I missed that we already agreed on `rerty-settings`. The reason I'm thinking `retry` is that [Pubsub](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/b46332b787032a91055a5ba25e6d001de6ac28fe/spring-cloud-gcp-pubsub/src/main/java/com/google/cloud/spring/pubsub/core/PubSubConfiguration.java#L617) already uses this name, the property looks like this `spring.cloud.gcp.pubsub.subscriber.retry.initial-rpc-timeout-seconds`. If we don't have a strong preference of one or the other, maybe we can keep it consistent?

src/main/java/com/google/api/generator/spring/composer/SpringPropertiesClassComposer.java
```diff
@@ -331,7 +307,18 @@ private static Map<String, TypeNode> createDynamicTypes(Service service, String
                 .setName(Utils.getServicePropertiesClassName(service))
                 .setPakkage(packageName)
                 .build());
+
+    // TODO: This should move to static types once class is added into spring-cloud-gcp-core
```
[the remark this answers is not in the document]
emmileaf: @zhumin8 On a second thought, do you think it would be easier to leave this as a dynamic type in generator code until the shared classes are published and available in maven central?  

Let me know if there's anything I'm missing here - currently, this use of dynamic type unblocks code generation, and the generated code is able to compile against a locally installed snapshot jar of spring-cloud-gcp.

src/test/java/com/google/api/generator/spring/composer/goldens/EchoSpringPropertiesGrpc.golden
zhumin8: I don't have strong option on either. We can go with `retry` if that's more consistent with the manual modules.

emmileaf: SGTM - updated to `retry`! Will follow-up with corresponding changes in the unit testing PR as well.

src/main/java/com/google/api/generator/spring/composer/SpringPropertiesClassComposer.java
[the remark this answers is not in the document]
zhumin8: I think we can leave as dynamic type temporarily and unblock. 
But we do need to switch to local repo for spring-cloud-gcp dependencies eventually to develop on the snapshot version.

> currently, this use of dynamic type unblocks code generation, and the generated code is able to compile against a locally installed snapshot jar of spring-cloud-gcp.

yes. But same argument on adding concrete types for other spring denpendencies, it's safer to have the imported class confirmed in the generator than validate via compile in generated module.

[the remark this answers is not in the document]
emmileaf: Gotcha, that makes sense. Thank you for the explanation!

meltsufin (approved):

PR #1163, opened 2022-12-19
build: use bazelisk avoid unexpected Bazel version upgrades

suztomo:
" ci / build (8) " has been failing in https://github.com/googleapis/gapic-generator-java/pull/1156. Creating an empty pull request to see the build failures.

This PR changes:

- The unit tests in Bazel and "coverage" on it have been removed. @blakeli0 intends to remove it in another PR anyway.
- GitHub Actions to use bazelisk command that keeps the Bazel version stable.
- The client library generation job is separated from the build job.

As per https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md, ubuntu-latest has bazelisk installed.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -110,19 +110,6 @@ jobs:
         ./gradlew clean build publishToMavenLocal sourcesJar allJars
         popd
 
-  coverage:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-go@v3
-        with:
-          java-version: 8
-      - run: java -version
-
-      - name: Generate Code Coverage Report
-        # Run only test targets, and not golden_update targets.
-        run: bazelisk coverage //:units --combined_report=lcov
-
   license-header:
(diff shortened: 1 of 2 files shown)


suztomo:
Bazel version 6.0.0 is suspicious.

Screen Shot 2022-12-19 at 4 21 58 PM


Indeed, I'm able to reproduce the issue in Bazel 6.0.0.

https://github.com/bazelbuild/bazel/releases 6.0.0 was released today.

.github/workflows/ci.yaml
```diff
@@ -75,20 +52,63 @@ jobs:
           retention-days: 5
 
       - name: Java Linter
-        run: bazel --batch run //:google_java_format_verification
+        run: bazelisk --batch run //:google_java_format_verification
```
blakeli0, line 55: I think we can remove Java linter as well if it's already included in the Maven build.

blakeli0 (approved):

.github/workflows/ci.yaml
suztomo: Removed.

PR #1192, opened 2022-12-21
ci: downstream testing for gax-java module

suztomo:
Moving gax-java's https://github.com/googleapis/gax-java/blob/main/.github/workflows/downstream-native-image.yaml to this gapic-generator-java repository.


.github/workflows/downstream-native-image.yaml
```diff
@@ -0,0 +1,41 @@
...
+  graalvm:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        graalvm: [22.3.0]
+        java: [11, 17]
+        repo:
+        # GAPIC library that doesn't use a real GCP project in integration tests
+        - orgpolicy
```
suztomo, line 22: @mpeddada1 This has been using the old orgpolicy repository as the canary. If you're good, I think it's better to replace this with the showcase module in this repository. With that, it won't need the pom manipulation in downstream-client-library-check.sh because showcase does not rely on the shared dependencies BOM.

mpeddada1: @suztomo Adding testing for the showcase module is a good idea! I'm curious, how big of a change would it be to replace the canary library with it?

mpeddada1 (approved):

.github/workflows/downstream-native-image.yaml
suztomo:
```
cd showcase
mvn test -P native
```

burkedavison: Unit tests are disabled for showcase due to issues with the generated REST unit tests. If they're needed for native testing, we will need to explicitly ignore them.

suztomo: Unit tests are excluded from native test already.

mpeddada1: +1 Native image tests opt-in classes based on the `IT*` and `*ClientTest` naming conventions.

PR #1199, opened 2022-12-27
chore: Changes for release job

ddixit14:
@blakeli0 @burkedavison I've opened a new PR. This does not include any deletions.


burkedavison (approved):

.kokoro/release/common.cfg
```diff
@@ -9,7 +9,7 @@ build_file: "gapic-generator-java/.kokoro/trampoline.sh"
 # Configure the docker image for kokoro-trampoline.
 env_vars: {
   key: "TRAMPOLINE_IMAGE"
-  value: "gcr.io/cloud-devrel-kokoro-resources/java8"
+  value: "gcr.io/cloud-devrel-kokoro-resources/java11"
```
blakeli0, line 12: Are we going to have a different release image for `gapic-generator-java` submodule?

.github/workflows/ci-maven.yaml
```diff
@@ -75,22 +75,21 @@ jobs:
...
-              -Dcheckstyle.skip -Dfmt.skip -DskipTests
+          mvn -V -B -ntp clean install -DskipTests
       - uses: actions/setup-java@v3
         with:
           java-version: 8
           distribution: temurin
           cache: maven
       - run: java -version
-      - name: Run test only for gapic-generator-java using Java 8
+      - name: Run tests in Java 8 with the source compiled in Java 11 for gapic-generator-java
```
ddixit14, line 88: @blakeli0 @burkedavison @meltsufin This test will now ensure that gapic-generator-java submodule is compiled with Java 11 and the sources+binaries generated are Java 8 compatible. Since we are checking this here only, there shouldn't be a need to handle this in the release job. 
Am I thinking clearly here?  
@suztomo ensured this kind of compatibility in gax (https://github.com/googleapis/gax-java/blob/main/.kokoro/build.sh) when it was a split-repo.

.github/workflows/ci-maven.yaml
```diff
@@ -75,22 +75,21 @@ jobs:
...
         shell: bash
         run: |
-          mvn -B -ntp verify --projects 'gapic-generator-java' \
-              -Dcheckstyle.skip -Dfmt.skip
+          mvn -V -B -ntp surefire:test --projects 'gapic-generator-java'
+          mvn -B -ntp clirr:check --projects 'gapic-generator-java'
```
blakeli0, line 92: We can remove this clirr check for `gapic-generator-java`. We should add clirr checks for all other modules.

.kokoro/release/common.cfg
blakeli0: Decided that we are going to use a single Java11 image. But we should add release flag to gapic-generator-java or the parent if possible, to guarantee the compatibility of Java 8.

meltsufin (approved):

blakeli0 (approved):

PR #1207, opened 2023-01-03
fix: add name to ConditionalOnMissingBean when creating bean

zhumin8:
Add name to `@ConditionalOnMissingBean` annotation when creating `TransportChannelProvider` bean so that it can be picked up with Qualifier in serviceSettings bean creation that follows.
This was missed in original [commit](https://github.com/googleapis/gapic-generator-java/commit/0643085e8e837f23d11b3bc18bded7e1551233ba) adding this bean and should fix error seen in https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1407#issuecomment-1370058527

### DIFF
--- a/src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java
+++ b/src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java
@@ -341,7 +341,7 @@ public class SpringAutoConfigClassComposer implements ClassComposer {
         AssignmentExpr.builder()
             .setVariableExpr(
                 VariableExpr.withVariable(
-                    Variable.builder().setName("value").setType(TypeNode.STRING).build()))
+                    Variable.builder().setName("name").setType(TypeNode.STRING).build()))
             .setValueExpr(ValueExpr.withValue(StringObjectValue.withValue(methodName)))
             .build();
     AnnotationNode conditionalOnMissingBean =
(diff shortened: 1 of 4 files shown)


meltsufin (approved):

src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java
```diff
@@ -330,19 +330,26 @@ private static List<AnnotationNode> createClassAnnotations(
...
             .setMethodName("defaultTransportChannelProvider")
             .setStaticReferenceType(types.get("ServiceSettings"))
             .setReturnType(STATIC_TYPES.get("TransportChannelProvider"))
             .build();
 
+    AssignmentExpr nameStringAssignmentExpr =
+        AssignmentExpr.builder()
+            .setVariableExpr(
+                VariableExpr.withVariable(
+                    Variable.builder().setName("value").setType(TypeNode.STRING).build()))
```
emmileaf, line 344: Should this be `name` (e.g. `name = "defaultEchoTransportChannelProvider"`) instead of `value`?

zhumin8: Yes! Thank you!

emmileaf (approved):

diegomarquezp (approved):
Thanks!

PR #1123, opened 2022-12-09
fix(spring): Replace global properties with CredentialsProvider bean

zhumin8:
As discussed off-line, instead of depending on a new `global-properties` module, use `CredentialsProvider` bean from `spring-cloud-gcp-autoconfigure` instead.
- Replace dependency from `core` to `autoconfigure` module.
- Revert #1071 and implement changes shown in [poc](https://github.com/zhumin8/language_poc1/pull/7)
- Minor code cleanups and util method extracted from `SpringPropertiesClassComposer.java`. 

Note: #1071 and https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1308 will be obsolete after this change.
Also, conflicts needs to be resolved between this pr and #1110 depending on merging order.

PR #1208, opened 2023-01-04
fix(spring): add AutoConfigureAfter annotation

emmileaf:
This PR adds `AutoConfigureAfter(GcpContextAutoConfiguration.class)` annotation to enforce configuration order. It also makes two related changes to enable this annotation:
- Switches dependency from `spring-cloud-gcp-core` to `spring-cloud-gcp-autoconfigure` (ref: [#1123](https://github.com/googleapis/gapic-generator-java/pull/1123)), and bumps version to latest
  * Corresponding generator script update needed: [TODO]
- Fixes `ImportWriter` to account for types introduced by annotation parameters - this change is made in d59b40c, and I plan to copy it to the main branch of `gapic-generator-java` following completion of repo restructure.


src/main/java/com/google/api/generator/engine/writer/ImportWriterVisitor.java
```diff
@@ -166,6 +166,9 @@ public void visit(ScopeNode scope) {
   @Override
   public void visit(AnnotationNode annotation) {
     annotation.type().accept(this);
+    if (annotation.descriptionExprs() != null) {
```
zhumin8, line 169: Ah, Good catch! I guess we missed this and just got lucky with 
```
@ConditionalOnClass(EchoClient.class)
@EnableConfigurationProperties(EchoSpringProperties.class)
```

zhumin8 (approved):
LGTM.

diegomarquezp (approved):

Issue #1117, opened 2022-12-09
Skip generating REST methods if the RPC does not have HTTP annotation.


blakeli0:
We should not generate REST methods if the RPC does not have HTTP annotation. See this [proto](https://github.com/googleapis/googleapis/blob/e69c47f438eb55caec51716fbc2b8a42313ca024/google/cloud/mediatranslation/v1beta1/media_translation.proto#L39) and the [proto](https://github.com/googleapis/googleapis/blob/e69c47f438eb55caec51716fbc2b8a42313ca024/google/cloud/language/v1/language_service.proto#L38) that has it.
In this case, we are already [not generating REST methods](https://github.com/googleapis/google-cloud-java/blob/main/java-mediatranslation/google-cloud-mediatranslation/src/main/java/com/google/cloud/mediatranslation/v1beta1/stub/HttpJsonSpeechTranslationServiceStub.java) because it's BIDI streaming, but we need to use the HTTP annotation as the filter instead of using RPC type.

In addition, we should not generating the client at all if there are no RPCs with HTTP annotation. See [go/actools-diregapic-parallelization-sync](http://go/actools-diregapic-parallelization-sync) for details.

There could be multiple reasons an RPC does not have HTTP annotation:

- The RPC is not supposed to be supported by REST. e.g. BIDI streaming methods
- The RPC is not yet supported by the backend service, but could be supported in the future
- The proto is not properly configured.

lqiu96:
Closing this issue as we've merged in both Part 1 and Part 2 of this issue.

PR #1211, opened 2023-01-04
feat: Add callable getters for non-eligible or non-enabled REST methods

lqiu96:
This is a PR for Part 1 of #1117 (Override method with clearer exception messages for non-eligible and non-enabled Service RPCs). Opening this while I look at other possible approaches: 

Other approaches looked at/ to revisit:
- Override createCallableClassMembers to return Map<String, Expr> (Return either VariableExpr or ThrowExpr)
- Create a duplicate createCallableClassMembers (i.e. createInvalidCallableClassMembers) that copies functionality of createCallableClassMembers but returns Map<String, ThrowExpr>

Both approaches above had an issue when setting the return type for the callableGetter. ThrowExpr's type is always set as `UnsupportedOperationException` but the MethodDefinition's return type is not (i.e. for Streams it would be ServerStreamCallable or ClientStreamCallable/ Unary is UnaryCallable vs. Operation, etc.). Would need potentially need another mapping between callableName and method return type for ThrowExprs.

This PR's implementation is copied from: https://github.com/googleapis/gapic-generator-java/blob/8c5e17ba325b7711f9ba9501992ab48e736ffc18/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceStubClassComposer.java#L284-L302
- Use getCallableType(protoMethod) to get the correct return type for the RPC
- ThrowExpr's return type is set to `UnsupportedOperationException`

### DIFF
--- a/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
+++ b/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
@@ -228,7 +228,9 @@ public abstract class AbstractTransportServiceStubClassComposer implements Class
     return String.format("Not implemented: %s()", callableName);
   }
 
-  protected abstract boolean isSupportedMethod(Method method);
+  protected boolean isSupportedMethod(Method method) {
+    return true;
+  }
 
   protected abstract Statement createMethodDescriptorVariableDecl(
       Service service,
--- a/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
+++ b/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
@@ -83,18 +83,18 @@ import javax.annotation.Nullable;
 public abstract class AbstractTransportServiceStubClassComposer implements ClassComposer {
   private static final Statement EMPTY_LINE_STATEMENT = EmptyLineStatement.create();
 
-  protected static final String METHOD_DESCRIPTOR_NAME_PATTERN = "%sMethodDescriptor";
-  protected static final String PAGED_RESPONSE_TYPE_NAME_PATTERN = "%sPagedResponse";
-  protected static final String PAGED_CALLABLE_CLASS_MEMBER_PATTERN = "%sPagedCallable";
+  private static final String METHOD_DESCRIPTOR_NAME_PATTERN = "%sMethodDescriptor";
+  private static final String PAGED_RESPONSE_TYPE_NAME_PATTERN = "%sPagedResponse";
+  private static final String PAGED_CALLABLE_CLASS_MEMBER_PATTERN = "%sPagedCallable";
 
-  protected static final String BACKGROUND_RESOURCES_MEMBER_NAME = "backgroundResources";
-  protected static final String CALLABLE_NAME = "Callable";
-  protected static final String CALLABLE_FACTORY_MEMBER_NAME = "callableFactory";
+  private static final String BACKGROUND_RESOURCES_MEMBER_NAME = "backgroundResources";
+  private static final String CALLABLE_NAME = "Callable";
+  private static final String CALLABLE_FACTORY_MEMBER_NAME = "callableFactory";
   protected static final String CALLABLE_CLASS_MEMBER_PATTERN = "%sCallable";
-  protected static final String OPERATION_CALLABLE_CLASS_MEMBER_PATTERN = "%sOperationCallable";
-  protected static final String OPERATION_CALLABLE_NAME = "OperationCallable";
+  private static final String OPERATION_CALLABLE_CLASS_MEMBER_PATTERN = "%sOperationCallable";
+  private static final String OPERATION_CALLABLE_NAME = "OperationCallable";
   // private static final String OPERATIONS_STUB_MEMBER_NAME = "operationsStub";
-  protected static final String PAGED_CALLABLE_NAME = "PagedCallable";
+  private static final String PAGED_CALLABLE_NAME = "PagedCallable";
 
   protected static final TypeStore FIXED_TYPESTORE = createStaticTypes();
 
--- a/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
+++ b/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
@@ -228,9 +228,7 @@ public abstract class AbstractTransportServiceStubClassComposer implements Class
     return String.format("Not implemented: %s()", callableName);
   }
 
-  protected boolean isSupportedMethod(Method method) {
-    return true;
-  }
+  protected abstract boolean isSupportedMethod(Method method);
 
   protected abstract Statement createMethodDescriptorVariableDecl(
       Service service,
--- a/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
+++ b/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
@@ -225,10 +225,6 @@ public abstract class AbstractTransportServiceStubClassComposer implements Class
     return GapicClass.create(kind, classDef);
   }
 
-  protected String getUnsupportedOperationExceptionReason(String callableName, Method protoMethod) {
-    return String.format("Not implemented: %s()", callableName);
-  }
-
   protected Transport getTransport() {
     return Transport.GRPC;
   }
--- a/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
+++ b/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
@@ -59,7 +59,6 @@ import com.google.api.generator.gapic.model.GapicServiceConfig;
 import com.google.api.generator.gapic.model.Message;
 import com.google.api.generator.gapic.model.Method;
 import com.google.api.generator.gapic.model.Service;
-import com.google.api.generator.gapic.model.Transport;
 import com.google.api.generator.gapic.utils.JavaStyle;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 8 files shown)


gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
```diff
@@ -204,20 +220,54 @@ public GapicClass generate(GapicContext context, Service service) {
...
+    if (protoMethod.stream() == Method.Stream.BIDI
+        || protoMethod.stream() == Method.Stream.CLIENT) {
+      return String.format(
+          "Not implemented: %s(). %s is not implemented for %s",
+          callableName, protoMethod.stream(), Transport.REST);
+    } else if (protoMethod.httpBindings() == null) {
+      return String.format(
+          "Not implemented: %s(). RPC is not enabled for %s", callableName, Transport.REST);
+    } else {
+      return String.format("Not implemented: %s()", callableName);
```
blakeli0, line 239: What could be in this else branch? 
Also I think a generic error message like `REST transport is not supported for this method yet` should be good enough, it's better than a message that was too generic before. If we really want to tell the differences, we probably want to change the first message to `Not supported:...` instead of `Not implemented:...` since we would never implement REST for BIDI and CLIENT streaming.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
```diff
@@ -190,6 +192,20 @@ public GapicClass generate(GapicContext context, Service service) {
             messageTypes,
             context.restNumericEnumsEnabled());
 
+    List<MethodDefinition> methodDefinitions =
```
blakeli0, line 195: Can we move this HttpJson/REST specific logic to `HttpJsonServiceStubClassComposer` instead of in this abstract composer?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
```diff
@@ -190,6 +192,20 @@ public GapicClass generate(GapicContext context, Service service) {
...
+        createClassMethods(
+            context,
+            service,
+            typeStore,
+            classMemberVarExprs,
+            callableClassMemberVarExprs,
+            protoMethodNameToDescriptorVarExprs,
+            classStatements);
+    methodDefinitions.addAll(createInvalidClassMethods(service));
```
blakeli0, line 204: Can we move this HttpJson/REST specific logic to HttpJsonServiceStubClassComposer instead of in this abstract composer?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
lqiu96: > What could be in this else branch?

Nothing as of now -- I just intended to give it a default message. I think `REST transport is not supported for this method yet` sounds good. On second thought, perhaps this function can exist in `HttpJsonTransportServiceStubClassComposer`.

> If we really want to tell the differences, we probably want to change the first message to Not supported:... instead of Not implemented:... since we would never implement REST for BIDI and CLIENT streaming.

Yep, I think `Not implemented: ...` makes more sense.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
lqiu96: I wasn't sure which part you were highlighting. Did you mean the entire MethodDefinition logic or just the `createInvalidClassMethods` function?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
```diff
@@ -121,6 +122,21 @@ protected boolean isSupportedMethod(Method method) {
...
 
+  protected String getUnsupportedOperationExceptionReason(String callableName, Method protoMethod) {
+    if (protoMethod.stream() == Method.Stream.BIDI
+        || protoMethod.stream() == Method.Stream.CLIENT) {
+      return String.format(
+          "Not supported: %s(). %s streaming is not implemented for %s",
+          callableName, protoMethod.stream(), Transport.REST);
+    } else if (protoMethod.httpBindings() == null) {
+      return String.format(
+          "Not implemented: %s(). %s transport is not supported for this method yet",
```
lqiu96, line 133: change to implemented

lqiu96:
@blakeli0 Can you take another look whenever you get a chance?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/model/Method.java
```diff
@@ -99,6 +99,14 @@ public boolean isOperationPollingMethod() {
     return operationPollingMethod();
   }
 
+  public boolean isSupportedByTransport(Transport transport) {
```
blakeli0, line 102: I know this may already be tested implicitly in some golden tests, since we refactored it to `Method` class, can we add a couple of true unit tests to `MethodTest`?

lqiu96: Added some tests for this in the units.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
```diff
@@ -1248,4 +1249,56 @@ protected List<Statement> createTypeRegistry(Service service) {
...
+            protoMethodNameToDescriptorVarExprs,
+            classStatements));
+    javaMethods.addAll(createInvalidClassMethods(service));
+    return javaMethods;
+  }
+
+  private List<MethodDefinition> createInvalidClassMethods(Service service) {
+    List<MethodDefinition> methodDefinitions = new ArrayList<>();
+    for (Method protoMethod : service.methods()) {
+      if (!protoMethod.isSupportedByTransport(getTransport())) {
```
blakeli0, line 1279: Usually for scenarios like this we can do something like 
```
if (protoMethod.isSupportedByTransport(getTransport())) {
  continue;
}
//do other logic...
```
to have one less indent so that the code is more readable.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
```diff
@@ -82,18 +83,18 @@
...
-  private static final String BACKGROUND_RESOURCES_MEMBER_NAME = "backgroundResources";
-  private static final String CALLABLE_NAME = "Callable";
-  private static final String CALLABLE_FACTORY_MEMBER_NAME = "callableFactory";
-  private static final String CALLABLE_CLASS_MEMBER_PATTERN = "%sCallable";
-  private static final String OPERATION_CALLABLE_CLASS_MEMBER_PATTERN = "%sOperationCallable";
-  private static final String OPERATION_CALLABLE_NAME = "OperationCallable";
+  protected static final String BACKGROUND_RESOURCES_MEMBER_NAME = "backgroundResources";
+  protected static final String CALLABLE_NAME = "Callable";
+  protected static final String CALLABLE_FACTORY_MEMBER_NAME = "callableFactory";
+  protected static final String CALLABLE_CLASS_MEMBER_PATTERN = "%sCallable";
```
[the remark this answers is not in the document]
blakeli0: It's better to just make one String protected since that's the only one you would need in this PR.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
```diff
@@ -1248,4 +1249,56 @@ protected List<Statement> createTypeRegistry(Service service) {
...
+                .setName(callableName)
+                .setReturnType(getCallableType(protoMethod))
+                .setBody(
+                    Arrays.asList(
+                        ExprStatement.withExpr(
+                            ThrowExpr.builder()
+                                .setType(FIXED_TYPESTORE.get("UnsupportedOperationException"))
+                                .setMessageExpr(
+                                    String.format(
+                                        "Not implemented: %s(). %s transport is not implemented for this method yet",
```
blakeli0, line 1296: Now that we only have one error message, we may want to be a little more careful with the message. `transport is not implemented for this method yet` sounds like we didn't get a chance to implement it yet which is not the case sometimes(for BIDI and CLIENT streaming). I don't have a better message either, but let's give it a little more thought, maybe we can look at what the message is for other languages as well.

gapic-generator-java/src/test/java/com/google/api/generator/gapic/model/MethodTest.java
```diff
@@ -112,4 +112,59 @@ public void shouldSetParamsExtractor_shouldReturnFalseIfHasNoHttpBindingsAndNoRo
...
+    assertThat(methodBiDiStreaming.isSupportedByTransport(Transport.GRPC)).isTrue();
+    Method methodNoStreamingNoHttpBindings =
+        METHOD.toBuilder().setStream(Method.Stream.NONE).build();
+    assertThat(methodNoStreamingNoHttpBindings.isSupportedByTransport(Transport.GRPC)).isTrue();
+    Method methodBiDiStreamingNoHttpBindings =
+        METHOD.toBuilder().setStream(Method.Stream.BIDI).build();
+    assertThat(methodBiDiStreamingNoHttpBindings.isSupportedByTransport(Transport.GRPC)).isTrue();
+  }
+
+  @Test(expected = IllegalArgumentException.class)
```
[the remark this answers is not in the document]
lqiu96: Makes sense, will update!

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
lqiu96: Gotcha, will update.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
lqiu96: Yeah, I'll try to think of some better wording to possibly capture both cases. It's initially why I wanted to capture two different error messages since they're technically two different reasons why we didn't generate the resulting callable/ code.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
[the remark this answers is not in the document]
lqiu96: Sounds good.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
lqiu96: Change it to `Not implemented: {callableName}(). REST transport is not implemented for this method yet.`

blakeli0 (approved):

lqiu96:
PR comments resolved. Updated message to `Not implemented: {callableName}(). REST transport is not implemented for this method yet.`. Adding automerge for this PR

PR #1212, opened 2023-01-04
fix(spring): limit retry settings configuration by method type

emmileaf:
This PR is a patch for the `RetrySettings` configuration issue discovered in https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1407#issuecomment-1363280649. It excludes streaming and LRO methods from retry configuration through spring properties for now. 
- Ref: [UnaryCallSettings](https://github.com/googleapis/gax-java/blob/main/gax/src/main/java/com/google/api/gax/rpc/UnaryCallSettings.java) vs. [StreamingCallSettings](https://github.com/googleapis/gax-java/blob/main/gax/src/main/java/com/google/api/gax/rpc/StreamingCallSettings.java); how gapic-generator [configures default retry settings per-method](https://github.com/googleapis/gapic-generator-java/blob/0fb58e0b459a5dea109afd8b41da57e48728deb3/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceStubSettingsClassComposer.java#L1447-L1497)
 - Future method type investigation/support needed to align with gapic generator logic: 
      - LRO methods (through [OperationCallSettings](https://github.com/googleapis/gax-java/blob/main/gax/src/main/java/com/google/api/gax/rpc/OperationCallSettings.java))
      - Server streaming methods (through [ServerStreamingCallSettings](https://github.com/googleapis/gax-java/blob/main/gax/src/main/java/com/google/api/gax/rpc/ServerStreamingCallSettings.java))

### DIFF
--- a/src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java
+++ b/src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java
@@ -659,10 +659,13 @@ public class SpringAutoConfigClassComposer implements ClassComposer {
     List<Statement> updateRetrySettingsStatementBody = new ArrayList<>();
 
     for (Method method : service.methods()) {
-      List<Statement> updateMethodWithServiceRetryStatments =
+      if (!method.stream().equals(Method.Stream.NONE)) {
+        continue;
+      }
+      List<Statement> updateMethodWithServiceRetryStatements =
           createUpdateRetrySettingsStatements(
               method.name(), settingBuilderVariable, serviceRetryPropertiesVar, types);
-      updateRetrySettingsStatementBody.addAll(updateMethodWithServiceRetryStatments);
+      updateRetrySettingsStatementBody.addAll(updateMethodWithServiceRetryStatements);
       updateRetrySettingsStatementBody.add(EMPTY_LINE_STATEMENT);
     }
 
@@ -680,6 +683,9 @@ public class SpringAutoConfigClassComposer implements ClassComposer {
 
     // If-blocks to update with method-level properties
     for (Method method : service.methods()) {
+      if (!method.stream().equals(Method.Stream.NONE)) {
+        continue;
+      }
       String methodName = CaseFormat.UPPER_CAMEL.to(CaseFormat.LOWER_CAMEL, method.name());
 
       Variable methodRetryPropertiesVar =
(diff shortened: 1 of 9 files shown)


src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java
```diff
@@ -659,10 +659,13 @@ private static MethodDefinition createSettingsBeanMethod(
     List<Statement> updateRetrySettingsStatementBody = new ArrayList<>();
 
     for (Method method : service.methods()) {
-      List<Statement> updateMethodWithServiceRetryStatments =
+      if (!method.stream().equals(Method.Stream.NONE) || method.hasLro()) {
```
diegomarquezp, line 662: I think it would make sense to filter the methods in the foreach declaration (e.g. `for (Method method : Utils.getFilteredMethods(service.methods))` to centralize/dereplicate the filtering logic (could also be done by adding a convenience method in `Service` but may be overkill)

diegomarquezp: Ah, I missed this was a draft. Hopefully it was still a useful catch!

emmileaf: Makes sense, given we expect this to be changed in the future. Updated and moved this logic to utils.

diegomarquezp (approved):
LGTM!

src/main/java/com/google/api/generator/spring/utils/Utils.java
```diff
@@ -81,6 +83,14 @@ public static String getSpringPropertyPrefix(String packageName, String serviceN
     return packageName + "." + CaseFormat.UPPER_CAMEL.to(CaseFormat.LOWER_HYPHEN, serviceName);
   }
 
+  public static List<Method> getMethodsForRetryConfiguration(Service service) {
+    // Returns list of methods with retry configuration support
+    // This currently excludes streaming and LRO methods
+    return service.methods().stream()
```
blakeli0, line 89: Just a note that we need to revisit this logic before GA, as we discussed, BIDI and Client streaming do not support retry, but Server streaming and LRO should support it.

blakeli0 (approved):

src/main/java/com/google/api/generator/spring/utils/Utils.java
emmileaf: Noted - thank you!

PR #1220, opened 2023-01-06
chore: updating release-please versions in WORKSPACE

ddixit14:


WORKSPACE
```diff
@@ -64,7 +64,9 @@ maven_install(
     repositories = ["https://repo.maven.apache.org/maven2/"],
 )
 
+<!--- {x-version-update-start:gapic-generator-java:released} -->
```
blakeli0, line 67: This comment style is for xmls only and it may not work for WORKSPACE

ddixit14: It works for a README.md file in another repo, so let's see if it works here as well. And this annotation is not for inline version updates like we do in pom files, this is a block update.

ddixit14: It gave an error 
```
(17:48:47) ERROR: /home/runner/work/gapic-generator-java/gapic-generator-java/WORKSPACE:67:1: syntax error at '<': expected expression
(17:48:47) ERROR: /home/runner/work/gapic-generator-java/gapic-generator-java/WORKSPACE:67:2: invalid character: '!'
(17:48:47) ERROR: /home/runner/work/gapic-generator-java/gapic-generator-java/WORKSPACE:6[9](https://github.com/googleapis/gapic-generator-java/actions/runs/3857289723/jobs/6574513783#step:7:10):2: invalid character: '!'
```

I've reverted it back to previous annotations. Let's see if these annotations work with specifying "extraFiles" in the yaml config.

blakeli0 (approved):

PR #1221, opened 2023-01-06
fix: Handle cancel in ReleasingClientCall and rethrow the exception in start

mutianf:
We got a customer issue and the stacktrace didn't show us where the call was cancelled (b/263968439)

When trying to reproduce this error, we noticed some inconsistent behaviors when we cancel the stream in onStart()

```
client.readRowsCallable().call(Query.create("t"), new ResponseObserver<Row>() {
    @Override
    public void onStart(StreamController streamController) {
        streamController.cancel();
    }
    @Override
    public void onResponse(Row row) {
    }
    @Override
    public void onError(Throwable throwable) {
    }
    @Override
    public void onComplete() {
    }
});
```
When the above code is called, we got exception:

```
Caused by: java.util.concurrent.CancellationException: User cancelled stream
	at com.google.api.gax.rpc.ServerStreamingAttemptCallable.onCancel(ServerStreamingAttemptCallable.java:309)
```

However, if we make another call before it:

```
Iterator<Row> stream = client.readRowsCallable().call(Query.create("t")).iterator();
client.readRowsCallable().call(Query.create("t"), new ResponseObserver<Row>() {
    @Override
    public void onStart(StreamController streamController) {
        streamController.cancel();
    }
   ...
}
```

We got exception:

```
java.lang.IllegalStateException: Not started
	at com.google.common.base.Preconditions.checkState(Preconditions.java:502)
	at io.grpc.internal.ClientCallImpl.sendMessageInternal(ClientCallImpl.java:511)
	at io.grpc.internal.ClientCallImpl.sendMessage(ClientCallImpl.java:504)
```
Which is very difficult to debug.

I think there are 2 problems in the ReleasingClientCall:

1. It's not tracking cancellation status: After the call is cancelled, it continued to go into start() logic and eventually will call `ClientCallImpl#startInternal` https://github.com/grpc/grpc-java/blob/v1.51.1/core/src/main/java/io/grpc/internal/ClientCallImpl.java#L199, where the `cancelCalled` will be true and the state check will fail. In this case we won't see the actual stacktrace that cancelled the stream.
2. It's not rethrowing the error in onStart(): The `IllegalStateException` thrown from `ClientCallImpl#startInternal` will be caught, `GrpcDirectStreamController` will move on to `ClientCallImpl#sendMessage` and eventually fail when checking if the stream is started https://github.com/grpc/grpc-java/blob/v1.51.1/core/src/main/java/io/grpc/internal/ClientCallImpl.java#L514.

### DIFF
--- a/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
+++ b/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
@@ -519,7 +519,7 @@ class ChannelPool extends ManagedChannel {
 
   /** ClientCall wrapper that makes sure to decrement the outstanding RPC count on completion. */
   static class ReleasingClientCall<ReqT, RespT> extends SimpleForwardingClientCall<ReqT, RespT> {
-    private CancellationException cancellationException;
+    @Nullable private CancellationException cancellationException;
     final Entry entry;
 
     public ReleasingClientCall(ClientCall<ReqT, RespT> delegate, Entry entry) {
--- a/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
+++ b/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
@@ -45,6 +45,7 @@ import io.grpc.Status;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.CancellationException;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
@@ -53,6 +54,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import javax.annotation.Nullable;
 import org.threeten.bp.Duration;
 
 /**
@@ -517,6 +519,7 @@ class ChannelPool extends ManagedChannel {
 
   /** ClientCall wrapper that makes sure to decrement the outstanding RPC count on completion. */
   static class ReleasingClientCall<ReqT, RespT> extends SimpleForwardingClientCall<ReqT, RespT> {
+    private CancellationException cancellationException;
     final Entry entry;
 
     public ReleasingClientCall(ClientCall<ReqT, RespT> delegate, Entry entry) {
@@ -526,6 +529,9 @@ class ChannelPool extends ManagedChannel {
 
     @Override
     public void start(Listener<RespT> responseListener, Metadata headers) {
+      if (cancellationException != null) {
+        throw new IllegalStateException("Call is already cancelled", cancellationException);
+      }
       try {
         super.start(
             new SimpleForwardingClientCallListener<RespT>(responseListener) {
@@ -542,7 +548,14 @@ class ChannelPool extends ManagedChannel {
       } catch (Exception e) {
         // In case start failed, make sure to release
         entry.release();
+        throw e;
       }
+    }
+
+    @Override
+    public void cancel(@Nullable String message, @Nullable Throwable cause) {
+      this.cancellationException = new CancellationException(message);
+      super.cancel(message, cause);
     }
   }
 }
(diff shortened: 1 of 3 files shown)


gax-java/gax-grpc/src/test/java/com/google/api/gax/grpc/ChannelPoolTest.java
```diff
@@ -595,4 +603,50 @@ public void removedActiveChannelsAreShutdown() throws Exception {
...
+            .setTransportChannel(GrpcTransportChannel.create(channelPool))
+            .setDefaultCallContext(GrpcCallContext.of(channelPool, CallOptions.DEFAULT))
+            .build();
+    ServerStreamingCallSettings settings =
+        ServerStreamingCallSettings.<Color, Money>newBuilder().build();
+    ServerStreamingCallable streamingCallable =
+        GrpcCallableFactory.createServerStreamingCallable(
+            GrpcCallSettings.create(METHOD_SERVER_STREAMING_RECOGNIZE), settings, context);
+    Color request = Color.newBuilder().setRed(0.5f).build();
+    try {
```
igorbernstein2, line 627: IllegalStateException e = assertThrows(() -> streamingCallable.call(..), IllegalStateException);
assertThat(e).hasMessage...
assertThat(e.getCause()).is...

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcDirectStreamController.java
```diff
@@ -74,8 +74,9 @@ public void run() {}
 
   @Override
   public void cancel() {
-    cancellationException = new CancellationException("User cancelled stream");
-    clientCall.cancel(null, cancellationException);
+    String message = "User cancelled stream";
+    cancellationException = new CancellationException(message);
+    clientCall.cancel(message, cancellationException);
```
igorbernstein2, line 79: what does the double message give us?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
```diff
@@ -517,6 +519,7 @@ public <RequestT, ResponseT> ClientCall<RequestT, ResponseT> newCall(
 
   /** ClientCall wrapper that makes sure to decrement the outstanding RPC count on completion. */
   static class ReleasingClientCall<ReqT, RespT> extends SimpleForwardingClientCall<ReqT, RespT> {
+    private CancellationException cancellationException;
```
igorbernstein2, line 522: `@Nullable`

igorbernstein2 (approved):
lgtm after nits

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
```diff
@@ -542,7 +548,14 @@ public void onClose(Status status, Metadata trailers) {
       } catch (Exception e) {
         // In case start failed, make sure to release
         entry.release();
+        throw e;
```
burkedavison, line 551: What's our definition of a 'breaking change'? Does a behavior change such as throwing an exception in a situation where it was previously swallowed count?

mutianf: I think normally this wouldn't happen unless something weird happens. In that case I think this will help us understand what the real problem is?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcDirectStreamController.java
mutianf: not much :/ reverted back

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
mutianf: And also, if start here failed, GrpcDirectStreamController will move on to `ClientCallImpl#sendMessage` which will likely fail there

burkedavison: Ah, and this is all in package-protected classes. Thanks for helping to clarify.

mutianf:
@burkedavison Do you think it's ok to merge this PR? Do you have any other concerns? Thanks!

burkedavison:
> @burkedavison Do you think it's ok to merge this PR? Do you have any other concerns? Thanks!

Hey @mutianf ,
LGTM, but I'd like to request approval from someone more familiar with gax-java development. @blakeli0 / @meltsufin ?

blakeli0 (approved):

PR #1226, opened 2023-01-06
fix(spring): fix method javadoc missing @param and @return

zhumin8:
Adds `@param` and `@return` to method javadoc comments in generated autoconfiguration classes. 
Also adding a change to `JavaDocComment` to enable `@return` in comment. (this change in 537e33c2cdaf5e79d29f12fe12d372e1f75a525d can be copied to main branch of `gapic-generator-java` if useful.)

### DIFF
--- a/src/main/java/com/google/api/generator/engine/ast/JavaDocComment.java
+++ b/src/main/java/com/google/api/generator/engine/ast/JavaDocComment.java
@@ -52,6 +52,7 @@ public abstract class JavaDocComment implements Comment {
     String throwsDescription = null;
     String deprecated = null;
     List<String> paramsList = new ArrayList<>();
+    String returnText = null;
     List<String> componentsList = new ArrayList<>();
     // Private accessor, set complete and consolidated comment.
     abstract Builder setComment(String comment);
@@ -66,6 +67,11 @@ public abstract class JavaDocComment implements Comment {
 
     public Builder setDeprecated(String deprecatedText) {
       deprecated = deprecatedText;
+      return this;
+    }
+
+    public Builder setReturn(String returnDescription) {
+      returnText = returnDescription;
       return this;
     }
 
@@ -130,12 +136,16 @@ public abstract class JavaDocComment implements Comment {
           && Strings.isNullOrEmpty(throwsDescription)
           && Strings.isNullOrEmpty(deprecated)
           && paramsList.isEmpty()
+          && Strings.isNullOrEmpty(returnText)
           && componentsList.isEmpty();
     }
 
     public JavaDocComment build() {
-      // @param, @throws and @deprecated should always get printed at the end.
+      // @param, @return, @throws and @deprecated should always get printed at the end.
       componentsList.addAll(paramsList);
+      if (!Strings.isNullOrEmpty(returnText)) {
+        componentsList.add(String.format("@return %s", returnText));
+      }
       if (!Strings.isNullOrEmpty(throwsType)) {
         componentsList.add(
             String.format("@throws %s %s", throwsType, HtmlEscaper.process(throwsDescription)));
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/api/generator/engine/ast/JavaDocComment.java
```diff
@@ -130,12 +136,16 @@ public boolean emptyComments() {
...
+          && Strings.isNullOrEmpty(returnText)
           && componentsList.isEmpty();
     }
 
     public JavaDocComment build() {
-      // @param, @throws and @deprecated should always get printed at the end.
+      // @param, @return, @throws and @deprecated should always get printed at the end.
       componentsList.addAll(paramsList);
+      if (!Strings.isNullOrEmpty(returnText)) {
+        componentsList.add(String.format("@return %s", returnText));
```
burkedavison, line 147: This is great, and will be a big win for for the main branch. Any chance we could get this implemented in main, then merge the update from main into autoconfig-gen-draft2? (It may not be that simple, though -- since it looks like this branch still has to sync with the multimodule update.)

diegomarquezp (approved):
Nice!

src/test/java/com/google/api/generator/engine/ast/JavaDocCommentTest.java
```diff
@@ -178,6 +178,30 @@ public void createJavaDocComment_throwsAndDeprecated() {
     assertEquals(expected, javaDocComment.comment());
   }
 
+  @Test
+  public void createJavaDocComment_paramsAndReturn() {
+    // No matter how many times or order `setThrows` and `setDeprecated` are called,
+    // only one @throws and @deprecated will be printed.
```
emmileaf, line 184: This comment probably slipped in from the previous unit test:
```suggestion
```

emmileaf (commented):
One small test comment fix, otherwise LGTM!

burkedavison:
@zhumin8 : Shouldn't this cause a very large scale change in the golden files?

src/main/java/com/google/api/generator/engine/ast/JavaDocComment.java
burkedavison: Shouldn't this cause a large scale change in the golden files, or am I misunderstanding the scope of this change?

emmileaf: @burkedavison with the multimodule update, it's not too feasible to merge `main` into `autoconfig-gen-draft2` (which will always stay as a separate branch) - but copying the relevant changes in a separate PR to `main` might be the easier way to go here? 

(For context: have a similar commit in this situation as part of #1208 - given that spring codegen plans to eventually move to use the gapic-generator-java jar, these changes do also need to be made in `main`; we are adding them directly to the spring branch for now given the current repo structure divergence.)

blakeli0: This change is enabling us to add `@return` to the JavaDocs. You would still need to set `returnText` in the composers class for each method to actually generate the JavaDocs.
@burkedavison We decided that the SpringCodeGen will continue to work on a branch until preview release to prevent disruptions. That being said, I agree this change is important to the main branch as well and we should backport it, @zhumin8 can you please add it to the TODO list? We can create a PR against main with just this change.

emmileaf: > Shouldn't this cause a large scale change in the golden files, or am I misunderstanding the scope of this change?

I think this change just enables setting return description in javadoc comments (a change that can be used by any of the gapic class comment composers, but the actual added calls to `setReturn` are scoped to the spring-specific ones).

burkedavison: Thanks for clarifying!

emmileaf (approved):

PR #1227, opened 2023-01-06
fix(spring): use newHttpJsonBuilder in settings bean for REST transport option

emmileaf:
This PR is a follow-up to https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1364#discussion_r1062017855. 

Updated behaviour (see #1078) for when `Transport.REST_GRPC` and `useRest = true` (through properties):
- The `ServiceSettings` bean uses `newHttpJsonBuilder`
- The `TransportChannelProvider` bean definition returns `defaultHttpJsonTransportProviderBuilder`

### DIFF
--- a/src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java
+++ b/src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java
@@ -484,7 +484,7 @@ public class SpringAutoConfigClassComposer implements ClassComposer {
     if (hasRestOption) {
       // For GRPC+REST libraries
       // LanguageServiceSettings.Builder clientSettingsBuilder;
-      //    if (clientProperties.isUseRest()) {
+      //    if (this.clientProperties.isUseRest()) {
       //      clientSettingsBuilder = LanguageServiceSettings.newHttpJsonBuilder();
       //    } else {
       //      clientSettingsBuilder = LanguageServiceSettings.newBuilder();
--- a/src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java
+++ b/src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java
@@ -464,12 +464,7 @@ public class SpringAutoConfigClassComposer implements ClassComposer {
                 .build());
 
     List<Statement> bodyStatements = new ArrayList<>();
-    //   LanguageServiceSettings.Builder clientSettingsBuilder =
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java
```diff
@@ -350,17 +350,75 @@ private static MethodDefinition createTransportChannelProviderBeanMethod(
...
+        MethodInvocationExpr.builder()
+            .setStaticReferenceType(types.get("ServiceSettings"))
+            .setMethodName("defaultHttpJsonTransportProviderBuilder")
+            .build();
+    MethodInvocationExpr defaultHttpJsonTransportProviderExpr =
+        MethodInvocationExpr.builder()
+            .setExprReferenceExpr(defaultTransportProviderExprChain)
+            .setMethodName("build")
+            .setReturnType(STATIC_TYPES.get("InstantiatingHttpJsonChannelProvider"))
+            .build();
```
blakeli0, line 383: Since this section is only used inside the if block, it would be better to move it into the if block, you could create a separate private method to improve the readability if you think it makes the if block too long.

src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java
```diff
@@ -350,17 +350,75 @@ private static MethodDefinition createTransportChannelProviderBeanMethod(
...
+
+    if (hasRestOption) {
```
blakeli0, line 385: This check is fine for preview release, but note that we need to come up with a strategy for REST only client like `compute`. Do we expect customers to always set this option to true or we automatically set this to true for REST only client? @zhumin8 @emmileaf

src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java
emmileaf: Updated, thanks for the suggestion!

src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java
zhumin8: I believe we are currently only populating `useRest` property in case `Transport.GRPC_REST`. So for REST only client, we could skip populating this property and set default to `useRest = True` behaviour

blakeli0 (approved):

PR #1233, opened 2023-01-10
feat: update JavaDoc to use @return

alicejli:
Adds the ability to create `@return` in the JavDocComments for the generator.

This is to follow standard JavaDoc documentation: https://engdoc.corp.google.com/eng/doc/devguide/java/practices/javadoc.md?cl=head#return . 

For use in both snippetGen (https://github.com/googleapis/gapic-generator-java/pull/1190/files#diff-66a3ecf48fe8e1eb956bf5fddbbeab290210f2c844434d391c9b2d9097c5fe3e) and Spring  CodeGen (https://github.com/googleapis/gapic-generator-java/pull/1226/files#diff-2859250f2b60ae040a98157043b7bb531840598b2db7b5058bf09569f7170e9e)

### DIFF
--- a/.kokoro/downstream-client-library-check.sh
+++ b/.kokoro/downstream-client-library-check.sh
@@ -31,25 +31,21 @@ cd "${scriptDir}"/..
 mvn -B -ntp install --projects '!gapic-generator-java' \
     -Dcheckstyle.skip -Dfmt.skip -DskipTests
 
-# Read current gax version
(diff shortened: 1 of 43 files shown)


meltsufin (approved):
So, for `void` methods, it will just be `null` and no `@return` would be added, right?

alicejli:
> So, for `void` methods, it will just be `null` and no `@return` would be added, right?

Yep; currently all the autogenerated samples from Phase 1 of SnippetGen do not even have JavaDocComments for the input parameters or return type. For the Phase 2 configured samples of SnippetGen, it will behave as you mention above: if the method is `void`, then it will not have a `@return` comment.

PR #1236, opened 2023-01-10
feat: Do not generate Service REST code if there are no matching RPC in a Service

lqiu96:
This is a PR for Part 2 of Issue #1117: Do not generate any Service related REST code if there are no enabled REST RPC in a Service

Part 1 (MERGED) : https://github.com/googleapis/gapic-generator-java/pull/1211

What this PR attempts to do:
- Do not generate `HttpJson{Service_Name}CallableFactory` (generated by AbstractServiceCallableFactoryClassComposer and the sub composers)
- Do not generate `HttpJson{Service_Name}Stub` (generated by AbstractTransportServiceStubClassComposer and the sub composers)
- Remove HttpJson references `{Service_Name}StubSettings` (generated by AbstractServiceStubSettingsClassComposer and the sub composers)
- Remove HttpJson references `{Service_Name}ServiceSettings` (generated by AbstractServiceSettingsClassComposer and the sub composers)
- Do not generate `{Service_Name}ClientHttpJsonTest` (generated by `AbstractServiceClientTestClassComposer` and sub composers)

Service class exposes a new `hasAnyEnabledMethodsForTransport()` to determine if the above logic is needed for the Transports (gRPC -> Always enabled vs HttpJson -> Needs to enabled/ eligible).


gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceCallableFactoryClassComposer.java
```diff
@@ -60,8 +61,17 @@ protected TransportContext getTransportContext() {
...
 
+  protected Transport getTransport() {
+    return Transport.GRPC;
+  }
+
   @Override
   public GapicClass generate(GapicContext context, Service service) {
+    // Do not generate the Callable Factory if there are no RPCs enabled for the Transport
+    if (!service.hasAnyEnabledMethodsForTransport(getTransport())) {
+      return GapicClass.createEmpty();
```
blakeli0, line 72: How bad is it gonna be if we return `null` here instead of creating an empty `GapicClass`? Or if we move the check to before calling this method? I would prefer to not returning anything or check before calling this method, because an `GapicClass.createEmpty()` could mean an empty file rather than no files. 
Another option is to change the return type to `Optional<GapicClass>` so that we can return `Optional.empty()` here but that apparently is a much larger effort.

blakeli0:
Can we have a golden test(either in the unit tests or integration tests) so that we can easily compare the diff introduced by this PR?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceCallableFactoryClassComposer.java
lqiu96: I don't think it's bad -- I just wanted to avoid passing around nulls if possible. Maybe this could be a possible use case for null?

> GapicClass.createEmpty() could mean an empty file rather than no files

I wonder if it would make sense if I created something like `GapicClass.shouldGenerateClass()` that is generated as `true` except for this case (i.e. in a `GapicClass.createNonGenerateableClass()` func or something).

> Another option is to change the return type to Optional<GapicClass> so that we can return Optional.empty() here but that apparently is a much larger effort.

I think that also makes sense, but yeah would require a larger change.

lqiu96:
> Can we have a golden test(either in the unit tests or integration tests) so that we can easily compare the diff introduced by this PR?

Yep! Working on it. I was playing around with the showcase tests but I'll sync up with Burke about getting this setup/ working.

WORKSPACE
```diff
@@ -26,7 +26,7 @@ load("@com_google_api_gax_java//:repositories.bzl", "com_google_api_gax_java_rep
 
 com_google_api_gax_java_repositories()
 
-_googleapis_commit = "44d6bef0ca6db8bba3fb324c8186e694bcc4829c"
+_googleapis_commit = "7438480b2a1bc6371d748e974f7a3647f90c4e8d"
```
lqiu96, line 29: Updating the googleapis commit to a later commit (It'll be from Jan 12th, 2023). I'm not sure if there is significance of this particular commit, but I'm using apigee for some golden integration tests and the grpc+rest transports for the library was enabled after this commit

lqiu96: If this commit can be updated, perhaps it would make sense to do this in a separate PR? This would require updating the goldens for every existing library.

lqiu96:
Updating the commit so that ApigeeConnect V1 Golden has gRPC+REST Transport: https://github.com/googleapis/gapic-generator-java/pull/1249

WORKSPACE
lqiu96: This been updated and merged in https://github.com/googleapis/gapic-generator-java/pull/1249

lqiu96:
TODO (as of 1/17/23): Look to remove the HttpJson sample that is being created by for each Service (if the service doesn't have a HttpJson RPC)

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/samplecode/ServiceClientHeaderSampleComposer.java
```diff
@@ -342,7 +342,7 @@ public static Sample composeTransportSample(
         RegionTag.builder()
             .setServiceName(service.name())
             .setRpcName(rpcName)
-            .setOverloadDisambiguation("setCredentialsProvider")
+            .setOverloadDisambiguation("composeTransport")
```
lqiu96, line 345: @blakeli0 I believe the name for this sample was wrong. It should be an example showcasing how to switch to the non-default transport (HttpJson).

showcase/BUILD.bazel
```diff
@@ -52,6 +65,8 @@ java_gapic_assembly_gradle_pkg(
     transport = "grpc+rest",
     deps = [
         ":showcase_java_gapic",
+        # TODO(lawrenceqiu): Not adding :showcase_java_grpc_extended dep as that includes WickedGrpc.java
+        # Need to figure out why it's being included
```
lqiu96, line 69: I'm not able to figure out why this is the case. Thanks @burkedavison for helping me with the rest of the bazel build rules 😄 !

showcase/gapic-showcase-extended/proto/wicked.proto
```diff
@@ -0,0 +1,49 @@
+// Copyright 2018 Google LLC
```
burkedavison, line 1:
```suggestion
// Copyright 2023 Google LLC
```

lqiu96: Thanks for the catch!

showcase/gapic-showcase-extended/proto/wicked.proto
```diff
@@ -0,0 +1,49 @@
...
+option go_package = "github.com/googleapis/gapic-showcase/server/genproto";
+option java_package = "com.google.showcase.v1beta1";
+option java_multiple_files = true;
+option ruby_package = "Google::Showcase::V1beta1";
+
+// This service is used to show a Service with either non-enabled or non-eligible
+// RPCs for HttpJson (Http 1.1).
+// Non-Enabled: Missing the (google.api.http) annotation to enabled it
+// Non-Eligible: BIDI and Client side streaming are not supported with Http 1.1
+// Service name is reference to `No REST for the Wicked` (Credit: @BurkeDavison)
```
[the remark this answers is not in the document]
lqiu96: Haha cheers :) Just wanted to give you a shoutout for the name

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/ServiceStubSettingsClassComposer.java
```diff
@@ -50,17 +51,26 @@ private static TypeStore createStaticTypes() {
             InstantiatingHttpJsonChannelProvider.class));
   }
 
+  @Override
```
blakeli0, line 54: Do we need to override the `getTransport` method here?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/model/GapicClass.java
```diff
@@ -26,7 +27,8 @@ public enum Kind {
     MAIN,
     STUB,
     TEST,
-    PROTO
+    PROTO,
+    NON_GENERATED
```
blakeli0, line 31: This is fine to me. Can you please add more comments to explain what `NON_GENERATED` means?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/ServiceStubSettingsClassComposer.java
blakeli0: Nevermind, noticed that we do need it.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceCallableFactoryClassComposer.java
```diff
@@ -60,8 +61,17 @@ protected TransportContext getTransportContext() {
     return transportContext;
   }
 
+  protected Transport getTransport() {
```
blakeli0, line 64: Noticed that we have a `getTransportContext` methods for most of the abstract composers, can we leverage `getTransportContext().transport()` for our use case instead of adding new methods?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/model/Transport.java
```diff
@@ -20,12 +20,13 @@ public enum Transport {
...
-   * Parse command line transport argument in the format `grpc+rest`.
+   * Parse command line transport argument in the format `grpc+rest`. Convert name String to
+   * lowercase as `gRPC` is not valid but `grpc` is
    *
    * @param name name of the transport. Valid inputs are "grpc", "rest", "grpc+rest"
    * @return the {@code Transport} enum matching the command line argument
    */
   public static Transport parse(String name) {
-    return valueOf(name.replace('+', '_').toUpperCase());
+    return valueOf(name.toLowerCase().replace('+', '_').toUpperCase());
```
blakeli0, line 30: Can we add a few unit tests to this method?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/model/Transport.java
```diff
@@ -20,12 +20,13 @@ public enum Transport {
   GRPC_REST;
 
   /**
-   * Parse command line transport argument in the format `grpc+rest`.
+   * Parse command line transport argument in the format `grpc+rest`. Convert name String to
+   * lowercase as `gRPC` is not valid but `grpc` is
```
blakeli0, line 24: This comment is a little confusing as `gRPC` is a valid input now, we might want to just remove this comment and update the valid inputs section in the javadoc below.

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpcrest/goldens/HttpJsonWickedCallableFactory.golden
```diff
@@ -0,0 +1,278 @@
...
+ * // https://cloud.google.com/java/docs/setup#configure_endpoints_for_the_client_library
+ * WickedSettings wickedSettings = WickedSettings.newBuilder().setEndpoint(myEndpoint).build();
+ * WickedClient wickedClient = WickedClient.create(wickedSettings);
+ * }</pre>
+ *
+ * <p>Please refer to the GitHub repository's samples for more quickstart code snippets.
+ */
+@BetaApi
+@Generated("by gapic-generator-java")
+public class WickedClient implements BackgroundResource {
```
blakeli0, line 89: A couple of the golden files have class name `WickedClient` generated which does not seem right, can you take a look?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/WickedClientTest.java
```diff
@@ -0,0 +1,213 @@
...
+import javax.annotation.Generated;
+import org.junit.After;
+import org.junit.AfterClass;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+@Generated("by gapic-generator-java")
+public class WickedClientTest {
```
blakeli0, line 47: Is there any manual showcase tests added? It's great that you added a new showcase proto, but if there is no interactions with the showcase server, I would suggest to use the unit tests or golden integration tests for simplicity.

test/integration/goldens/asset/samples/snippets/generated/main/java/com/google/cloud/asset/v1/assetservice/create/SyncCreateComposeTransport.java
```diff
@@ -16,17 +16,17 @@
...
-public class SyncCreateSetCredentialsProvider1 {
+public class SyncCreateComposeTransport {
 
   public static void main(String[] args) throws Exception {
-    syncCreateSetCredentialsProvider1();
+    syncCreateComposeTransport();
   }
 
-  public static void syncCreateSetCredentialsProvider1() throws Exception {
+  public static void syncCreateComposeTransport() throws Exception {
```
blakeli0, line 29: This does not look correct to me, can we leave the SnippetGen code out of this PR for now? Then create a separate PR for the SnippetGen changes and ask @alicejli to review it?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/WickedClientTest.java
burkedavison: @blakeli0 : I requested we go this route to be explicit about what the golden tests are verifying with the goal to eventually decouple ourselves from specific APIs that may change over time outside of our control. If a service today only supports gRPC but later supports REST we will have lost our golden test for this behavior.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/model/GapicClass.java
lqiu96: Sure, will add.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceCallableFactoryClassComposer.java
lqiu96: I was thinking of using this, except it doesn't work for grpc+rest workflows. When I call `getTransportContext().transport()`, it would return `Transport.GRPC_REST`. If I pass that to the sub-composers, the sub-composers wouldn't know if it was REST or GRPC.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/model/Transport.java
lqiu96: Sure, will add.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/model/Transport.java
lqiu96: Ah I see. I added that comment to explain why I was calling `toLowercase()`, but I see why that's confusing now. I'll change.

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpcrest/goldens/HttpJsonWickedCallableFactory.golden
lqiu96: Sure, taking a look.

test/integration/goldens/asset/samples/snippets/generated/main/java/com/google/cloud/asset/v1/assetservice/create/SyncCreateComposeTransport.java
lqiu96: The only reason I included this in this PR was because there was samples being generated that use the `newHttpTransport()` code for services/clients that were not enabled for HTTP. Fixing this snippetgen logic removed generating the http code for non-enabled services.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/WickedClientTest.java
lqiu96: @blakeli0 I'll forward you the email where we had some discussions with Noah and Victor. I added a `showcase-extended` module to hold some non-api specific cases that aren't covered by the existing showcase test cases.

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpcrest/goldens/HttpJsonWickedCallableFactory.golden
lqiu96: Should be updated correctly now.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceCallableFactoryClassComposer.java
blakeli0: I would consider `Transport.GRPC_REST` the same as `Transport.REST` when we are checking if there is any REST supported methods, as that's practically what we are doing right now. Otherwise having two methods `getTransportContext()` and `getTransport()` serve similar purposes are confusing.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/WickedClientTest.java
blakeli0: > If a service today only supports gRPC but later supports REST we will have lost our golden test for this behavior.

Yes that's a good test case. However, this test case does not require us to interact with showcase server, and we can achieve the same thing in unit golden tests. Currently we have two identical `wicked.proto` in both unit test proto and showcase proto, and they are testing the same thing from what I can see.

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpcrest/goldens/HttpJsonWickedCallableFactory.golden
lqiu96: Might not have refreshed for you. The new changes have deleted `HttpJsonWickedCallableFactory.golden`

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceCallableFactoryClassComposer.java
lqiu96: I agree that for this feat that `Transport.GRPC_REST the same as Transport.REST`. The issue I came across was that there is a bunch of logic in the AbstractComposer classes that shouldn't generate if is REST transport. But both sub-composers have a TransportContext of Transport.GRPC_REST. Since both sub-composers technically have the same Transport, it's impossible to build logic where `GRPC_REST == REST` in the Abstract class.

I was trying to find a generic solution that wouldn't require me to hard code `Transport.REST` : `service.hasAnyEnabledMethodsForTransport(Transport.REST)`

I was thinking that `service.hasAnyEnabledMethodsForTransport(getTransport())` would be clearer for the Abstract Classes as it shouldn't be tied to any transport, but if it's confusing to add a similar method then I can just hardcode the REST transport.

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpcrest/goldens/HttpJsonWickedCallableFactory.golden
blakeli0: Thanks, I see the change now.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/WickedClientTest.java
lqiu96: I agree it might be a bit overkill to have this multiple times. The `wicked.proto` in unit tests was duplicated to help ensure sonarcode coverage (as none of the existing unit tests protos covered this path). `apigee connect v1` was used as an existing library for the (legacy?) golden ITs and the showcase `wicked.proto` was built as a initial framework for expanding showcase coverage (being able to test for generated client behavior and generator generation behavior).

@burkedavison @blakeli0  Do you think it would make sense to just include this as part of the unit tests proto and not showcase proto?

burkedavison: I agree they only need to be in one place. No strong opinion on which.

blakeli0: > I would eventually like to move away from golden ITs

I agree. But I think we should have the majority test cases in unit tests(golden or non-golden) instead of in showcase tests. Showcase tests are a great replacement for the integration tests that interact with servers, but they are still more expensive than unit tests.
> I agree they only need to be in one place. No strong opinion on which.

Thanks Burke! Also thanks for helping Lawrence setting up showcase tests for this extended proto, even though we may not use it this time, but we will definitely need it for the upcoming Dynamic Routing Header features.

lqiu96: Understood, I can leave the framework for extending showcase (keeping the showcase-extendend module and the related bazel build files). I'll use the golden units for now with `wicked.proto`.

lqiu96:
Will need to do some more local testing to ensure if Http methods are properly generated and that I'm not breaking the generated client libraries. 

Changes from showcase and goldens look good.

PR #1238, opened 2023-01-11
fix: Rename gapic-parent-pom to groupId to google.cloud.api

lqiu96:
The gapic-generator powers non-Cloud code generation. It may make sense for the groupId to be under `com.google.api` instead of `com.google.cloud` and for the gapic monorepo related modules to have a consistent groupId.


lqiu96:
Ah, I broke a bunch of things...

gapic-generator-java-pom-parent/pom.xml
```diff
@@ -3,7 +3,7 @@
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" child.project.url.inherit.append.path="false">
   <modelVersion>4.0.0</modelVersion>
-  <groupId>com.google.cloud</groupId>
+  <groupId>com.google.api</groupId>
```
suztomo, line 6: Memo. This is the main change.

suztomo (approved):

PR #1245, opened 2023-01-12
fix(spring): switch camelcase conversion logic to use JavaStyle

emmileaf:
This PR switches camelcase handling to use align with logic used in generating client libraries, e.g. [here](https://github.com/googleapis/gapic-generator-java/blob/8c1facf7a2ad2e42d6de490aebbb671439d4f1f2/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceSettingsClassComposer.java#L254). However, it introduces no updates to golden tests for the spring composers, and my local test to regenerate https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1449 using this branch also showed no diff.

I think `spring-cloud-gcp` enforces stricter checkstyle rules ([fail on warning](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/a5a346600e4b123643d9a507144d9167280c8870/pom.xml#L273-L274)) than what is used for the client libraries, so a corresponding change has been made on that side to suppress this for a number of files with abbreviation handling edge cases: https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1455
- Some examples of edge cases observed: [getOSPolicyAssignmentSettings](https://github.com/googleapis/google-cloud-java/blob/dedef71f600e85b1c38e7110f5ffd44bf2ba32b4/java-os-config/google-cloud-os-config/src/main/java/com/google/cloud/osconfig/v1/OsConfigZonalServiceSettings.java#L118),  [IDSSettings](https://github.com/googleapis/google-cloud-java/blob/main/java-ids/google-cloud-ids/src/main/java/com/google/cloud/ids/v1/IDSSettings.java)

### DIFF
--- a/src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java
+++ b/src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java
@@ -111,7 +111,7 @@ public class SpringAutoConfigClassComposer implements ClassComposer {
     Expr thisExpr = ValueExpr.withValue(ThisObjectValue.withType(dynamicTypes.get(className)));
     Transport transport = context.transport();
     boolean hasRestOption = transport.equals(Transport.GRPC_REST);
-    String serviceSettingsMethodName = JavaStyle.toLowerCamelCase(service.name() + "Settings");
+    String serviceSettingsMethodName = JavaStyle.toLowerCamelCase(service.name()) + "Settings";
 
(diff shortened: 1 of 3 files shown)


zhumin8 (approved):
Thanks for making these changes. Although no difference in generated code is made, it's good to be in sync with client library for maintenance.

meltsufin (approved):

PR #1247, opened 2023-01-13
ci: Add downstream native image testing

burkedavison:


.kokoro/presubmit/downstream-build.sh
```diff
@@ -0,0 +1,96 @@
+#!/bin/bash
```
mpeddada1, line 1: Are we planning to have this run as a Kokoro job or would the next step be to modify https://github.com/googleapis/gapic-generator-java/blob/main/.github/workflows/downstream-native-image.yaml to run this script?

burkedavison: Kokoro. See private chat.

suztomo (approved):

.kokoro/presubmit/downstream-build.sh
```diff
@@ -0,0 +1,96 @@
...
+cd ../x:version
+set ${SHARED_DEPS_VERSION}
+save pom.xml
+EOF
+
+echo "Modifications to google-cloud-java:"
+git diff
+echo
+
+MODULES_UNDER_TEST=java-os-login,java-asset
```
suztomo, line 87: How might we parametrize this values?

suztomo, line 87: Would you add source code comment why we're choosing these two?

burkedavison: Done -- environment variable set in `.kokoro/presubmit/common.cfg`

burkedavison: Done -- comments provided next to new environment variable location in `.kokoro/presubmit/common.cfg`

PR #1249, opened 2023-01-13
chore: Update Bazel WORKSPACE googleapi commit

lqiu96:
Update the _googleapis_commit value to a new commit.


burkedavison:
WORKSPACE changes LGTM, but I don't have any context for the changes in the goldens. 
@meltsufin , @alicejli , @ddixit14 : Do you have enough awareness of the expected updates to approve?

lqiu96:
To give some more context (repeating from the chats): 

I'm thinking about adding Apigee Connect V1 as part of the IT Goldens since it covers a new REGAPIC feature in the generator (not generating certain HttpJson Client/ Services). I believe the old commit was before this library was enabled for gRPC+REST so the using the old commit would only generate gRPC related code. 

I've updated the goldens since I think the new commit adds in new features.

meltsufin (approved):

PR #1271, opened 2023-01-24
chore: update README to include instructions on bazelisk installation

mpeddada1:


DEVELOPMENT.md
```diff
@@ -140,6 +140,13 @@ $ gapic-showcase run
 
 ### Running the Integration Tests
 
+Before running the integration tests, make sure to install bazelisk:
+
+```shell
+$ go install github.com/bazelbuild/bazelisk@latest
```
burkedavison, line 146: Hm. I'm surprised you needed to explicitly install bazelisk -- I think it came with my bazel installation, or I'm forgetting the manual install...

mpeddada1: oh interesting! For me, it was still missing despite install bazel:

```
mpeddada@mpeddada:~/IdeaProjects/gapic-generator-java/showcase$ sudo apt-get install bazel
[sudo] password for mpeddada: 
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
bazel is already the newest version (6.0.0).
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
mpeddada@mpeddada:~/IdeaProjects/gapic-generator-java/showcase$ bazelisk
bash: bazelisk: command not found
```

burkedavison: Ah. Yes. You are correct.

```sh
$ which bazelisk
/opt/homebrew/bin/bazelisk
$ brew list
==> Formulae
bazelisk [...]
```
So at some point I invoked `brew install bazelisk`

blakeli0: I think you do have to install bazelisk separately, I learned that the hard way.
Another nit, this would require us to install `go` as well, I know `go` is already required for showcase testing so this is easy for us to set up the environment, but do we want to add more info(e.g. [bazelisk Github page](https://github.com/bazelbuild/bazelisk)) for what bazelisk is and how to set it up in case the `go install` does not work.

suztomo: @mpeddada1  Add a link to bazelisk documentation. I don't want users relying on 'go'. (I use `npm` to use bazelisk.). Move that paragraph to "## Set Up" section.

mpeddada1: Done. Replaced the instruction with a link to the Github page and moved it to "Set Up". Thank you!

burkedavison (approved):

PR #1280, opened 2023-01-26
ci: split showcase golden and integration test phases

burkedavison:
> @suztomo : The failure is unclear. In another pull request, would you separate golden test and integration tests?

CC: @mpeddada1

### DIFF
--- a/.github/workflows/ci-maven.yaml
+++ b/.github/workflows/ci-maven.yaml
@@ -13,7 +13,7 @@ jobs:
       matrix:
         java: [ 11, 17 ]
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - uses: actions/setup-java@v3
         with:
           java-version: ${{ matrix.java }}
@@ -139,7 +139,7 @@ jobs:
   lint:
     runs-on: ubuntu-22.04
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - uses: actions/setup-java@v3
         with:
           java-version: 11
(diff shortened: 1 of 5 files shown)


.github/workflows/ci-maven.yaml
```diff
@@ -156,20 +156,24 @@ jobs:
...
           java-version: ${{ matrix.java }}
           distribution: temurin
-      - run: java -version
+      - run: mvn -version
       - name: Install maven modules
         run: |
           mvn install -B -ntp -DskipTests -Dclirr.skip -Dcheckstyle.skip
 
-      - name: Install system dependencies
+      - name: Showcase golden tests
```
suztomo, line 169: That's clear. Thanks.

suztomo (approved):

.github/workflows/ci-maven.yaml
```diff
@@ -156,20 +156,24 @@ jobs:
...
+        working-directory: showcase
         run: |
-          sudo apt-get update
-          sudo apt-get install -y curl unzip
+          mvn test \
+            -P enable-golden-tests \
+            --batch-mode \
+            --no-transfer-progress
```
suztomo, line 175: Do you really need maven that runs bazel that runs a shell script?

suztomo (changes requested):
Examine the use of the Maven in showcase golden test.

.github/workflows/ci-maven.yaml
mpeddada1, line 169: Would this need to be omitted from https://github.com/googleapis/gapic-generator-java/blob/main/.github/workflows/downstream-native-image.yaml as well?

.github/workflows/ci-maven.yaml
burkedavison: Showcase has been set up to be invoked with all `mvn` commands to avoid needing to change developer workflows once @blakeli0 has completed https://github.com/googleapis/gapic-generator-java/issues/1069

If this is something you'd like to revisit, I would request that we open a ticket and handle it separately.

suztomo: Fair enough.

suztomo (approved):

.github/workflows/ci-maven.yaml
blakeli0: FYI, I just closed https://github.com/googleapis/gapic-generator-java/issues/1069 as the generator process is already using the gapic-generator-java jar. 
I agree with @suztomo in long term, but since we are still in the middle of mono repo migration cleanup/enhancement, I think this is good for now and we can revisit it once everything is stable.

burkedavison: Tracking with new issue: https://github.com/googleapis/gapic-generator-java/issues/1282

Issue #1279, opened 2023-01-26
REST LRO fails for certain calls


lqiu96:
Java Sample:
```
package org.example;
 
import com.google.api.gax.longrunning.OperationFuture;
import com.google.cloud.speech.v1.LongRunningRecognizeMetadata;
import com.google.cloud.speech.v1.LongRunningRecognizeRequest;
import com.google.cloud.speech.v1.LongRunningRecognizeResponse;
import com.google.cloud.speech.v1.RecognitionAudio;
import com.google.cloud.speech.v1.RecognitionConfig;
import com.google.cloud.speech.v1.RecognizeRequest;
import com.google.cloud.speech.v1.RecognizeResponse;
import com.google.cloud.speech.v1.SpeechClient;
import com.google.cloud.speech.v1.SpeechRecognitionResult;
import com.google.cloud.speech.v1.SpeechSettings;
import java.io.IOException;
import java.util.List;
import java.util.concurrent.ExecutionException;
 
public class Main {
  public static void main(String[] args)
      throws IOException, ExecutionException, InterruptedException {
    String gcsUri = "gs://cloud-samples-data/speech/brooklyn_bridge.raw";
    RecognitionConfig recognitionConfig =
        RecognitionConfig.newBuilder()
            .setEncoding(RecognitionConfig.AudioEncoding.LINEAR16)
            .setSampleRateHertz(16000)
            .setLanguageCode("en-US")
            .build();
    RecognitionAudio recognitionAudio = RecognitionAudio.newBuilder().setUri(gcsUri).build();
 
    System.out.println("Running with gRPC...");
    try (SpeechClient speechClient = SpeechClient.create()) {
      testSpeechRecognize(speechClient, recognitionConfig, recognitionAudio);
    }
 
    System.out.println("Running with REST...");
    SpeechSettings speechSettings = SpeechSettings.newHttpJsonBuilder().build();
    try (SpeechClient speechClient = SpeechClient.create(speechSettings)) {
      testSpeechRecognize(speechClient, recognitionConfig, recognitionAudio);
    }
  }
 
  public static void testSpeechRecognize(
      SpeechClient speechClient,
      RecognitionConfig recognitionConfig,
      RecognitionAudio recognitionAudio)
      throws ExecutionException, InterruptedException {
    RecognizeResponse recognizeResponse =
        speechClient.recognize(
            RecognizeRequest.newBuilder()
                .setConfig(recognitionConfig)
[...]

PR #1288, opened 2023-01-31
fix: Allow custom HttpRules for REST LROs

lqiu96:
Fixes: #1279 

This is both a gapic-generator and gax-java change:
- Gapic Generator has a map of the Custom Http Bindings inside HttpJson{Service}Stub.java and passes it to the HttpJsonOperationStub. The generator parses the ServiceYaml file and recreates the HttpRules to be sent over.
- Gax Java will update the rawPath and additional binding values for Operation (based on values received via HttpJsonOperationClient)

For example in Java-Speech generated client:
```
    this.httpJsonOperationsStub =
        HttpJsonOperationsStub.create(
            clientContext,
            callableFactory,
            typeRegistry,
            ImmutableMap.<String, HttpRule>builder()
                .put(
                    "google.longrunning.Operations.GetOperation",
                    HttpRule.newBuilder().setGet("/v1/operations/{name=**}").build())
                .put(
                    "google.longrunning.Operations.ListOperations",
                    HttpRule.newBuilder().setGet("/v1/operations").build())
                .build());
...
```

Output:
```
Running with REST...
Transcript: how old is the Brooklyn Bridge -> No longer errors out
```


gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/rest/goldens/HttpJsonComplianceStub.golden
```diff
@@ -350,6 +351,9 @@ public class HttpJsonComplianceStub extends ComplianceStub {
                   .build())
           .build();
 
+  private static final Map<String, String> operationCustomHttpBindings =
```
blakeli0, line 354: We probably don't want to create unused variables if we can avoid it in the generator.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStub.java
```diff
@@ -72,8 +72,14 @@
 public class HttpJsonOperationsStub extends OperationsStub {
   private static final Pattern CLIENT_PACKAGE_VERSION_PATTERN =
       Pattern.compile("\\.(?<version>v\\d+[a-zA-Z]*\\d*[a-zA-Z]*\\d*)\\.[\\w.]*stub");
-
-  private static final ApiMethodDescriptor<ListOperationsRequest, ListOperationsResponse>
+  private static final String LRO_LIST_OPERATIONS = "google.longrunning.Operations.ListOperations";
```
blakeli0, line 75: Are these all possible operations? Judging from the code in the composer, looks like there could be patch etc.?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
```diff
@@ -1103,6 +1143,43 @@ protected List<Expr> createOperationsStubInitExpr(
...
+
+  private Map<String, String> filterCustomHttpBindingsMap(
+      Map<String, String> customHttpBindings, Predicate<Map.Entry<String, String>> predicate) {
+    return customHttpBindings.entrySet().stream()
+        .filter(predicate)
+        .collect(Collectors.toMap(x -> x.getKey(), x -> x.getValue()));
+  }
+
+  private String getValueBasedOnPatternCase(HttpRule httpRule) {
+    switch (httpRule.getPatternCase().getNumber()) {
```
blakeli0, line 1165: It would be good if we can provide a reference to the http proto for these mappings.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
```diff
@@ -1103,6 +1143,43 @@ protected List<Expr> createOperationsStubInitExpr(
             .build());
   }
 
+  private Map<String, String> parseCustomHttpBindings(GapicContext context) {
+    Map<String, String> customHttpBindings = new HashMap<>();
```
blakeli0, line 1147: We can add all the default `operation -> url` mapping to this map and override then if needed. Basically we need to move as much code as possible to the generator from gax, as gax is a runtime dependency and it would be much harder to make changes there later.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ProtoMessageRequestFormatter.java
```diff
@@ -147,6 +147,12 @@ public Builder<RequestT> setAdditionalPaths(String... rawAdditionalPaths) {
       return this;
     }
 
+    @InternalApi
+    public Builder<RequestT> updateRawPath(String rawPath) {
```
blakeli0, line 151: Is it possible to initialize the `ProtoMessageRequestFormatter` with the updated path instead of overriding it later?

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStub.java
```diff
@@ -223,21 +230,70 @@ public static final HttpJsonOperationsStub create(
...
-        OperationsStubSettings.newBuilder().build(), clientContext, callableFactory, typeRegistry);
+        OperationsStubSettings.newBuilder().build(),
+        clientContext,
+        callableFactory,
+        typeRegistry,
+        new HashMap<>());
+  }
+
+  public static final HttpJsonOperationsStub create(
+      OperationsStubSettings settings, Map<String, String> customOperationHttpBindings)
```
blakeli0, line 241: This approach is fine but I would try to explore the possibility of add the `customOperationHttpBindings` to `settings` or `clientContext`. It would be cleaner and we don't have to have so many overloaded methods.

lqiu96: (From what I understand) The core of the logic which we can change in the generated library comes from:
```
  protected HttpJsonSpeechStub(
      SpeechStubSettings settings,
      ClientContext clientContext,
      HttpJsonStubCallableFactory callableFactory)
      throws IOException {
    this.callableFactory = callableFactory;
    this.httpJsonOperationsStub =
        HttpJsonOperationsStub.create(
            clientContext,
            callableFactory,
            typeRegistry,
...
```
at https://github.com/googleapis/google-cloud-java/blob/9c0ddf2879b0e3ddf29b5d67cd44f0336f492134/java-speech/google-cloud-speech/src/main/java/com/google/cloud/speech/v1/stub/HttpJsonSpeechStub.java#L181-L188

The `SpeechStubSettings settings` parameter is different from `OperationsStubSettings` which the `HttpJsonOperationsStub.create()` is expecting. The other approach I'm looking at is to try and add it either the ClientContext (currently leaning towards not putting it here) or the `HttpJsonStubCallableFactory` (Need to do a bit more digging into this).

The overloaded methods were the easiest way I could get the CI builds to pass. Ideally, I would just have each of the `.create()` methods take in another `Map` parameter, but I'm having issue with the build passing. I can show you this tomorrow.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
lqiu96: We can add the defaults for bunch of mixins (or only Operation). I just saw that it is possible to have way more mixins (https://github.com/googleapis/gapic-showcase/blob/b94ecfc51059b49770e5bdb6f0d7ea07903158e8/schema/google/showcase/v1beta1/showcase_v1beta1.yaml#L46-L85) so I didn't want to just create a list that needed to be manually updated later on.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
lqiu96: I'll remember to add it in later. I'm not sure if this is the best approach yet, so I haven't been adding any docs or tests yet.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStub.java
lqiu96: There are only 5 operations (and 4 of them have http annotations): https://github.com/googleapis/googleapis/blob/c8a382b9980b59a241a561aef1599980209a5e87/google/longrunning/operations.proto#L67-L122

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
blakeli0: I see, I think that makes sense, we could do it the other way also, construct a map with all the bindings from the yaml, and populate default for the LRO operations. Or if we are confident about the list of mixins, we could pre-populate them as well, a related question, did you get a chance to see how mixins work for location and iam? Are we going to have similar problem for them?

lqiu96: From what I can see, I think only the Operation proto is currently being used. I would imagine if Location and IAM is used, the would be created the same way the Operations Client was in GAX: https://github.com/googleapis/gax-java/pull/1456 and we would reference it via https://github.com/googleapis/google-cloud-java/blob/10bb0cb494f64b864408ede46834e1046351370c/java-speech/google-cloud-speech/src/main/java/com/google/cloud/speech/v1/SpeechClient.java#L164-L166.

I was running under the assumption that they would be added in sometime in the future, but I'm not sure about it. Plus each generated client (operation/ iam/ location) would have the initial defaults set from the proto file: https://github.com/googleapis/gapic-generator-java/blob/d1a16195937df041f65a52717ddf9dc6ceb09b4f/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStub.java#L84

lqiu96:
Error in CI:
```
(20:48:13) FAIL: //test/integration:redis (see /home/runner/.cache/bazel/_bazel_runner/1f815583ebc8da6875b9074f31ba9a5d/execroot/gapic_generator_java/bazel-out/k8-fastbuild/testlogs/test/integration/redis/test.log)
```

Taking another look. I'm unable to replicate this locally...

lqiu96:
The issue seems to be due to the changes in the pom.xml
```
    <grpc.version>1.52.1</grpc.version>
    <google.auth.version>1.15.0</google.auth.version>
```

I'm not sure why this is the case. Downgrading the the deps version resolves the CI issue.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStub.java
```diff
@@ -223,21 +230,58 @@ public static final HttpJsonOperationsStub create(
...
+        OperationsStubSettings.newBuilder().build(),
+        clientContext,
+        callableFactory,
+        typeRegistry,
+        new HashMap<>());
+  }
+
+  // We choose to only override these two constructors
+  // The generator can generate these two combinations with custom HttpRules
+  public static final HttpJsonOperationsStub create(
```
blakeli0, line 242: Calling one seems can be replaced by calling the one below with `TypeRegistry.getEmptyTypeRegistry()`. If there is no other major benefit, I would remove this constructor.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStub.java
```diff
@@ -223,21 +230,58 @@ public static final HttpJsonOperationsStub create(
...
+        customHttpBindings);
   }
 
   /**
    * Constructs an instance of HttpJsonOperationsStub, using the given settings. This is protected
    * so that it is easy to make a subclass, but otherwise, the static factory methods should be
    * preferred.
    */
-  protected HttpJsonOperationsStub(OperationsStubSettings settings, ClientContext clientContext)
+  protected HttpJsonOperationsStub(
```
blakeli0, line 274: Adding parameters to protected methods are breaking as well, as customers can extend this public class and override it.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStub.java
```diff
@@ -249,38 +293,30 @@ protected HttpJsonOperationsStub(
       OperationsStubSettings settings,
       ClientContext clientContext,
       HttpJsonStubCallableFactory callableFactory,
-      TypeRegistry typeRegistry)
+      TypeRegistry typeRegistry,
```
blakeli0, line 296: Same as above, adding parameters to protected methods are breaking changes.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
```diff
@@ -1103,6 +1151,124 @@ protected List<Expr> createOperationsStubInitExpr(
...
+   * the setting function name.
+   *
+   * <p>This ignores the `Custom` HttpRule. Operation proto only supports GET, POST, DELETE and
+   * Custom requires special logic.
+   *
+   * @param httpRule HttpRule to parse
+   * @return HttpRule.Builder's setter function name when setting for an HttpVerb
+   */
+  private String getHttpVerbSetterFromHttpRule(HttpRule httpRule) {
+    switch (httpRule.getPatternCase().getNumber()) {
```
blakeli0, line 1230: Can we use `httpRule.getPatternCase().toString()` for this case?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
```diff
@@ -1103,6 +1151,124 @@ protected List<Expr> createOperationsStubInitExpr(
             .build());
   }
 
+  /**
+   * Build an Expr that creates an HttpRule. Creates a builder and adds the http verb, custom path,
+   * and any additional bindings.
+   *
+   * <p>`additional_bindings` can only be nested one layer deep, so we only check once
+   *
+   * @param httpRule HttpRule to be re-created for the library to send over
```
blakeli0, line 1160: A general tip for Javadoc, it is not [required](go/java-style#s7-javadoc) for private methods and I would suggest not to add it because it's more maintenance in the future. Comments are always welcome though.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
```diff
@@ -1103,6 +1151,124 @@ protected List<Expr> createOperationsStubInitExpr(
...
+      case 3:
+        return httpRule.getPut();
+      case 4:
+        return httpRule.getPost();
+      case 5:
+        return httpRule.getDelete();
+      case 6:
+        return httpRule.getPatch();
+      default:
+        return null;
```
blakeli0, line 1268: This condition should never happen, but in case it happens, can we throw a `IllegalArgumentException` instead of returning null and getting a `NullPointerException` instead?

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStub.java
lqiu96: I've kept this one in because it is currently possible for the generator to generate this `create()` function with only two parameters. Though, searching in the codebase only suggests that three parameters has only ever been generated, I wasn't sure if that would always be the case going forward.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStub.java
lqiu96: Ah. I'll try to overload the constructor then.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStub.java
lqiu96: I'll try to overload instead

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
lqiu96: Ah yeah, I think so. Let me try changing to that.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
lqiu96: Ok, I'll just change this to comments. I wanted to document this since this is a one-off addition just to handle special cases for Operations.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
lqiu96: Yeah makes sense. Since this is really only meant for handling Operations, do you think I should also only catch for GET, POST, DELETE and throw `IllegalArgumentException` for any other HttpVerb?

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStub.java
lqiu96: On a second look, I can probably just remove the other constructor and add the param when calling the constructor.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ProtoMessageRequestFormatter.java
```diff
@@ -141,12 +141,24 @@ public Builder<RequestT> setPath(
       return this;
     }
 
+    @BetaApi
+    public Builder<RequestT> setAdditionalPaths(List<String> rawAdditionalPaths) {
```
blakeli0, line 145: I think all usages of this method should already be covered by the method below? If that's the case, I would not add this new public method even though it is marked as `BetaAPI`.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStub.java
```diff
@@ -309,22 +345,128 @@ protected HttpJsonOperationsStub(
...
+  }
+
+  /**
+   * Wrapper to get a list of the MethodDescriptors similar to `getMethodDescriptors()`. Difference
+   * is that this gets the modified MethodDescriptors with any customized values
+   *
+   * @return List of the custom MethodDescriptors
+   */
+  @InternalApi
+  public List<ApiMethodDescriptor> getAllMethodDescriptors() {
```
blakeli0, line 464: Not sure if I get the differences between this method and the one below, I think they should have the same behavior other other than the mutability of the list.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
blakeli0: Yes, I agree.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ProtoMessageRequestFormatter.java
lqiu96: Gotcha. I'll remove this one and just convert from List<String> to String[].

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStub.java
lqiu96: I'll remove this one. 

It was intended for `static -> default method descriptors` and `non-static -> customized/ modified method descriptors`, but I think that just ends up making this confusing. I was worried that running these tests in parallel would mean that the method descriptors would be updated and different test cases would pulling in modified values, but the MethodDescriptors are static anyways it wouldn't resolve the issue.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStub.java
blakeli0: Yes, as the protected constructor could be called by the subclass's constructor as well.

lqiu96: Ok in that case, I'll keep these changes as is. Given the generator is able to call both:
- ClientContext clientContext, HttpJsonStubCallableFactory callableFactory, Map<String, HttpRule> customHttpBindings
- ClientContext clientContext, HttpJsonStubCallableFactory callableFactory, TypeRegistry typeRegistry, Map<String, HttpRule> customHttpBindings

blakeli0: Maybe I'm missing something, theoretically we only need to expose one public constructor that takes the additional map, why do we have two new public constructors and two new protected constructors? Is it just for future convenience?

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStub.java
blakeli0: Can this constructor be private?

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStub.java
```diff
@@ -309,22 +355,98 @@ protected HttpJsonOperationsStub(
     backgroundResources = new BackgroundResourceAggregation(clientContext.getBackgroundResources());
   }
 
-  private static <RequestT extends Message, ResponseT>
-      ApiMethodDescriptor<RequestT, ResponseT> getApiVersionedMethodDescriptor(
-          ApiMethodDescriptor<RequestT, ResponseT> methodDescriptor, String apiVersion) {
-    if (apiVersion == null) {
-      return methodDescriptor;
+  /* OperationsClient's RPCs are mapped to GET/POST/DELETE and this function only expects those HttpVerbs to be used */
+  private static String getValueBasedOnPatternCase(HttpRule httpRule) {
```
blakeli0, line 359: The duplicate logic here and in the composer really bothers me, as the OperationClient should not exist in gax, but I guess we have to live with them for now unless we do a major refactoring.

gapic-generator-java/src/test/resources/showcase_v1beta1.yaml
```diff
@@ -0,0 +1,93 @@
...
+        - post: '/v1beta1/{resource=rooms/*}:testIamPermissions'
+          body: '*'
+        - post: '/v1beta1/{resource=rooms/*/blurbs/*}:testIamPermissions'
+          body: '*'
+        - post: '/v1beta1/{resource=sequences/*}:testIamPermissions'
+          body: '*'
+    - selector: google.longrunning.Operations.ListOperations
+      get: '/v1beta1/operations'
+      additional_bindings:
+        - get: '/v1/operations'
```
blakeli0, line 81: Is it possible to have multiple additional bindings here? If yes, can we add more here to cover the test case?

blakeli0: In addition, do you know how `additional_bindings` are used in LROs or is it even used? If not, we probably can get rid of the duplicated logics.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStub.java
lqiu96: The generator currently can create two variants of `HttpJsonOperationStub.create()` at https://github.com/googleapis/gapic-generator-java/blob/b43d9fc7f25800861847d4324298f1e16da5a508/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java#L1086-L1090
1. HttpJsonOperationStub.create(clientContext, callableFactory)
2. HttpJsonOperationStub.create(clientContext, callableFactory, typeRegistry)

Though, it seems that every version in google-cloud-java has the 2nd option and I'm not sure if there is a use case for the 1st option. I chose to implement both since I'm not 100% sure that the 1st option would never be called. Now there are 4 possibilities (and realistically probably only 2 used -- the 2nd and 4th):
1. HttpJsonOperationStub.create(clientContext, callableFactory)
2. HttpJsonOperationStub.create(clientContext, callableFactory, typeRegistry)
3. HttpJsonOperationStub.create(clientContext, callableFactory, ImmutableMap.of(...))
4. HttpJsonOperationStub.create(clientContext, callableFactory, typeRegistry, ImmutableMap.of(...))

I think I can probably make the newly added constructor private though.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStub.java
lqiu96: Ah yeah, I believe so. I'll make it private.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStub.java
lqiu96: Agreed. I'm not sure if there is a cleaner/ better way. Maybe a static function in a util class for Operations that the generator can reference (?), but that seems like overkill.

gapic-generator-java/src/test/resources/showcase_v1beta1.yaml
lqiu96: Yep, I'll add more bindings for this. Thanks!

lqiu96: > In addition, do you know how `additional_bindings` are used in LROs or is it even used? If not, we probably can get rid of the duplicated logics.

I've actually just been doing a bit of digging into as I've been running into some showcase issues with `additional_bindings`. For httpjson, I don't think there currently is any logic to handles checking other bindings once the first one doesn't match: https://github.com/googleapis/gapic-generator-java/blob/c23f981e2ac3c573bed51e725dc7061551179400/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpRequestRunnable.java#L184

I don't see any logic that handles checking the other paths/ additional_bindings which I think is a bug. For grpc, I'll need to do some more digging.

lqiu96: Filed a bug here: https://github.com/googleapis/gapic-generator-java/issues/1440

blakeli0: Thanks Lawrence, that bug looks generic for all `additional_bindings`, anything special for the OperationClient? Or as long as it's REST, there is no differences between OperationClient and other clients?

lqiu96: Yeah, the bug is generic for all `additional_bindings` and it's a very specific edge case. 

There is nothing special for OperationsClient, but I just added multiple `additional_bindings` to the Operations Client for showcase.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStub.java
blakeli0: I agree. Also if there is no usage of `HttpJsonOperationsStub.create(clientContext, callableFactory)`, I don't see why we have to keep maintaining `HttpJsonOperationsStub.create(clientContext, callableFactory)`, we can deprecate it and refactor the code you mentioned above to always use `HttpJsonOperationStub.create(clientContext, callableFactory, typeRegistry, ImmutableMap.of(...))`. It could be a separate PR though.

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStubTest.java
```diff
@@ -0,0 +1,153 @@
...
+import com.google.longrunning.CancelOperationRequest;
+import com.google.longrunning.DeleteOperationRequest;
+import com.google.longrunning.GetOperationRequest;
+import com.google.longrunning.ListOperationsRequest;
+import com.google.protobuf.TypeRegistry;
+import java.util.List;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class HttpJsonOperationsStubTest {
```
blakeli0, line 50: This test can probably use some refactoring. One best practice for sharing test object is that create it every time in `setUp` and reference the object later, similar to [this](https://github.com/googleapis/gapic-generator-java/blob/883d2e169c8d3b15b93acf36bc847587c8b21db3/gax-java/gax-grpc/src/test/java/com/google/api/gax/grpc/GrpcApiExceptionFactoryTest.java#L77) test file.

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStubTest.java
```diff
@@ -0,0 +1,153 @@
...
+            (ProtoMessageRequestFormatter<ListOperationsRequest>)
+                apiMethodDescriptorList.get(0).getRequestFormatter();
+    assertThat(listOperationsRequestProtoMessageRequestFormatter.getAdditionalPathTemplates())
+        .hasSize(1);
+    assertThat(
+            listOperationsRequestProtoMessageRequestFormatter
+                .getAdditionalPathTemplates()
+                .get(0)
+                .toRawString())
+        .isEqualTo("testList2");
```
blakeli0, line 115: Verifying a value that was set up in `setUp` method is also not a good practice, see [here](go/unit-testing-practices#setUp)

showcase/gapic-showcase/src/main/java/com/google/showcase/v1beta1/stub/HttpJsonEchoStub.java
```diff
@@ -366,7 +368,24 @@ protected HttpJsonEchoStub(
       throws IOException {
     this.callableFactory = callableFactory;
     this.httpJsonOperationsStub =
-        HttpJsonOperationsStub.create(clientContext, callableFactory, typeRegistry);
+        HttpJsonOperationsStub.create(
```
blakeli0, line 371: I guess the actual showcase integration tests that verified the interactions will be done in this [PR](https://github.com/googleapis/gapic-generator-java/pull/1431/files#diff-22c84236080a30cc7ccf014fadf1a077821e97f1a4b46f1543e25ccc1075425b)?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
```diff
@@ -1103,6 +1150,73 @@ protected List<Expr> createOperationsStubInitExpr(
...
+  private String getOperationsURIValueFromHttpRule(HttpRule httpRule) {
+    switch (httpRule.getPatternCase().getNumber()) {
+      case 2:
+        return httpRule.getGet();
+      case 4:
+        return httpRule.getPost();
+      case 5:
+        return httpRule.getDelete();
+      default:
+        throw new IllegalArgumentException(
```
blakeli0, line 1214: I would probably write some true unit tests for this method and the method above, as the input and output are very easy to mock and verify, but I know they are mostly already covered in the golden unit tests, so as long as we covered all the cases, it's fine to me. 
However, this branch couldn't be covered by golden tests, so we should add a unit test for it(by changing the scope of this method to default).

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStub.java
lqiu96: Sounds good. I'll remove `HttpJsonOperationsStub create(ClientContext clientContext, HttpJsonStubCallableFactory callableFactory, Map<String, HttpRule> customHttpBindings)` any anything else not needed. 

Since `HttpJsonOperationsStub.create(clientContext, callableFactory)` is already public at: https://github.com/googleapis/gapic-generator-java/blob/0ccf457a774bd3dbe3b101d82222bdc39f6e5b0e/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStub.java#L211-L218, I'll just leave it as is.

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStubTest.java
lqiu96: I originally had an HttpJsonOperationStub object inside this test, but the method to get the MethodDescriptors is static: https://github.com/googleapis/gapic-generator-java/blob/0ccf457a774bd3dbe3b101d82222bdc39f6e5b0e/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStub.java#L331 and is invoked by `HttpJsonOperationStub.getMethodDescriptors()`.

Is there a way to test this so that I can create an object and reference it in my tests?

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStubTest.java
lqiu96: This test tries to cover the `updateDefaultApiMethodDescriptors` function. Perhaps the easier/ cleaner way for this is to just open method as default scope and test it that way?

showcase/gapic-showcase/src/main/java/com/google/showcase/v1beta1/stub/HttpJsonEchoStub.java
lqiu96: Going forward, the interactions will be covered in that PR. For this PR, I just generated the speech client locally and tested that Operations works.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
lqiu96: > However, this branch couldn't be covered by golden tests, so we should add a unit test for it

I'm a bit lost of this part. The Echo goldens for rest should have it right?
https://github.com/googleapis/gapic-generator-java/blob/88487e38f86494e2ba18d4bf7c7f23c6671aeb24/gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/rest/goldens/HttpJsonEchoStub.golden#L360-L405

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpcrest/GrpcRestTestProtoLoader.java
```diff
@@ -58,6 +60,13 @@ public GapicContext parseShowcaseEcho() {
     ServiceDescriptor echoServiceDescriptor = echoFileDescriptor.getServices().get(0);
     assertEquals("Echo", echoServiceDescriptor.getName());
 
+    String serviceYamlFileName = "showcase_v1beta1.yaml";
```
blakeli0, line 63: I probably missed this in previous reviews, instead of `showcase_v1beta1.yaml`, can we rename it to reflect the fact that this yaml is only used by Echo at this point?

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStub.java
```diff
@@ -309,32 +325,113 @@ protected HttpJsonOperationsStub(
...
+    }
+  }
 
-    return descriptorBuilder
-        .setRequestFormatter(
-            requestFormatter.toBuilder().updateRawPath("/v1/", '/' + apiVersion + '/').build())
-        .build();
+  /* This function returns the list of method descriptors (custom or default) */
+  @InternalApi
+  public List<ApiMethodDescriptor> getAllMethodDescriptors() {
```
blakeli0, line 424: We can make remove the `public` here since it's only used in tests, we should also make that clear in the comment.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStub.java
```diff
@@ -309,32 +325,113 @@ protected HttpJsonOperationsStub(
     backgroundResources = new BackgroundResourceAggregation(clientContext.getBackgroundResources());
   }
 
-  private static <RequestT extends Message, ResponseT>
-      ApiMethodDescriptor<RequestT, ResponseT> getApiVersionedMethodDescriptor(
-          ApiMethodDescriptor<RequestT, ResponseT> methodDescriptor, String apiVersion) {
-    if (apiVersion == null) {
-      return methodDescriptor;
+  /* OperationsClient's RPCs are mapped to GET/POST/DELETE and this function only expects those HttpVerbs to be used */
+  private static String getValueBasedOnPatternCase(HttpRule httpRule) {
```
blakeli0, line 329: This method does not have to be `static` now.

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStubTest.java
```diff
@@ -0,0 +1,168 @@
...
+                HttpRule.newBuilder().setGet("testGet").build(),
+                "google.longrunning.Operations.DeleteOperation",
+                HttpRule.newBuilder().setDelete("testDelete").build(),
+                "google.longrunning.Operations.CancelOperation",
+                HttpRule.newBuilder().setPost("testCancel").build()));
+    // The order is: List, Get, Delete, Cancel
+    List<ApiMethodDescriptor> apiMethodDescriptorList =
+        httpJsonOperationsStub.getAllMethodDescriptors();
+    assertThat(apiMethodDescriptorList.get(0).getRequestFormatter().getPathTemplate().toRawString())
+        .isEqualTo("testList");
```
blakeli0, line 75: We could extract this string to a variable in the method or static field in the test class, since they are reused in the setup and verification.

blakeli0 (commented):
LGTM

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpcrest/GrpcRestTestProtoLoader.java
lqiu96: Sure, can rename to `echo_v1beta.yaml`. We can rename it back to `showcase_v1beta1.yaml` if other tests need to use this file again for other showcase protos.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStub.java
lqiu96: Done

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStubTest.java
lqiu96: Done

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/longrunning/stub/HttpJsonOperationsStub.java
lqiu96: MockHttpService requires this in OperationsClientTest. I'll add `@VisibleForTesting` annotation here to mark its intended use case.

blakeli0 (approved):

Issue #1270, opened 2023-01-24
Monorepo migration breaks the pre-commit mvn test step


burkedavison:
Although #1253 has been resolved, the `mvn test` step is now failing. 

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.10.1:compile (default-compile) on project gapic-generator-java: Compilation failure: Compilation failure: 
[ERROR] /gapic-generator-java/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/ServiceClientTestClassComposer.java:[19,43] package com.google.api.gax.httpjson.testing does not exist
[ERROR] /gapic-generator-java/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/ServiceClientTestClassComposer.java:[24,38] package com.google.api.gax.rpc.testing does not exist
[ERROR] /gapic-generator-java/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/grpc/ServiceClientTestClassComposer.java:[18,39] package com.google.api.gax.grpc.testing does not exist
[ERROR] /gapic-generator-java/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/grpc/ServiceClientTestClassComposer.java:[19,39] package com.google.api.gax.grpc.testing does not exist
[ERROR] /gapic-generator-java/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/grpc/ServiceClientTestClassComposer.java:[20,39] package com.google.api.gax.grpc.testing does not exist
[ERROR] /gapic-generator-java/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/grpc/ServiceClientTestClassComposer.java:[21,39] package com.google.api.gax.grpc.testing does not exist
```

Attempting to fix this also unveils further issues with the next bazel integration test invocation.

blakeli0:
Thanks Burke. I discussed this with a few folks as well, I personally don't rely on pre-commit but some folks are still using it, fixing `pre-commit` is on my TODO list. For now, you can always manually run unit tests and integration tests.

lqiu96:
I think this is related to https://github.com/googleapis/gapic-generator-java/issues/1144 -- I think the temp workaround (which doesn't solve the `mvn test` issue at the root) is to cd into each module and run `mvn test` when needed.

PR #1294, opened 2023-02-03
chore: Fix pre-commit.

blakeli0:
fixes: #1270 

Update pre-commit to use `mvn install` instead of `mvn test` as `gapic-generator-java` is dependent on gax test jars. This is also mimicking the same behavior in [ci](https://github.com/googleapis/gapic-generator-java/blob/d1a16195937df041f65a52717ddf9dc6ceb09b4f/.github/workflows/ci-maven.yaml#L25).


lqiu96:
Ran this locally with some Java changes:
```
(14:11:28) ERROR: /private/var/tmp/_bazel_lawrenceqiu/848933e5d291df54c266044a93c65004/external/local_config_cc/BUILD:48:19: in cc_toolchain_suite rule @local_config_cc//:toolchain: cc_toolchain_suite '@local_config_cc//:toolchain' does not contain a toolchain for cpu 'darwin_arm64'
```

Also would it be possible to just run `mvn fmt:format` in the git pre-commit hooks instead of having it fail and then running it seperately?

blakeli0:
> Also would it be possible to just run `mvn fmt:format` in the git pre-commit hooks instead of having it fail and then running it seperately?

I would keep them separately as the `pre-commit` is supposed to be checks only, doing formatting also in `pre-commit` would mixing concerns. Also you would need to run `git add .` again to add the formatted files, which is something I don't think we should do as well in `pre-commit`   

```
(14:11:28) ERROR: /private/var/tmp/_bazel_lawrenceqiu/848933e5d291df54c266044a93c65004/external/local_config_cc/BUILD:48:19: in cc_toolchain_suite rule @local_config_cc//:toolchain: cc_toolchain_suite '@local_config_cc//:toolchain' does not contain a toolchain for cpu 'darwin_arm64'
```
This might be an M1 only issue. Which step did it fail? Integration tests?

lqiu96:
> > Also would it be possible to just run `mvn fmt:format` in the git pre-commit hooks instead of having it fail and then running it seperately?
> 
> I would keep them separately as the `pre-commit` is supposed to be checks only, doing formatting also in `pre-commit` would mixing concerns. Also you would need to run `git add .` again to add the formatted files, which is something I don't think we should do as well in `pre-commit`

I see -- that's a valid point. The only reason I bring this up is because when I had formatting issue in the past, I've just ended up running `mvn fmt:format` immediately after. Unless there is a reason to spot check the formatting issues, I think it *could* be a nice QOL for the pre-commit hooks to just format it.

> ```
> (14:11:28) ERROR: /private/var/tmp/_bazel_lawrenceqiu/848933e5d291df54c266044a93c65004/external/local_config_cc/BUILD:48:19: in cc_toolchain_suite rule @local_config_cc//:toolchain: cc_toolchain_suite '@local_config_cc//:toolchain' does not contain a toolchain for cpu 'darwin_arm64'
> ```
> 
> This might be an M1 only issue. Which step did it fail? Integration tests?


More from the logs:
```
Repository rule jvm_import_external defined at:
  /private/var/tmp/_bazel_lawrenceqiu/848933e5d291df54c266044a93c65004/external/bazel_tools/tools/build_defs/repo/jvm.bzl:217:38: in <toplevel>
(14:11:28) ERROR: Analysis of target '//test/integration:com.google.cloud.iam.credentials.v1.IAMCredentialsClientTest' failed; build aborted: Analysis of target '@local_config_cc//:toolchain' failed
```
Looks to be the golden ITs

lqiu96:
This was resolved when I used v5.x+ of bazel. Possible solution found via: https://github.com/bazelbuild/bazel/issues/13514

lqiu96 (approved):

PR #1343, opened 2023-02-09
fix(spring): update hasRestOption to exclude services with no rest-supported methods

emmileaf:
This PR aligns the spring generator with main branch changes from https://github.com/googleapis/gapic-generator-java/issues/1117. I copied the relevant `isSupportedByTransport(Transport.REST)` logic as a quick fix instead of cherry-picking the commits, since the branches have diverged a bit since the monorepo migration. 

Future note: as part of migrating this spring generator code away from its separate branch and depending on the gapic generator jar, I'd like to revisit this logic to call the corresponding helpers in gapic-generator-java main.

### DIFF
--- a/src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java
+++ b/src/main/java/com/google/api/generator/spring/composer/SpringAutoConfigClassComposer.java
@@ -110,7 +110,17 @@ public class SpringAutoConfigClassComposer implements ClassComposer {
 
     Expr thisExpr = ValueExpr.withValue(ThisObjectValue.withType(dynamicTypes.get(className)));
     Transport transport = context.transport();
-    boolean hasRestOption = transport.equals(Transport.GRPC_REST);
+    // TODO(emmwang): this condition is adapted from latest gapic-generator-java changes as part of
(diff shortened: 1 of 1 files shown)


blakeli0 (approved):
LGTM. Let's try to get this in first and unblock the release, can you please add a test for it in a separate PR? So that we know this functionality is not breaking when we do the migration.

lqiu96 (approved):

emmileaf:
> LGTM. Let's try to get this in first and unblock the release, can you please add a test for it in a separate PR? So that we know this functionality is not breaking when we do the migration.

Sounds good, will merge and follow-up on this since the existing golden tests don’t cover for the additional REST-related logic.

@lqiu96  I saw that you had added `wicked.proto` as part of your changes, and was wondering if you could help me understand additional context on this test service?

emmileaf:
@lqiu96 Gotcha, thanks for the explanation. I think I'll try to re-use as much as this unit testing setup as I can for testing generation behaviour on the spring side as well. Really appreciate all the work you've already laid out here!

lqiu96:
@emmileaf Sounds good! There may be a few lingering issues with the showcase-extended module. Let me know and I can try to help resolve them

PR #1348, opened 2023-02-13
fix(spring): extend hasRestOption fix to properties composer and add golden tests

emmileaf:
This PR is a follow-up on #1343:
- Extends fix to `SpringPropertiesClassComposer`, so that for services without REST-enabled rpcs, the unused `useRest` property is also not generated 
- Adds golden tests for the updated `hasRestOption` scenario using the wicked proto
- Updates `SpringAutoconfigCommentComposer` for javadoc comments alluding to the `useRest` option

### DIFF
--- a/src/main/java/com/google/api/generator/spring/composer/SpringPropertiesClassComposer.java
+++ b/src/main/java/com/google/api/generator/spring/composer/SpringPropertiesClassComposer.java
@@ -75,7 +75,18 @@ public class SpringPropertiesClassComposer implements ClassComposer {
     String className = Utils.getServicePropertiesClassName(service);
     GapicServiceConfig gapicServiceConfig = context.serviceConfig();
     Map<String, TypeNode> dynamicTypes = createDynamicTypes(service, packageName);
-    boolean hasRestOption = context.transport().equals(Transport.GRPC_REST);
+    // TODO(emmwang): this condition is adapted from latest gapic-generator-java changes as part of
+    // https://github.com/googleapis/gapic-generator-java/issues/1117, but should be updated to use
+    // the gapic-implemented helpers once spring generator code is migrated.
+    boolean hasRestSupportedMethod =
+        service.methods().stream()
+            .anyMatch(
+                x ->
+                    x.httpBindings() != null
+                        && x.stream() != Method.Stream.BIDI
+                        && x.stream() != Method.Stream.CLIENT);
+    boolean hasRestOption =
+        context.transport().equals(Transport.GRPC_REST) && hasRestSupportedMethod;
 
     // TODO: this is the prefix user will use to set properties, may need to change depending on
     // branding.
--- a/src/test/java/com/google/api/generator/gapic/composer/grpcrest/GrpcRestTestProtoLoader.java
+++ b/src/test/java/com/google/api/generator/gapic/composer/grpcrest/GrpcRestTestProtoLoader.java
@@ -31,6 +31,7 @@ import com.google.protobuf.Descriptors.FileDescriptor;
 import com.google.protobuf.Descriptors.ServiceDescriptor;
 import com.google.protobuf.StructProto;
 import com.google.showcase.grpcrest.v1beta1.EchoGrpcrest;
+import com.google.showcase.v1beta1.WickedOuterClass;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.HashSet;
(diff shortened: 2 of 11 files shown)


src/test/java/com/google/api/generator/gapic/composer/grpcrest/GrpcRestTestProtoLoader.java
```diff
@@ -82,4 +83,35 @@ public GapicContext parseShowcaseEcho() {
         .setTransport(getTransport())
         .build();
   }
+
+  public GapicContext parseShowcaseWicked() {
```
burkedavison, line 87: Are we going to need to add an entry in this class for every Showcase client that we generate?

src/test/java/com/google/api/generator/gapic/composer/grpcrest/GrpcRestTestProtoLoader.java
```diff
@@ -82,4 +83,35 @@ public GapicContext parseShowcaseEcho() {
...
+    FileDescriptor fileDescriptor = WickedOuterClass.getDescriptor();
+    ServiceDescriptor messagingService = fileDescriptor.getServices().get(0);
+    assertEquals("Wicked", messagingService.getName());
+
+    Map<String, Message> messageTypes = Parser.parseMessages(fileDescriptor);
```
burkedavison, line 92: This logic looks duplicated with the above method's. If this pattern will be repeated for further clients, consider refactoring this common logic out into something like `createGapicContext(fileDescriptor)`

src/test/java/com/google/api/generator/gapic/composer/grpcrest/GrpcRestTestProtoLoader.java
blakeli0: We do need an entry in the test loader for every new test proto as the unit tests need to access the protoc generated Java files. 
The name is a little confusing in this test due to historic reasons, but I don't consider this proto a showcase proto that will be used by showcase clients, this proto is used for unit tests only.

src/main/java/com/google/api/generator/spring/composer/SpringPropertiesClassComposer.java
```diff
@@ -75,7 +75,18 @@ public GapicClass generate(GapicContext context, Service service) {
     String className = Utils.getServicePropertiesClassName(service);
     GapicServiceConfig gapicServiceConfig = context.serviceConfig();
     Map<String, TypeNode> dynamicTypes = createDynamicTypes(service, packageName);
-    boolean hasRestOption = context.transport().equals(Transport.GRPC_REST);
```
blakeli0, line 78: How is this property used in this composer? Why we don't need this logic in https://github.com/googleapis/gapic-generator-java/pull/1343?

emmileaf: @blakeli0 The changes from #1343 removed the generation of SpringAutoConfiguration code that condition on this property (`if (this.clientProperties.getUseRest())`) to set rest-specific defaults. (e.g. [this change to TetherSpringAutoconfiguration](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1586/files#diff-332aa1300b6692a2a98b02f50e233405f3a483a752f68f02878f5a2fb5e41119))

The [property](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/main/spring-cloud-previews/google-cloud-apigee-connect-spring-starter/src/main/java/com/google/cloud/apigeeconnect/v1/spring/TetherSpringProperties.java#L41-L42) itself is still there (just ignored by the autoconfig), and will be removed as part of this PR's changes. I realized when updating these tests that we shouldn't be generating the property either, and it was something I had overlooked in #1343's patch.

blakeli0: I see, thanks for the explanation! It was an unused property hence not causing any compilation issues. 

I did a quick search for `rest` and see that it is mentioned in this [comment](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/c215b9fa8105920fccf09d3be78fc79d195d6089/spring-cloud-previews/google-cloud-apigee-connect-spring-starter/src/main/java/com/google/cloud/apigeeconnect/v1/spring/TetherSpringAutoConfiguration.java#L88) , we may want to update it to something like `unless the useRest option is supported and provided`, as not every client library support rest.

emmileaf: That makes sense - I've updated the comment composer correspondingly. Thanks for the suggestion!

blakeli0 (approved):

PR #1378, opened 2023-02-24
chore: Trying to create issues in nightly build

ddixit14:


.github/workflows/nightly.yaml
```diff
@@ -25,6 +31,14 @@ jobs:
       - run: mvn -version
       - name: Unit Tests
         run: mvn install --errors --batch-mode --no-transfer-progress -Dcheckstyle.skip -T 1C
+      - name: Create issue
```
suztomo, line 34: My guess: the execution does not reach "Create` issue" step if the "mvn install ..." above fails.

.github/workflows/nightly.yaml
```diff
@@ -25,6 +31,14 @@ jobs:
...
+        if: ${{ failure() }}
```
suztomo, line 35: Where is this defined?

.github/workflows/nightly.yaml
```diff
@@ -15,6 +15,12 @@ jobs:
         java: [ 11, 17 ]
     runs-on: ${{ matrix.os }}
     steps:
+      - name: Set up Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: '14.x'
+      - name: Install github-cli
```
suztomo, line 22: I think github-cli, "gh" command, is already available in the runtime (macOS-12, ubuntu-22.04, windows-2022). Did you observe otherwise?

.github/workflows/nightly.yaml
suztomo, line 34:
```suggestion
      - name: Create issue if previous step failed
```

suztomo (commented):
Cam you add the matrix value (os java version) in the title?

.github/workflows/nightly.yaml
ddixit14: You're correct. I've removed the setup steps now.

.github/workflows/nightly.yaml
ddixit14: I think it is default. Any failure in this job would make it true.

suztomo: Thank you. This is the document https://docs.github.com/en/actions/learn-github-actions/expressions#failure.

suztomo (approved):

PR #1425, opened 2023-02-27
feat: add stream methods for `Page`

### LINKED ISSUE #1426: `Page` do not easily support Java Streams
Add stream support for `com.google.api.gax.paging.Page`

JoeWang1127:
Fixes #1426  ☕️


JoeWang1127:
Convert to draft for adding integration tests.

suztomo:
Don't merge this yet. We want to control when to introduce this new feature.

gax-java/gax/clirr-ignored-differences.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- see http://www.mojohaus.org/clirr-maven-plugin/examples/ignored-differences.html -->
+<differences>
+  <difference>
+    <differenceType>7012</differenceType>
```
suztomo, line 5: Add a comment why this is ok.

JoeWang1127: done

JoeWang1127:
I'll merge main branch into this branch when it's ready to merge.

gax-java/gax/clirr-ignored-differences.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- see http://www.mojohaus.org/clirr-maven-plugin/examples/ignored-differences.html -->
+<differences>
+  <difference>
+    <!-- add default stream methods to `Page` interface -->
+    <differenceType>7012</differenceType>
+    <className>com/google/api/gax/paging/Page</className>
+    <method>* stream*(*)</method>
```
suztomo, line 8: It's strange that CLIRR is complaining the change. Adding default method in interface shouldn't cause a breaking changes.

suztomo: @JoeWang1127 Just in case, would you build this gax-java locally (with -SNAPSHOT version) and try to use in java-storage to see anything breaks? You ran some storage samples that you used to understand the request.

JoeWang1127: In [Specifying ignored differences](https://www.mojohaus.org/clirr-maven-plugin/examples/ignored-differences.html) of Clirr Maven Plugin doc, 7012 has the following definition:
```
7012 (Method Added to Interface): className, method
```
So adding a method to interface, default or not, is a breaking change to this rule.

Also, there's 7013:
```
7013 (Abstract Method Added to Class): className, method
```
which is not complain.

JoeWang1127:
@suztomo I've verified the change in the demo project (on branch [`use-stream`](https://github.com/JoeWang1127/storage-demo/tree/use-stream)).

suztomo:
Thank you.

gax-java/gax/src/main/java/com/google/api/gax/paging/Page.java
```diff
@@ -52,12 +55,26 @@
...
   /** Returns an iterable over the elements in this page. */
   Iterable<ResourceT> getValues();
+
+  /**
+   * Returns a stream that traverses all the elements of the underlying data source. The data is
+   * fetched lazily page by page, where each page may contain multiple elements. A new page is
+   * fetched whenever the elements of any particular page are exhausted.
+   */
+  default Stream<ResourceT> streamAll() {
+    return StreamSupport.stream(iterateAll().spliterator(), false);
```
suztomo, line 73: @BenWhitehead Would you elaborate what's not ideal? Parallel Stream.

gax-java/gax/clirr-ignored-differences.xml
BenWhitehead: Clirr isn't aware of java 8 default methods. It is looking at the bytecode from the perspective of java7 binary compatibility. When adding a default method it is safe to specify an ignore rule.

gax-java/gax/src/main/java/com/google/api/gax/paging/Page.java
BenWhitehead: I think this is great for a first step of getting it on the api and saving our customers from having to do the same hoop jumping.

In the case of future improvements, right now this approach will not allow of parallel processing the same way a natively parallel friendly stream would. In all actuality, Page isn't fully parallelizable due to the fact that each page contains the `nextPageToken` necessary in order to fetch the next page. But, Page could be more parallel friendly where processing of the actual resources can happen on a thread separate from the thread performing the actual get.

suztomo: @JoeWang1127 Do you have question or comment about parallelism? (I don't think of an implementation as of now)

BenWhitehead: I agree, parallelism does not need to block this PR and can be an additional improvement in the future.

JoeWang1127: @suztomo I'm not understand about the separate thread in Ben's comment:
```
Page could be more parallel friendly where processing of the actual resources can happen on a thread separate from the thread performing the actual get.
```

suztomo: This https://docs.oracle.com/javase/tutorial/collections/streams/parallelism.html#executing_streams_in_parallel

gax-java/gax/clirr-ignored-differences.xml
JoeWang1127: @BenWhitehead Thanks for the clarification.

gax-java/gax/src/main/java/com/google/api/gax/paging/Page.java
JoeWang1127: Thanks for the reference.

I don't have a question about it.

gax-java/gax/src/test/java/com/google/api/gax/rpc/PagingTest.java
```diff
@@ -133,6 +133,54 @@ public void pagedByPage() {
     Truth.assertThat(requestCapture.getAllValues()).containsExactly(0, 2, 4).inOrder();
   }
 
+  @Test
+  public void streamedByPage() {
```
blakeli0, line 137: Can we update the test names based on the [best practices](https://engdoc.corp.google.com/eng/doc/devguide/testing/unit/best-practices.md?cl=head&polyglot=java#naming) and a recent [TotT](https://engdoc.corp.google.com/eng/doc/tott/episodes/660.md?cl=head)?

JoeWang1127: done

blakeli0 (approved):

PR #1430, opened 2023-03-02
chore: add aggregate test coverage collection of gax within the showcase and gax modules

mpeddada1:
### Coverage
In order to get test coverage of gax within the showcase and gax modules, a new module `coverage-report` needed to be created. The coverage module declares gapic-showcase (with a test scope), gax, gax-grpc, gax-httpjson as dependencies and relies on `jacoco:report-aggregate` to collect aggregate metrics. 

#### Unit test coverage 
Gathering just unit test metrics requires `mvn clean test -DenableTestCoverage` to be called. 
Screenshot 2023-03-03 at 6 32 50 PM

#### Integration test coverage
Gathering just integration test metrics requires `mvn clean verify -DskipUnitTests -DenableTestCoverage -Penable-integration-tests` to be called. The `DskipUnitTests` is needed skip the maven-surefire-plugin. 
Screenshot 2023-03-03 at 6 36 26 PM


mpeddada1:
graalvm downstream check is resulting in `Error:  Test configuration file wasn't found. Make sure that test execution wasn't skipped.` The new maven surefire and maven failsafe settings seem to be overriding the one in the `native` plugin from shared-config.

pom.xml
```diff
@@ -21,6 +21,8 @@
     <module>java-common-protos</module>
     <module>java-iam</module>
     <module>gapic-generator-java-bom</module>
+    <module>showcase</module>
```
burkedavison, line 24: Does this have any impact on the recommended workflows in `DEVELOPMENT.md`? (Particularly, are there going to be any issues with the `mvn install` command -- or will showcase testing largely just be skipped? Will this add any significant delay to the development cycle now that Showcase is being generated, or will generation not be a burden?)

gax-java/pom.xml
```diff
@@ -222,6 +222,7 @@
         <artifactId>maven-jar-plugin</artifactId>
         <executions>
           <execution>
+            <phase>test</phase>
```
burkedavison, line 225: Could you clarify what this is doing?

@blakeli0 , @suztomo : Is it possible that the `<skip>true</skip>` three lines down from here is the reason why we have to execute `mvn install` rather than `mvn test` to avoid linking errors?

gapic-generator-java-pom-parent/pom.xml
```diff
@@ -95,8 +95,86 @@
         </plugins>
       </build>
     </profile>
+
+    <profile>
+      <id>test-coverage</id>
```
burkedavison, line 100: The PR description does a great job clarifying how this is meant to be used, but that may get lost in obscurity once the PR is merged. Could you add some information in an appropriate `.md` to explain the different expected workflows + report options?

gax-java/pom.xml
mpeddada1: Sure! This line helps resolve the compilation error we see when running `mvn test` at the root. The [PR description](https://github.com/googleapis/gapic-generator-java/pull/1430#issue-1607483042) goes over this briefly but upon comparing the `mvn test` with `mvn install`, I noticed that the `jar:test-jar` step which creates the `testlib` JAR that gapic-generator-java depends on was being executed when `mvn install` was called but **after** `surefire:test` step. On the other hand, it was not being invoked at all when `mvn test` was executed. This change forces the maven-jar-plugin of GAX to be executed at the `test` phase.

pom.xml
burkedavison: Sounds great, thank you.

gax-java/pom.xml
burkedavison: @blakeli0 and @suztomo for visibility. Great work!

gapic-generator-java-pom-parent/pom.xml
mpeddada1: Good idea, done. I've added a README to the `coverage-report` module.

pom.xml
blakeli0, line 24: Adding these two modules to the root pom here may have impact to our release, that they might be released automatically. @suztomo Do you have any concerns here?

suztomo: @mpeddada1 To address the concern of accidental releases, can we define a profile so that the showcase module is part of the root project only when the profile is enabled?

mpeddada1: Good idea! Done.

mpeddada1:
@suztomo @blakeli0 Ready for another round of review.

blakeli0 (approved):

suztomo:
> As described in #1270, calling mvn test was resulting in compilation errors for gax test-jars.

https://github.com/googleapis/gapic-generator-java/pull/1474 might resolve this problem. If you had to implement some workaround for GAX test JARs, would you revisit it? If not applicable, then that's good.

mpeddada1:
> #1474 might resolve this problem. If you had to implement some workaround for the GAX test JARs, would you revisit it? If not applicable, then that's good.

Awesome! Thank you. It works without the workaround:)

coverage-report/README.md
```diff
@@ -0,0 +1,16 @@
+## Coverage Report
+
+This module is meant to gather aggregate jacoco test coverage metrics across the `gax-java` and `showcase` modules.
```
suztomo, line 3:
```suggestion
This module gathers aggregate jacoco test coverage metrics across the `gax-java` and `showcase` modules.
```

coverage-report/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+### Unit Test Coverage
+In order to view aggregate unit test coverage of GAX in both `gax-java` and `showcase`:
+
+1. At the root of the repository, run `mvn clean test -DenableTestCoverage`.
+2. The metrics can be found at `${HOME}/gapic-generator-java/coverage-report/target/site/jacoco-aggregate/index.html`
```
suztomo, line 9: I believe the home directory is irrelevant?

coverage-report/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+### Integration Test Coverage
+
+In order to view aggregate integration test coverage of GAX in both `gax-java` and `showcase`:
+
+1. At the root of the repository, run `mvn clean verify -DskipUnitTests -DenableTestCoverage -Penable-integration-tests`.
+2. The metrics can be found at `${HOME}/gapic-generator-java/coverage-report/target/site/jacoco-aggregate/index.html`
```
suztomo, line 16: I believe the home directory is irrelevant?

coverage-report/README.md
```diff
@@ -0,0 +1,16 @@
+## Coverage Report
```
suztomo, line 1: Would you add the nice screenshots you prepared in this pull request description?

coverage-report/README.md
suztomo, line 3: Would you add why the coverage of the 2 modules, gax-java and showcase, is important?

gapic-generator-java-pom-parent/pom.xml
```diff
@@ -95,8 +95,91 @@
...
+              </execution>
+            </executions>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
   </profiles>
 
+
+
```
suztomo, line 182: Remove unnecessary lines.

suztomo (changes requested):
Would you add a check in .github/workflows (either existing one or new file) to run the two commands for unit tests and integration tests.

suztomo (approved):

coverage-report/README.md
mpeddada1: Done, thank you!

coverage-report/README.md
mpeddada1: Agreed, starting with `gapic-generator-java` should be sufficient.

coverage-report/README.md
mpeddada1: Added more info on the purpose of the metrics being collected.

gapic-generator-java-pom-parent/pom.xml
mpeddada1: Done.

coverage-report/README.md
mpeddada1: Good idea, done.

mpeddada1:
> Would you add a check in .github/workflows (either existing one or new file) to run the two commands for unit tests and integration tests.

Yup! I modified the `Unit Tests` check in ci-maven.yaml to use `mvn test ` instead of `mvn install` now that the issue has been fixed. Additionally, the PR after this (https://github.com/googleapis/gapic-generator-java/pull/1482) is going to add both commands  (`mvn test` and `mvn verify`) on the whole repo to `sonar.yaml` for piping the metrics to Sonarcloud.

.github/workflows/ci-maven.yaml
```diff
@@ -22,9 +22,12 @@ jobs:
       - run: java -version
       - name: Unit Tests
         run: |
-          mvn install --batch-mode --no-transfer-progress -Dcheckstyle.skip \
+          mvn test --batch-mode --no-transfer-progress -Dcheckstyle.skip \
```
suztomo, line 25: `-DenableTestCoverage `?

suztomo (commented):
I meant a check to ensure “ -DenableTestCoverage” works fine. Does it exist already via SonarCloud?

.github/workflows/ci-maven.yaml
mpeddada1: I see! thanks for clarifying. That check will be included as part of the sonar job in the follow-up PR: https://github.com/googleapis/gapic-generator-java/pull/1482/

PR #1473, opened 2023-03-09
feat: make stream wait timeout a first class citizen

mutianf:
Port change https://github.com/googleapis/gax-java/pull/1409 because gax is going to have a major version bump. 
The original PR was closed because remapping rpcTimeout is too risky. But now gax is having a major version bump soon we're hoping to fix this behavior.

From the original comment:

For a server streaming api, there are 4 conceptual timeouts:

1. overall operation timeout - the maximum amount time that passes from the user invoking a method until that method exits
2. attempt/rpc timeout - if retries are enabled the maximum amount of time that passes for each attempt in an operation
3. message wait timeout - the maximum amount of time to wait for the next message from the server
4. idle timeout - how long to wait before considering the stream orphaned by the user and closing it
Each has a usecase:

1. operation timeout is useful for users to fulfill their own slo guarantees
2. attempt timeout are useful when a client developer knows the absolute limit of an rpc but that limit happens to be shorter than the required slo for a customer. For example a point read of a bigtable key shouldnt ever take more than 100ms. If it does, then we can assume something is wrong (GFE died without sending a FIN packet) and abort the request and retry.
3. message wait timeouts have a similar use as attempt timeouts with tighter guarantees
4. idle timeouts are useful to reduce buffer bloat on the server
Currently all of them are implemented in gax but the delineation was muddied by me a while back. This PR tries to fix the situation. In the current world, operation timeout is defined by RetrySettings#totalTimeout, idle timeout is defined by ServerStreamingCallSettings#idleTimeout. However RetrySettings#rpcTimeout is mapped to message wait timeout and attempt timeout is only configureable per call using ApiCallContext#withTimeout.

This PR cleans up the situation by mapping RetrySettings#rpcTimeout to attempt timeouts and exposes a new setting for wait timeout on ServerStreamingCallSettings.

### DIFF
--- a/gax-java/gax-grpc/src/test/java/com/google/api/gax/grpc/TimeoutTest.java
+++ b/gax-java/gax-grpc/src/test/java/com/google/api/gax/grpc/TimeoutTest.java
@@ -242,7 +242,12 @@ public class TimeoutTest {
             .build();
     Duration newTimeout = Duration.ofSeconds(5);
     RetrySettings contextRetrySettings =
-        retrySettings.toBuilder().setTotalTimeout(newTimeout).setMaxAttempts(3).build();
+        retrySettings
+            .toBuilder()
+            .setInitialRpcTimeout(newTimeout)
+            .setMaxRpcTimeout(newTimeout)
+            .setMaxAttempts(3)
+            .build();
     GrpcCallContext retryingContext =
         defaultCallContext
             .withRetrySettings(contextRetrySettings)
--- a/gax-java/gax/src/main/java/com/google/api/gax/rpc/Callables.java
+++ b/gax-java/gax/src/main/java/com/google/api/gax/rpc/Callables.java
@@ -115,7 +115,8 @@ public class Callables {
         callable.withDefaultCallContext(
             clientContext
                 .getDefaultCallContext()
-                .withStreamIdleTimeout(callSettings.getIdleTimeout()));
+                .withStreamIdleTimeout(callSettings.getIdleTimeout())
+                .withStreamWaitTimeout(callSettings.getWaitTimeout()));
 
     return callable;
   }
--- a/gax-java/gax/src/main/java/com/google/api/gax/rpc/ServerStreamingCallSettings.java
+++ b/gax-java/gax/src/main/java/com/google/api/gax/rpc/ServerStreamingCallSettings.java
@@ -47,9 +47,11 @@ import org.threeten.bp.Duration;
  * <p>This class includes settings that are applicable to all server streaming calls, which
  * currently just includes retries and watchdog timers.
  *
- * <p>The watchdog timer is configured via {@code idleTimeout}. The watchdog will terminate any
- * stream that has not has seen any demand (via {@link StreamController#request(int)}) in the
- * configured interval. To turn off idle checks, set the interval to {@link Duration#ZERO}.
+ * <p>The watchdog timer is configured via {@code idleTimeout} and {@code waitTimeout}. The watchdog
+ * will terminate any stream that has not has seen any demand (via {@link
+ * StreamController#request(int)}) in the configured interval or has not seen a message from the
+ * server in {@code waitTimeout}. To turn off idle checks, set the interval to {@link
+ * Duration#ZERO}.
  *
  * <p>Retry configuration allows for the stream to be restarted and resumed. It is composed of 3
  * parts: the retryable codes, the retry settings and the stream resumption strategy. The retryable
@@ -79,12 +81,14 @@ public final class ServerStreamingCallSettings<RequestT, ResponseT>
   @Nonnull private final StreamResumptionStrategy<RequestT, ResponseT> resumptionStrategy;
 
   @Nonnull private final Duration idleTimeout;
+  @Nonnull private final Duration waitTimeout;
 
   private ServerStreamingCallSettings(Builder<RequestT, ResponseT> builder) {
     this.retryableCodes = ImmutableSet.copyOf(builder.retryableCodes);
     this.retrySettings = builder.retrySettingsBuilder.build();
     this.resumptionStrategy = builder.resumptionStrategy;
     this.idleTimeout = builder.idleTimeout;
+    this.waitTimeout = builder.waitTimeout;
   }
 
   /**
@@ -123,6 +127,15 @@ public final class ServerStreamingCallSettings<RequestT, ResponseT>
     return idleTimeout;
   }
 
+  /**
+   * See the class documentation of {@link ServerStreamingCallSettings} for a description of what
+   * the {@link #waitTimeout} does.
+   */
+  @Nonnull
+  public Duration getWaitTimeout() {
+    return waitTimeout;
+  }
+
   public Builder<RequestT, ResponseT> toBuilder() {
     return new Builder<>(this);
   }
@@ -135,6 +148,7 @@ public final class ServerStreamingCallSettings<RequestT, ResponseT>
   public String toString() {
     return MoreObjects.toStringHelper(this)
         .add("idleTimeout", idleTimeout)
+        .add("waitTimeout", waitTimeout)
         .add("retryableCodes", retryableCodes)
         .add("retrySettings", retrySettings)
         .toString();
@@ -148,6 +162,8 @@ public final class ServerStreamingCallSettings<RequestT, ResponseT>
 
     @Nonnull private Duration idleTimeout;
 
+    @Nonnull private Duration waitTimeout;
+
     /** Initialize the builder with default settings */
     private Builder() {
       this.retryableCodes = ImmutableSet.of();
@@ -155,6 +171,7 @@ public final class ServerStreamingCallSettings<RequestT, ResponseT>
       this.resumptionStrategy = new SimpleStreamResumptionStrategy<>();
 
       this.idleTimeout = Duration.ZERO;
+      this.waitTimeout = Duration.ZERO;
     }
 
     private Builder(ServerStreamingCallSettings<RequestT, ResponseT> settings) {
@@ -164,6 +181,7 @@ public final class ServerStreamingCallSettings<RequestT, ResponseT>
       this.resumptionStrategy = settings.resumptionStrategy;
 
       this.idleTimeout = settings.idleTimeout;
+      this.waitTimeout = settings.waitTimeout;
     }
 
     /**
@@ -233,9 +251,9 @@ public final class ServerStreamingCallSettings<RequestT, ResponseT>
               .setInitialRetryDelay(Duration.ZERO)
               .setRetryDelayMultiplier(1)
               .setMaxRetryDelay(Duration.ZERO)
-              .setInitialRpcTimeout(Duration.ZERO)
+              .setInitialRpcTimeout(timeout)
               .setRpcTimeoutMultiplier(1)
-              .setMaxRpcTimeout(Duration.ZERO)
+              .setMaxRpcTimeout(timeout)
               .setMaxAttempts(1)
               .build());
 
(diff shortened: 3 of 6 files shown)


suztomo:
@mutianf We planned a major release but it turned out that we can implement many of the ideas without introducing breaking changes. So GAX is not going to have a major release.

However, I think your idea here is worth discussing. Would you add it into the wishlist? https://docs.google.com/document/d/1oEzsIEr7DFoyG6-QXXZyIyjIZkbe8i2obLOG90RnVKk/edit?resourcekey=0-T1uuNtaLJBuOtIlcRfe2iQ (internal doc)

mutianf:
Thanks @suztomo, I added to the bottom of the list. Feel free to add me to meetings that discuss these issues!

gax-java/gax/src/main/java/com/google/api/gax/rpc/ServerStreamingCallSettings.java
```diff
@@ -79,12 +81,14 @@
   @Nonnull private final StreamResumptionStrategy<RequestT, ResponseT> resumptionStrategy;
 
   @Nonnull private final Duration idleTimeout;
+  @Nonnull private final Duration waitTimeout;
```
blakeli0, line 84: Can we use `java.time.Duration` instead? We no longer support Java 7 and plan to deprecate `org.threeten.bp.Duration` soon.

gax-java/gax/src/main/java/com/google/api/gax/rpc/Callables.java
```diff
@@ -115,7 +115,8 @@ public static <RequestT, ResponseT> ServerStreamingCallable<RequestT, ResponseT>
         callable.withDefaultCallContext(
             clientContext
                 .getDefaultCallContext()
-                .withStreamIdleTimeout(callSettings.getIdleTimeout()));
+                .withStreamIdleTimeout(callSettings.getIdleTimeout())
+                .withStreamWaitTimeout(callSettings.getWaitTimeout()));
```
lqiu96, line 119: Do you know what the behavior is when either (or both) the idleTimeout and waitTimeout values are greater than the Watchdog's scheduleInterval value? I'm guessing that the Watchdog does nothing and makes me wonder if there should (or needs to) be a check to ensure that the scheduleInterval is less than both of these timeout values (when the values are non-zero).

Kinda like how we have a check if RPC timeout is greater than total timeout: `java.lang.IllegalStateException: max rpc timeout must not be shorter than initial timeout`

gax-java/gax/src/main/java/com/google/api/gax/rpc/ServerStreamingCallSettings.java
```diff
@@ -233,9 +251,9 @@ public Builder<RequestT, ResponseT> setSimpleTimeoutNoRetries(@Nonnull Duration
               .setInitialRetryDelay(Duration.ZERO)
               .setRetryDelayMultiplier(1)
               .setMaxRetryDelay(Duration.ZERO)
-              .setInitialRpcTimeout(Duration.ZERO)
+              .setInitialRpcTimeout(timeout)
               .setRpcTimeoutMultiplier(1)
-              .setMaxRpcTimeout(Duration.ZERO)
+              .setMaxRpcTimeout(timeout)
```
lqiu96, line 256: I might be missing something. Does this need to change? The ServerStreamAtttemptCallable only sets a timeout if the RPC Timeout is non-zero. For this case (the simple timeout + no retry), the RPC Timeout would be zero (from the initial RPC timeout value) and no timeout would be set (which *should* cancel once we hit the total timeout).

lqiu96:
I think the changes make sense. Just left a few questions (trying to get more familiar with the streaming retry logic).

gax-java/gax/src/main/java/com/google/api/gax/rpc/ServerStreamingCallSettings.java
```diff
@@ -272,6 +290,19 @@ public Builder<RequestT, ResponseT> setIdleTimeout(@Nonnull Duration idleTimeout
...
+    @Nonnull
+    public Duration getWaitTimeout() {
+      return waitTimeout;
+    }
+
+    /**
+     * See the class documentation of {@link ServerStreamingCallSettings} for a description of what
+     * the {@link #waitTimeout} does. {@link Duration#ZERO} disables the watchdog.
+     */
+    public void setWaitTimeout(@Nonnull Duration waitTimeout) {
```
blakeli0, line 302: How do we plan to use it? Are we always expecting this field to be set by customers? Introducing a `Nonnull` public field to an existing setting is a breaking change because customers would have to update their code. We should have a default value for these timeout.

blakeli0 (commented):
Thanks Mattie for opening this PR! It would be great if we can add a showcase integration test to visualize how customers would use it, and maybe test the behavior of `streamWaitTimeout` as well. We already have a basic [showcase integration tests](https://github.com/googleapis/gapic-generator-java/blob/main/showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITServerSideStreaming.java) for server side streaming. 
See the [README](https://github.com/googleapis/gapic-generator-java/blob/main/showcase/README.md) of showcase module for more details. Ley me know if you have any additional questions regarding showcase integration tests, @burkedavison /@lqiu96 / @mpeddada1 can provide more info as well.

gax-java/gax/src/main/java/com/google/api/gax/rpc/ServerStreamingCallSettings.java
mutianf: Can this be a separate follow up PR to update the Duration type? I think it'll be cleaner that way.

gax-java/gax/src/main/java/com/google/api/gax/rpc/Callables.java
mutianf: I think watchdog will kill the stream anyways whenever it sees the idleTimeout and waitTimeout are expired. So I don't feel it's necessary to compare timeouts with checkInterval here?

mutianf: Also the default checkInterval is pretty small (10 seconds). So I don't think it'll be a problem.

gax-java/gax/src/main/java/com/google/api/gax/rpc/ServerStreamingCallSettings.java
mutianf: Builder() initialize it with 0 (line 174) and we don't always expect user to set it.

mutianf:
Thanks for reviewing! I'll add showcase integration test and address other comments later.

gax-java/gax-grpc/src/test/java/com/google/api/gax/grpc/TimeoutTest.java
```diff
@@ -242,7 +242,12 @@ public void testNonRetryServerStreamingSettingsContextWithRetry() {
             .build();
     Duration newTimeout = Duration.ofSeconds(5);
     RetrySettings contextRetrySettings =
-        retrySettings.toBuilder().setTotalTimeout(newTimeout).setMaxAttempts(3).build();
+        retrySettings
+            .toBuilder()
+            .setInitialRpcTimeout(newTimeout)
+            .setMaxRpcTimeout(newTimeout)
```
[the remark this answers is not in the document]
mutianf: Yes, this is testing resetting the deadline from the context rpc timeout: https://github.com/googleapis/gax-java/blob/5e76b4b4487d90623dd8daa0b0249505fa97c99d/gax/src/main/java/com/google/api/gax/rpc/ServerStreamingAttemptCallable.java#L223-L228

gax-java/gax/src/main/java/com/google/api/gax/rpc/ServerStreamingCallSettings.java
blakeli0: I'm OK with it if the follow up PR is merged before this PR gets released. Otherwise it would be a breaking change to change the type.

gax-java/gax/src/main/java/com/google/api/gax/rpc/ServerStreamingCallSettings.java
blakeli0: > Builder() initialize it with 0 (line 174) and we don't always expect user to set it.

Thanks, what does `Duration.ZERO` mean in this case? We wait indefinitely in between messages for ServerSreaming RPCs? Do we still expect customers to set it sometimes?

gax-java/gax/src/main/java/com/google/api/gax/rpc/Callables.java
blakeli0, line 119: Do we have a unit test verifies that we are getting `waitTimeout` from `callSettings` now?

gax-java/gax/src/main/java/com/google/api/gax/rpc/ServerStreamingCallSettings.java
lqiu96: Ah, gotcha. I missed the part where the totalTimeout was removed. Makes sense that it uses the RpcTimeout value now.

gax-java/gax/src/main/java/com/google/api/gax/rpc/Callables.java
lqiu96: Actually, on a second look .. I think may be misunderstanding how Watchdog works. I think this is fine.

gax-java/gax/src/main/java/com/google/api/gax/rpc/ServerStreamingCallSettings.java
mutianf: 0 means disable watchdog. This default is override in the client library https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java#LL641C50-L641C50 

we expect customer to set waitTimeout if they expect longer wait time between responses.

gax-java/gax/src/main/java/com/google/api/gax/rpc/ServerStreamingCallSettings.java
mutianf: I see, but idleTimeout is already using `org.threeten.bp.Duration`. So maybe it's worth updating this in the next release? I'm not sure if having mixed Duration types in the same class is a good idea :(

gax-java/gax/src/main/java/com/google/api/gax/rpc/Callables.java
mutianf: Added a test in ServerStreamingCallSettingsTest, similar to idleTimeout.

gax-java/gax/src/main/java/com/google/api/gax/rpc/ServerStreamingCallSettings.java
blakeli0: Usually it's not, we plan to deprecate `org.threeten.bp.Duration` soon and add an alternative method with `java.time.Duration` for all threeten methods, it would be great if we can change it now, especially if we don't expect customers to set it often.
e.g.  For
```
public Duration getWaitTimeout() {
    return waitTimeout;
  }
```
We will have to add an alternative method
```
public java.time.Duration getWaitTimeoutDuration() {
    return waitTimeout;
  }
```
with a slightly different name as well, which is not ideal.
I'm OK with keeping it this way as well if you feel strongly, but want to share with you our plan and potential consequences.

gax-java/gax/src/main/java/com/google/api/gax/rpc/ServerStreamingCallSettings.java
mutianf: Make sense, I updated the javadoc

gax-java/gax/src/main/java/com/google/api/gax/rpc/ServerStreamingCallSettings.java
mutianf: Chatted offline, we'll keep the threeten type for now :)

blakeli0 (approved):

igorbernstein2 (approved):
Awesome! thanks for fixing this! and I apologize for the original misfeature

Issue #1440, opened 2023-03-07
GAPIC HttpJson additional_bindings annotation is not being used


lqiu96:
Discovered while implementing Showcase testing for the ComplianceSuite. This fails the `Compliance.RepeatDataPathTrailingResource` test case: `Testing group: `Binding selection testing`- RPC Name: `Compliance.RepeatDataPathResource` with Request Name: `Binding testing additional binding`.

The path template for `Compliance.RepeatDataPathResource`:
`/v1beta1/repeat/{info.fString=first/*}/{info.fChild.fString=second/*}/bool/{info.fBool}:pathresource`. The data is 
```
"info": {
            "fString": "second/greetings",
            "pBool": true,

            "fChild": {
              "fString": "first/hello"
            }
          }
```
Since `info.fString` exists, it spliced in as `second/greetings` and `info.fChild.fString` exists, so it's spliced in as `first/hello`. But the path template is expecting something matching `first/*` which `second/greetings` doesn't match. This is same issue with `second/**` not matching with `first/hello`. 

The additional_binding for this RPC is `/v1beta1/repeat/{info.fChild.fString=first/*}/{info.fString=second/*}/bool/{info.fBool}:childfirstpathresource` and this will match the pathTemplate correctly.

Our logic is wrong in that:
1. It doesn't validate the match
2. Doesn't even have logic to check the additional_bindings

Error Logs:
Java side:
```
Caused by: com.google.api.client.http.HttpResponseException: 400 Bad Request
GET http://localhost:7469/v1beta1/repeat/second/greetings/first/hello/bool/false:pathresource?fInt64=0&info.pBool=true&intendedBindingUri=/v1beta1/repeat/%7Binfo.f_child.f_string%3Dfirst/*%7D/%7Binfo.f_string%3Dsecond/*%7D/bool/%7Binfo.f_bool%7D:childfirstpathresource&serverVerify=true&name=Binding%20testing%20additional%20binding&info.fString=second/greetings&fInt32=0&fDouble=0.0&info.fChild.fString=first/hello
[...]

PR #1522, opened 2023-03-21
fix: Validate paths and check additionalPathTemplates

lqiu96:

Part of #1440

---
We validate that the default path or any additional path matches. Otherwise, it is possible to have an invalid path returned and will result in a 404 when making the call.

### DIFF
--- a/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ProtoMessageRequestFormatter.java
+++ b/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ProtoMessageRequestFormatter.java
@@ -116,7 +116,8 @@ public class ProtoMessageRequestFormatter<RequestT extends Message>
         return additionalPath;
       }
     }
-    // If there are no matches, we return the default path
+    // If there are no matches, we return the default path, this is for backwards compatibility.
+    //TODO: Log this scenario once we implemented the Cloud SDK logging.
     return path;
   }
 
--- a/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ProtoMessageRequestFormatter.java
+++ b/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ProtoMessageRequestFormatter.java
@@ -117,7 +117,7 @@ public class ProtoMessageRequestFormatter<RequestT extends Message>
       }
     }
     // If there are no matches, we return the default path, this is for backwards compatibility.
-    //TODO: Log this scenario once we implemented the Cloud SDK logging.
+    // TODO: Log this scenario once we implemented the Cloud SDK logging.
     return path;
   }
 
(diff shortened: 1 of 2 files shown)


gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ProtoMessageRequestFormatter.java
```diff
@@ -95,10 +95,28 @@ public String getRequestBody(RequestT apiMessage) {
...
   @Override
   public String getPath(RequestT apiMessage) {
-    return pathTemplate.instantiate(pathVarsExtractor.extract(apiMessage));
+    String path = pathTemplate.instantiate(pathVarsExtractor.extract(apiMessage));
+    if (pathTemplate.matches(path)) {
+      return path;
+    }
+    for (PathTemplate additionalPathTemplate : additionalPathTemplates) {
+      String additionalPath =
+          additionalPathTemplate.instantiate(pathVarsExtractor.extract(apiMessage));
```
burkedavison, line 114: Is `pathVarsExtractor.extract(apiMessage)` expected to return different values each time?

If not, please save the result to a local variable in the first line of this method rather than repeating this call.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ProtoMessageRequestFormatter.java
```diff
@@ -95,10 +95,28 @@ public String getRequestBody(RequestT apiMessage) {
...
+      if (additionalPathTemplate.matches(additionalPath)) {
+        return additionalPath;
+      }
+    }
+    throw new IllegalArgumentException("No matching paths for Request: " + apiMessage);
```
blakeli0, line 119: Was it silently failing before? If it was, throwing an exception now could be considered a breaking change.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ProtoMessageRequestFormatter.java
lqiu96: Ah, will update! I was prototyping with part 2 which has `pathVarsExtractor.extract(apiMessage)` return different values each time.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ProtoMessageRequestFormatter.java
lqiu96: Yeah, it would making a request to the invalid URL and would result in a 404. I can update this return the default path's raw string instead of the exception.

blakeli0 (approved):

lqiu96:
Failing for `[java_showcase_integration_tests] SonarCloud Code Analysis`. This new code change will be covered by the compliance proto. I plan to add it once I resolve part 2 of this issue.

PR #1535, opened 2023-03-22
ci: installing shared dependencies BOM upper-bound check

suztomo:
Moving the removed job from https://github.com/googleapis/google-cloud-java/pull/9266/files#r1145267872 to gapic-generator-java.

### DIFF
--- a/.github/workflows/shared_dependencies.yaml
+++ b/.github/workflows/shared_dependencies.yaml
@@ -4,6 +4,8 @@ on:
     - main
   pull_request:
     paths:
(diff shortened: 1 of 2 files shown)


.github/workflows/shared_dependencies.yaml
suztomo, line 1: This is from https://github.com/googleapis/google-cloud-java/pull/9266/files#r1145267872

ddixit14 (approved):

burkedavison (approved):

PR #1553, opened 2023-03-24
Print watchdog timeout in error message

kongweihan:
So that we can easily debug whether the client has set the correct timeout.


gax-java/gax/src/main/java/com/google/api/gax/rpc/Watchdog.java
```diff
@@ -318,14 +318,14 @@ boolean cancelIfStale() {
         switch (this.state) {
           case IDLE:
             if (!idleTimeout.isZero() && waitTime >= idleTimeout.toMillis()) {
-              myError = new WatchdogTimeoutException("Canceled due to idle connection", false);
+              myError = new WatchdogTimeoutException("Canceled due to idle connection. idle timeout is " + idleTimeout.toMillis() + " ms", false);
```
mutianf, line 321: idle timeout is always 5 minutes, so I don't think we need to log it here.

kongweihan: Sure, but the people seeing this log ling likely wouldn't know. And it's possible that we'll change it in the future?

mutianf (approved):

mutianf:
Please fix the commit message

gax-java/gax/src/main/java/com/google/api/gax/rpc/Watchdog.java
mutianf: Fair. I don't think we ever want to let customer configure idleTimeout though.

gax-java/gax/src/main/java/com/google/api/gax/rpc/Watchdog.java
```diff
@@ -318,14 +318,26 @@ boolean cancelIfStale() {
...
           case IDLE:
             if (!idleTimeout.isZero() && waitTime >= idleTimeout.toMillis()) {
-              myError = new WatchdogTimeoutException("Canceled due to idle connection", false);
+              myError =
+                  new WatchdogTimeoutException(
+                      "Canceled due to idle connection. waitTime="
+                          + waitTime
+                          + "ms idleTimeout="
+                          + idleTimeout.toMillis()
+                          + "ms",
```
igorbernstein2, line 327: I would rephrase this:

"Canceled idle RPC after X ms (threshold: Y ms)

gax-java/gax/src/main/java/com/google/api/gax/rpc/Watchdog.java
```diff
@@ -318,14 +318,26 @@ boolean cancelIfStale() {
...
           case WAITING:
             if (!waitTimeout.isZero() && waitTime >= waitTimeout.toMillis()) {
               myError =
                   new WatchdogTimeoutException(
-                      "Canceled due to timeout waiting for next response", true);
+                      "Canceled due to timeout waiting for next response. waitTime="
+                          + waitTime
+                          + "ms waitTimeout="
+                          + waitTimeout.toMillis()
+                          + "ms",
```
igorbernstein2, line 339: Similar as above:

"Cancel RPC after waiting for a server response for X ms (threshold: Y ms)

gax-java/gax/src/main/java/com/google/api/gax/rpc/Watchdog.java
kongweihan: Done.

gax-java/gax/src/main/java/com/google/api/gax/rpc/Watchdog.java
kongweihan: Done.

gax-java/gax/src/main/java/com/google/api/gax/rpc/Watchdog.java
```diff
@@ -318,14 +318,26 @@ boolean cancelIfStale() {
         switch (this.state) {
           case IDLE:
             if (!idleTimeout.isZero() && waitTime >= idleTimeout.toMillis()) {
-              myError = new WatchdogTimeoutException("Canceled due to idle connection", false);
+              myError =
+                  new WatchdogTimeoutException(
+                      "Canceled idle connection after "
+                          + waitTime
+                          + " ms (threshold: "
+                          + idleTimeout.toMillis()
```
blakeli0, line 326: Now that the error message is not a static String anymore, can you please add some test for it? We already have some [test cases](https://github.com/googleapis/gapic-generator-java/blob/59efbbbd485282e37abbc5eed87b72a15a33240e/gax-java/gax/src/test/java/com/google/api/gax/rpc/WatchdogTest.java#L154) that verify the `WatchdogTimeoutException`, you could add assertions for the error message.

PR #1555, opened 2023-03-24
chore: coverage-report/pom.xml to have api-common annotation

suztomo:
This will fix https://github.com/googleapis/gapic-generator-java/pull/1511#issuecomment-1483254025


mpeddada1 (approved):
Thank you for the fix!

Issue #1439, opened 2023-03-06
REGAPIC Showcase Coverage


lqiu96:
Implement initial coverage for REGAPIC via Gapic-Showcase. We want to try and reach some parity with the tests that are implemented in Golang, but we may not strive to implement everything done in Golang and may not copy their structure.

Plans for implementation:
- [x] Callables (Echo)
  - [x] Unary - PR at https://github.com/googleapis/gapic-generator-java/pull/1483
  - [x] Server Side Streaming - PR at https://github.com/googleapis/gapic-generator-java/pull/1588
- [x] Behaviors (Echo)
  - [x] Deadline - PR at https://github.com/googleapis/gapic-generator-java/pull/1603
  - [x] Pagination - PR at https://github.com/googleapis/gapic-generator-java/pull/1635
  - [x] Retry - PR at https://github.com/googleapis/gapic-generator-java/pull/1603
- [x] Http Annotation - query/ path/ body combinations (Compliance) - PR: https://github.com/googleapis/gapic-generator-java/pull/1568
- [x] CRUD (Identity) - PR: https://github.com/googleapis/gapic-generator-java/pull/1589
   - [x] Get
   - [x] List
   - [x] Create
   - [x] Delete
   - [x] Update
- [x] Mixins (Echo)
   - [x] Location - PR: https://github.com/googleapis/gapic-generator-java/pull/1656
   - [x] IAM - PR: https://github.com/googleapis/sdk-platform-java/pull/1789
   - [x] Operations - PR at https://github.com/googleapis/gapic-generator-java/pull/1603

Bugs/ Blockers:
- [x] https://github.com/googleapis/gapic-generator-java/issues/1440 (Part 1 completed, Part 2 is not required)
- [x] https://github.com/googleapis/gapic-generator-java/issues/1437
- [x] https://github.com/googleapis/gapic-generator-java/issues/1504
- [x] https://github.com/googleapis/gapic-showcase/issues/1263
- [x] https://github.com/googleapis/gapic-showcase/issues/1261
- [x] https://github.com/googleapis/gapic-generator-java/issues/1547
- [ ] https://github.com/googleapis/gapic-showcase/issues/1286
- [x] https://github.com/googleapis/gapic-showcase/issues/1299
- [x] https://github.com/googleapis/gapic-showcase/issues/1338

lqiu96:
Closing this issue. The only thing left is a small bug for the HttpJson server side streaming test.

PR #1568, opened 2023-03-28
chore: Add Showcase Http Annotations Test

lqiu96:

Part of #1439 ☕️

No longer blocked as https://github.com/googleapis/gapic-generator-java/pull/1522 has been merged in.


showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITHttpAnnotation.java
```diff
@@ -0,0 +1,98 @@
...
+                    .getClassLoader()
+                    .getResourceAsStream("compliance_suite.json")),
+            builder);
+    complianceSuite = builder.build();
+
+    ComplianceSettings httpjsonComplianceSettings =
+        ComplianceSettings.newHttpJsonBuilder()
+            .setCredentialsProvider(NoCredentialsProvider.create())
+            .setTransportChannelProvider(
+                EchoSettings.defaultHttpJsonTransportProviderBuilder()
```
blakeli0, line 50: This should be `ComplianceSettings` instead of `EchoSettings`? Surprised it worked though.

.gitignore
```diff
@@ -16,3 +16,5 @@ target/
 .vscode/settings.json
 
 *.iml
+
+showcase/gapic-showcase/src/test/resources/compliance_suite.json
```
blakeli0, line 20: I think we should check it in if it is required for testing.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITHttpAnnotation.java
```diff
@@ -0,0 +1,98 @@
...
+            x -> complianceClient.repeatDataPathResource(x));
+  }
+
+  @After
+  public void destroyClient() {
+    complianceClient.close();
+  }
+
+  @Test
+  public void testCompliance() {
```
blakeli0, line 83: I'm actually not sure what we are verifying, can you make it clear with a better name and/or comments?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITHttpAnnotation.java
lqiu96: Ah yep, updated!

.gitignore
lqiu96: Ok, I can add it in.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITHttpAnnotation.java
lqiu96: Added some comments and updated the test function's name.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITHttpAnnotation.java
```diff
@@ -0,0 +1,103 @@
...
+  public void testHttpAnnotations() {
+    for (ComplianceGroup compliancegroup : complianceSuite.getGroupList()) {
+      List<String> validRpcList =
+          compliancegroup.getRpcsList().stream()
+              .filter(complianceValidRpcMap::containsKey)
+              .collect(Collectors.toList());
+      for (String rpcName : validRpcList) {
+        for (RepeatRequest repeatRequest : compliancegroup.getRequestsList()) {
+          ComplianceData expectedData = repeatRequest.getInfo();
+          RepeatResponse response = complianceValidRpcMap.get(rpcName).apply(repeatRequest);
```
burkedavison, line 97: I like how you've structured this with the map.

Minor suggestion: I got caught on this line having to figure out what was going on. Consider the slightly more explicit:

```java
for (String rpcName : validRpcList) {
    Function<Request, Response> rpc = complianceValidRpcMap.get(rpcName);
    for (Request request : getRequestsList()) {
        Response response = rpc.apply(request);
        ...
    }

Comment (lqiu96) [reply showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITHttpAnnotation.java:97]: Sure, will do!

**Comment by lqiu96** [comment]
@blakeli0 From our discussion earlier, I think we could split it up by ComplianceGroup and have a @Test function for each group. Each group would be testing features like values and bindings against a series of RPCs. WDYT?

File: showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITHttpAnnotation.java [hunk 5]
```diff
@@ -0,0 +1,104 @@
...
+      List<String> validRpcList =
+          compliancegroup.getRpcsList().stream()
+              .filter(complianceValidRpcMap::containsKey)
+              .collect(Collectors.toList());
+      for (String rpcName : validRpcList) {
+        Function<RepeatRequest, RepeatResponse> rpc = complianceValidRpcMap.get(rpcName);
+        for (RepeatRequest repeatRequest : compliancegroup.getRequestsList()) {
+          ComplianceData expectedData = repeatRequest.getInfo();
+          RepeatResponse response = rpc.apply(repeatRequest);
+          assertThat(response.getRequest().getInfo()).isEqualTo(expectedData);
```
mpeddada1, line 99: Could a helper method help reduce the number of nested for loops needed here? Reducing the amount of logic done in a test method can help with debugging in the future and prevent the test itself from developing bugs.

mpeddada1, line 99: This may require a big change but was wondering - would it be possible to explicitly specify the expected value is without computing it?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITHttpAnnotation.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.stream.Collectors;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+public class ITHttpAnnotation {
+
+  private ComplianceSuite complianceSuite;
+  private ComplianceClient complianceClient;
+  private Map<String, Function<RepeatRequest, RepeatResponse>> complianceValidRpcMap;
```
mpeddada1, line 31: Nice! this is a pretty cool use of the map.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITHttpAnnotation.java
burkedavison: As written, the compliance_suite.json is defined in a language-neutral way and can be (should?*) be owned by the gapic-showcase repo. If we hardcode expected values in the java test that duplicates the expected input, I would think this would eliminate our ability to have a common test suite defined elsewhere.

(*) Is there a way to get the compliance_suite.json from the gapic-showcase dependency rather than copying it into this repo?

lqiu96: Would it make more sense to just have each ComplianceGroup as its own test then? It could remove the outer most for loop if that's cleaner.

For each group, we test the product of rpcName and requestList.

Or did you mean something like:
```
  @Test
  public void testHttpAnnotations() {
    for (ComplianceGroup compliancegroup : complianceSuite.getGroupList()) {
      List<String> validRpcList =
          compliancegroup.getRpcsList().stream()
              .filter(complianceValidRpcMap::containsKey)
      helper(compliancegroup, validRpcList);
    }
  }

  private void helper(ComplianceGroup compliancegroup, List<String> validRpcList) {
    for (String rpcName : validRpcList) {
      Function<RepeatRequest, RepeatResponse> rpc = complianceValidRpcMap.get(rpcName);
      for (RepeatRequest repeatRequest : compliancegroup.getRequestsList()) {
        ComplianceData expectedData = repeatRequest.getInfo();
        RepeatResponse response = rpc.apply(repeatRequest);
        assertThat(response.getRequest().getInfo()).isEqualTo(expectedData);
      }
    }
  }
```

burkedavison: Great - this is perfect.

I vote to .gitignore it since it's a "generated" file to avoid confusion over the source of truth; but the main concern you've already covered.

Thank you.

lqiu96: Sounds good. Going to loop in @blakeli0 again if he has some other concerns that might be missed. If not, I can add the file back into the .gitignore

mpeddada1: I do like the idea of splitting the test based on the ComplianceGroup since it would reduce cognitive load needed in understanding the behavior the test is trying to verify. Also, in the event that a test fails it will be easier to determine where/what behavior is failing.  

Perhaps something similar to how [ComplianceClientTest](https://github.com/googleapis/gapic-generator-java/blob/820f4571af45b0d26ce3b883b27ebba5ac6f2270/showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/ComplianceClientTest.java#L41) is structured. 

Will also let @burkedavison and @blakeli0 weigh in.

mpeddada1: re comment on hard-coding the expected value: Thanks for the clarification! I didn't take this aspect into account: `I would think this would eliminate our ability to have a common test suite defined elsewhere.` Given this reasoning, computing the value instead of hard-cording looks like the way to go.

lqiu96: Sure, makes sense. I've split them by ComplianceGroup (unless there is also hesitancy about this). The groups are split by behavior (i.e. bindings, weird/extreme values, templating, etc.) so a failure for a test case in testing one group should alert us to the problematic behavior.

lqiu96:
@blakeli0 @burkedavison @mpeddada1 Could you all give a re-review whenever you all get a chance?

burkedavison (approved):

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITHttpAnnotation.java
```diff
@@ -0,0 +1,119 @@
...
+    List<String> validRpcList =
+        complianceGroup.getRpcsList().stream()
+            .filter(complianceValidRpcMap::containsKey)
+            .collect(Collectors.toList());
+    for (String rpcName : validRpcList) {
+      Function<RepeatRequest, RepeatResponse> rpc = complianceValidRpcMap.get(rpcName);
+      for (RepeatRequest repeatRequest : complianceGroup.getRequestsList()) {
+        ComplianceData expectedData = repeatRequest.getInfo();
+        RepeatResponse response = rpc.apply(repeatRequest);
+        assertThat(response.getRequest().getInfo()).isEqualTo(expectedData);
```
mpeddada1, line 115: nit: For future readers, perhaps we can add a comment before this block? "Verify that ... contains the expected ..."

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITHttpAnnotation.java
```diff
@@ -0,0 +1,119 @@
+package com.google.showcase.v1beta1.it;
```
mpeddada1, line 1: We may be missing a copyright header for this file?

Hm I wonder why our checkstyle didn't fail.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITHttpAnnotation.java
blakeli0: > I vote to .gitignore it since it's a "generated" file to avoid confusion over the source of truth; but the main concern you've already covered.

I feel it might be better to include this `compliance_suite.json` file in our repo and maybe the protos as well, as it would be easier to read and troubleshoot the tests. Even though they are "generated", but I consider them as part of the whole gapic-showcase test suite, just like the "generated" gapic client libraries, which are already included in our repo.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITHttpAnnotation.java
lqiu96: Thanks! Added comments.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITHttpAnnotation.java
lqiu96: Added the copyright header. I'll need to check why it didn't run for the test files.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITHttpAnnotation.java
lqiu96: Sure, added the json file in.

lqiu96:
Let me know if the changes look fine @mpeddada1 @blakeli0

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITHttpAnnotation.java
mpeddada1: Thanks @lqiu96!

mpeddada1 (approved):

PR #1572, opened 2023-03-28
chore: add script to create a native image submodule for testing

mpeddada1:
In order to test for changes in gapic-generator-java, we need to do the following:

1) Modify graal-sdk version in gapic-generator-java/gax
2) Modify the gapic-generator-bom version in java-shared-dependencies. Install it to the local maven cache.
3) Ensure that java-shared-config contains the latest version of the native-maven-plugin.
4) Modify shared-dependencies in the desired handwritten library.

This script automates this process by creating a submodule project with all the necessary repositories. In order to use it, we need to do the following:

1) Create a new github repository from https://github.com/new
2) Clone the repo created in step 1 on local machine.
3) From the root of the repo, call:
```
GRAALVM_VERSION=22.2.0 NATIVE_MAVEN_PLUGIN=0.9.20 ORGANIZATION=mpeddada1 bash -x ${HOME}/scripts/create_native_image_test_env.sh
```
4) The submodule will be populated with all the repos necessary for testing including, gapic-generator-java, java-shared-config, java-pubsub, java-bigquery, java-spanner-jdbc and java-bigtable. 

**Note**: The script will create branches named ${GRAALVM_VERSION}_update in the forked repos listed. We need to make sure that forks of all the necessary repos are present and synced before running the script. 

Sample repo created from this script: https://github.com/mpeddada1/graalvm-submodule-test2


burkedavison:
Is this solution the result of a discussion with @suztomo ? (What are your thoughts, Tomo?)

I am still on the fence + processing my opinion on this. Perhaps some additional discussion is needed.

scripts/create_native_image_test_env.sh
```diff
@@ -0,0 +1,215 @@
+#!/bin/bash
+
```
suztomo, line 2: Can you add source code comment that explains the purpose of this script?

suztomo (changes requested):

scripts/create_native_image_test_env.sh
mpeddada1: Good idea, done. I was also thinking of adding a few instructions on usage. Do you think this would be a good location? https://github.com/googleapis/gapic-generator-java/blob/main/DEVELOPMENT.md

burkedavison:
LGTM

scripts/create_native_image_test_env.sh
```diff
@@ -0,0 +1,225 @@
...
+# This script creates a submodule project with all the necessary repositories and changes needed to test
```
suztomo, line 3: This explains what it does. Can you add _why_ you had to create this script?

scripts/create_native_image_test_env.sh
```diff
@@ -0,0 +1,228 @@
...
+fi
+
+if [ -z "$ORGANIZATION" ]; then
+  echo "Please specify the ORGANIZATION where your updates need to be stored."
+  exit 1
+fi
+
+GRAALVM_BRANCH="${GRAALVM_VERSION}_update"
+
+## Round 1: Add gapic-generator-java and update graal-sdk version in GAX.
```
suztomo, line 55: It seems this script assumes this to run in a newly created repository with "git init"?

If so, can you add if statement to confirm the current working directory has ".git" directory?

Add source code comment about the assumption.

mpeddada1: Yeah it assumes that a repo is created with either 'git clone <link to remote>' or `git init` and `git remote add <link to remote>`.

Done.

PR #1575, opened 2023-03-28
feat: add stream method for `ServerStream`

### LINKED ISSUE #1576: `ServerStream` do not easily support Java Streams
Add stream support for `com.google.api.gax.rpc.ServerStream`.

JoeWang1127:
Fixes #1576 ☕️


burkedavison:
LGTM

suztomo (approved):
Let's merge this at the same time as the Page's

gax-java/gax/src/main/java/com/google/api/gax/rpc/ServerStream.java
```diff
@@ -89,6 +91,10 @@ public Iterator<V> iterator() {
     return iterator;
   }
 
+  public Stream<V> stream() {
```
blakeli0, line 94: Can you please add Javadoc? All public methods are required to have Javadoc.

gax-java/gax/src/main/java/com/google/api/gax/rpc/ServerStream.java
```diff
@@ -89,6 +91,10 @@ public Iterator<V> iterator() {
...
+    return StreamSupport.stream(this.spliterator(), false);
```
blakeli0, line 95: I see that the second argument is set to false, which means this is a sequential stream, do we want to expose another `parallelStream` method as well?

JoeWang1127: I'm not sure whether a deterministic result is required as parallel stream can randomize the result.

gax-java/gax/src/main/java/com/google/api/gax/rpc/ServerStream.java
JoeWang1127: done

PR #1582, opened 2023-03-29
chore: add client-side streaming showcase integration tests for gRPC support

mpeddada1:
Verifies:
- OK Server response when client sends streaming requests to server.
- Handling of Server Error response.

### DIFF
--- a/showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITClientSideStreaming.java
+++ b/showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITClientSideStreaming.java
@@ -21,7 +21,6 @@ import static org.junit.Assert.assertThrows;
 
 import com.google.api.core.SettableApiFuture;
 import com.google.api.gax.core.NoCredentialsProvider;
-import com.google.api.gax.grpc.InstantiatingGrpcChannelProvider;
 import com.google.api.gax.rpc.ApiStreamObserver;
 import com.google.api.gax.rpc.CancelledException;
 import com.google.api.gax.rpc.StatusCode;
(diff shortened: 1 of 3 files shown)


showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITClientSideStreaming.java
```diff
@@ -0,0 +1,133 @@
...
+  private EchoClient grpcClient;
+
+  @Before
+  public void createClients() throws IOException {
+    // Create gRPC Echo Client
+    EchoSettings grpcEchoSettings =
+        EchoSettings.newBuilder()
+            .setCredentialsProvider(NoCredentialsProvider.create())
+            .setTransportChannelProvider(
+                InstantiatingGrpcChannelProvider.newBuilder()
```
burkedavison, line 53: What's the recommended way to initialize channel providers?

Should we be using the default method for this `https://github.com/googleapis/gapic-generator-java/blob/main/showcase/gapic-showcase/src/main/java/com/google/showcase/v1beta1/EchoSettings.java#L163` rather than a specific implementation?

burkedavison (approved):

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITClientSideStreaming.java
mpeddada1: Good point! Hm I think it is better to use the default unless we need to provide some special settings that are specifically tailored to a test. It can help avoid false positives where test failures that are caused due to the implementation details of the tests rather than the behavior of the methods being tested.

PR #1583, opened 2023-03-29
chore: Fix testCancelOuterFutureBeforeStart flaky test

### LINKED ISSUE #1342: gax-java(mac): Flakey ScheduledRetryingExecutorTest.testCancelOuterFutureBeforeStart
Test [flaked](https://github.com/googleapis/gapic-generator-java/actions/runs/4130064207/jobs/7136363769) during nightly run. Java 17.

```
Error:  Tests run: 30, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 9.284 s <<< FAILURE! - in com.google.api.gax.retrying.ScheduledRetryingExecutorTest
Error:  com.google.api.gax.retrying.ScheduledRetryingExecutorTest.testCancelOuterFutureBeforeStart[Custom retry settings: true]  Time elapsed: 0.99 s  <<< FAILURE!
org.mockito.exceptions.verification.NoInteractionsWanted: 

No interactions wanted here:
-> at com.google.api.gax.retrying.AbstractRetryingExecutorTest.testCancelOuterFutureBeforeStart(AbstractRetryingExecutorTest.java:247)
But found this interaction on mock 'tracer':
-> at com.google.api.gax.retrying.FailingCallable.call(FailingCallable.java:85)
Actually, above is the only interaction with this mock.
	at com.google.api.gax.retrying.AbstractRetryingExecutorTest.testCancelOuterFutureBeforeStart(AbstractRetryingExecutorTest.java:247)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
	at org.junit.runners.Suite.runChild(Suite.java:128)
	at org.junit.runners.Suite.runChild(Suite.java:27)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
	at org.junit.runners.Suite.runChild(Suite.java:128)
	at org.junit.runners.Suite.runChild(Suite.java:27)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
	at org.apache.maven.surefire.junitcore.JUnitCore.run(JUnitCore.java:55)
	at org.apache.maven.surefire.junitcore.JUnitCoreWrapper.createRequestAndRun(JUnitCoreWrapper.java:137)
	at org.apache.maven.surefire.junitcore.JUnitCoreWrapper.executeEager(JUnitCoreWrapper.java:107)
	at org.apache.maven.surefire.junitcore.JUnitCoreWrapper.execute(JUnitCoreWrapper.java:83)
	at org.apache.maven.surefire.junitcore.JUnitCoreWrapper.execute(JUnitCoreWrapper.java:75)
	at org.apache.maven.surefire.junitcore.JUnitCoreProvider.invoke(JUnitCoreProvider.java:158)
	at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:456)
	at org.apache.maven.surefire.booter.ForkedBooter.execute(ForkedBooter.java:169)
	at org.apache.maven.surefire.booter.ForkedBooter.run(ForkedBooter.java:595)
	at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:581)
```

lqiu96:

Fixes #1342☕️


gax-java/gax/src/test/java/com/google/api/gax/retrying/AbstractRetryingExecutorTest.java
```diff
@@ -223,14 +223,14 @@ public void testTotalTimeoutExceeded() throws Exception {
 
   @Test
   public void testCancelOuterFutureBeforeStart() throws Exception {
-    FailingCallable callable = new FailingCallable(4, "request", "SUCCESS", tracer);
+    FailingCallable callable = new FailingCallable(0, "request", "SUCCESS", tracer);
```
lqiu96, line 226: The parameter is `expectedFailuresCount` and it should be `0`. The callable's `call()` function should not be invoked.

lqiu96:
I believe the issue is that there was a slight timing issue with when the tracer would run. The ScheduledExectur would schedule the callable to run immediately. The subsequent retries would not run since the future has been cancelled, but the first one always runs (as the first run doesn't [check if the future is cancelled](https://github.com/googleapis/gapic-generator-java/blob/3f6c4c04fde254cfc3c1441b44beeb2bbe0c7e03/gax-java/gax/src/main/java/com/google/api/gax/retrying/ScheduledRetryingExecutor.java#L115-L119).

burkedavison:
LGTM

lqiu96:
Thanks @burkedavison. Talked with @blakeli0 about this and there are some concerns about this being a behavior breaking change. There some underlying details that I'm not fully certain about, so I'm going to hold off on merging this in.

---
Interestingly, the only non-testing related call of `executor.submit()` is here: https://github.com/googleapis/gapic-generator-java/blob/3f6c4c04fde254cfc3c1441b44beeb2bbe0c7e03/gax-java/gax/src/main/java/com/google/api/gax/retrying/CallbackChainRetryingFuture.java#L136-L138 which checks if the future is not done. 

I wonder if we can do a similar check inside the test, something like:
```
    if (!future.isDone()) {
      future.setAttemptFuture(executor.submit(future));
    }
    assertEquals(0, future.getAttemptSettings().getAttemptCount());

    // Assert that the callable's call() is not invoked
    verifyNoInteractions(tracer);
```
The gax implementation doesn't seem to run if it's been cancelled.

lqiu96:
Actually, on a second look... maybe I can fix this via the tests itself?

FailingCallable could mimic what we have in AttemptCallable:
https://github.com/googleapis/gapic-generator-java/blob/3f6c4c04fde254cfc3c1441b44beeb2bbe0c7e03/gax-java/gax/src/main/java/com/google/api/gax/rpc/RetryingCallable.java#L61-L63

We can set the externalFuture in the FailingCallable and check if the `externalFuture.isDone()` is true before calling the tracer logic.

gax-java/gax/src/test/java/com/google/api/gax/retrying/AbstractRetryingExecutorTest.java
```diff
@@ -222,29 +223,32 @@ public void testTotalTimeoutExceeded() throws Exception {
...
 
+    callable.setExternalFuture(future);
     future.setAttemptFuture(executor.submit(future));
-
-    assertFutureCancel(future);
     assertEquals(0, future.getAttemptSettings().getAttemptCount());
 
-    verifyNoMoreInteractions(tracer);
+    // Assert that no tracer attempt has run
+    verifyNoInteractions(tracer);
```
blakeli0, line 251: I wonder how useful is this verification now as we are basically verifying the behavior of our own `MockCallable`. As you mentioned in other comments, we are always checking that [the future is not done](https://github.com/googleapis/gapic-generator-java/blob/3f6c4c04fde254cfc3c1441b44beeb2bbe0c7e03/gax-java/gax/src/main/java/com/google/api/gax/retrying/CallbackChainRetryingFuture.java#L136-L138) before submitting it to an retrying executor, so the current flakiness is actually expected since we are not checking if the future is done again in the [submit](https://github.com/googleapis/gapic-generator-java/blob/3f6c4c04fde254cfc3c1441b44beeb2bbe0c7e03/gax-java/gax/src/main/java/com/google/api/gax/retrying/ScheduledRetryingExecutor.java#L115-L119) method itself. 
That being said, I don't think we want to mimic the behavior in `MockCallable` as that's outside the scope of the unit tests of  `AbstractRetryingExecutorTest`. My suggestion is that we keep this test case but remove this line of verification if we don't find any other benefit.

gax-java/dependencies.properties
```diff
@@ -75,7 +75,7 @@ maven.org_graalvm_sdk=org.graalvm.sdk:graal-sdk:22.3.1
 
 # Testing maven artifacts
 maven.junit_junit=junit:junit:4.13.2
-maven.org_mockito_mockito_core=org.mockito:mockito-core:2.28.2
+maven.org_mockito_mockito_core=org.mockito:mockito-core:4.11.0
```
blakeli0, line 78: Is this a required change? If not, I would leave it out of this PR, and create another PR for updating the mockito version.

gax-java/gax/src/test/java/com/google/api/gax/retrying/FailingCallable.java
```diff
@@ -66,6 +66,8 @@ class FailingCallable implements Callable<String> {
   private final String result;
   private final CountDownLatch firstAttemptFinished = new CountDownLatch(1);
 
+  private volatile RetryingFuture<String> externalFuture;
```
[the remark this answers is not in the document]
lqiu96: Thanks! Will remove.

gax-java/gax/src/test/java/com/google/api/gax/retrying/AbstractRetryingExecutorTest.java
```diff
@@ -222,29 +223,32 @@ public void testTotalTimeoutExceeded() throws Exception {
...
-            .setTotalTimeout(Duration.ofMillis(10_0000L))
+            .setTotalTimeout(Duration.ofMillis(10_000L))
             .build();
     RetryingExecutorWithContext<String> executor =
         getExecutor(getAlgorithm(retrySettings, 0, null));
     RetryingFuture<String> future = executor.createFuture(callable, retryingContext);
     boolean res = future.cancel(false);
-
     assertTrue(res);
+    assertTrue(future.isCancelled());
```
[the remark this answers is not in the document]
lqiu96: I believe `assertFutureCancel()` is used to verify that the future was cancelled after the executorService has submitted the callable for execution and waits to ensure that it was cancelled properly. Since this was cancelled prior to being executed, we don't need to wait to verify that the future is cancelled.

gax-java/dependencies.properties
lqiu96: I believe this was failing the CI when running the integration tests. I can upgrade this in a separate PR.

gax-java/gax/src/test/java/com/google/api/gax/retrying/AbstractRetryingExecutorTest.java
lqiu96: MockCallable's tracer is to mimic the functionality inside AttemptCallable (which is what I believe FailingCallable is based on): https://github.com/googleapis/gapic-generator-java/blob/4bb0a5e3ce78878841e87a47cd457995ea9ed87e/gax-java/gax/src/main/java/com/google/api/gax/rpc/AttemptCallable.java#L78-L85

The tracer's functionality for the MockCallable can also be covered with the assertion below: `assertEquals(0, future.getAttemptSettings().getAttemptCount());`

so I'm fine with removing.

gax-java/gax/src/test/java/com/google/api/gax/retrying/AbstractRetryingExecutorTest.java
```diff
@@ -222,29 +222,29 @@ public void testTotalTimeoutExceeded() throws Exception {
   }
 
   @Test
-  public void testCancelOuterFutureBeforeStart() throws Exception {
-    FailingCallable callable = new FailingCallable(4, "request", "SUCCESS", tracer);
+  public void testCancelOuterFutureBeforeStart() {
+    // Use MockCallable instead of FailingCallable as we do not need the request to be re-tried.
+    // For this test, the callable should attempt to run and then see that the external future
+    // is cancelled and that the callable's execution has ended.
+    MockCallable callable = new MockCallable("SUCCESS");
```
blakeli0, line 229: If we don't verify the tracer anymore, do we still need this new `MockCallable`? Can we keep using `FailingCallable`? I think the flakiness was only from `verifyNoMoreInteractions()`?

gax-java/gax/src/test/java/com/google/api/gax/retrying/FailingCallable.java
```diff
@@ -66,20 +67,29 @@ class FailingCallable implements Callable<String> {
...
+  }
+
   CountDownLatch getFirstAttemptFinishedLatch() {
     return firstAttemptFinished;
   }
 
   @Override
   public String call() throws Exception {
     try {
+      if (externalFuture.isDone()) {
```
blakeli0, line 90:
```suggestion
      if (externalFuture != null && externalFuture.isDone()) {
```
We can add a null check so that we don't have to set it for existing tests.

gax-java/gax/src/test/java/com/google/api/gax/retrying/AbstractRetryingExecutorTest.java
```diff
@@ -222,29 +227,27 @@ public void testTotalTimeoutExceeded() throws Exception {
...
-            .setTotalTimeout(Duration.ofMillis(10_0000L))
+            .setTotalTimeout(Duration.ofMillis(10_000L))
             .build();
     RetryingExecutorWithContext<String> executor =
         getExecutor(getAlgorithm(retrySettings, 0, null));
     RetryingFuture<String> future = executor.createFuture(callable, retryingContext);
+    callable.setExternalFuture(future);
     boolean res = future.cancel(false);
-
+    verifyNoMoreInteractions(tracer);
```
blakeli0, line 245: Can we still verify it as the last step like previously? Or it has to be verified earlier now?

lqiu96: I think it makes sense to verify it here (before the executor submits the future) to ensure that the callable's logic isn't invoked (not invoking == exiting once it realizes the external future is done) after it's been submitted for execution.

Otherwise we can use verifyNoInteractions(tracer)  and put that at the end which requires an updating mockito version in dependencies.properties. Functionality-wise, they should be the same so I just used `verifyNoMoreInteractions` which already exists

gax-java/gax/src/test/java/com/google/api/gax/retrying/FailingCallable.java
lqiu96: Yeah that is an option too. I was thinking that all existing and future tests *should* pass in the external future (which more accurately mimics the AttemptCallable) so it shouldn't be null. WDYT?

blakeli0: I see, yes that makes sense, can you add some comments to reflect the idea?

gax-java/gax/src/test/java/com/google/api/gax/retrying/AbstractRetryingExecutorTest.java
blakeli0: Are they the same functional-wise? The callable would never run until `executor.submit(future)`, and I think the purpose of this test is to verify that the inner callable would never run if the future is already cancelled, despite that the future is submitted. So verifying after `executor.submit(future)` makes more sense to me.

gax-java/gax/src/test/java/com/google/api/gax/retrying/FailingCallable.java
lqiu96: Sure, will do.

gax-java/gax/src/test/java/com/google/api/gax/retrying/AbstractRetryingExecutorTest.java
lqiu96: The inner callable does run, but does not execute any logic -- `call()` is invoked by then immediately exited as the external future is always done. Since this is the first attempt, the `random retry delay` is set to 0 and is scheduled to execute immediately. I was thinking that this would create a possible race between when the callable exits and when  we run `verifyNoMoreInteractions`.

If `call()` runs before `verifyNoMoreInteractions` then there wouldn't be a point of running `verifyNoMoreInteractions` as the callable has already exited
If `call()` runs after `verifyNoMoreInteractions`, then this line would confirm the behavior. 

(From my understanding) We can't guarantee when the scheduleExecutor executes the callable (either before or after `verifyNoMoreInteractions` in the test) so asserting `verifyNoMoreInteractions` before the execution would confirm that the `call()` ran and that the tracer never ran.

blakeli0: Which line would invoke `call()` immediately? I still think the callable is not invoked until `executor.submit(future)` based on the [Javadoc](https://github.com/googleapis/gapic-generator-java/blob/59efbbbd485282e37abbc5eed87b72a15a33240e/gax-java/gax/src/main/java/com/google/api/gax/retrying/RetryingExecutor.java#L67)

lqiu96: Er* It *should* be invoked immediately once `executor.submit()` runs:
https://github.com/googleapis/gapic-generator-java/blob/59efbbbd485282e37abbc5eed87b72a15a33240e/gax-java/gax/src/main/java/com/google/api/gax/retrying/ScheduledRetryingExecutor.java#L115-L119 as `retryingFuture.getAttemptSettings().getRandomizedRetryDelay().toMillis()` -> 0ms for the first attempt

I looked up the `verifyNoMoreInteractions` javadoc and I misunderstood the use case. I assumed it was a point-in-time verification to ensure there was no more interactions once it was invoked (the race logic I mentioned above), but  it's meant to ensure that there were no more interactions after we've verified that things have already run/ executed/ invoked a certain number of times.

I think having either `verifyNoInteractions` or `verifyNoMoreInteractions` at the end of the file *should* work. May make more sense to have this be `verifyNoInteractions` instead of `verifyNoMoreInteractions` for clarity (tracer should have 0 interactions and `no more` implies more than 0) but that would require Mockito 4.x in the `dependencies.properties` file.

blakeli0 (approved):

PR #1584, opened 2023-03-29
chore: remove java-common-protos/.kokoro and enable renovate bot for requirements file. 

mpeddada1:
Fixes https://github.com/googleapis/gapic-generator-java/issues/1485

Notable changes:

- Removed java-common-protos/.kokoro which was carried over from the repo's migration to the gapic-generator-java repo.
-  Unpins dependencies in .kokoro/requirements.txt. Similar to the changes applied in https://github.com/googleapis/synthtool/pull/1741, now that our jobs have been upgraded to Python 3.9, we no longer have to pin the dependencies in requirement.txt.
- Enables renovate bot to update requirements.txt when new version is found. We are currently managing the versions in the requirement.txt file manually (unless [there is a dependabot security update](https://github.com/googleapis/gapic-generator-java/pull/1297)). See https://github.com/googleapis/gapic-generator-java/pull/1370.


suztomo (changes requested):
We don't use java-common-protos/.kokoro. (Let me know if you think otherwise)


Would you add a check (either this PR or another PR) to ensure these requirements files can be installed. It's great if we can use the same container image as we use in the release job.

mpeddada1:
> We don't use java-common-protos/.kokoro. (Let me know if you think otherwise)

Not sure if this is expected but I see that the nightly and presubmit jobs are still running but the release jobs haven't been executed since Nov, 2022. Does that mean that the `.kokoro/release` scripts can be removed? 

Screenshot 2023-03-30 at 10 41 50 AM

> Would you add a check (either this PR or another PR) to ensure these requirements files can be installed. It's great if we can use the same container image as we use in the release job.

Sure! I'll work on adding the check in.

mpeddada1:
@suztomo I've removed the `.kokoro` directory from java-common-protos and added a check for requirements.txt. Could I get another review from you?

.github/workflows/ci.yaml
```diff
@@ -166,6 +166,14 @@ jobs:
         #  testlib modules of gax
         run: mvn package clirr:check -DskipTests
 
+  requirements-kokoro:
+    runs-on: ubuntu-latest
+    container: gcr.io/cloud-devrel-public-resources/java11
+    steps:
+      - uses: actions/checkout@v3
+      - run: python3 --version
+      - run: python3 -m pip install --require-hashes -r .kokoro/requirements.txt
```
suztomo, line 175: Great. Can you also add check to confirm `releasetool` module import? 

- https://github.com/googleapis/gapic-generator-java/blob/main/.kokoro/release/stage.sh#LL23C3-L23C25
- https://github.com/googleapis/synthtool/blob/master/docker/owlbot/java/container_test.yaml#L36

suztomo (approved):
Thank you. Merge once all are green.

Issue #1504, opened 2023-03-17
HttpJson Unary Callables not following RetrySetting's TotalTimeout value


lqiu96:
Sample Code run against Gapic-Showcase:

Initialize the Echo Client
```
    EchoSettings httpjsonEchoSettings =
        EchoSettings.newHttpJsonBuilder()
            .setCredentialsProvider(NoCredentialsProvider.create())
            .setTransportChannelProvider(
                EchoSettings.defaultHttpJsonTransportProviderBuilder()
                    .setHttpTransport(
                        new NetHttpTransport.Builder().doNotValidateCertificate().build())
                    .setEndpoint("http://localhost:7469")
                    .build())
            .build();
    httpjsonClient = EchoClient.create(httpjsonEchoSettings);
```

Make the call:
```
    // Default timeout for UnaryCall is 5 seconds -- We want to ensure a long enough delay for
    // this test
    int delayInSeconds = 10;
    assertThrows(
        DeadlineExceededException.class,
        () ->
            httpjsonClient.block(
                BlockRequest.newBuilder()
                    .setSuccess(
                        BlockResponse.newBuilder().setContent("httpjsonBlockContent_10SecDelay1"))
                    .setResponseDelay(
                        com.google.protobuf.Duration.newBuilder()
                            .setSeconds(delayInSeconds)
                            .build())
                    .build()));
    try {
      httpjsonClient.block(
          BlockRequest.newBuilder()
              .setSuccess(BlockResponse.newBuilder().setContent("httpjsonBlockContent_10SecDelay2"))
              .setResponseDelay(
                  com.google.protobuf.Duration.newBuilder().setSeconds(delayInSeconds).build())
              .build());
    } catch (DeadlineExceededException e) {
      assertThat(e.getStatusCode().getCode()).isEqualTo(StatusCode.Code.DEADLINE_EXCEEDED);
    }
```

Error Message:
```
[ERROR] Tests run: 10, Failures: 1, Errors: 0, Skipped: 1, Time elapsed: 45.797 s <<< FAILURE! - in com.google.showcase.v1beta1.it.ITEcho
[ERROR] com.google.showcase.v1beta1.it.ITEcho.testBlockHttpJson_throwsDeadlineExceededException  Time elapsed: 10.019 s  <<< FAILURE!
java.lang.AssertionError: expected com.google.api.gax.rpc.DeadlineExceededException to be thrown, but nothing was thrown
[...]

PR #1603, opened 2023-04-03
fix: Unary Callables Deadline values respect the TotalTimeout in RetrySettings

lqiu96:
We should be able to cancel the runnable/ close the call exactly when the RPC timeout has elapsed.

Fixes: #1504


gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonClientCallImpl.java
```diff
@@ -88,6 +93,7 @@
   private final ApiMethodDescriptor<RequestT, ResponseT> methodDescriptor;
   private final HttpTransport httpTransport;
   private final Executor executor;
+  private final ScheduledExecutorService deadlineCancellationExecutor;
```
blakeli0, line 96: Can we try to reuse the `executor` above? I know it's not a `ScheduledExecutorService` at this point, but the what's being passed in is from [here](https://github.com/googleapis/gapic-generator-java/blob/4bb0a5e3ce78878841e87a47cd457995ea9ed87e/gax-java/gax/src/main/java/com/google/api/gax/core/ExecutorProvider.java#L43), which is indeed a `ScheduledExecutorService`, so we should be able to change the type and reuse it. I did a quick search and I don't think most of the change would be breaking, there is one public [builder method](https://github.com/googleapis/gapic-generator-java/blob/4bb0a5e3ce78878841e87a47cd457995ea9ed87e/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ManagedHttpJsonChannel.java#L125) that could be considered breaking.
The reason we may want to reuse the existing executor is that we can leverage the existing client lifecycles for shutting down the executor, otherwise we need to explicitly shut down this new `deadlineCancellationExecutor` somewhere to prevent from resource leaking, which is missing from this PR currently. Reusing the existing executor may also have other issues like the ThreadPool is not large enough so new requests may have to wait for new thread. The bottom line is that we need to be careful with anything related to executors, so we have enough resource for large number of concurrent requests and we don't leak resource accidentally.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonClientCallImpl.java
```diff
@@ -161,6 +168,36 @@ public void start(Listener<ResponseT> responseListener, HttpJsonMetadata request
...
+          this::closeAndNotifyListeners, timeoutNanos, TimeUnit.NANOSECONDS);
+    }
+  }
+
+  // No need to trigger the deliver() loop again as we have already closed the runnable
+  // task and added the OnCloseNotificationTask. We notify the FutureListener that the
+  // there is a timeout exception from this RPC call (DEADLINE_EXCEEDED)
+  private void closeAndNotifyListeners() {
+    synchronized (lock) {
+      close(
```
blakeli0, line 193: This [comment](https://github.com/googleapis/gapic-generator-java/blob/4bb0a5e3ce78878841e87a47cd457995ea9ed87e/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonClientCallImpl.java#L393-L396) in `close` method makes me worried that we may still have to wait for the server to fully send the response.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpRequestRunnable.java
```diff
@@ -191,13 +190,27 @@ HttpRequest createHttpRequest() throws IOException {
...
+          && httpRequest.getReadTimeout() < timeoutMs
+          && timeoutMs < Integer.MAX_VALUE) {
+        httpRequest.setReadTimeout((int) timeoutMs);
+      }
+
+      // Connect timeout is the time allowed for establishing the connection.
+      // This is updated to match the RPC timeout as we do not want a shorter
+      // connect timeout to preemptively throw a ConnectExcepetion before
+      // we've reached the RPC timeout
+      if (httpRequest.getConnectTimeout() > 0
```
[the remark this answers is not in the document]
lqiu96: Sure, will update. I'll add a test for the timeout values.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpRequestRunnable.java
```diff
@@ -191,13 +190,27 @@ HttpRequest createHttpRequest() throws IOException {
 
     HttpRequest httpRequest = buildRequest(requestFactory, url, jsonHttpContent);
 
-    Instant deadline = httpJsonCallOptions.getDeadline();
-    if (deadline != null) {
-      long readTimeout = Duration.between(Instant.now(), deadline).toMillis();
+    Duration timeout = httpJsonCallOptions.getTimeout();
```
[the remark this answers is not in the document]
lqiu96: Make sense, will update.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonClientCallImpl.java
lqiu96: Gotcha, makes sense. I'll take another look into how grpc handles the multiple executors. I'd imagine they would also possibly run into this issue.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonClientCallImpl.java
lqiu96: Yeah the cancellation in the runnable doesn't seem like it's actually cancel's the runnable. I'll look into the possibility of being able to actually disconnect the connection instead of waiting for the socket timeout.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonClientCalls.java
```diff
@@ -134,19 +137,29 @@ public void onMessage(T message) {
...
+      // the trailer exceptions are set via parameters in there.
+      //
+      // A successful response will have a null exception in the trailers.
+      // The trailer's exception CAN be a DEADLINE_EXCEEDED exception set
+      // from the deadlineScheduler and we must return it as an exception.
+      //
+      // Note: An exception has the highest priority. We must respect the timeout
+      // value set in the RetrySettings even if the message barely misses the
+      // cutoff time.
+      if (trailers.getException() != null) {
```
blakeli0, line 151: Why do we need to change this logic? Is there a case that `isMessageReceived` is true and `DEADLINE_EXCEEDED` exception is set?

.github/workflows/sonar.yaml
```diff
@@ -32,6 +32,9 @@ jobs:
           path: ~/.m2
           key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
           restore-keys: ${{ runner.os }}-m2
+      - name: Install maven modules
```
blakeli0, line 35: Is this a required change?

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonClientCallImpl.java
```diff
@@ -114,19 +119,21 @@
   private ProtoMessageJsonStreamIterator responseStreamIterator;
 
   @GuardedBy("lock")
-  private boolean closed;
+  private volatile boolean closed;
```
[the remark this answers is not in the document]
lqiu96: Yep

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonCallOptions.java
```diff
@@ -45,6 +46,9 @@ public abstract class HttpJsonCallOptions {
   @Nullable
   public abstract Instant getDeadline();
```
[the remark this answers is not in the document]
lqiu96: I don't believe it is. I'll remove it.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonClientCalls.java
lqiu96: There is a (very small) possibility that this can happen:
1. Enqueue a message/ response for call
2. `notifyListeners()` that a message has been sent which updates `isMessageReceived` = true
3. `timeout()` runs and enqueues the `DEADLINE_EXCEEDED` exception
4. Last loop of `deliver()` runs executes the the `onclose()` method

There is now a state where both a message and `DEADLINE_EXCEEDED` have been received.

lqiu96:
> Looking good overall! Two general questions:
> 
> 1. How does this change affect server streaming calls? As I believe most of the code touched are used by server streaming as well.

I'll add some more tests (and showcase tests) for Server-Side streaming calls. I'll also add some tests for LROs as I believe the the RetryAlgorithms affect their logic.

> 2. Do we know how much performance overhead it might incur? e.g. Call the same RPC 1000 times with and without the fix, compare the differences.

Sure, I can create a benchmark for these RPCs with various RetrySettings.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonClientCalls.java
blakeli0: This is a flag to me that we may want to change our implementation, as now we have two source of truth to determine if we received response or not, `isMessageReceived` and `trailers.getException()`. Let's discuss more in our meeting.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonCallOptions.java
```diff
@@ -34,6 +34,7 @@
 import com.google.auto.value.AutoValue;
 import com.google.protobuf.TypeRegistry;
 import javax.annotation.Nullable;
+import org.threeten.bp.Duration;
```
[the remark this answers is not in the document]
blakeli0: We don't have to change the existing implementation, but for newly added field, we can convert `org.threeten.bp.Duration` to `java.time.Duration` and expose `java.time.Duration` only on the surface.

lqiu96:
@blakeli0 -- I took a look at the comments above:

1. How does this change affect server streaming calls? As I believe most of the code touched are used by server streaming as well.

I looked into the showcase tests for Server-Side Streaming and there isn't a client/ rpc that blocks for streaming calls. For server-side stream's listener, I believe it will operate the same way: https://github.com/googleapis/gapic-generator-java/blob/c2d6d159871b827987640a947a35c9aaa0dea800/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonDirectStreamController.java#L116-L123. I'll have to try and implement the the Sequence client to be able to properly test for retries with unary and streaming rpcs.

2. Do we know how much performance overhead it might incur? e.g. Call the same RPC 1000 times with and without the fix, compare the differences.

I have a bit of difficultly retrying the RPC x1000 times for HttpJson. The previous implementation would either return the result directly if RPC timeout is configured to be sub-20 seconds or would throw a SocketTimeoutException. I'm not sure how I can configure it to be able to properly retry.

Though, when running this for the fix, I saw that the cancellation occurs at around the TotalTimeout. Running the the RPC x1000 with RPC timeout 100ms and total timeout 100000ms, it was able to return the result (DEADLINE_EXCEEDED) at 100023ms

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonCallOptions.java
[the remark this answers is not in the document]
lqiu96: Sounds good. Added this as `java.time.Duration` and made the manual conversion back to `org.threeten.bp.Duration` where needed.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonClientCalls.java
lqiu96: Per internal discussion, we've migrated this back to our old implementation.

gax-java/gax/src/main/java/com/google/api/gax/retrying/ExponentialRetryAlgorithm.java
```diff
@@ -232,6 +225,28 @@ public boolean shouldRetry(TimedAttemptSettings nextAttemptSettings) {
...
+    // it would mean that the Callable does not set an explicit timeout and that the request
+    // would continue until it either receives a response or an exception from the socket.
+    //
+    // This is *hopefully* a rare case that may only occur when there it is close to the
+    // totalTimeout deadline, but this should try to account for the differences in units
+    // being used for calculations (nanos vs ms). An RPC Timeout may not need nanosecond
+    // precision as it is possible that it would send the request and then immediately times
+    // out.
+    return totalTimeSpentNanos >= totalTimeoutNanos || rpcTimeout.isZero();
+  }
```
lqiu96, line 248: @blakeli0 This was the section in the retryAlgorithm that I modified. Adding in `rpcTimeout.isZero()` for non-LRO logic. Wanted to double check that this is fine.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonClientCalls.java
```diff
@@ -50,12 +49,25 @@ public static <RequestT, ResponseT> HttpJsonClientCall<RequestT, ResponseT> newC
...
+      // This is temporary here as we plan to migrate to java.util.Duration
+      if (callOptions.getTimeout() == null
+          || httpJsonContext
+                  .getTimeout()
+                  .compareTo(Duration.ofMillis(callOptions.getTimeout().toMillis()))
+              < 0) {
+        callOptions =
+            callOptions
+                .toBuilder()
+                .setTimeout(java.time.Duration.ofMillis(httpJsonContext.getTimeout().toMillis()))
```
lqiu96, line 69: @blakeli0 FYI, added some manual conversion between java.time.Duration and threetenbp.Duration

lqiu96:
@blakeli0 Took a look at the streaming side of things for deadlines. Seems like there are a few different deadlines able to be configured.

Stream Wait Timeout - Managed by Watchdog and tries the stream if the duration between messages is too long
Stream Idle Timeout - Managed by Watchdog and closes the stream if not more messages
RPC Timeout - Managed by HttpJsonClientCallImpl and the deadlineScheduler
Total Timeout - Helps determine if the RPC should be retried for each RPC call

This change should properly manage the that the entire result from the server-side streaming call finishes within the RPC timeout. It's not the timeout between messages as that is managed by a different timeout service (Watchdog).

gax-java/gax/src/main/java/com/google/api/gax/retrying/ExponentialRetryAlgorithm.java
blakeli0: I need to understand this scenario better, let's discuss it in Wednesday's meeting.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ManagedHttpJsonChannel.java
```diff
@@ -79,6 +88,7 @@ public synchronized void shutdown() {
       return;
     }
     try {
+      deadlineScheduledExecutorService.shutdown();
```
blakeli0, line 91: Noticed that `awaitTermination()` method below has a _TODO_ and it always returns false. It's a bug we should fix but not in scope of this PR, can you please create an issue to track it?

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonCallOptions.java
```diff
@@ -43,7 +43,7 @@ public abstract class HttpJsonCallOptions {
   public static final HttpJsonCallOptions DEFAULT = newBuilder().build();
 
   @Nullable
-  public abstract Instant getDeadline();
+  public abstract java.time.Duration getTimeout();
```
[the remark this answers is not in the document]
lqiu96: Done

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/testing/MockHttpService.java
```diff
@@ -46,6 +46,7 @@
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Queue;
+import org.threeten.bp.Duration;
```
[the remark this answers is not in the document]
lqiu96: Done

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/testing/MockHttpService.java
```diff
@@ -182,16 +188,33 @@ private static class MessageResponseFactory implements HttpResponseFactory {
...
       this.serviceMethodDescriptors = ImmutableList.copyOf(serviceMethodDescriptors);
       this.response = response;
+      this.delay = delay;
     }
 
     @Override
     public MockLowLevelHttpResponse getHttpResponse(String httpMethod, String fullTargetUrl) {
+      // We use Thread.sleep to mimic a long server response. Most tests should not
+      // require a sleep and can return a response immediately.
+      try {
```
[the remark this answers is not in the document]
lqiu96: Done

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ManagedHttpJsonChannel.java
lqiu96: Created here: https://github.com/googleapis/gapic-generator-java/issues/1663

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITLongRunningOperation.java
```diff
@@ -0,0 +1,266 @@
...
+import org.threeten.bp.temporal.ChronoUnit;
+
+/**
+ * For this test, we test a combination of various LRO RetrySettings and try to ensure that the
+ * calls are polling correctly. Each test attempts to test the number of attempts done in each call.
+ */
+public class ITLongRunningOperation {
+
+  @Test
+  public void testGRPC_LROSuccessfulResponse_NoDeadlineExceeded()
```
blakeli0, line 52: Does the test name still reflect what we are testing? Or maybe we need to make it clearer? 
I see `NoDeadlineExceeded` which assumes not retry, but I see an assertion in the end that asserts `attemptCount` to be at least 2.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITUnaryDeadline.java
```diff
@@ -0,0 +1,448 @@
...
+    RetrySettings defaultNoRetrySettings =
+        RetrySettings.newBuilder()
+            .setInitialRpcTimeout(Duration.ofMillis(5000L))
+            .setRpcTimeoutMultiplier(1.0)
+            .setMaxRpcTimeout(Duration.ofMillis(5000L))
+            .setTotalTimeout(Duration.ofMillis(5000L))
+            // Explicitly set retries as disabled (maxAttempts == 1)
+            .setMaxAttempts(1)
+            .build();
+    EchoStubSettings.Builder grpcEchoSettingsBuilder = EchoStubSettings.newBuilder();
```
blakeli0, line 63: Same thing for the setups in this class, it would be great if we can simplify them. Maybe we can expose a method in [TestClientInitializer](https://github.com/googleapis/gapic-generator-java/blob/main/showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/util/TestClientInitializer.java) that takes `RetrySettings` and `RetryableCodes` as parameters.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITUnaryDeadline.java
```diff
@@ -0,0 +1,448 @@
...
+                    .setChannelConfigurator(ManagedChannelBuilder::usePlaintext)
+                    .build())
+            .build();
+    try (EchoClient grpcClient = EchoClient.create(grpcEchoSettings)) {
+      BlockRequest blockRequest =
+          BlockRequest.newBuilder()
+              .setSuccess(BlockResponse.newBuilder().setContent("gRPCBlockContent_3sDelay_noRetry"))
+              .setResponseDelay(com.google.protobuf.Duration.newBuilder().setSeconds(3).build())
+              .build();
+      RetryingFuture<BlockResponse> retryingFuture =
```
blakeli0, line 81: Is this how customers would make use of retrys? If it is, we need to document it in the retry guide as well. I might be wrong but I thought we just make a unary call as usual and the retries would happen behind the scene.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITUnaryDeadline.java
```diff
@@ -0,0 +1,448 @@
...
+import org.threeten.bp.Duration;
+
+/**
+ * For this test, we test a combination of various RetrySettings to try and ensure that the timeouts
+ * set by the customer are cancelled when timeouts have exceeded their limits
+ *
+ * <p>Each test attempts to get the number of attempts done in each call. The attemptCount is
+ * incremented by 1 as the first attempt is zero indexed.
+ */
+public class ITUnaryDeadline {
```
blakeli0, line 49: Not part of this PR. It would be great if we can programmatically start/stop the showcase server so we can test things like connect/socket timeouts.

gax-java/gax/src/main/java/com/google/api/gax/retrying/ExponentialRetryAlgorithm.java
```diff
@@ -232,6 +229,15 @@ public boolean shouldRetry(TimedAttemptSettings nextAttemptSettings) {
     return true;
   }
 
+  // For non-LRO RPCs, do not attempt to retry if the totalTime spend is over
+  // the totalTimeout (timeout is in the past) or at the totalTimeout (timeout
+  // will occur immediately). For any other value, the deadlineScheduler will
+  // terminate in the future (even if the timeout is small).
+  @InternalApi
+  protected boolean shouldRPCTerminate(long timeLeftMs) {
```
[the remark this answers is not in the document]
lqiu96: Yeah, this is here due the the very *subtle* difference between unary/streaming vs LRO retry logic. Polling from LRO's should retry even if timeLeftMs == 0 (as it would be the final polling attempt), but unary/streaming callables should not (and it would signify an rpc timeout attempt of 0)

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITLongRunningOperation.java
```diff
@@ -0,0 +1,266 @@
...
+      throws IOException, ExecutionException, InterruptedException {
+    EchoStubSettings.Builder grpcEchoSettingsBuilder = EchoStubSettings.newBuilder();
```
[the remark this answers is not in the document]
lqiu96: Ah yep, will do.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITLongRunningOperation.java
lqiu96: Ah yep, I can update the test names to better clarity. Some tests cases are testing that the deadline exceeded and that it throws a DEADLINE_EXCEEDED exception and it's very easy to confuse.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITUnaryDeadline.java
lqiu96: It's not how customers should be using these callables by default. I've only exposed the RetryingFuture myself since that's the only way I could track the number of attempts. I'll add it to the guide as a way to get the number of attempts.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITUnaryDeadline.java
lqiu96: Might be something I can raise in the showcase repo and see what the other think (or see if that's possible / easily done).

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonCallOptions.java
```diff
@@ -64,9 +64,11 @@ public HttpJsonCallOptions merge(HttpJsonCallOptions inputOptions) {
 
     Builder builder = this.toBuilder();
 
-    Instant newDeadline = inputOptions.getDeadline();
-    if (newDeadline != null) {
-      builder.setDeadline(newDeadline);
+    if (inputOptions.getTimeout() != null) {
+      Duration newTimeout = java.time.Duration.ofNanos(inputOptions.getTimeout().getNano());
```
blakeli0, line 68: We can use ms instead of nanos now?

lqiu96: Done

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITUnaryDeadline.java
blakeli0: I meant more like doing it in our own repo, but yeah asking others if it has been done before is a good idea.

blakeli0 (approved):

blakeli0 (approved):

PR #1604, opened 2023-04-04
test: Add BIDI streaming showcase test.

blakeli0:
This PR does the following two things:
1. Add a basic BIDI streaming showcase test. 
2. Introduce a util class `TestClientInitializer` to centralize the initialization of test clients. 

This PR does not intend to test all scenarios of BIDI streaming, we may need more test cases for BIDI steaming.

### DIFF
--- a/showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITBidiStreaming.java
+++ b/showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITBidiStreaming.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.showcase.v1beta1.it;
+
+import com.google.api.core.SettableApiFuture;
+import com.google.api.gax.rpc.ClientStream;
+import com.google.api.gax.rpc.ResponseObserver;
+import com.google.api.gax.rpc.StreamController;
+import com.google.common.truth.Truth;
+import com.google.showcase.v1beta1.EchoClient;
+import com.google.showcase.v1beta1.EchoRequest;
+import com.google.showcase.v1beta1.EchoResponse;
+import com.google.showcase.v1beta1.it.util.TestClientInitializer;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+public class ITBidiStreaming {
+
+  private EchoClient grpcClient;
+
+  @Before
+  public void setUp() throws Exception {
+    // Create gRPC Echo Client
+    grpcClient = TestClientInitializer.createGrpcEchoClient();
+  }
+
+  // The current implementation of BIDI streaming on Echo showcase server is that it would echo the
+  // request content back on every request, so this test verifies that the response content is
+  // exactly the same as request content.
+  // Ideally we should make the BIDI streaming server more generic, e.g. only respond when there are
+  // three requests, respond twice for every request etc. If that happens, the response content may
+  // not be exactly the same as request content.
+  @Test
+  public void testGrpc_splitCall_shouldListensToResponse() throws Exception {
+    // given
+    List<String> expected = Arrays.asList("The rain in Spain stays mainly on the plain".split(" "));
+    TestResponseObserver responseObserver = new TestResponseObserver();
+
+    // when
+    ClientStream<EchoRequest> clientStream = grpcClient.chatCallable().splitCall(responseObserver);
+    expected.forEach(
+        requestContent -> {
+          EchoRequest request = EchoRequest.newBuilder().setContent(requestContent).build();
+          clientStream.send(request);
+        });
+    clientStream.closeSend();
+
+    // then
+    List<String> actual = responseObserver.getFuture().get();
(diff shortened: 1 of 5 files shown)


showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITBidiStreaming.java
```diff
@@ -0,0 +1,104 @@
...
+
+  // The current implementation of BIDI streaming on Echo showcase server is that it would echo the
+  // request content back on every request, so this test verifies that the response content is
+  // exactly the same as request content.
+  // Ideally we should make the BIDI streaming server more generic, e.g. only respond when there are
+  // three requests, respond twice for every request etc. If that happens, the response content may
+  // not be exactly the same as request content.
+  @Test
+  public void testGrpc_splitCall_shouldListensToResponse() throws Exception {
+    // given
```
mpeddada1, line 53: nit: The spacing you've added between the `arrange`, `act` and `assert` blocks makes the separation pretty clear so we may not need this comment. Same below.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITBidiStreaming.java
```diff
@@ -0,0 +1,104 @@
...
+    expected.forEach(
+        requestContent -> {
+          EchoRequest request = EchoRequest.newBuilder().setContent(requestContent).build();
+          clientStream.send(request);
+        });
+    clientStream.closeSend();
+
+    // then
+    List<String> actual = responseObserver.getFuture().get();
+    Truth.assertThat(actual).containsExactlyElementsIn(expected);
```
mpeddada1, line 68: How about we make `Truth.assertThat` a static import?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/util/TestClientInitializer.java
```diff
@@ -0,0 +1,52 @@
...
+
+package com.google.showcase.v1beta1.it.util;
+
+import com.google.api.client.http.javanet.NetHttpTransport;
+import com.google.api.gax.core.NoCredentialsProvider;
+import com.google.showcase.v1beta1.EchoClient;
+import com.google.showcase.v1beta1.EchoSettings;
+import io.grpc.ManagedChannelBuilder;
+
+public class TestClientInitializer {
```
burkedavison, line 25: +1

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITServerSideStreaming.java
```diff
@@ -46,29 +41,11 @@ public class ITServerSideStreaming {
   private EchoClient httpjsonClient;
 
   @Before
-  public void createClients() throws IOException, GeneralSecurityException {
+  public void createClients() throws Exception {
```
mpeddada1, line 44: I think the general recommendation is to keep the exceptions as specific as possible. In terms of documenting, it makes it easier for readers to understand what the method throws/why.
(https://effectivejava.googleplex.com/third-edition#page=325).

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITBidiStreaming.java
burkedavison, line 68: Should this have `inOrder()` attached to the end, or will order vary?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITBidiStreaming.java
```diff
@@ -0,0 +1,104 @@
...
+    // Create gRPC Echo Client
+    grpcClient = TestClientInitializer.createGrpcEchoClient();
+  }
+
+  // The current implementation of BIDI streaming on Echo showcase server is that it would echo the
+  // request content back on every request, so this test verifies that the response content is
+  // exactly the same as request content.
+  // Ideally we should make the BIDI streaming server more generic, e.g. only respond when there are
+  // three requests, respond twice for every request etc. If that happens, the response content may
+  // not be exactly the same as request content.
```
mpeddada1, line 50: @blakeli0 would you be able to elaborate on this? Would the # response per request be something that is configurable?

burkedavison (approved):

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITBidiStreaming.java
blakeli0: Thanks, just added `inOrder()`, I thought `containsExactlyElementsIn()` already does that.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITServerSideStreaming.java
blakeli0: I completely agree with the recommendation for methods in non-test classes. However, for test methods, especially in this case is a `setup` method, we don't need to understand which methods throws what checked exception, all we care is to have a client created. So it could be a distraction to have a lot of specific checked exceptions, by declaring a generic `Exception` could help testers to focus on the scenario as well.

Of course this could be subjective, there are some good [discussions](https://stackoverflow.com/questions/2496525/is-it-against-best-practice-to-throw-exception-on-most-junit-tests?rq=3) on StackOverFlow as well.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITBidiStreaming.java
blakeli0: Thanks for the suggestion! I think this would still help other developers to read the tests so I'll keep it.

lqiu96 (approved):

mpeddada1 (approved):

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITBidiStreaming.java
mpeddada1: Thanks for the context! Definitely agree with you, the callable showcase tests are doing very basic verifications at the moment so we should keep an eye out for complex use cases.

PR #1609, opened 2023-04-05
fix: add javadoc for `ApiFutures`

### LINKED ISSUE #1239: Missing Javadoc over methods in ApiFutures
Hello,

In one of our projects we're using some methods from the [`ApiFutures`](https://github.com/googleapis/api-common-java/blob/main/src/main/java/com/google/api/core/ApiFutures.java) class, which basically provide implementations of the methods from Guava's [`Futures`](https://github.com/google/guava/blob/master/guava/src/com/google/common/util/concurrent/Futures.java) class, but accept `ApiFuture`s, instead of `ListenableFuture`s.  Most (all?) of the implementations in `ApiFutures` are just simple delegators to the equivalent methods in the `Futures` class, after they wrapped the `ApiFuture` argument in a `ListenableFuture` object.

However, when we tried to understand the behavior of the methods in `ApiFutures`, without looking at their implementations, we noticed that they lack Javadoc, not even simple `@see` references to the equivlaent method in the `Futures` class. People who are not familiar with Java's `Future`s, and read code that uses methods from `ApiFutures` for the first time, struggle to understand their semantics without looking at the implementations.

Can you please add the missing Javadoc over the methods in this class?

Thanks!

JoeWang1127:
Fixes #1239 ☕️

### DIFF
--- a/api-common-java/src/main/java/com/google/api/core/ApiFutures.java
+++ b/api-common-java/src/main/java/com/google/api/core/ApiFutures.java
@@ -46,6 +46,11 @@ public final class ApiFutures {
   private ApiFutures() {}
 
   /**
+   * Registers separate success and failure callbacks to be run when the {@link ApiFuture}'s
+   * computation is complete or, if the computation is already complete, immediately.
+   *
+   * @param future The future attach the callback to
+   * @param callback The callback to invoke when future is completed
    * @deprecated Use {@linkplain #addCallback(ApiFuture, ApiFutureCallback, Executor) the overload
    *     that requires an executor}. For identical behavior, pass {@link
    *     com.google.common.util.concurrent.MoreExecutors#directExecutor}, but consider whether
@@ -64,9 +69,9 @@ public final class ApiFutures {
    * <p>Note that this method is a delegate of {@link Futures#addCallback(ListenableFuture,
    * FutureCallback, Executor)}.
    *
-   * @param future The future attach the callback to.
-   * @param callback The callback to invoke when future is completed.
-   * @param executor The executor to run callback when the future completes.
+   * @param future The future attach the callback to
+   * @param callback The callback to invoke when future is completed
+   * @param executor The executor to run callback when the future completes
    * @see Futures#addCallback(ListenableFuture, FutureCallback, Executor)
    */
   public static <V> void addCallback(
@@ -88,6 +93,15 @@ public final class ApiFutures {
   }
 
   /**
+   * Returns an {@link ApiFuture} whose result is taken from the given primary input or, if the
+   * primary input fails with the given exceptionType, from the result provided by the callback.
+   *
+   * @param input The primary input {@code ApiFuture}
+   * @param exceptionType The exception type that triggers use of {@code fallback}
+   * @param callback The {@link ApiFunction} to be called if input fails with the expected exception
+   *     type
+   * @return A future whose result is taken either from the given {@code input} or by the {@code
+   *     callback}
    * @deprecated Use {@linkplain #catching(ApiFuture, Class, ApiFunction, Executor) the overload
    *     that requires an executor}. For identical behavior, pass {@link
    *     com.google.common.util.concurrent.MoreExecutors#directExecutor}, but consider whether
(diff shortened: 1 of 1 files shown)


api-common-java/src/main/java/com/google/api/core/ApiFutures.java
```diff
@@ -159,57 +232,89 @@ public static <V, X> ApiFuture<X> transform(
...
     return new ListenableFutureToApiFuture<>(
         Futures.allAsList(
             Iterables.transform(
                 futures,
-                new Function<ApiFuture<? extends V>, ListenableFuture<? extends V>>() {
-                  public ListenableFuture<? extends V> apply(ApiFuture<? extends V> apiFuture) {
-                    return listenableFutureForApiFuture(apiFuture);
-                  }
-                })));
+                (Function<ApiFuture<? extends V>, ListenableFuture<? extends V>>)
```
burkedavison: Please remove the cast. It shouldn't be needed.

api-common-java/src/main/java/com/google/api/core/ApiFutures.java
```diff
@@ -159,57 +232,89 @@ public static <V, X> ApiFuture<X> transform(
...
-                  }
-                })));
-  }
-
-  /*
-   * @deprecated Use {@linkplain #transformAsync(ApiFuture, ApiFunction, Executor) the
-   * overload that requires an executor}. For identical behavior, pass {@link
-   * com.google.common.util.concurrent.MoreExecutors#directExecutor}, but consider whether
-   * another executor would be safer.
+                (Function<ApiFuture<? extends V>, ListenableFuture<? extends V>>)
```
burkedavison: Please remove the cast. It shouldn't be needed.

api-common-java/src/main/java/com/google/api/core/ApiFutures.java
```diff
@@ -45,18 +45,30 @@
...
-   * another executor would be safer.
+  /**
+   * @deprecated Use {@linkplain #addCallback(ApiFuture, ApiFutureCallback, Executor) the overload
+   *     that requires an executor}. For identical behavior, pass {@link
+   *     com.google.common.util.concurrent.MoreExecutors#directExecutor}, but consider whether
+   *     another executor would be safer.
    */
   @Deprecated
   public static <V> void addCallback(
       final ApiFuture<V> future, final ApiFutureCallback<? super V> callback) {
```
burkedavison, line 56: One major intent of this ticket is to improve the reference API documentation found at https://cloud.google.com/java/docs/reference/api-common/latest/com.google.api.core.ApiFutures, so please make sure that each of these methods have entries for `@param`, `@return`, and `@throws` as appropriate.

burkedavison (changes requested):

api-common-java/src/main/java/com/google/api/core/ApiFutures.java
JoeWang1127: This is outdated.

api-common-java/src/main/java/com/google/api/core/ApiFutures.java
JoeWang1127: This is outdated.

api-common-java/src/main/java/com/google/api/core/ApiFutures.java
JoeWang1127: done.

api-common-java/src/main/java/com/google/api/core/ApiFutures.java
```diff
@@ -45,18 +45,35 @@
 public final class ApiFutures {
   private ApiFutures() {}
 
-  /*
-   * @deprecated Use {@linkplain #addCallback(ApiFuture, ApiFutureCallback, Executor) the
-   * overload that requires an executor}. For identical behavior, pass {@link
-   * com.google.common.util.concurrent.MoreExecutors#directExecutor}, but consider whether
-   * another executor would be safer.
+  /**
+   * Registers separate success and failure callbacks to be run when the {@link ApiFuture}'s
```
blakeli0, line 49:
```suggestion
   * Registers a callback to be run when the {@link ApiFuture}'s
```
It's just one callback being passed in, the original doc could confuse people that there are multiple callbacks passed in.

api-common-java/src/main/java/com/google/api/core/ApiFutures.java
```diff
@@ -45,18 +45,35 @@
...
+   *     another executor would be safer.
    */
   @Deprecated
   public static <V> void addCallback(
       final ApiFuture<V> future, final ApiFutureCallback<? super V> callback) {
     addCallback(future, callback, directExecutor());
   }
 
+  /**
+   * Registers separate success and failure callbacks to be run when the {@link ApiFuture}'s
```
blakeli0, line 66:
```suggestion
   * Registers a callback to be run when the {@link ApiFuture}'s
```
Same as the above.

api-common-java/src/main/java/com/google/api/core/ApiFutures.java
JoeWang1127: done

api-common-java/src/main/java/com/google/api/core/ApiFutures.java
JoeWang1127: done

blakeli0 (approved):

burkedavison:
LGTM

burkedavison (approved):

PR #1622, opened 2023-04-11
chore: test protobuf 23.x

burkedavison:
This is a test PR only. Do not merge.


burkedavison:
@blakeli0 : I'm surprised that Showcase required a golden file update, but the other golden files didn't. Have we intentionally removed the ("legacy") golden file tests?

blakeli0:
> @blakeli0 : I'm surprised that Showcase required a golden file update, but the other golden files didn't. Have we intentionally removed the ("legacy") golden file tests?

It actually makes sense and is working as intended. All the required golden files update are from the proto folder, which are generated from protoc. Our golden unit/integration tests are only for the files generated by `gapic-generator-java`, not the files generated by protoc.

.bazeliskrc
```diff
@@ -1,2 +1,2 @@
 # See https://github.com/bazelbuild/bazelisk
-USE_BAZEL_VERSION=6.0.0
```
blakeli0, line 2: BTW, we are still using 6.0.0 in [googleapis](https://github.com/googleapis/googleapis/blob/288aa7fb71c9b6244db1a816cfd4108f811e6049/.bazeliskrc#L2), so it's better to keep in sync with it.

burkedavison: Reverted.

PR #1642, opened 2023-04-20
feat: support GDC-H Credentials 

diegomarquezp:
This PR adds support for GDC-H credentials.
Mainly, it adds the option to have `gdchApiAudience` in `ClientContext` and `StubSettings`.
`ClientContext.create(StubSettings)` will verify that:

- if `gdchApiAudience` is set, then only a `GdchCredential` returned from the `CredentialProvider` is accepted
- `gdchApiAudience` is a valid URI


showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
```diff
@@ -0,0 +1,183 @@
...
+    assertNull(unexpected);
+    GdchCredentials fromClient =
+            (GdchCredentials) client.getSettings().getCredentialsProvider().getCredentials();
+    URI audienceFromClientProvider = fromClient.getApiAudience();
+    assertNotNull(audienceFromClientProvider);
+    assertTrue(audienceFromClientProvider.equals(new URI(audience)));
+    assertTrue(
+            "GDCH credentials with audience should be GdchCredentials",
+            credentials instanceof GdchCredentials);
+    GdchCredentials fromContext = null;
```
diegomarquezp, line 133: @blakeli0 I'm looking for a way to access the final credentials that are created with `ClientContext` as in [here](https://github.com/googleapis/sdk-platform-java/blob/ea05922a7010a274db9e38be3940cc54a02a585b/gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java#L185-L186). 

I believe subclassing `EchoSettings` or `EchoStub[Settings]` to capture the context and hold it so I can assert some aspects of it may be the way, but I wonder if there is a more straightforward way of accessing the call context. Thanks!

gax-java/gax/src/test/java/com/google/api/gax/rpc/ClientContextTest.java
```diff
@@ -30,10 +30,7 @@
 package com.google.api.gax.rpc;
 
 import static com.google.common.truth.Truth.assertThat;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
+import static org.junit.Assert.*;
```
blakeli0, line 33: You may need to adjust the intellij settings to not import `*`, see this [guide](go/java-style#s3.3.1-wildcard-imports) for no wildcard imports.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
```diff
@@ -0,0 +1,183 @@
+package com.google.showcase.v1beta1.it;
+
+import static org.junit.Assert.*;
```
blakeli0, line 3: Same thing as previously, no wildcard imports.

gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
```diff
@@ -325,6 +348,8 @@ public abstract static class Builder {
     @BetaApi("The surface for tracing is not stable yet and may change in the future.")
     public abstract Builder setTracerFactory(ApiTracerFactory tracerFactory);
 
+    public abstract Builder setGdchApiAudience(String credentialsApiAudience);
```
blakeli0, line 351: It's better to keep the field name and getter name consistent.

gax-java/gax/src/main/java/com/google/api/gax/rpc/StubSettings.java
```diff
@@ -289,6 +300,7 @@ protected Builder(ClientContext clientContext) {
         this.streamWatchdogCheckInterval = clientContext.getStreamWatchdogCheckInterval();
         this.tracerFactory = clientContext.getTracerFactory();
         this.quotaProjectId = getQuotaProjectIdFromClientContext(clientContext);
+        this.gdchApiAudience = clientContext.getGdchApiAudience();
```
blakeli0, line 303: This is probably a larger design issue, I'm not sure why we need to create `StubSettings` from `ClientContext`. Technically we only need to set `gdchApiAudience` to `StubSettings`, use it to get credentials when create `ClientContext`. If we don't need to support this case, we wouldn't need to expose getter and setters for `gdchApiAudience` in `ClientContext`.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
blakeli0: I don't think we need to verify the credential from `CientContext` directly, verify it from the client is good enough. In addition, I know you probably want to test the scenario that a `StubSetting` can be created from `ClientContext` but as I mentioned in previous comment, I'm not sure that's a valid case even though we currently support it.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
```diff
@@ -0,0 +1,183 @@
...
+            .setGdchApiAudience("any-audience")
+            .build();
+    Exception expected = getExceptionFromClientCreation();
+    assertNotNull(expected);
+    assertTrue(expected instanceof IllegalArgumentException);
+    assertTrue(
+        expected.getMessage().contains("audience can only be set when using GdchCredentials"));
+  }
+
+  private Exception getExceptionFromClientCreation() {
```
blakeli0, line 161: This helper method is actually more confusing to me. Since most of the cases are not throwing, I don't think we need this and we can `assertThrows` for cases that need it, which is more readable as well.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
```diff
@@ -0,0 +1,183 @@
...
+  @Rule public TemporaryFolder tempFolder = new TemporaryFolder();
+
+  private EchoClient client;
+  private EchoSettings settings;
+  private Credentials credentials;
+
+  @Before
+  public void setup() throws IOException, URISyntaxException {
+    prepareCredentials();
+    client = null;
```
blakeli0, line 47: We don't have to set it to null, the default value is null.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
```diff
@@ -0,0 +1,183 @@
...
+    Exception expected = getExceptionFromClientCreation();
+    assertNotNull(expected);
+    assertTrue(expected instanceof IllegalArgumentException);
+    assertTrue(
+        expected.getMessage().contains("audience can only be set when using GdchCredentials"));
+  }
+
+  private Exception getExceptionFromClientCreation() {
+    try {
+      client = EchoClient.create(settings);
```
blakeli0, line 163: We probably want to close the client in a clean up method.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
```diff
@@ -0,0 +1,183 @@
...
+  public void testClientWithGdchCredentialWithValidAudience_correct()
+      throws IOException, URISyntaxException {
+    String audience = "valid-audience";
+    settings = settings.toBuilder().setGdchApiAudience(audience).build();
+    Exception unexpected = getExceptionFromClientCreation();
+    assertNull(unexpected);
+    GdchCredentials fromClient =
+            (GdchCredentials) client.getSettings().getCredentialsProvider().getCredentials();
+
+    assertSame(credentials, fromClient);
```
blakeli0, line 107: I thought the credential would be different since it now has an additional field `gdchApiAudience`?

diegomarquezp: So the provider is not modified when creating the `ClientContext`. In fact, `ClientContext.create()` retrieves the `Credentials` from the provider, then modifies them accordingly (e.g. re-create with `gdchApiAudience` or any other type of `Credentials) and stores them as an internal property. 

I'm just describing what happens with `ClientContext` creation. 
For example, 
https://github.com/googleapis/sdk-platform-java/blob/ea05922a7010a274db9e38be3940cc54a02a585b/gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java#L193-L199

Modifies the `Credentials` object and then a new `ClientContext` is created with such `Credentials`

https://github.com/googleapis/sdk-platform-java/blob/ea05922a7010a274db9e38be3940cc54a02a585b/gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java#L261-L264

Should `ClientContext.create()` alter the `Credentials` provided by `CredentialsProvider`?

gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
```diff
@@ -167,6 +173,23 @@ public static ClientContext create(StubSettings settings) throws IOException {
 
     Credentials credentials = settings.getCredentialsProvider().getCredentials();
 
+    String gdhcApiAudience = settings.getGdchApiAudience();
+    if (gdhcApiAudience != null && credentials instanceof GdchCredentials) {
```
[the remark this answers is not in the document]
diegomarquezp: The audience is optional. I agree with this logic being more readable!

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
diegomarquezp: Right, my purpose was to confirm that a new `Credentials` is created if I set `gdchApiAudience`. Since the `CredentialsProvider` will always return the initial `Credentials` (without audience), 

I was just looking for a way that `Context.create()` is actually using `Credentials` with audience when specified. 

(`ClientContext.create()` logic on recreating `Credentials` explained [here](https://github.com/googleapis/sdk-platform-java/pull/1642/files#r1204754450))

gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
```diff
@@ -167,6 +173,23 @@ public static ClientContext create(StubSettings settings) throws IOException {
...
+      // We recompute the GdchCredentials with the audience
+      URI gdchAudienceUri;
+      try {
+        gdchAudienceUri = URI.create(gdhcApiAudience);
+      } catch (IllegalArgumentException ex) { // thrown when passing a malformed uri string
+        throw new IllegalArgumentException("The GDC-H API audience string is not a valid URI", ex);
+      }
+      credentials =
+          ((GdchCredentials) credentials).toBuilder().setGdchAudience(gdchAudienceUri).build();
```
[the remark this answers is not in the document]
diegomarquezp: That method makes more sense. I changed it

gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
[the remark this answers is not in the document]
diegomarquezp: Modified accordingly

gax-java/gax/src/test/java/com/google/api/gax/rpc/ClientContextTest.java
diegomarquezp: Thanks, removing

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
diegomarquezp: Removed

gax-java/gax/src/main/java/com/google/api/gax/rpc/StubSettings.java
```diff
@@ -435,6 +447,11 @@ public B setStreamWatchdogCheckInterval(@Nonnull Duration checkInterval) {
       return self();
     }
 
+    public B setGdchApiAudience(String gdchApiAudience) {
```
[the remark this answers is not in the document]
diegomarquezp: Added javadoc on public getters and setters involving `gdchApiAudience`

gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
diegomarquezp: Done

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
diegomarquezp: Done!

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
diegomarquezp: added tear down method to the test suite that closes the client

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
diegomarquezp: `assertThrows` indeed seems more convenient. Using it now!

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
diegomarquezp: As discussed today, it makes more sense to test credentials from `ClientContext` in a unit test since it's more about the internal workings of gax. I added a special test case

diegomarquezp:
@TimurSadykov hoping you could take a look at this PR

gax-java/gax/src/main/java/com/google/api/gax/rpc/StubSettings.java
TimurSadykov: I don't have context on why, but seems like creation of the StubSettings from ClientContext is an existing historical thing and therefore it probably make sense to carry forward?

blakeli0: Yes, Diego and I discussed offline and we agreed to keep it.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
TimurSadykov: While the logic above is correct, the test seems not validating what it declares (Blake's point from different angle). The test description WithAudience - is is not tested. 

Instead, test validates that credentials stay in credential provider as is. 

This is effectively a unit test, not integration

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
```diff
@@ -0,0 +1,138 @@
...
+   * Confirms setting a valid audience is successful.
+   *
+   * <p>The provider's credentials do not get altered when adding audience. They are recreated and
+   * used only inside `ClientContext`
+   *
+   * @throws IOException
+   * @throws URISyntaxException
+   */
+  @Test
+  public void testClientWithGdchCredentialWithValidAudience_correct()
```
TimurSadykov, line 106: Can we have an integration test that trigger refresh and check that aud exists in the request via fake transport? Probably other test cases could be validated similarly via refresh as well

TimurSadykov (changes requested):
Overall looks good, comments on integration test

gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
```diff
@@ -167,6 +177,23 @@ public static ClientContext create(StubSettings settings) throws IOException {
 
     Credentials credentials = settings.getCredentialsProvider().getCredentials();
 
+    String gdhcApiAudience = settings.getGdchApiAudience();
```
blakeli0, line 180: nit: `gdchApiAudience` not `gdhcApiAudience` ;)

gax-java/gax/src/test/java/com/google/api/gax/rpc/ClientContextTest.java
```diff
@@ -790,4 +797,109 @@ public void testExecutorSettings() throws Exception {
...
+    return creds;
+  }
+
+  private TransportChannelProvider getFakeTransportChannelProvider() {
+    return new FakeTransportProvider(
+        FakeTransportChannel.create(new FakeChannel()), null, true, null, null);
+  }
+
+  @Test
+  public void testGdchCredentialNoAudience_correct() throws IOException {
```
blakeli0, line 821: See this [guide](https://engdoc.corp.google.com/eng/doc/devguide/testing/unit/best-practices.md?cl=head&polyglot=java#naming) for naming patterns, this test name could be `testCreateClientContext_withGDCHCredentialAndNoAudience`, the error test case could be `testCreateClientContext_withGDCHCredentialAndInvalidAudience_throwsException`. Or something like `shouldThrowsExceptionOnCreatingClientContextWithGDCHCredentialAndInvalidAudience`

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
blakeli0: > Also, this is effectively a unit test, not integration, right?

We want to test that the audience set through `ClientSetting` in a client library is being properly propagated to `ClientContext` and set to `Credentials`, which we cannot test through unit tests, only through showcase clients. So yeah this is not a true integration tests that interact with servers, more like a component test only verifying the client behaviors.
That being said, I agree with you that this test does not verify the audience, which we should.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
diegomarquezp: I just modified the tests to confirm audience works by refreshing the token. I imported some test classes from `google-auth-library`. I had to propose a PR for this https://github.com/googleapis/google-auth-library-java/pull/1224

diegomarquezp:
Now depends on https://github.com/googleapis/google-auth-library-java/pull/1224

gax-java/gax/src/test/java/com/google/api/gax/rpc/ClientContextTest.java
diegomarquezp: Corrected names

showcase/gapic-showcase/pom.xml
```diff
@@ -229,5 +229,13 @@
       <classifier>testlib</classifier>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
blakeli0, line 232: Even though it is a test jar that is only used in showcase, we may still want to manage it in a centralized place like [here](https://github.com/googleapis/sdk-platform-java/blob/main/gapic-generator-java/pom.xml), WDYT @suztomo?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
```diff
@@ -0,0 +1,196 @@
...
+  public void testCreateClient_withGdchCredentialAndNoAudience() throws IOException {
+    context = ClientContext.create(settings);
+    stubSettings = EchoStubSettings.newBuilder(context).build();
+    client = EchoClient.create(stubSettings.createStub());
+    Credentials fromContext = context.getCredentials();
+    Credentials fromClient = initialCredentials;
+    assertNotSame(fromContext, fromClient);
+    NullPointerException expectedEx =
+        assertThrows(NullPointerException.class, () -> initialCredentials.refresh());
+    assertTrue(
```
blakeli0, line 121: This is verifying an exception from auth library, which I don't think we would want to test. The test case should focus on what we added in client libraries.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
```diff
@@ -0,0 +1,196 @@
...
+        expectedEx.getMessage().contains("Audience are not configured for GDCH service account"));
+    Exception unexpected = null;
+    try {
+      registerCredential(fromContext);
+      ((GdchCredentials) fromContext).refreshAccessToken();
+    } catch (Exception ex) {
+      unexpected = ex;
+    }
+    assertNull(unexpected);
```
blakeli0, line 130: Usually we don't have to specifically verify that no exception being thrown. The test case would fail naturally if there is an exception thrown, just `((GdchCredentials) fromContext).refreshAccessToken();` with a comment is more readable.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
```diff
@@ -0,0 +1,196 @@
...
+    Credentials fromClient = initialCredentials;
+    assertNotSame(fromContext, fromClient);
+    NullPointerException expectedEx =
+        assertThrows(NullPointerException.class, () -> initialCredentials.refresh());
+    assertTrue(
+        expectedEx.getMessage().contains("Audience are not configured for GDCH service account"));
+    Exception unexpected = null;
+    try {
+      registerCredential(fromContext);
+      ((GdchCredentials) fromContext).refreshAccessToken();
```
blakeli0, line 126: I was expecting refresh token would fail without audience being set through ClientSettings?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
```diff
@@ -0,0 +1,196 @@
...
+            .getMessage()
+            .contains("Audience are not configured for GDCH service account"));
+    Exception unexpected = null;
+    try {
+      registerCredential(fromContext);
+      ((GdchCredentials) fromContext).refreshAccessToken();
+    } catch (Exception ex) {
+      unexpected = ex;
+    }
+    assertNull(unexpected);
```
blakeli0, line 162: The behavior looks identical to me with or without audience, am I missing something?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  @Test
+  public void testCreateClient_withGdchCredentialWithInvalidAudience_throws() throws IOException {
```
blakeli0, line 166: This and the next test cases are already tested through unit tests, I don't think we have to test them again in showcase.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
diegomarquezp: The audience will be equal to the `endpoint` of the `ClientContext` if no audience is explicitly set. This confirms that the refresh mechanism should work anyways.

https://github.com/googleapis/sdk-platform-java/blob/6f9f4c55dfc60047e29e6d041538691d9e1ca33a/gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java#L185-L191

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
diegomarquezp: What changes is the underlying audience being used in the `GdchCredentials`. The `audience` of the credentials is private and we cannot inspect them. I thought confirming both ways of instantiating a client context may work

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
diegomarquezp: removed

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
diegomarquezp: added a simple call with comment instead. Thanks!

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
```diff
@@ -0,0 +1,196 @@
...
+
+  /**
+   * {@link com.google.api.gax.rpc.ClientContext} will create a new {@link GdchCredentials} with an
+   * audience defaulted to the endpoint if the audience is not manually passed. This test confirms
+   * that a new credential is created from the context and can be refreshed
+   *
+   * @throws IOException
+   */
+  @Test
+  public void testCreateClient_withGdchCredentialAndNoAudience() throws IOException {
```
diegomarquezp, line 112: Maybe add more inline comments

diegomarquezp: Also add outcome to test name

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
diegomarquezp: Added a transport interceptor to confirm which audience is used

showcase/gapic-showcase/pom.xml
suztomo: Use the google-cloud-shared-dependencies. Then it matches what a normal client library would work.

diegomarquezp: Thanks - I added the test-jar reference to shared dependencies

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
```diff
@@ -0,0 +1,206 @@
...
+    assertTrue(
+        expectedEx.getMessage().contains("Audience are not configured for GDCH service account"));
+
+    // However, the credentials prepared in ClientContext should be able to refresh since the
+    // audience would be
+    // internally defaulted the endpoint of the StubSettings
+    registerCredential(fromContext);
+    ((GdchCredentials) fromContext).refreshAccessToken();
+    String usedAudience = transportFactory.transport.getLastAudienceSent();
+    assertNotNull(usedAudience);
```
blakeli0, line 146: This is redundant as we are already verifying that the audience is equal to endpoint.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
```diff
@@ -0,0 +1,206 @@
...
+            .contains("Audience are not configured for GDCH service account"));
+    Exception unexpected = null;
+
+    // But the credentials prepared in ClientContext should be able to refresh since the audience
+    // would be internally
+    // set to the one passed in stub settings
+    registerCredential(fromContext);
+    ((GdchCredentials) fromContext).refreshAccessToken();
+    String usedAudience = transportFactory.transport.getLastAudienceSent();
+    assertNotNull(usedAudience);
```
blakeli0, line 193: Same as the above test.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/util/InterceptingMockTokenServerTransport.java
```diff
@@ -0,0 +1,38 @@
...
+
+    public String getLastAudienceSent() {
+        try {
+            String contentString = lastRequest.getContentAsString();
+            Map<String, String> query = TestUtils.parseQuery(contentString);
+            String assertion = query.get("assertion");
+            JsonWebSignature signature = JsonWebSignature.parse(JSON_FACTORY, assertion);
+            String foundTargetAudience = (String) signature.getPayload().get("api_audience");
+            return foundTargetAudience;
+        } catch (Exception ex) {
```
blakeli0, line 34: What's the exception you are expecting? If it's a checked exception, just add it to the method signature is good enough.

java-shared-dependencies/first-party-dependencies/pom.xml
```diff
@@ -92,6 +92,14 @@
         <version>2.20.1-SNAPSHOT</version><!-- {x-version-update:google-cloud-core:current} -->
         <classifier>tests</classifier>
       </dependency>
+      <dependency>
+        <groupId>com.google.auth</groupId>
+        <artifactId>google-auth-library-oauth2-http</artifactId>
+        <version>1.17.2-SNAPSHOT</version>
```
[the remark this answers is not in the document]
diegomarquezp: @blakeli0 I think that variable is unrelated. It is used in 
```
        <groupId>com.google.oauth-client</groupId>
        <artifactId>google-oauth-client-bom</artifactId>

```

gax-java/gax/src/test/java/com/google/api/gax/rpc/ClientContextTest.java
```diff
@@ -790,4 +798,156 @@ public void testExecutorSettings() throws Exception {
...
+    TransportChannelProvider transportChannelProvider = getFakeTransportChannelProvider();
+    Credentials creds = getMockGdchCredentials();
+
+    CredentialsProvider provider = FixedCredentialsProvider.create(creds);
+    StubSettings settings = new FakeStubSettings.Builder().setGdchApiAudience(null).build();
+    FakeClientSettings.Builder clientSettingsBuilder = new FakeClientSettings.Builder(settings);
+    clientSettingsBuilder.setCredentialsProvider(provider);
+    clientSettingsBuilder.setTransportChannelProvider(transportChannelProvider);
+
+    // should not throw
```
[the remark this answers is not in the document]
diegomarquezp: Changed

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
```diff
@@ -0,0 +1,206 @@
...
+
+    // Again, since the initial credentials don't have an audience, we should not be able to refresh
+    // them
+    NullPointerException thrownByClientCreds =
+        assertThrows(NullPointerException.class, () -> initialCredentials.refresh());
+    assertTrue(
+        thrownByClientCreds
+            .getMessage()
+            .contains("Audience are not configured for GDCH service account"));
+    Exception unexpected = null;
```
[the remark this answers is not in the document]
diegomarquezp: Removed

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
diegomarquezp: Removed

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
diegomarquezp: Removed

java-shared-dependencies/first-party-dependencies/pom.xml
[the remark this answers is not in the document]
blakeli0: The auth [bom](https://github.com/googleapis/google-auth-library-java/blob/12e8db6025e0263b801d5385844924a4f5ff7b7e/bom/pom.xml#L59) includes `google-auth-library-oauth2-http` and they have the same version number. We should be able to either not define a separate version for it, or reuse the existing property.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/util/InterceptingMockTokenServerTransport.java
diegomarquezp: Right - moved to method signature

gapic-generator-java-pom-parent/pom.xml
```diff
@@ -27,7 +27,7 @@
         consistent across modules in this repository -->
     <javax.annotation-api.version>1.3.2</javax.annotation-api.version>
     <grpc.version>1.55.1</grpc.version>
-    <google.auth.version>1.18.0</google.auth.version>
+    <google.auth.version>1.19.0</google.auth.version>
```
blakeli0, line 30: We can revert this now that we have a separate deps PR for it.

java-shared-dependencies/first-party-dependencies/pom.xml
```diff
@@ -92,6 +92,13 @@
         <version>2.20.1-SNAPSHOT</version><!-- {x-version-update:google-cloud-core:current} -->
         <classifier>tests</classifier>
       </dependency>
+      <dependency>
```
blakeli0, line 95: I don't think we need it now that we use `${google.auth.version}` in showcase pom.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
```diff
@@ -0,0 +1,202 @@
+package com.google.showcase.v1beta1.it;
```
blakeli0, line 1: Missing copyright headers for this file and all other new files.

showcase/pom.xml
```diff
@@ -53,7 +53,7 @@
         <version>4.13.2</version>
         <scope>test</scope>
       </dependency>
-    </dependencies>
+  </dependencies>
```
blakeli0, line 56: This change should be reverted.

java-shared-dependencies/first-party-dependencies/pom.xml
diegomarquezp: removed

showcase/pom.xml
diegomarquezp: reverted

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
diegomarquezp: added copyright to the 3 new files

gapic-generator-java-pom-parent/pom.xml
diegomarquezp: reverted and merged `main` into this PR

showcase/gapic-showcase/src/test/resources/META-INF/native-image/resource-config.json
```diff
@@ -1,3 +1,3 @@
 {
-  "resources":[{"pattern": ".*.json"}]
+  "resources":[{"pattern": ".*.json"}, {"pattern":  ".*.pem"}]
```
blakeli0, line 2: Noticed that this is in `native-image` folder, is this change required for native image tests?

blakeli0 (approved):
LGTM with one minor comment. Feel free to merge it after the comment is resolved.

java-shared-dependencies/first-party-dependencies/pom.xml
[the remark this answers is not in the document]
blakeli0: Decide to use `${google.auth.version}` from parent pom in showcase pom.

showcase/gapic-showcase/src/test/resources/META-INF/native-image/resource-config.json
diegomarquezp: It was not necessary!

PR #1667, opened 2023-05-05
Main dynamic routing headers

### LINKED ISSUE #1666: Implement Dynamic Routing Headers for HttpJson
Implement similar functionality with gRPC's Dynamic Routing Header implementation.

Try to re-use as much code as possible from gRPC's implementation.

lqiu96:
[WIP] This is a *currently* a draft PR from the implementation plan created a few days ago.

Fixes #1666 ☕️

### DIFF
--- a/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
+++ b/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
@@ -23,6 +23,7 @@ import com.google.api.gax.rpc.ClientContext;
 import com.google.api.gax.rpc.ClientStreamingCallable;
 import com.google.api.gax.rpc.LongRunningClient;
 import com.google.api.gax.rpc.OperationCallable;
+import com.google.api.gax.rpc.RequestParamsBuilder;
 import com.google.api.gax.rpc.RequestParamsExtractor;
 import com.google.api.gax.rpc.ServerStreamingCallable;
 import com.google.api.gax.rpc.UnaryCallable;
@@ -34,13 +35,18 @@ import com.google.api.generator.engine.ast.ConcreteReference;
 import com.google.api.generator.engine.ast.EmptyLineStatement;
 import com.google.api.generator.engine.ast.Expr;
 import com.google.api.generator.engine.ast.ExprStatement;
+import com.google.api.generator.engine.ast.IfStatement;
 import com.google.api.generator.engine.ast.JavaDocComment;
+import com.google.api.generator.engine.ast.LambdaExpr;
+import com.google.api.generator.engine.ast.LogicalOperationExpr;
 import com.google.api.generator.engine.ast.MethodDefinition;
 import com.google.api.generator.engine.ast.MethodInvocationExpr;
 import com.google.api.generator.engine.ast.NewObjectExpr;
 import com.google.api.generator.engine.ast.ReferenceConstructorExpr;
+import com.google.api.generator.engine.ast.RelationalOperationExpr;
 import com.google.api.generator.engine.ast.ScopeNode;
 import com.google.api.generator.engine.ast.Statement;
+import com.google.api.generator.engine.ast.StringObjectValue;
 import com.google.api.generator.engine.ast.ThisObjectValue;
 import com.google.api.generator.engine.ast.ThrowExpr;
 import com.google.api.generator.engine.ast.TryCatchStatement;
@@ -56,11 +62,15 @@ import com.google.api.generator.gapic.model.GapicClass;
 import com.google.api.generator.gapic.model.GapicClass.Kind;
 import com.google.api.generator.gapic.model.GapicContext;
 import com.google.api.generator.gapic.model.GapicServiceConfig;
+import com.google.api.generator.gapic.model.HttpBindings;
 import com.google.api.generator.gapic.model.Message;
 import com.google.api.generator.gapic.model.Method;
+import com.google.api.generator.gapic.model.RoutingHeaderRule;
 import com.google.api.generator.gapic.model.Service;
 import com.google.api.generator.gapic.utils.JavaStyle;
+import com.google.api.pathtemplate.PathTemplate;
 import com.google.common.base.Preconditions;
+import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.longrunning.Operation;
@@ -1167,5 +1177,249 @@ public abstract class AbstractTransportServiceStubClassComposer implements Class
     }
 
     return transportOpeationsStubType;
+  }
+
+  protected LambdaExpr createRequestParamsExtractorClassInstance(
+      Method method, List<Statement> classStatements) {
+    List<Statement> bodyStatements = new ArrayList<>();
+    VariableExpr requestVarExpr =
+        VariableExpr.withVariable(
+            Variable.builder().setType(method.inputType()).setName("request").build());
+    TypeNode returnType =
+        TypeNode.withReference(
+            ConcreteReference.builder()
+                .setClazz(Map.class)
+                .setGenerics(TypeNode.STRING.reference(), TypeNode.STRING.reference())
+                .build());
+    MethodInvocationExpr.Builder returnExpr =
+        MethodInvocationExpr.builder().setReturnType(returnType);
+    // If the google.api.routing annotation is present(even with empty routing parameters),
+    // the implicit routing headers specified in the google.api.http annotation should not be sent
+    if (method.routingHeaderRule() == null) {
+      createRequestParamsExtractorBodyForHttpBindings(
+          method, requestVarExpr, bodyStatements, returnExpr);
+    } else {
+      createRequestParamsExtractorBodyForRoutingHeaders(
+          method, requestVarExpr, classStatements, bodyStatements, returnExpr);
+    }
+
+    // Overrides extract().
+    // https://github.com/googleapis/gax-java/blob/8d45d186e36ae97b789a6f89d80ae5213a773b65/gax/src/main/java/com/google/api/gax/rpc/RequestParamsExtractor.java#L55
+    return LambdaExpr.builder()
+        .setArguments(requestVarExpr.toBuilder().setIsDecl(true).build())
+        .setBody(bodyStatements)
+        .setReturnExpr(returnExpr.build())
+        .build();
+  }
+
+  private void createRequestParamsExtractorBodyForHttpBindings(
+      Method method,
+      VariableExpr requestVarExpr,
+      List<Statement> bodyStatements,
+      MethodInvocationExpr.Builder returnExprBuilder) {
+    TypeNode paramsVarType =
+        TypeNode.withReference(
+            ConcreteReference.builder()
+                .setClazz(ImmutableMap.Builder.class)
+                .setGenerics(TypeNode.STRING.reference(), TypeNode.STRING.reference())
+                .build());
+    VariableExpr paramsVarExpr =
+        VariableExpr.withVariable(
+            Variable.builder().setName("params").setType(paramsVarType).build());
+
+    Expr paramsAssignExpr =
+        AssignmentExpr.builder()
+            .setVariableExpr(paramsVarExpr.toBuilder().setIsDecl(true).build())
+            .setValueExpr(
+                MethodInvocationExpr.builder()
+                    .setStaticReferenceType(FIXED_TYPESTORE.get("ImmutableMap"))
+                    .setMethodName("builder")
+                    .setReturnType(paramsVarType)
+                    .build())
+            .build();
+    bodyStatements.add(ExprStatement.withExpr(paramsAssignExpr));
+
+    for (HttpBindings.HttpBinding httpBindingFieldBinding :
+        method.httpBindings().pathParameters()) {
+      MethodInvocationExpr requestBuilderExpr =
+          createRequestFieldGetterExpr(requestVarExpr, httpBindingFieldBinding.name());
+      Expr valueOfExpr =
+          MethodInvocationExpr.builder()
+              .setStaticReferenceType(TypeNode.STRING)
+              .setMethodName("valueOf")
+              .setArguments(requestBuilderExpr)
+              .build();
+
+      Expr paramsPutExpr =
+          MethodInvocationExpr.builder()
+              .setExprReferenceExpr(paramsVarExpr)
+              .setMethodName("put")
+              .setArguments(
+                  ValueExpr.withValue(StringObjectValue.withValue(httpBindingFieldBinding.name())),
+                  valueOfExpr)
+              .build();
+      bodyStatements.add(ExprStatement.withExpr(paramsPutExpr));
+    }
+
+    returnExprBuilder.setExprReferenceExpr(paramsVarExpr).setMethodName("build");
+  }
+
+  private void createRequestParamsExtractorBodyForRoutingHeaders(
+      Method method,
+      VariableExpr requestVarExpr,
+      List<Statement> classStatements,
+      List<Statement> bodyStatements,
+      MethodInvocationExpr.Builder returnExprBuilder) {
+    TypeNode routingHeadersBuilderType =
+        TypeNode.withReference(
+            ConcreteReference.builder().setClazz(RequestParamsBuilder.class).build());
+    VariableExpr routingHeadersBuilderVarExpr =
+        VariableExpr.builder()
+            .setVariable(
+                Variable.builder().setName("builder").setType(routingHeadersBuilderType).build())
+            .setIsDecl(true)
+            .build();
+    MethodInvocationExpr routingHeaderBuilderInvokeExpr =
+        MethodInvocationExpr.builder()
+            .setStaticReferenceType(routingHeadersBuilderType)
+            .setMethodName("create")
+            .setReturnType(routingHeadersBuilderType)
+            .build();
+    Expr routingHeadersBuilderInitExpr =
+        AssignmentExpr.builder()
+            .setVariableExpr(routingHeadersBuilderVarExpr)
+            .setValueExpr(routingHeaderBuilderInvokeExpr)
+            .build();
+    bodyStatements.add(ExprStatement.withExpr(routingHeadersBuilderInitExpr));
+    List<RoutingHeaderRule.RoutingHeaderParam> routingHeaderParams =
+        method.routingHeaderRule().routingHeaderParams();
+    VariableExpr routingHeadersBuilderVarNonDeclExpr =
+        VariableExpr.builder()
+            .setVariable(
+                Variable.builder().setName("builder").setType(routingHeadersBuilderType).build())
+            .build();
+    for (int i = 0; i < routingHeaderParams.size(); i++) {
+      RoutingHeaderRule.RoutingHeaderParam routingHeaderParam = routingHeaderParams.get(i);
+      MethodInvocationExpr requestFieldGetterExpr =
+          createRequestFieldGetterExpr(requestVarExpr, routingHeaderParam.fieldName());
+      Expr routingHeaderKeyExpr =
+          ValueExpr.withValue(StringObjectValue.withValue(routingHeaderParam.key()));
+      String pathTemplateName =
+          String.format("%s_%s_PATH_TEMPLATE", JavaStyle.toUpperSnakeCase(method.name()), i);
+      TypeNode pathTemplateType =
+          TypeNode.withReference(ConcreteReference.withClazz(PathTemplate.class));
+      Variable pathTemplateVar =
+          Variable.builder().setType(pathTemplateType).setName(pathTemplateName).build();
+      Expr routingHeaderPatternExpr = VariableExpr.withVariable(pathTemplateVar);
+      Statement pathTemplateClassVar =
+          createPathTemplateClassStatement(routingHeaderParam, pathTemplateType, pathTemplateVar);
+      classStatements.add(pathTemplateClassVar);
+      MethodInvocationExpr addParamMethodExpr =
+          MethodInvocationExpr.builder()
+              .setExprReferenceExpr(routingHeadersBuilderVarNonDeclExpr)
+              .setMethodName("add")
+              .setArguments(requestFieldGetterExpr, routingHeaderKeyExpr, routingHeaderPatternExpr)
+              .build();
+
+      ExprStatement addParamStatement = ExprStatement.withExpr(addParamMethodExpr);
+      // No need to add null check if there is no nested fields
+      if (routingHeaderParam.getDescendantFieldNames().size() == 1) {
+        bodyStatements.add(addParamStatement);
+      } else {
+        IfStatement ifStatement =
+            IfStatement.builder()
+                .setConditionExpr(
+                    fieldValuesNotNullConditionExpr(
+                        requestVarExpr, routingHeaderParam.getDescendantFieldNames()))
+                .setBody(ImmutableList.of(addParamStatement))
+                .build();
+        bodyStatements.add(ifStatement);
+      }
+    }
+    returnExprBuilder
+        .setExprReferenceExpr(routingHeadersBuilderVarNonDeclExpr)
+        .setMethodName("build");
+  }
+
+  private Statement createPathTemplateClassStatement(
+      RoutingHeaderRule.RoutingHeaderParam routingHeaderParam,
+      TypeNode pathTemplateType,
+      Variable pathTemplateVar) {
+    VariableExpr pathTemplateVarExpr =
+        VariableExpr.builder()
+            .setVariable(pathTemplateVar)
+            .setIsDecl(true)
+            .setIsStatic(true)
+            .setIsFinal(true)
+            .setScope(ScopeNode.PRIVATE)
+            .build();
+    ValueExpr valueExpr =
+        ValueExpr.withValue(StringObjectValue.withValue(routingHeaderParam.pattern()));
+    Expr pathTemplateExpr =
+        AssignmentExpr.builder()
+            .setVariableExpr(pathTemplateVarExpr)
+            .setValueExpr(
+                MethodInvocationExpr.builder()
+                    .setStaticReferenceType(pathTemplateType)
+                    .setMethodName("create")
+                    .setArguments(valueExpr)
+                    .setReturnType(pathTemplateType)
+                    .build())
+            .build();
+    return ExprStatement.withExpr(pathTemplateExpr);
+  }
+
+  private Expr fieldValuesNotNullConditionExpr(
+      VariableExpr requestVarExpr, List<String> fieldNames) {
+    MethodInvocationExpr.Builder requestFieldGetterExprBuilder =
+        MethodInvocationExpr.builder().setExprReferenceExpr(requestVarExpr);
+    Expr fieldValuesNotNullExpr = null;
+    for (int i = 0; i < fieldNames.size() - 1; i++) {
+      String currFieldName = fieldNames.get(i);
+      String bindingFieldMethodName =
+          String.format("get%s", JavaStyle.toUpperCamelCase(currFieldName));
+      requestFieldGetterExprBuilder =
+          requestFieldGetterExprBuilder.setMethodName(bindingFieldMethodName);
+      // set return type of each method invocation to String just to pass the validation for
+      // RelationalOperationExpr that both side of relational operation needs to be a valid equality
+      // type
+      MethodInvocationExpr requestGetterExpr =
+          requestFieldGetterExprBuilder.setReturnType(TypeNode.STRING).build();
+      Expr currentValueNotNullExpr =
+          RelationalOperationExpr.notEqualToWithExprs(
+              requestGetterExpr, ValueExpr.createNullExpr());
+      if (fieldValuesNotNullExpr == null) {
+        fieldValuesNotNullExpr = currentValueNotNullExpr;
+      } else {
+        fieldValuesNotNullExpr =
+            LogicalOperationExpr.logicalAndWithExprs(
+                fieldValuesNotNullExpr, currentValueNotNullExpr);
+      }
+      requestFieldGetterExprBuilder =
+          MethodInvocationExpr.builder().setExprReferenceExpr(requestGetterExpr);
+    }
+    return fieldValuesNotNullExpr;
+  }
+
+  private MethodInvocationExpr createRequestFieldGetterExpr(
+      VariableExpr requestVarExpr, String fieldName) {
+    MethodInvocationExpr.Builder requestFieldGetterExprBuilder =
+        MethodInvocationExpr.builder().setExprReferenceExpr(requestVarExpr);
+    List<String> descendantFields = Splitter.on(".").splitToList(fieldName);
+    // Handle foo.bar cases by descending into the subfields.
+    // e.g. foo.bar -> request.getFoo().getBar()
+    for (int i = 0; i < descendantFields.size(); i++) {
+      String currFieldName = descendantFields.get(i);
+      String bindingFieldMethodName =
+          String.format("get%s", JavaStyle.toUpperCamelCase(currFieldName));
+      requestFieldGetterExprBuilder =
+          requestFieldGetterExprBuilder.setMethodName(bindingFieldMethodName);
+      if (i < descendantFields.size() - 1) {
+        requestFieldGetterExprBuilder =
+            MethodInvocationExpr.builder()
+                .setExprReferenceExpr(requestFieldGetterExprBuilder.build());
+      }
+    }
+    return requestFieldGetterExprBuilder.build();
   }
 }
--- a/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
+++ b/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
@@ -237,7 +237,7 @@ public class HttpJsonServiceStubClassComposer extends AbstractTransportServiceSt
             .setArguments(Arrays.asList(TYPE_REGISTRY_VAR_EXPR))
             .build();
 
-    if (method.shouldSetParamsExtractor()) {
+    if (method.hasRoutingHeaderParams()) {
       callSettingsBuilderExpr =
           MethodInvocationExpr.builder()
               .setExprReferenceExpr(callSettingsBuilderExpr)
--- a/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
+++ b/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
@@ -237,6 +237,15 @@ public class HttpJsonServiceStubClassComposer extends AbstractTransportServiceSt
             .setArguments(Arrays.asList(TYPE_REGISTRY_VAR_EXPR))
             .build();
 
+    if (method.shouldSetParamsExtractor()) {
+      callSettingsBuilderExpr =
+          MethodInvocationExpr.builder()
+              .setExprReferenceExpr(callSettingsBuilderExpr)
+              .setMethodName("setParamsExtractor")
+              .setArguments(createRequestParamsExtractorClassInstance(method, classStatements))
+              .build();
+    }
+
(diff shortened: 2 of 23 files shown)


gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
```diff
@@ -1168,4 +1178,254 @@ protected TypeNode getTransportOperationsStubType(Service service) {
...
+      MethodInvocationExpr.Builder returnExprBuilder) {
+    TypeNode routingHeadersBuilderType =
+        TypeNode.withReference(
+            ConcreteReference.builder().setClazz(RequestParamsBuilder.class).build());
+    VariableExpr routingHeadersBuilderVarExpr =
+        VariableExpr.builder()
+            .setVariable(
+                Variable.builder().setName("builder").setType(routingHeadersBuilderType).build())
+            .setIsDecl(true)
+            .build();
```
lqiu96, line 1228: Use RequestParamBuilder to percent-encode the values for http bindings. This use case is for services that need routing headers and backwards-compatibility.

gax-java/gax/src/main/java/com/google/api/gax/rpc/RequestParamsBuilder.java
```diff
@@ -56,19 +59,49 @@ public static RequestParamsBuilder create() {
...
-   * in such case.
+   * in such case. This method will percent encode both the header key and header value.
    *
    * @param fieldValue the field value from a request
    * @param headerKey the header key for the routing header param
    * @param pathTemplate {@link PathTemplate} the path template used for match-and-extract
    */
   public void add(String fieldValue, String headerKey, PathTemplate pathTemplate) {
-    if (fieldValue == null) {
+    if (fieldValue == null || fieldValue.isEmpty()) {
```
lqiu96, line 69: Adding this b/c it failed the showcase test. Checks that the value is non-null and the matched value is non-empty.

gax-java/gax/src/main/java/com/google/api/gax/rpc/RequestParamsBuilder.java
```diff
@@ -56,19 +59,49 @@ public static RequestParamsBuilder create() {
...
+
+  // Percent encode the value passed in
+  private String percentEncodeString(String value) {
+    try {
+      return URLEncoder.encode(value, StandardCharsets.UTF_8.toString());
+    } catch (UnsupportedEncodingException e) {
+      // Default to use the un-encoded value if there is an encoding issue
+      // TODO: Log this scenario once we implemented the Cloud SDK logging.
+      return value;
     }
```
lqiu96, line 105: Most, if not all, use cases (most header keys and header values) will not be affected. A lot of services require that the inputs match: `[a-zA-Z0-9]`.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonCallableFactory.java
```diff
@@ -60,8 +60,16 @@ private HttpJsonCallableFactory() {}
...
-        httpJsonCallSettings.getMethodDescriptor(), httpJsonCallSettings.getTypeRegistry());
+    UnaryCallable<RequestT, ResponseT> callable =
+        new HttpJsonDirectCallable<>(
+            httpJsonCallSettings.getMethodDescriptor(), httpJsonCallSettings.getTypeRegistry());
+
+    if (httpJsonCallSettings.getParamsExtractor() != null) {
+      callable =
+          new HttpJsonUnaryRequestParamCallable<>(
+              callable, httpJsonCallSettings.getParamsExtractor());
+    }
```
lqiu96, line 71: I believe this should always be true. Params Extractor is set for (google.api.http) annotation for backwards compatibility or via the (google.api.routing) annotation

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonCallableFactory.java
```diff
@@ -180,6 +188,12 @@ ServerStreamingCallable<RequestT, ResponseT> createServerStreamingCallable(
     ServerStreamingCallable<RequestT, ResponseT> callable =
         new HttpJsonDirectServerStreamingCallable<>(httpJsoncallSettings.getMethodDescriptor());
 
+    if (httpJsoncallSettings.getParamsExtractor() != null) {
+      callable =
+          new HttpJsonServerStreamingRequestParamCallable<>(
+              callable, httpJsoncallSettings.getParamsExtractor());
+    }
```
lqiu96, line 195: Same reason as above

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpRequestRunnable.java
```diff
@@ -196,6 +197,13 @@ HttpRequest createHttpRequest() throws IOException {
           httpRequest.getHeaders(), entry.getKey(), (String) entry.getValue());
     }
 
+    if (httpJsonCallOptions.getRequestParamsHeader() != null) {
+      HttpHeadersUtils.setHeader(
+          httpRequest.getHeaders(),
+          DYNAMIC_ROUTING_HEADER_KEY,
+          httpJsonCallOptions.getRequestParamsHeader());
+    }
```
lqiu96, line 205: Adds the header to the request

lqiu96:
Ruby manually encodes the header: https://github.com/googleapis/gapic-generator-ruby/blob/e4984a472bc989553dbaa620e564e4c5b5982a66/shared/output/gapic/templates/testing/lib/testing/routing_headers/service_explicit_headers/client.rb#L229
Go does the same: https://github.com/googleapis/google-cloud-go/blob/9018b8398889167e5d91a21f62cb271ed250d814/run/apiv2/services_client.go#L480-L485

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
```diff
@@ -237,6 +237,18 @@ protected Expr createTransportSettingsInitExpr(
             .setArguments(Arrays.asList(TYPE_REGISTRY_VAR_EXPR))
             .build();
 
+    // This generates the Request Params Extractors twice (once in the APIMethodDescriptor
+    // and once in the CallSettings). This is to ensure backwards compatibility and allow
+    // for implicit routing headers with just HttpBindings.
+    if (method.shouldSetParamsExtractor()) {
```
blakeli0, line 243: I think this should be already handled [here](https://github.com/googleapis/sdk-platform-java/blob/cc5724f04722e3d60eb9b4e02ecc1bdf5dd7a7fa/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/grpc/GrpcServiceStubClassComposer.java#L273-L281)?

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonCallContext.java
```diff
@@ -469,6 +469,15 @@ public HttpJsonCallContext withDeadline(Instant newDeadline) {
     return withCallOptions(builder.setDeadline(newDeadline).build());
   }
 
+  public HttpJsonCallContext withRequestParamsDynamicHeaderOption(String requestParams) {
```
blakeli0, line 472: We already have a `withExtraHeaders` method above which seems a better fit, this also makes us not have to add an extra field to `HttpJsonCallOptions`

gax-java/gax/src/main/java/com/google/api/gax/rpc/RequestParamsBuilder.java
```diff
@@ -56,19 +59,49 @@ public static RequestParamsBuilder create() {
...
+    if (headerKey == null || headerKey.isEmpty() || fieldValue == null || fieldValue.isEmpty()) {
+      return;
+    }
+    String encodedKey = percentEncodeString(headerKey);
+    String encodedValue = percentEncodeString(fieldValue);
+    paramsBuilder.put(encodedKey, encodedValue);
+  }
+
+  // Percent encode the value passed in
+  private String percentEncodeString(String value) {
```
blakeli0, line 98: [RequestUrlParamsEncoder](https://github.com/googleapis/sdk-platform-java/blob/cc5724f04722e3d60eb9b4e02ecc1bdf5dd7a7fa/gax-java/gax/src/main/java/com/google/api/gax/rpc/RequestUrlParamsEncoder.java#L79) might be doing the same thing already.

gax-java/gax/src/main/java/com/google/api/gax/rpc/RequestParamsBuilder.java
blakeli0: matchedValues should be empty if `fieldValue.isEmpty()`, which line is failing in showcase test without this change?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
lqiu96: Yeah, I believe that is for grpc. From the proposal, I think backwards compatibility (google.api.http) applies for both transports, so I've added it to the HttpJsonServiceStubClassComposer class as well.

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/GrpcLoggingStub.golden
```diff
@@ -173,9 +173,9 @@ public class GrpcLoggingServiceV2Stub extends LoggingServiceV2Stub {
             .setMethodDescriptor(deleteLogMethodDescriptor)
             .setParamsExtractor(
                 request -> {
-                  ImmutableMap.Builder<String, String> params = ImmutableMap.builder();
-                  params.put("log_name", String.valueOf(request.getLogName()));
-                  return params.build();
+                  RequestParamsBuilder builder = RequestParamsBuilder.create();
```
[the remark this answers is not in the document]
lqiu96: Yeah, url encoding is done in the RequestParamsBuilder class

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonCallContext.java
lqiu96: Ah I see. Let me try and use that.

gax-java/gax/src/main/java/com/google/api/gax/rpc/RequestParamsBuilder.java
lqiu96: This was my worry: https://github.com/googleapis/sdk-platform-java/blob/cc5724f04722e3d60eb9b4e02ecc1bdf5dd7a7fa/gax-java/gax/src/main/java/com/google/api/gax/rpc/RequestUrlParamsEncoder.java#L72-L73

I don't think it is encoding the values

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonCallableFactory.java
lqiu96: Ah no. This is wrong. I didn't realize that `Method.hasHttpBindings()` also checks that there is a pathParameter.

gax-java/gax/src/main/java/com/google/api/gax/rpc/RequestParamsBuilder.java
lqiu96: There are a few tests that fail (both transports). For example: `testGrpc_matchesOtherHeaderName`

This sets the Other_Header to be `instances/456` and would expect that the header to only be `baz=instances%2F456`.

The result of the header is: `header=&routing_id=&baz=instances%2F456` as it also matches these two routing header values: https://github.com/googleapis/gapic-showcase/blob/c41cdbd3a8ebf3c30ffa00dbeee1d96a6ae5fbcc/schema/google/showcase/v1beta1/echo.proto#L52-L58

From the routing.proto, I don't think sending an empty string is beneficial: https://github.com/googleapis/googleapis/blob/53a0be29c4a95a1d3b4c0d3a7a2ac8b52af2a3c0/google/api/routing.proto#L109-L110

gax-java/gax/src/main/java/com/google/api/gax/rpc/RequestParamsBuilder.java
blakeli0: I see, the name is really confusing, it is more like a `paramsConcatenater`. This [isValid](https://github.com/googleapis/sdk-platform-java/blob/cc5724f04722e3d60eb9b4e02ecc1bdf5dd7a7fa/gax-java/gax/src/main/java/com/google/api/gax/rpc/RequestUrlParamsEncoder.java#L110-L121) method seems to be doing some validation but `validateExtractedParameters` is always false.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
blakeli0: I think it was [migrated](https://github.com/googleapis/sdk-platform-java/blob/e7b2dbf24c4ecad5948340bc6078818167c059d5/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java#L1198-L1204) to the abstract composer as part of this PR? Any case that it couldn't handle?

gax-java/gax/src/main/java/com/google/api/gax/rpc/RequestParamsBuilder.java
blakeli0: I'm kind of surprised the result is `header=&routing_id=&baz=instances%2F456`. I agree with the expected behavior, but I would expect the current behavior should match the expected behavior because of this logic 
https://github.com/googleapis/sdk-platform-java/blob/e7b2dbf24c4ecad5948340bc6078818167c059d5/gax-java/gax/src/main/java/com/google/api/gax/rpc/RequestParamsBuilder.java#L73
`matchedValues` would be empty if `fieldValue` is empty and I just verified with unit tests, so theoretically no headers should be added to the builder.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
lqiu96: I believe it's still an abstract method inside the Abstract class: https://github.com/googleapis/sdk-platform-java/blob/7b6c9522fa0c16c5f98507bae9afff9a05bf7882/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java#L273

HttpJson also adds the `setParamsExtractor` only if `method.shouldSetParamsExtractor()`

gax-java/gax/src/main/java/com/google/api/gax/rpc/RequestParamsBuilder.java
lqiu96: It seems to be with some specific cases (i.e. not setting a header value when the pathtemplate is {header=**}). The matchedValues map has one entry (header -> "") and it'll match anything. I believe for most cases the logic will work properly.

blakeli0: I see, that makes sense and good to know. So a test case like this would fail without this change? 
https://github.com/googleapis/sdk-platform-java/blob/7b6c9522fa0c16c5f98507bae9afff9a05bf7882/gax-java/gax/src/test/java/com/google/api/gax/rpc/RequestParamsBuilderTest.java#L136-L140

lqiu96: Took another look at this case and I'm not sure it's the best test case for scenario above. I'll rename to test case to reflect what I think is causing this: protobuf's default value for strings. I think the test case should be: 

```
    Map<String, String> actual = getRoutingHeaders("{table_location=**}", "");
    assertThat(actual).isEmpty();
```
where protobuf puts `""` as the value when table_location routing param wasn't set in the request message.

gax-java/gax/src/main/java/com/google/api/gax/rpc/RequestParamsBuilder.java
```diff
@@ -56,19 +59,49 @@ public static RequestParamsBuilder create() {
...
+    try {
+      return URLEncoder.encode(value, StandardCharsets.UTF_8.toString());
```
blakeli0, line 100: Noticed that there are some [recommendations](https://stackoverflow.com/questions/29276432/what-character-set-should-be-used-for-url-encoding) against URLEncoder, can you take a look?

lqiu96: I see. I think we could use Guava's [PercentEncoder](https://github.com/google/guava/blob/master/guava/src/com/google/common/net/PercentEscaper.java). This should properly encode space characters and other chars (i.e. `+` and `/` and `.`)

lqiu96: On a second look, it seems that I'll need to percent-encode the reserved characters: https://en.wikipedia.org/wiki/URL_encoding

The PercentEscaper seems to encode every chars so I'll whitelist these tokens `.-~_` using:
`new PercentEscaper(".-~_" , false)` and I believe it should work.

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/GrpcLoggingStub.golden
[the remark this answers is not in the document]
blakeli0: Confirmed that it was not encoded.

gax-java/gax/src/main/java/com/google/api/gax/rpc/RequestParamsBuilder.java
blakeli0: > Unable to create tables with the unreserved characters and ~ and .

Is it because the backend does not allow these characters to be used for creating tables or the encoding? I guess the former but want to make sure.

lqiu96: Yep, the former. This is the bigquery restriction that I see: `Unicode letters, marks, numbers, connectors, dashes or spaces allowed.`

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java
lqiu96: > I know where my confusion was coming from now, I was expecting this logic to be moved to the parent abstract class as well. I checked the logic in grpc composer and they look identical, can we move this common logic to the parent class? 

I believe there is a small different with the `TypeRegistry`, but I'll see if I can work around it.

> Also what does the comment mean `This generates the Request Params Extractors twice (once in the APIMethodDescriptor and and once in the CallSettings)`, I thought we only set it in the `CallSettings`? Is there a golden file with changes in `APIMethodDescriptor`?

I believe this was part of the original implementation and I forgot to remove.

gax-java/gax/src/main/java/com/google/api/gax/rpc/RequestParamsBuilder.java
```diff
@@ -56,20 +59,44 @@ public static RequestParamsBuilder create() {
...
+
+  /**
+   * Add an entry to paramsBuilder with key-value pairing of (headerKey, fieldValue). The only
+   * validation done is to ensure the headerKey and fieldValue are not null and non-empty. This
+   * method will percent encode both the header key and header value.
+   *
+   * @param headerKey the header key for the routing header param
+   * @param fieldValue the field value from a request
+   */
+  public void add(String headerKey, String fieldValue) {
```
blakeli0, line 88: Now that we know we have to encode the headers, I think we should do it in a more generic way, like in `RequestUrlParamsEncoder` or even encode all `extraHeaders`.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonClientCalls.java
```diff
@@ -48,12 +50,19 @@ public static <RequestT, ResponseT> HttpJsonClientCall<RequestT, ResponseT> newC
       ApiMethodDescriptor<RequestT, ResponseT> methodDescriptor, ApiCallContext context) {
 
     HttpJsonCallContext httpJsonContext = HttpJsonCallContext.createDefault().nullToSelf(context);
+    HttpJsonCallOptions callOptions = httpJsonContext.getCallOptions();
+
+    Map<String, String> headerMap = callOptions.getRequestHeaderMap();
+    for (Map.Entry<String, List<String>> extraHeaderEntrySet :
+        httpJsonContext.getExtraHeaders().entrySet()) {
```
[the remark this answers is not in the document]
lqiu96: > We can change this behavior if we want but I would see if it there is a way to set the header to HttpJsonMetadata directly.

I was actually looking into this before and I wasn't able to think of a *clean* way to do this. I believe the headers are for the class are initially set empty: https://github.com/googleapis/sdk-platform-java/blob/ef4a7ff91b259f0b650cc06bd664eadca7bd17c9/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonClientCalls.java#L83

Ideally, I would be able to just populate the headers there, but I couldn't find a clean way to access the context.

ManagedHttpJsonChannel's newCall only takes in callOptions and I stored it there: https://github.com/googleapis/sdk-platform-java/blob/ef4a7ff91b259f0b650cc06bd664eadca7bd17c9/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ManagedHttpJsonChannel.java#L73-L83

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonClientCalls.java
```diff
@@ -48,12 +50,19 @@ public static <RequestT, ResponseT> HttpJsonClientCall<RequestT, ResponseT> newC
...
+      headerMap.put(extraHeaderEntrySet.getKey(), extraHeaderEntrySet.getValue().get(0));
```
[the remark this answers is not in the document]
lqiu96: I believe the Headers class (used when merging the headers in `withExtraHeaders`) has a check for non-null: https://github.com/googleapis/sdk-platform-java/blob/ef4a7ff91b259f0b650cc06bd664eadca7bd17c9/gax-java/gax/src/main/java/com/google/api/gax/rpc/internal/Headers.java#L50-L52

I can add a check so that there is one element inside.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonUnaryRequestParamCallable.java
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * A {@code UnaryCallable} that extracts values from the fields of the request and inserts them into
+ * headers.
+ *
+ * <p>Package-private for internal usage.
+ */
+class HttpJsonUnaryRequestParamCallable<RequestT, ResponseT>
+    extends UnaryCallable<RequestT, ResponseT> {
+  private static final String DYNAMIC_ROUTING_HEADER_KEY = "x-goog-request-params";
```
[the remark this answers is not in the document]
lqiu96: `CallOptionsUtil` exists inside the gax-grpc module and can't be accessed from gax-httpjson. It might make sense to move this to the Headers class inside the gax module and both transport modules can access.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonMetadata.java
```diff
@@ -51,7 +51,7 @@ public abstract class HttpJsonMetadata {
   public abstract Builder toBuilder();
 
   public static HttpJsonMetadata.Builder newBuilder() {
-    return new AutoValue_HttpJsonMetadata.Builder().setHeaders(Collections.emptyMap());
+    return new AutoValue_HttpJsonMetadata.Builder().setHeaders(new HashMap<>());
```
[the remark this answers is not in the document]
lqiu96: I don't think this is needed. Removing.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonUnaryRequestParamCallable.java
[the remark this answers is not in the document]
lqiu96: `CallOptionsUtil` exists inside the gax-grpc module and can't be accessed from gax-httpjson. It might make sense to move this to the Headers class inside the gax module and both transport modules can access.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITDynamicRoutingHeaders.java
```diff
@@ -0,0 +1,308 @@
...
+import io.grpc.Metadata;
+import io.grpc.MethodDescriptor;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+public class ITDynamicRoutingHeaders {
```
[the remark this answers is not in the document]
lqiu96: I'll look to add explicit cases for multiple matches. There is some differences between the echo rpc and the routing proto test cases, but I'll try to add as much as I can.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonUnaryRequestParamCallable.java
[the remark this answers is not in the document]
blakeli0: Yep, extract and move to a util class in gax sounds good to me.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonUnaryRequestParamCallable.java
```diff
@@ -0,0 +1,74 @@
...
+    this.callable = Preconditions.checkNotNull(callable);
+    this.paramsEncoder =
+        new RequestUrlParamsEncoder<>(Preconditions.checkNotNull(paramsExtractor), false);
+  }
+
+  @Override
+  public ApiFuture<ResponseT> futureCall(RequestT request, ApiCallContext context) {
+    ApiCallContext newCallContext = context;
+    String encodedHeader = paramsEncoder.encode(request);
+    if (encodedHeader != null && !encodedHeader.isEmpty()) {
```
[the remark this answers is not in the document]
blakeli0: Can you confirm with Slava? You are probably right and I agree that we should not send header if it's empty, but want to make sure it would not break existing behavior(in case some server are expecting empty header).

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonClientCalls.java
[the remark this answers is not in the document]
blakeli0: In addition, I'm not sure why the value is of type `List<String>`, my guess is that we did it on grpc side and just replicated the same pattern, and the only reason we did it on grpc side is that it's what `grpc-java` needed, maybe they concatenate the values like how we did it in `RequestUrlParamsEncoder`.
Maybe not the scope of this PR, but if we want to start support this `extraHeaders`, it's better to support it properly, either just always use the first element, change the value to `String` or concatenate the values.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITDynamicRoutingHeaders.java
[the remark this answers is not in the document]
blakeli0: Sounds good thanks!

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonClientCalls.java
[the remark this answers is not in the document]
lqiu96: Actually, I think I have a bit of leeway here. I see that HttpJsonClientCalls `futureUnaryCall` is package-private: https://github.com/googleapis/sdk-platform-java/blob/ef4a7ff91b259f0b650cc06bd664eadca7bd17c9/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonClientCalls.java#L79-L80

I'm thinking about just adding HttpJsonCallContext as a third param so I can have access to the headers without having to add it to the call options. I can populate the HttpJsonMetadata with the extraHeaders in the callcontext.

gax-java/gax/src/main/java/com/google/api/gax/rpc/RequestParamsBuilder.java
blakeli0: >Do we need to encode all headers?

That's a great question. I think we should but I'm not sure. There are some security risks of not encoding urls, so I was under the impression that headers should be encoded and are already encoded as well by the downstream grpc or http libraries. Probably not part of this PR but we can do a little more investigation on the headers encoding in general. As part of this PR, I think we should move this logic to `RequestUrlParamsEncoder` because that class is specific for the routing headers and the name is perfect for this logic.

blakeli0: >Do we need to encode all headers?

That's a great question. I think we should but I'm not sure. There are some security risks of not encoding urls, so I was under the impression that headers should be encoded and are already encoded as well by the downstream grpc or http libraries. Probably not part of this PR but we can do a little more investigation on the headers encoding in general. As part of this PR, I think we should move this logic to `RequestUrlParamsEncoder` because that class is specific for the routing headers and the name is perfect for this logic.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonUnaryRequestParamCallable.java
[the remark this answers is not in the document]
lqiu96: Yep will do.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonClientCalls.java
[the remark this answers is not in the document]
lqiu96: > it's better to support it properly, either just always use the first element, change the value to String or concatenate the values.

Yeah, agreed. I don't fully understand the `List<String>` pattern for headers either. I wanted to update to just a (K,V) of <String, String> but that is a breaking change on a `@BetaApi`. I'll do some more digging into this and see if I can find a reason for this. Depending on the context, do you think this would be something that we *could* consider modifying?

gax-java/gax/src/main/java/com/google/api/gax/rpc/RequestParamsBuilder.java
lqiu96: > Probably not part of this PR but we can do a little more investigation on the headers encoding in general.

Sure! I will create an issue in the backlog to investigate this. 

> As part of this PR, I think we should move this logic to RequestUrlParamsEncoder because that class is specific for the routing headers and the name is perfect for this logic.

Yep, I agree. The only concern I have is that changing `RequestUrlParamsEncoder` to now encode the header values might be a behavior breaking change: https://github.com/googleapis/sdk-platform-java/blob/ef4a7ff91b259f0b650cc06bd664eadca7bd17c9/gax-java/gax/src/main/java/com/google/api/gax/rpc/RequestUrlParamsEncoder.java#L68-L79 as it expects the params to the percent encoded already.

blakeli0: This class is only used for encoding the request params(which is effective routing headers), and it is marked as `@InternalAPI` so we should be able to change it at anytime.

lqiu96: Sounds good. I'll migrate the encoding logic here.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITDynamicRoutingHeaders.java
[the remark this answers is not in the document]
blakeli0: If there is any test case that we think is missing from the showcase protos, we should definitely add them to the protos in `gapic-showcase`. It could be a separate PR also if the effort is large.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonClientCalls.java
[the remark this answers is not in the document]
blakeli0: This could be a follow up to this PR, I'm OK with always getting the first element for now in this PR. We can have a separate issue just for supporting `extraHeaders`.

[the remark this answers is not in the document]
lqiu96: Doing a little digging and I believe this is probably the intended behavior for grpc (https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md#constructing-metadata) - One header key can support multiple values. Seems like we copied the behavior for httpjson and that's why we have a `Map<String, List<String>>`.

I can create a separate issue for looking into `extraHeaders` and seeing if copying grpc's behavior is needed and/or how we can better support it if it is.

gax-java/gax/src/main/java/com/google/api/gax/rpc/RequestParamsBuilder.java
lqiu96: @blakeli0 I took out the logic for doing validation inside this class. It's an `@InternalApi` and only used by the {Transport}{Request_Type}RequestParamCallable.

Let me know you're fine with that. Otherwise I can revert that part of the change.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonUnaryRequestParamCallable.java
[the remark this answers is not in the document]
lqiu96: Conversation with Slava:
```
the behaviour should be to NOT send the x-goog-request-params at all
server-side will likely do the right thing even if you send an empty value
```

gax-java/gax/src/main/java/com/google/api/gax/rpc/RequestParamsBuilder.java
blakeli0: SGTM.

blakeli0 (approved):

gax-java/gax/src/main/java/com/google/api/gax/rpc/RequestParamsBuilder.java
meltsufin: @blakeli0 @lqiu96 Did this effectively change the behavior of implicit routing headers from not encoding params to now URL encoding them?

blakeli0: Yes, it did. There was another [discussion](https://github.com/googleapis/sdk-platform-java/pull/1667#discussion_r1199286063) regarding should we add it or not for existing routing headers, we discussed offline and confirmed with Slava that all routing header should be encoded.

lqiu96: Yep, there was a discussion with Slava about the need to encode both the header key and value. We saw there was customer issue (b/283866374) regarding having an unencoded value.

PR #1669, opened 2023-05-09
chore: Shutdown and awaitTermination for showcase clients

lqiu96:
Seeing error messages regarding orphaned channels and not cleaning up the queues for gRPC.
Error:
```
io.grpc.internal.ManagedChannelOrphanWrapper$ManagedChannelReference cleanQueue
SEVERE: *~*~*~ Previous channel ManagedChannelImpl{logId=23, target=localhost:7469} was not shutdown properly!!! ~*~*~*
    Make sure to call shutdown()/shutdownNow() and wait until awaitTermination() returns true.
```

---
Few changes in this PR:
1. Create a single client to be used for each file (no longer for each test case).
2. Run the tests sequentially to not overwhelm the showcase server
3. await for the client to terminate (wait 10 seconds)

Fixes: #1680


lqiu96:
`httpjsonClient.awaitTermination()` does nothing as of now.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITCommonServiceMixins.java
```diff
@@ -63,6 +65,16 @@ public void createClients() throws Exception {
     httpjsonClient = TestClientInitializer.createHttpJsonEchoClient();
   }
 
+  @After
+  public void destroyClients() throws InterruptedException {
+    grpcClient.close();
+    grpcClient.awaitTermination(TestClientInitializer.AWAIT_TERMINATION_SECONDS, TimeUnit.SECONDS);
+
+    httpjsonClient.close();
+    httpjsonClient.awaitTermination(
```
blakeli0, line 74: This method is not implemented yet https://github.com/googleapis/gapic-generator-java/issues/1663

blakeli0 (commented):
Do we know why the deadline PR caused flakyness? The deadline PR should only affect http/json stuff in theory.

lqiu96:
> Do we know why the deadline PR caused flakyness? The deadline PR should only affect http/json stuff in theory.

I believe it's just addition of a bunch of tests added for the deadline PR. These tests run in parallel and the showcase server was never built for performance or for handling being spammed with retries. I added a bit of wait for the tests and have the tests sequentially to try and help with the CI jobs.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ManagedHttpJsonChannel.java
```diff
@@ -47,23 +47,24 @@
 @BetaApi
 public class ManagedHttpJsonChannel implements HttpJsonChannel, BackgroundResource {
 
-  private static final ExecutorService DEFAULT_EXECUTOR =
-      InstantiatingExecutorProvider.newBuilder().build().getExecutor();
```
lqiu96, line 51: Static `DEFAULT_EXECUTOR` error'd out on the unit tests. Same executor was being used and it was shutdown after the first test.

showcase/pom.xml
```diff
@@ -88,10 +88,6 @@
           <plugin>
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-failsafe-plugin</artifactId>
-            <configuration>
-              <forkCount>1C</forkCount>
-              <reuseForks>true</reuseForks>
-            </configuration>
```
lqiu96, line 94: Run the tests sequentially for now to see if it helps with the CI jobs.

lqiu96:
I'll try to break this PR into two PRs (one for `awaitTermination()` implementation and another into showcase fixes). I tried to group them together since the new PR might fail the due to the flaky CI jobs.

lqiu96:
Separate PR created here: https://github.com/googleapis/sdk-platform-java/pull/1677

showcase/pom.xml
blakeli0: Can we revert this change with the fix?

blakeli0:
> Few changes in this PR:
> 1. Create a single client to be used for each file (no longer for each test case).
> 2. Run the tests sequentially to not overwhelm the showcase server
> 3. await for the client to terminate (wait 10 seconds)

1 and 3 sounds good to me, with both 1 and 3 in, I think we may not need 2 anymore.

lqiu96:
> > Few changes in this PR:
> > 
> > 1. Create a single client to be used for each file (no longer for each test case).
> > 2. Run the tests sequentially to not overwhelm the showcase server
> > 3. await for the client to terminate (wait 10 seconds)
> 
> 1 and 3 sounds good to me, with both 1 and 3 in, I think we may not need 2 anymore.

I kept 2 in to help the server with any flakes (especially for the deadline tests that spam the server). The overall runtime increases from 45 seconds to ~2 minutes which I don't think is too bad. @blakeli0 Would you be fine with keeping this in for a few weeks to try and observe some CI stability. If this helps mitigate the issues, I can change it to be parallel again?

blakeli0:
> > > Few changes in this PR:
> > > 
> > > 1. Create a single client to be used for each file (no longer for each test case).
> > > 2. Run the tests sequentially to not overwhelm the showcase server
> > > 3. await for the client to terminate (wait 10 seconds)
> > 
> > 
> > 1 and 3 sounds good to me, with both 1 and 3 in, I think we may not need 2 anymore.
> 
> I kept 2 in to help the server with any flakes (especially for the deadline tests that spam the server). The overall runtime increases from 45 seconds to ~2 minutes which I don't think is too bad. @blakeli0 Would you be fine with keeping this in for a few weeks to try and observe some CI stability. If this helps mitigate the issues, I can change it to be parallel again?

SGTM, with more showcase tests being added, we should revisit this soon though. CC: @burkedavison in case there is any concerns.

blakeli0 (approved):
LGTM if we can get a green light from Burke on changing the test to sequential.

burkedavison (approved):

lqiu96:
> SGTM. Could we create a "reminder" issue to revisit the parallel / sequential decision in a few weeks?

Yep. I will revisit this July 6th.

PR #1670, opened 2023-05-09
chore: update protobuf dependency to match workspace

alicejli:
Updating dependencies.properties protobuf version to match workspace version. Looks like renovate-bot doesn't update these automatically.

parent-pom is updated to match workspace: https://github.com/googleapis/gapic-generator-java/blob/main/gapic-generator-java-pom-parent/pom.xml#LL33C24-L33C24


ddixit14:
Adding reference: https://github.com/googleapis/googleapis/blob/master/WORKSPACE#L162

PR #1673, opened 2023-05-10
deps: Update showcase version to 0.28.1

blakeli0:


lqiu96 (approved):

WORKSPACE
```diff
@@ -107,17 +107,13 @@ http_archive(
 )
 
 # Showcase
-_showcase_commit = "656e5f46d125a69c82c0cb7edcfcd8b03ed77b89"
-
-_showcase_sha256 = "26d4b71ac31cbca5e4ed4cdcb5bfeca185e405392d70d2020b4528b5b47c8022"
+_showcase_version = "0.28.1"
```
blakeli0, line 110: @burkedavison @mpeddada1 I'm changing this to use showcase version instead of committish as I think this is more intuitive, let me know if you have any concerns.

Issue #1663, opened 2023-05-01
Implement `awaitTermination()` for HttpJson's Channel


lqiu96:
Inside ManagedHttpJsonChannel's `awaitTermination()` method,
https://github.com/googleapis/gapic-generator-java/blob/4054bc668daf797a8fce798843e0ec4cda669faa/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ManagedHttpJsonChannel.java#L105-L108

there is a `TODO` that simply returns false. Implement this functionality.

PR #1677, opened 2023-05-12
feat: Implement awaitTermination() for MangedHttpJsonChannel

lqiu96:

Fixes: https://github.com/googleapis/sdk-platform-java/issues/1663


gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ManagedHttpJsonChannel.java
```diff
@@ -47,23 +47,24 @@
 @BetaApi
 public class ManagedHttpJsonChannel implements HttpJsonChannel, BackgroundResource {
 
-  private static final ExecutorService DEFAULT_EXECUTOR =
-      InstantiatingExecutorProvider.newBuilder().build().getExecutor();
-
```
lqiu96, line 52: Static DEFAULT_EXECUTOR error'd out on the unit tests. Same executor was being used and it was shutdown after the first unit test.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ManagedHttpJsonChannel.java
```diff
@@ -113,15 +130,31 @@ public void shutdownNow() {
...
   }
 
   @Override
-  public void close() {}
+  public void close() {
+    shutdown();
+  }
 
   public static Builder newBuilder() {
-    return new Builder().setExecutor(DEFAULT_EXECUTOR);
```
lqiu96, line 124: Logic which executor to use is determined in `build()` for the Builder. The behavior is the same and should not be breaking.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ManagedHttpJsonChannel.java
```diff
@@ -150,8 +183,16 @@ public Builder setHttpTransport(HttpTransport httpTransport) {
     public ManagedHttpJsonChannel build() {
       Preconditions.checkNotNull(endpoint);
 
+      boolean usingDefaultExecutor = executor == null;
```
blakeli0, line 186: I'm not sure this is where we should manage the executor, I know it's existing code that creates a default executor within the channel, but maybe we should create it outside and pass it in. We may already have logics to handle the lifecycle of the default executor, e.g. code in [ClientContext](https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java#L231-L233), can you investigate more on how we does it for gRPC channel?

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ManagedHttpJsonChannel.java
```diff
@@ -98,12 +105,22 @@ public synchronized void shutdown() {
 
   @Override
   public boolean isShutdown() {
-    return isTransportShutdown;
+    boolean isShutdown = deadlineScheduledExecutorService.isShutdown();
+    // Check that the Gax's ExecutorService is shutdown as well
+    if (usingDefaultExecutor) {
+      isShutdown = isShutdown && ((ExecutorService) executor).isShutdown();
+    }
+    return isShutdown;
```
blakeli0, line 113: I think we still need to take `isTransportShutdown` into consideration because `httpTransport.shutdown()` could fail even if `deadlineScheduledExecutorService.shutdown()` succeeded.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ManagedHttpJsonChannel.java
lqiu96: I'm not sure that the default executor (which I believe is used for background resources) is the same one for the channel. 

This comment inside `InstantiatingHttpJsonChannelProvider`: https://github.com/googleapis/sdk-platform-java/blob/bc8a4ad864f4d6f2b9d98e5fe296320fea19e74b/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/InstantiatingHttpJsonChannelProvider.java#L247-L258 would create a new Executor when the channel is created.

The `withExecutor` call only gets invoked when this if block is true: https://github.com/googleapis/sdk-platform-java/blob/bc8a4ad864f4d6f2b9d98e5fe296320fea19e74b/gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java#L179-L184

And that is true when the `StubSettings` is using the `deprecatedExecutorProviderSet`: https://github.com/googleapis/sdk-platform-java/blob/bc8a4ad864f4d6f2b9d98e5fe296320fea19e74b/gax-java/gax/src/main/java/com/google/api/gax/rpc/StubSettings.java#L108-L112

It seems for backwards-compatibility, it is still allowed: https://github.com/googleapis/sdk-platform-java/blob/bc8a4ad864f4d6f2b9d98e5fe296320fea19e74b/gax-java/gax/src/main/java/com/google/api/gax/rpc/StubSettings.java#L304-L326

I believe this change would have the channel create the a new executor and then shut it down when the channel is closing. If they provide their own Executor, then the customer would have to manage it since `usingDefaultExecutor` would not be true.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ManagedHttpJsonChannel.java
lqiu96: Ah, makes sense. I'll add that logic in again. Thanks!

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ManagedHttpJsonChannel.java
blakeli0: Thanks Lawrence! If we want to support the feature of a `default executor` in `ManagedHttpJsonChannel`, then your implementation makes perfect sense. 
However, I think we may not have to support it. `ManagedHttpJsonChannel` is basically trying to mimic [ManagedChannelImpl](https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java), where it makes sense to support a default executor because there are customer trying to create a gRPC channel from scratch. While for this `ManagedHttpJsonChannel`, correct me if I'm wrong, I don't think we want customers to create a channel by themselves, we are always creating a new http channel from `InstantiatingHttpJsonChannelProvider` and pass it from [here](https://github.com/googleapis/sdk-platform-java/blob/cc5724f04722e3d60eb9b4e02ecc1bdf5dd7a7fa/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/InstantiatingHttpJsonChannelProvider.java#L192) . Technically customers can create their own channel provider and extend `ManagedHttpJsonChannel`, but even that we don't have to support a default executor.
I think the Bigtable team added them around two years ago, can you please check with them and see if there is any special use case for it? If not, my preference is that we can just remove it and add doc saying that executors are expected to be passed from outside.

lqiu96: I am probably missing context/ understanding about the original implementation. Why would customers be creating a gRPC channel from scratch? If they use our client libraries with default configs (no executors passed in), gax would create a channel with an executor for them.

From my understanding (and I could be wrong on this): By default, we don't require a customer to pass in an external executor provider. If a customer *really* wants to use their own executors, they can pass it in via StubSettings and we'll use it.  I believe that is what this line is for: https://github.com/googleapis/sdk-platform-java/blob/cc5724f04722e3d60eb9b4e02ecc1bdf5dd7a7fa/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/InstantiatingHttpJsonChannelProvider.java#L192 since `newBuilder()` by default uses `DEFAULT_EXECUTOR`: https://github.com/googleapis/sdk-platform-java/blob/cc5724f04722e3d60eb9b4e02ecc1bdf5dd7a7fa/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ManagedHttpJsonChannel.java#L123-L125

Grpc-java have their own executors and its lifecycle is managed there. But, httpjson is implemented by us and we should be managing it (i.e. managing the `DEFAULT_EXECUTOR`). 

> I think the Bigtable team added them around two years ago, can you please check with them and see if there is any special use case for it? 

CC'ing @mutianf. Could I get your review/ feedback for this PR? I saw you originally made the changes related to this code here: https://github.com/googleapis/gax-java/pull/1355. I think I am missing some context behind these changes and was hoping you could help clear this up for us. Thanks!

mutianf: A bit background on the change:
Gax used to override grpc default executor with it's own default `InstantiatingExecutorProvider.newBuilder().build()`, which creates a pool of only 4 executors. This significantly affected the performance (of Bigtable), because all the background tasks (batching, retrying, watchdog) and grpc async calls are all scheduled on these 4 executors. Grpc's default executor doesn't limit the thread and is a lot more efficient.

The core change in my PR is that we deprecated needExecutors so an executor is not required for grpcChannels https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L367. When grpc creates a ManagedChannel, if executor is null, it uses the default grpcChannel https://github.com/grpc/grpc-java/blob/v1.37.x/core/src/main/java/io/grpc/internal/ManagedChannelImplBuilder.java#L304-L311.

Deprecating `needsExecutor` implies that we expect all channels, grpc or http, to have a default executor. And I decided to use  gax `InstantiatingExecutorProvider` as the default executor for http channels. I also made the decision that executors for background tasks should be different from executors for async calls. However, I don't have a preference on if customers should pass in their own executor for http channels, or we create and manage the executors for them.

lqiu96: Thanks for the context! 

> When grpc creates a ManagedChannel, if executor is null, it uses the default grpcChannel

I believe this is the default behavior now. Creating a GRPC client with `{Service}Client client = {Service}Client.create()` would have the executor as null (and same with a HttpJson client).

> Deprecating needsExecutor implies that we expect all channels, grpc or http, to have a default executor. And I decided to use gax InstantiatingExecutorProvider as the default executor for http channels. I also made the decision that executors for background tasks should be different from executors for async calls because background task should not block async work in any case.

Make sense from what I've seen. The `backgroundExecutorProvider` is also created via `InstantiatingExecutorProvider.newBuilder().build()` and that is managed by the BackgroundResources (which I believe is separate from the default executors created in httpjson's channel).

I believe this change should still allow customers to pass in their own executors and the passed in executor is still respected by both grpc and http channels. The default behavior is that if they don't, we will create a new executor (one executor per channel) and have the executor be managed that channel. 

I think this behavior should make sense:
- Background executor is managed by BackgroundResource
- Channel executor is managed by the channel

WDYT @blakeli0

blakeli0: Thanks @mutianf! It's good to know that it is an intentional change to not use the default executor created by `InstantiatingExecutorProvider` for gRPC channels, and `InstantiatingExecutorProvider` is only used by background resources now, however we probably need more docs to clarify this behavior.
> Deprecating needsExecutor implies that we expect all channels, grpc or http, to have a default executor.

I agree and I think we need to document this as well.
@lqiu96 Do you mind working with Matti to document the existing behavior in Javadocs/comments/internal docs? It could be a separate PR or part of this PR.

In addition, why do we still use `InstantiatingExecutorProvider` to create default executors for http channels if we already observed the performance issue for gRPC? Is there known performance issue for http? Or maybe there are not many people using http so we haven't noticed yet?

lqiu96: Sure, I can work on documenting this behavior.

> In addition, why do we still use InstantiatingExecutorProvider to create default executors for http channels if we already observed the performance issue for gRPC?

I don't think that `InstantiatingExecutorProvider` used solely in a httpjson channel has any performance issues. I think there was performance issue for gRPC when it's used for all the async calls + retries + other background activities. The original PR split the executors so that the channel executor and background executors are separate.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ManagedHttpJsonChannel.java
```diff
@@ -150,8 +183,20 @@ public Builder setHttpTransport(HttpTransport httpTransport) {
...
       Preconditions.checkNotNull(endpoint);
 
+      // If the executor provided for this channel is null, gax will provide a
+      // default executor to used for the calls. Only the default executor's
+      // lifecycle will be managed by the channel. Any external executor needs to
+      // managed by the user.
+      boolean usingDefaultExecutor = executor == null;
+      if (usingDefaultExecutor) {
+        executor = InstantiatingExecutorProvider.newBuilder().build().getExecutor();
+      }
```
blakeli0, line 193: The following is more readable to me, we need to add `usingDefaultExecutor` as a field to the builder as well, which is also a best practice for using builders.
```suggestion
      if (executor == null) {
        executor = InstantiatingExecutorProvider.newBuilder().build().getExecutor();
        usingDefaultExecutor = false;
      }
```

blakeli0 (commented):
Thanks Lawrence, the general approach makes sense if we can confirm that using a default executor from `InstantiatingHttpJsonChannelProvider` is fine. Also can you please add some tests to it? I know we didn't have any unit tests for it previously and it would be tricky to test executors.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ManagedHttpJsonChannel.java
```diff
@@ -84,10 +85,16 @@ public <RequestT, ResponseT> HttpJsonClientCall<RequestT, ResponseT> newCall(
...
   public synchronized void shutdown() {
+    // Calling shutdown() twice should no-op
     if (isTransportShutdown) {
       return;
     }
     try {
+      // Only shutdown the executor if it was created by Gax. External executors
+      // should be managed by the user.
+      if (usingDefaultExecutor) {
+        ((ExecutorService) executor).shutdown();
```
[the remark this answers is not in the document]
lqiu96: The defaultExecutor's type *currently* is always set to be an ExecutorService since it's created via `InstantiatingExecutorProvider`: https://github.com/googleapis/sdk-platform-java/blob/cc5724f04722e3d60eb9b4e02ecc1bdf5dd7a7fa/gax-java/gax/src/main/java/com/google/api/gax/core/InstantiatingExecutorProvider.java#LL62C3-L62C3

I can add a quick check with `instanceOf` to ensure that it won't error out. I think changing the type from Executor to ExecutorService impacts a bunch of core classes (TransportChannelProvider, InstantiatingGrpcChannelProvider, InstantiatingHttpJsonChannelProvider)

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ManagedHttpJsonChannel.java
```diff
@@ -98,12 +105,22 @@ public synchronized void shutdown() {
...
-    return isTransportShutdown;
+    boolean isTerminated = deadlineScheduledExecutorService.isTerminated();
+    // Check that the Gax's ExecutorService is terminated as well
+    if (usingDefaultExecutor) {
+      isTerminated = isTerminated && ((ExecutorService) executor).isTerminated();
+    }
+    return isTerminated;
   }
 
   @Override
```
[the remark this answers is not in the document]
lqiu96: Ah right, will update.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ManagedHttpJsonChannel.java
```diff
@@ -84,10 +85,16 @@ public <RequestT, ResponseT> HttpJsonClientCall<RequestT, ResponseT> newCall(
...
     }
     try {
+      // Only shutdown the executor if it was created by Gax. External executors
+      // should be managed by the user.
+      if (usingDefaultExecutor) {
+        ((ExecutorService) executor).shutdown();
+      }
       deadlineScheduledExecutorService.shutdown();
       httpTransport.shutdown();
       isTransportShutdown = true;
```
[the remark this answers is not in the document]
lqiu96: Sure, I'll create an issue to add try and add that functionality into the http-client library.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ManagedHttpJsonChannel.java
lqiu96: Sure, will update! I had a bit of hesitation with adding a setter for `usingDefaultExecutor` since I was afraid people would do some combination of setting:
- `usingDefaultExecutor` = false && `executor` = null
- `usingDefaultExecutor` = true && `executor` = non-null value
which are kinda invalid states.

Then `build()` would override the `usingDefaultExecutor` value anyways. Probably not best practice for builder, but I think `usingDefaultExecutor` should be internal anyways.

blakeli0: We can add `usingDefaultExecutor` as a private field but not have to expose it through a setter.

lqiu96: Ah, I misunderstood. I can add a field for it.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ManagedHttpJsonChannel.java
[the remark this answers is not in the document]
lqiu96: Issue created here: https://github.com/googleapis/google-http-java-client/issues/1850. Need to do some investigating on how to implement it properly.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ManagedHttpJsonChannel.java
lqiu96: I believe both gRPC and HttpJson use `InstantiatingExecutorProvider ` for the background tasks: https://github.com/googleapis/sdk-platform-java/blob/745ea58281c59b78f3fb490852718712eb081763/gax-java/gax/src/main/java/com/google/api/gax/rpc/StubSettings.java#L258

Unless I misunderstood the comment above, isn't the issue that gRPC experienced due the fact that they were sharing this background executor for both async calls and background tasks? They've opted to split the async calls to the executors provided by grpc-java. Httpjson also splits the async calls to a new instance of `InstantiatingExecutorProvider`. Neither transport would have the async calls and background tasks share the same executor since they're created with new ExecutorService instances: https://github.com/googleapis/sdk-platform-java/blob/745ea58281c59b78f3fb490852718712eb081763/gax-java/gax/src/main/java/com/google/api/gax/core/InstantiatingExecutorProvider.java#L61-L64

lqiu96: After a second look, I believe it creates a pool with a *minimum* (not maximum) of 4 cores: https://github.com/googleapis/sdk-platform-java/blob/745ea58281c59b78f3fb490852718712eb081763/gax-java/gax/src/main/java/com/google/api/gax/core/InstantiatingExecutorProvider.java#L80-L81

I wonder if it makes sense to try and limit this pool for the background executor (say like 4 cores) and have an unbounded pool for the async executor.

mutianf: > why do we still use InstantiatingExecutorProvider to create default executors for http channels if we already observed the performance issue for gRPC? Is there known performance issue for http?

I didn't want to touch the http implementations and also I didn't know if any customers experienced performance issue with http executors so I decided to stick with what we had. "it's just we haven't got enough adoption for http client yet." could be a reason why we haven't heard any complaints.

> I wonder if it makes sense to try and limit this pool for the background executor (say like 4 cores) and have an unbounded pool for the async executor.

I think this sounds good.

grpc uses cachedThreadPool https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/internal/GrpcUtil.java#L555 which doesn't limit the number of threads. So if we want unbounded async executor we can follow what grpc did. We should also give customers flexibility to override the default executor (for example to use a ForkJoinPool) if they want better performance and is suitable for their application.

blakeli0: > After a second look, I believe it creates a pool with a minimum (not maximum) of 4 cores:

Yeah this is exactly my concern, the fact that we were sharing executors between calls and background tasks is one aspect, but even with the `InstantiatingExecutorProvider` alone I think we may have performance issues. The current max number of threads is the same of number of CPUs, which is a good practice for non-I/O bound tasks, but if the thread is also used for making calls, then it is likely too small. This [article](https://engineering.zalando.com/posts/2019/04/how-to-set-an-ideal-thread-pool-size.html) has a good rule of thumb for I/O bound tasks. Making the thread pool unlimited also has disadvantages, because creating threads has overhead, the max number of threads a CPU can handle is the number of cores, so if we have more threads live than the cores, CPU switching between threads also has overhead and could degrade the performance. I think we may need more investigation to determine the right number of threads for making async calls, and in the mean time, use `InstantiatingExecutorProvider` for only the background tasks seems the right approach to me.

 > grpc uses cachedThreadPool https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/internal/GrpcUtil.java#L555 which doesn't limit the number of threads

@mutianf Maybe I'm wrong and admitted that I'm not very familiar with gRPC code, but this does not seem to be the thread pool used by making calls, because they are all set as daemon thread which does not prevent the JVM from exiting when the program finishes but the thread is still running.

lqiu96: I see. I believe we still need to provide an executor for customers using httpjson (as now requiring customers to provide an executor may be a breaking change). Using the current `InstantiatingExecutorProvider` implementation might have some performance impact. For now, would it make sense to create an unbounded executor for async calls and have the background executor be limited to the number of CPUs? Or should I do something different to create the executors?

lqiu96: > Yes. In terms providing the feature for customers to pass their own executors for making async calls, I don't think we have that for gRPC, so we don't want to add it for Http yet, it could be a good FR in the backlog. I think the current behavior of a custom executor provider is that if it is set in client settings, then it would be used for both async calls and background tasks, ...  As long as we keep the same behavior for http, we are good enough.

Makes sense for FR and I can create an issue for that. I believe that behavior is correct -- I was referencing this comment which I should have quoted above: 

> in the mean time, use InstantiatingExecutorProvider for only the background tasks seems the right approach to me.

The channel's default executor is created as `InstantiatingExecutorProvider` for async calls : https://github.com/googleapis/sdk-platform-java/blob/df1bab0bb9139774b6d6a39842e27f0d23a6a356/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ManagedHttpJsonChannel.java#L215-L218 

I wasn't sure if you meant to remove the `InstantiatingExecutorProvider` in the channel. I was thinking you meant that we should use the default `newBuilder()` behavior for background tasks: https://github.com/googleapis/sdk-platform-java/blob/a988fe7dcfc8568c9256c22b5df907b16e5eea5a/gax-java/gax/src/main/java/com/google/api/gax/rpc/StubSettings.java#L258 and that we should create a new method inside `InstantiatingExecutorProvider` (like newAsyncBuilder or w/e) that allows us to customize with a different amount of cpus for the async calls (i.e. allow it being unbound or fixed or w/e).

I'll play around with thread values and investigate how gRPC sets it.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ManagedHttpJsonChannel.java
```diff
@@ -82,58 +84,116 @@ public <RequestT, ResponseT> HttpJsonClientCall<RequestT, ResponseT> newCall(
...
-    // TODO
-    return false;
+    long endTimeNanos = System.nanoTime() + unit.toNanos(duration);
+    long awaitTimeNanos = endTimeNanos - System.nanoTime();
+    if (awaitTimeNanos <= 0) {
+      return false;
+    }
+    // Only awaitTermination for the executor if it was created by Gax. External executors
+    // should be managed by the user.
+    if (usingDefaultExecutor && executor instanceof ExecutorService) {
```
blakeli0, line 164: This `usingDefaultExecutor && executor instanceof ExecutorService` is used 4 times and can be extracted to a private method like `shouldManageExecutor()`.

gax-java/gax/src/main/java/com/google/api/gax/core/InstantiatingExecutorProvider.java
```diff
@@ -54,6 +54,9 @@ public Thread newThread(Runnable runnable) {
           return thread;
         }
       };
+  private static final int MIN_CPU_AMOUNT = 4;
```
blakeli0, line 57: This should be `MIN_THREAD_AMOUNT`?

lqiu96: Sure. Updating.

gax-java/gax/src/main/java/com/google/api/gax/core/InstantiatingExecutorProvider.java
```diff
@@ -76,9 +79,22 @@ public boolean shouldAutoClose() {
 
   public abstract Builder toBuilder();
 
+  // Used for CPU Bound tasks as the thread count is at max the number of processors
+  // Thread count minimum is at least `MIN_CPU_AMOUNT`
   public static Builder newBuilder() {
     int numCpus = Runtime.getRuntime().availableProcessors();
-    int numThreads = Math.max(4, numCpus);
+    int numThreads = Math.max(MIN_CPU_AMOUNT, numCpus);
```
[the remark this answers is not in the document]
lqiu96: Makes sense.I created a new builder: `newIOBuilder` that should only be used for HttpJson's async calls and wouldn't change the background executor.

blakeli0 (approved):

PR #1691, opened 2023-05-22
chore: Clean up resources in gRPCStreamDirectController and ChannelPool tests

lqiu96:
Clean up the resources that are created in the the GrpcDirectStreamControllerTest and ChannelPool tests.

Fixes: https://github.com/googleapis/sdk-platform-java/issues/1704


gax-java/gax-grpc/src/test/java/com/google/api/gax/grpc/ChannelPoolTest.java
```diff
@@ -241,12 +253,14 @@ public void channelPrimerIsCalledPeriodically() throws IOException {
     // 1 more call during channel refresh
     Mockito.verify(mockChannelPrimer, Mockito.times(3))
         .primeChannel(Mockito.any(ManagedChannel.class));
+    pool.shutdown();
```
blakeli0, line 256: Since this test is a unit test for `ChannelPool`, I would assume every test is creating its own channel pool, so it's worth it to extract `pool` to a field, the ndo shutdown and awaitTermination in a clean up method.

lqiu96: Ah! Good point. Updating.

PR #1692, opened 2023-05-22
feat: Create additional tag on release for google-cloud-shared-dependencies

alicejli:
For the Javadocs generation workflow, it will be useful to have a separate tag for each sdk-platform-java release that corresponds to a specific `google-cloud-shared-dependencies` version.


.github/workflows/create_shared_dependencies_release_tag.yaml
```diff
@@ -0,0 +1,33 @@
...
+        token: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Set up Git
+      run: |
+        git config --local user.email "action@github.com"
+        git config --local user.name "GitHub Action"
+
+    - name: Create additional tag
+      run: |
+        ARTIFACT_ID='google-cloud-shared-dependencies'
```
suztomo, line 30: for ARTIFACT_ID in gax google-cloud-shared-dependencies api-common`

.github/workflows/create_shared_dependencies_release_tag.yaml
```diff
@@ -0,0 +1,33 @@
+name: Create additional tag for each release with google-cloud-shared-dependencies version
```
suztomo, line 1: Would you make it generic so that gax and api-common-java versions are also tagged?

```
for ARTIFACT_ID in gax google-cloud-shared-dependencies api-common; do
...
```

(Approving, just in case you don't have time for it.)

suztomo (approved):

.github/workflows/create_shared_dependencies_release_tag.yaml
alicejli: good idea - updated. thanks!

PR #1695, opened 2023-05-22
fix: update workflow for refs and multiple artifacts

alicejli:
Follow up to https://github.com/googleapis/sdk-platform-java/pull/1692
This fixes an error in the workflow if it is triggered via release, and also corrects for multiple artifacts that begin with `gax`.

### DIFF
--- a/.github/workflows/create_additional_release_tags.yaml
+++ b/.github/workflows/create_additional_release_tags.yaml
@@ -3,11 +3,7 @@ name: Create additional tags for each release
 on:
   release:
     types: [published]
-  workflow_dispatch: # If manually triggered, clarify which release to create the additional tags for
-    inputs:
-      releaseTag:
-        description: 'Release Tag'
-        required: true
+  workflow_dispatch:
 
 jobs:
   build:
@@ -17,7 +13,6 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v3
       with:
-        ref: ${{ github.event.release.tag_name || github.event.inputs.releaseTag }}
         token: ${{ secrets.GITHUB_TOKEN }}
 
     - name: Set up Git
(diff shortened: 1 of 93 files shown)


.github/workflows/create_additional_release_tags.yaml
```diff
@@ -17,7 +17,7 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v3
       with:
-        ref: ${{ github.event.inputs.releaseTag }}
+        ref: ${{ github.event.release.tag_name || github.event.inputs.releaseTag }}
```
suztomo, line 20: Can we use the default value?


https://github.com/actions/checkout/blob/main/README.md

    # The branch, tag or SHA to checkout. When checking out the repository that
    # triggered a workflow, this defaults to the reference or SHA for that event.
    # Otherwise, uses the default branch.
    ref: ''

.github/workflows/create_additional_release_tags.yaml
```diff
@@ -27,10 +27,17 @@ jobs:
...
       run: |
+        REF=${{ github.event.release.tag_name || github.event.inputs.releaseTag }}
         ARTIFACT_IDS=('google-cloud-shared-dependencies' 'api-common' 'gax')
         for ARTIFACT_ID in "${ARTIFACT_IDS[@]}"
         do
-          VERSION=$(grep "${ARTIFACT_ID}" versions.txt | cut -d':' -f2)
-          git tag ${ARTIFACT_ID}/$VERSION ${{ github.event.inputs.releaseTag }}
-          git push origin ${ARTIFACT_ID}/$VERSION
+          VERSION=$(grep "^${ARTIFACT_ID}:" versions.txt | cut -d':' -f2 | tr -d '[:space:]')
+          TEMP_BRANCH="temp_branch_${ARTIFACT_ID}_$VERSION"
```
suztomo, line 35: I don't think you need a branch to create tags.

suztomo: Can you show me the permission error that was resolved by creating a branch or deleting the yaml file?

.github/workflows/create_additional_release_tags.yaml
suztomo: Workflow_dispatch lets you specify a tag when invoking. Try that.

suztomo: So you don't even need releaseTag argument.

.github/workflows/create_additional_release_tags.yaml
```diff
@@ -27,10 +27,17 @@ jobs:
...
+          git checkout -b $TEMP_BRANCH
+          git rm .github/workflows/create_additional_release_tag.yaml
+          git commit -m "Remove workflow file"
+          git tag "${ARTIFACT_ID}/$VERSION"
```
suztomo, line 39: Can you name tag with "v"?

(It follows convention. https://github.com/googleapis/google-cloud-go/tags)

.github/workflows/create_additional_release_tags.yaml
suztomo: > If it's removed, then the workflow definition file must exist at the tag.

So the use of `${{ github.event.release.tag_name || github.event.inputs.releaseTag }}` is fine.

alicejli:
I messed up fixing merge conflicts in this PR. Going to open a new one.

PR #1699, opened 2023-05-23
fix: update tag workflow

alicejli:
Replacement for https://github.com/googleapis/sdk-platform-java/pull/1695

This fixes a couple issues with the original workflow:
1) Skips tags that are already created instead of failing
2) Uses exact matches for artifact IDs (multiple artifacts that begin with `gax` were causing issues)
3) Removes manual input for manual trigger; will automatically take the latest release which is what we want.

### DIFF
--- a/.github/workflows/create_additional_release_tag.yaml
+++ b/.github/workflows/create_additional_release_tag.yaml
@@ -23,15 +23,13 @@ jobs:
     - name: Create additional tags
       run: |
         ARTIFACT_IDS=('google-cloud-shared-dependencies' 'api-common' 'gax')
-        for ARTIFACT_ID in "${ARTIFACT_IDS[@]}"
-        do
+        for ARTIFACT_ID in "${ARTIFACT_IDS[@]}"; do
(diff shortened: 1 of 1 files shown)


.github/workflows/create_additional_release_tag.yaml
```diff
@@ -30,7 +25,13 @@ jobs:
         ARTIFACT_IDS=('google-cloud-shared-dependencies' 'api-common' 'gax')
         for ARTIFACT_ID in "${ARTIFACT_IDS[@]}"
         do
-          VERSION=$(grep "${ARTIFACT_ID}" versions.txt | cut -d':' -f2)
-          git tag ${ARTIFACT_ID}/$VERSION ${{ github.event.inputs.releaseTag }}
-          git push origin ${ARTIFACT_ID}/$VERSION
+          VERSION=$(grep "^${ARTIFACT_ID}:" versions.txt | cut -d':' -f2 | tr -d '[:space:]')
```
suztomo, line 28: Can you format the indentation to follow https://google.github.io/styleguide/shellguide.html#loops

alicejli: Thanks for the reference! Updated.

suztomo (approved):

PR #1700, opened 2023-05-23
fix: fix release tag workflow to fetch all tags first

alicejli:
Hopefully final follow-up to: https://github.com/googleapis/sdk-platform-java/pull/1699 

Workflow currently fails if tag is already created. This is because the actions/checkout checks out the code in a fresh workspace that does not contain tags by default. This adds in the step of fetching tags.


suztomo (approved):
Alternatively we can make it override old tags. Either handling is fine, because I don't expect this workflow to run against an old release.

suztomo:
@mpeddada1 This change doesn't affect the release. This is just an error handling in case there's existing tags.

Issue #1708, opened 2023-05-24
gapic-generator-java: discrepancy in writers handling deeply nested types


emmileaf:
Suppose nested type `Outer.Middle.Inner`: in some scenarios the generated code will not compile because of nesting level mismatch between type and import:

```
// from ImportWriterVisitor
import com.google.testgapic.v1beta1.Outer.Middle; 

// from JavaWriterVisitor
Outer.Middle.Inner myVariable; 

```

There are possibly other discrepancies with nested class handling (e.g. for ConcreteReference vs. VaporReference) to be investigated. (See https://github.com/googleapis/sdk-platform-java/pull/1709/commits/f505036ad894b8abc1db25c8b4881f2b7009e61d for additional tests illustrating unintended behavior)

PR #1709, opened 2023-05-24
fix: [gapic-generator-java] align writer behavior for nested types

emmileaf:
WIP PR for https://github.com/googleapis/sdk-platform-java/issues/1708.

- [x] Reproduce generation error through test setup
- [ ] Implement fix
- [ ] Additional tests to verify intended behavior

### DIFF
--- a/gapic-generator-java/src/test/java/com/google/api/generator/engine/writer/ImportWriterVisitorTest.java
+++ b/gapic-generator-java/src/test/java/com/google/api/generator/engine/writer/ImportWriterVisitorTest.java
@@ -58,6 +58,7 @@ import com.google.api.generator.engine.ast.VariableExpr;
 import com.google.api.generator.test.utils.LineFormatter;
 import com.google.common.base.Function;
 import com.google.common.base.Strings;
+import com.google.testgapic.v1beta1.Outer;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
@@ -426,6 +427,32 @@ public class ImportWriterVisitorTest {
             "import java.lang.annotation.Repeatable;\n",
             "import java.util.List;\n\n"),
         writerVisitor.write());
+  }
+
+  @Test
+  public void writeImport_nestedClasses() {
+
+    VaporReference nestedVaporReference =
+        VaporReference.builder()
+            .setName("Inner")
+            .setEnclosingClassNames(Arrays.asList("Outer", "Middle"))
+            .setPakkage("com.google.testgapic.v1beta1")
+            .build();
+
+    TypeNode.withReference(nestedVaporReference).accept(writerVisitor);
+    // TODO: The import below causes compilation errors,
+    // since JavaWriterVisitor writes the corresponding type as Outer.Middle.Inner
+    // Need to change either Writer's behavior to align with the other
+    assertEquals("import com.google.testgapic.v1beta1.Outer.Middle;\n\n", writerVisitor.write());
+
+    writerVisitor.clear();
+
+    ConcreteReference nestedConcreteReference =
+        ConcreteReference.withClazz(Outer.Middle.Inner.class);
+    TypeNode.withReference(nestedConcreteReference).accept(writerVisitor);
+    // TODO: This is currently writing import com.google.testgapic.v1beta1.Middle.Outer;
+    // which is incorrect - uncomment assertion along with fix
+    // assertEquals("import com.google.testgapic.v1beta1.Outer.Middle;\n\n", writerVisitor.write());
   }
 
   @Test
--- a/gapic-generator-java/src/test/java/com/google/api/generator/engine/writer/JavaWriterVisitorTest.java
+++ b/gapic-generator-java/src/test/java/com/google/api/generator/engine/writer/JavaWriterVisitorTest.java
@@ -74,6 +74,7 @@ import com.google.api.generator.engine.ast.WhileStatement;
 import com.google.api.generator.test.utils.LineFormatter;
 import com.google.api.generator.test.utils.TestExprBuilder;
 import com.google.common.base.Function;
+import com.google.testgapic.v1beta1.Outer;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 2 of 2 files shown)


burkedavison:
> @meltsufin : No impact on the goldens?

If so, perhaps we can create a (very) simple extended showcase proto to exercise this behavior -- similar to how we did for validating REST-only client generation behavior?
https://github.com/googleapis/sdk-platform-java/tree/main/showcase/gapic-showcase-extended/proto

emmileaf:
> No impact on the goldens?

@meltsufin Not for this change. For a bit more context - this nested class import issue was initially uncovered when the generator mistakenly tried to pull in `DescriptorProtos.GeneratedCodeInfo.Annotation`, and this behavior is being fixed in https://github.com/googleapis/sdk-platform-java/pull/1726 (workaround in the specific [proto](https://github.com/googleapis/googleapis/blob/640f7439448629156b3ff30adc0332cc8b7ce52c/google/cloud/rapidmigrationassessment/v1/rapidmigrationassessment.proto#L66) was to fully specify package name). 

> If so, perhaps we can create a (very) simple extended showcase proto to exercise this behavior

@burkedavison I was considering this as well - actually tried doing this for validating the change in https://github.com/googleapis/sdk-platform-java/pull/1726, but this fix was easier to unit test so I just went with those instead. I can set up something similarly here if the approach taken in the other PR makes sense?

meltsufin (approved):

gapic-generator-java/src/main/java/com/google/api/generator/engine/ast/ConcreteReference.java
```diff
@@ -103,14 +103,15 @@ public ImmutableList<String> enclosingClassNames() {
...
+    // Builds list in order of inner to outer.
+    // Return the reversed list, since the outermost type is expected to lie at index 0.
     ImmutableList.Builder<String> listBuilder = new ImmutableList.Builder<>();
     Class<?> currentClz = clazz();
     while (currentClz.getEnclosingClass() != null) {
       listBuilder.add(currentClz.getEnclosingClass().getSimpleName());
       currentClz = currentClz.getEnclosingClass();
     }
-    return listBuilder.build();
+    return listBuilder.build().reverse();
```
blakeli0, line 114: Can we achieve the same thing without this change? Seems we can try to get the last element from the list in the writer?

gapic-generator-java/src/test/java/com/google/api/generator/engine/writer/ImportWriterVisitorTest.java
```diff
@@ -428,6 +429,26 @@ public void writeVariableExprImports_annotationsWithDescription() {
...
+            .setEnclosingClassNames(Arrays.asList("Outer", "Middle"))
+            .setPakkage("com.google.testgapic.v1beta1")
+            .build();
+
+    TypeNode.withReference(nestedVaporReference).accept(writerVisitor);
+    assertEquals("import com.google.testgapic.v1beta1.Outer;\n\n", writerVisitor.write());
+
+    writerVisitor.clear();
+
+    ConcreteReference nestedConcreteReference =
```
blakeli0, line 446: This should be break into two test cases if we want to test both type of references. However, I think we only need to test one of them if the only change is in the writer.

blakeli0:
> > No impact on the goldens?
> 
> @meltsufin Not for this change. For a bit more context - this nested class import issue was initially uncovered when the generator mistakenly tried to pull in `DescriptorProtos.GeneratedCodeInfo.Annotation`, and this behavior is being fixed in #1726 (workaround in the specific [proto](https://github.com/googleapis/googleapis/blob/640f7439448629156b3ff30adc0332cc8b7ce52c/google/cloud/rapidmigrationassessment/v1/rapidmigrationassessment.proto#L66) was to fully specify package name).
> 
> > If so, perhaps we can create a (very) simple extended showcase proto to exercise this behavior
> 
> @burkedavison I was considering this as well - actually tried doing this for validating the change in #1726, but this fix was easier to unit test so I just went with those instead. I can set up something similarly here if the approach taken in the other PR makes sense?

I agree with Mike and Burke that it would be better to have a golden test to easily see the changes. Showcase may be a little overkill since this does not interact with the server, a golden unit test would be a great fit.

gapic-generator-java/src/test/java/com/google/api/generator/engine/writer/ImportWriterVisitorTest.java
emmileaf: I think the additional change in `ConcreteReference` is still needed given https://github.com/googleapis/sdk-platform-java/pull/1709#discussion_r1214453516. Breaking this into separate test cases makes sense though - will update.

emmileaf:
> I agree with Mike and Burke that it would be better to have a golden test to easily see the changes. Showcase may be a little overkill since this does not interact with the server, a golden unit test would be a great fit.

@blakeli0 Makes sense. Looks like there's actually a [nested_message.proto](https://github.com/googleapis/sdk-platform-java/blob/main/gapic-generator-java/src/test/proto/nested_message.proto) that exists and is currently used in `TypeParserTest`, just without goldens. I'll try leveraging that to add some golden unit tests here.

emmileaf:
Added a test proto and new golden unit test in https://github.com/googleapis/sdk-platform-java/pull/1709/commits/6ab1628d1cc3da5b8b3fa22f89a9a96bb7eb4db3.

(Had some uncertainty deciding where best to put this golden test since the existing ones are mostly separate for each `Composer`, and all of these exercise the writer behavior implicitly. However, this fix is more specific to writer logic and could be applicable to a number of Composers and the class files they generate. 

I ended up adding one of these into `JavaWriterVisitorTest` that uses the grpc `ServiceClientClassComposer` to write an example file with nested types and imports - please let me know if this approach makes sense!)

blakeli0:
> Added a test proto and new golden unit test in [6ab1628](https://github.com/googleapis/sdk-platform-java/commit/6ab1628d1cc3da5b8b3fa22f89a9a96bb7eb4db3).
> 
> (Had some uncertainty deciding where best to put this golden test since the existing ones are mostly separate for each `Composer`, and all of these exercise the writer behavior implicitly. However, this fix is more specific to writer logic and could be applicable to a number of Composers and the class files they generate.
> 
> I ended up adding one of these into `JavaWriterVisitorTest` that uses the grpc `ServiceClientClassComposer` to write an example file with nested types and imports - please let me know if this approach makes sense!)

Thanks Emily! I think your approach is fine, it's very specific to what we want to test. As you mentioned, our golden tests are currently assuming writers behave perfectly so it is a little weird to add a golden test just for writers. If you prefer, we can add this test RPC to one of the existing protos, so that it's tested through other golden tests and don't have to load it separately.

blakeli0 (approved):
LGTM. We could move the golden tests to composers or add the test case to one of the existing protos, but I don't think it should block this PR.

blakeli0 (approved):
LGTM. We could move the golden tests to composers or add the test case to one of the existing protos, but I don't think it should block this PR.

emmileaf:
Sounds good - I'll merge this in since the fix is independent of https://github.com/googleapis/sdk-platform-java/pull/1726, so that it can make it into the upcoming release!

PR #1726, opened 2023-05-31
fix: [gapic-generator-java] handle response and metadata type ambiguity in LRO parsing

### LINKED ISSUE #1707: gapic-generator-java: LRO parsing can match wrong response or metadata type
This issue happens in [Parser.parseLro()](https://github.com/googleapis/sdk-platform-java/blob/cc5724f04722e3d60eb9b4e02ecc1bdf5dd7a7fa/gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java#L784): if only shortname is provided, and has ambiguity with message types from other protos, the parser currently naively matches on first encountered type, without throwing exception or indicating conflict.

https://github.com/googleapis/sdk-platform-java/blob/cc5724f04722e3d60eb9b4e02ecc1bdf5dd7a7fa/gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java#L833-L841

Example proto to reproduce:
* Annotation type is ambiguous with `DescriptorProtos.GeneratedCodeInfo.Annotation`
* Location type is ambiguous with `DescriptorProtos.SourceCodeInfo.Location`

```
service TestService {
 ...
  rpc annotationTest(Request) returns (google.longrunning.Operation) {
    option (google.longrunning.operation_info) = {
      response_type: "Annotation" 
      metadata_type: "Location"
    };
  }
}

message Annotation {
  string name = 1;
}

message Location {
  string name = 1;
}
```

emmileaf:
This PR:
* Updates `Parser.parseLro()` to additionally match response and metadata type on same package as rpc when only shortname is provided. (https://github.com/googleapis/sdk-platform-java/pull/1726/commits/76d8fbaeb6594cfdc3e43fd91d3a54d7320d1b98) 
   - Adds unit tests for relevant logic in `TypeParserTest`, using newly added `collisions.proto`

* Fixes `JavaWriterVisitor` logic for generics on `MethodInvocationExpr` (e.g. `ImmutableList.<another.package.Foo>of()`) so that the fully qualified type is written when needed. (https://github.com/googleapis/sdk-platform-java/pull/1726/commits/67fececd180a7753abb71104f25ab5efd9e10984)
  -   Adds unit tests for relevant logic in `JavaWriterVisitorTest`

* ~Enables showcase-extended module (from [#1236](https://github.com/googleapis/sdk-platform-java/pull/1236)) with golden tests, adding `collisions.proto` to capture these edge cases in end-to-end generation and verify intended behavior~ 

Fixes #1707 ☕️


gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -824,39 +824,34 @@ static LongrunningOperation parseLro(
     boolean isResponseTypeNameShortOnly = lastDotIndex < 0;
     String responseTypeShortName =
         lastDotIndex >= 0 ? responseTypeName.substring(lastDotIndex + 1) : responseTypeName;
+    // When only shortname is provided, match on same proto package as method (See
+    // https://aip.dev/151)
+    String responseTypeFullName =
+        isResponseTypeNameShortOnly
+            ? methodDescriptor.getFile().getPackage() + "." + responseTypeShortName
+            : responseTypeName;
```
lqiu96, line 832: nit: Maybe we could refactor L823 - L843 to have a private helper method that checks for if the response/ metadata name is a shortname and returns the valid full type name back?

i.e.
```
String responseTypeFullName = getTypeFullName(packageName, responseTypeName);
String metadataTypeFullName = getTypeFullName(packageName, metadataTypeName);
```

and the helper func has the logic for checking the dot index, if it's a shortname, and concating the full type name back?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -824,39 +824,34 @@ static LongrunningOperation parseLro(
...
 
     lastDotIndex = metadataTypeName.lastIndexOf('.');
     boolean isMetadataTypeNameShortOnly = lastDotIndex < 0;
     String metadataTypeShortName =
         lastDotIndex >= 0 ? metadataTypeName.substring(lastDotIndex + 1) : metadataTypeName;
+    // When only shortname is provided, match on same proto package as method (See
+    // https://aip.dev/151)
+    String metadataTypeFullName =
```
blakeli0, line 840: What if the short name is invalid? Can we throw an exception in the generator? Or we are going to generate code that is not compilable?

emmileaf: Note from discussion offline: if the short name is invalid (no match) then the exception is handled by existing logic here:
https://github.com/googleapis/sdk-platform-java/blob/b73d72aad94f0250841c1c8fcb0c2d3144237eea/gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java#L858-L868

But this behavior should be tested (through true unit tests rather than goldens). Will look into replacing showcase-extended setup in this PR with true unit testing for `Parser`.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
emmileaf: Ah you're right - the code is doing very similar things for response and metadata type here. I'll take a look at refactoring!

emmileaf: Updated in https://github.com/googleapis/sdk-platform-java/pull/1726/commits/de5e9c39c0d757200bddba4878c533f3104a8df3.

blakeli0 (approved):

PR #1777, opened 2023-06-15
chore: eliminate duplicate guava version declarations

burkedavison:
The guava version is now exclusively defined at https://github.com/googleapis/sdk-platform-java/blob/main/gapic-generator-java-pom-parent/pom.xml#L32


java-shared-dependencies/dependency-convergence-check/pom.xml
```diff
@@ -33,6 +44,7 @@
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
+      <version>4.13.2</version>
```
blakeli0, line 47: Is this change required?

PR #1779, opened 2023-06-16
chore: copy and update PROPERTIES using widely supported syntax

burkedavison:
The `dict | dict` syntax is only supported in Python 3.9+ which is not available in all test environments.

This PR uses a syntax that is supported more widely.


blakeli0 (approved):
LGTM. For my own curiosity, what testing environments do not have Python 3.9?

burkedavison:
> LGTM. For my own curiosity, what testing environments do not have Python 3.9?

Discussed offline

Issue #1547, opened 2023-03-23
com.google.showcase.v1beta1.MessagingClientHttpJsonTest#updateBlurbTest failing with 404 error


mpeddada1:
Stacktrace:

```
com.google.api.gax.rpc.NotFoundException: com.google.api.client.http.HttpResponseException: 404
POST https://localhost:7469/v1beta1/users/%5BUSER%5D/profile/blurbs/legacy/%5BLEGACY_USER%5D~%5BBLURB%5D?updateMask
Method not found for path 'https:///v1beta1/users/%5BUSER%5D/profile/blurbs/legacy/%5BLEGACY_USER%5D~%5BBLURB%5D'

	at com.google.api.gax.rpc.ApiExceptionFactory.createException(ApiExceptionFactory.java:90)
	at com.google.api.gax.rpc.ApiExceptionFactory.createException(ApiExceptionFactory.java:41)
	at com.google.api.gax.httpjson.HttpJsonApiExceptionFactory.createApiException(HttpJsonApiExceptionFactory.java:75)
	at com.google.api.gax.httpjson.HttpJsonApiExceptionFactory.create(HttpJsonApiExceptionFactory.java:54)
	at com.google.api.gax.httpjson.HttpJsonExceptionCallable$ExceptionTransformingFuture.onFailure(HttpJsonExceptionCallable.java:97)
	at com.google.api.core.ApiFutures$1.onFailure(ApiFutures.java:67)
	at com.google.common.util.concurrent.Futures$CallbackListener.run(Futures.java:1132)
	at com.google.common.util.concurrent.DirectExecutor.execute(DirectExecutor.java:31)
	at com.google.common.util.concurrent.AbstractFuture.executeListener(AbstractFuture.java:1270)
	at com.google.common.util.concurrent.AbstractFuture.complete(AbstractFuture.java:1038)
	at com.google.common.util.concurrent.AbstractFuture.setException(AbstractFuture.java:808)
	at com.google.api.core.AbstractApiFuture$InternalSettableFuture.setException(AbstractApiFuture.java:92)
	at com.google.api.core.AbstractApiFuture.setException(AbstractApiFuture.java:74)
	at com.google.api.gax.httpjson.HttpJsonClientCalls$HttpJsonFuture.setException(HttpJsonClientCalls.java:112)
	at com.google.api.gax.httpjson.HttpJsonClientCalls$FutureListener.onClose(HttpJsonClientCalls.java:146)
[...]

PR #1789, opened 2023-06-20
chore: Implement gRPC and HttpJson showcase tests for IAM

lqiu96:

Part of #1439 ☕️

These showcase tests were modeled after gapic-generator-go's IAM showcase tests.


showcase/gapic-showcase/pom.xml
```diff
@@ -229,5 +229,10 @@
       <classifier>testlib</classifier>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.google.api.grpc</groupId>
+      <artifactId>grpc-google-iam-v1</artifactId>
+      <scope>test</scope>
+    </dependency>
```
lqiu96, line 236: Added since there are compilation issues for MockIAMPolicy and MockIAMPolicyImpl

showcase/BUILD.bazel
```diff
@@ -19,7 +19,8 @@ proto_library_with_info(
         "@com_google_gapic_showcase//schema/google/showcase/v1beta1:showcase_proto",
 #        "//showcase/gapic-showcase-extended/proto:showcase_proto_extended",
         "@com_google_googleapis//google/cloud:common_resources_proto",
-        "@com_google_googleapis//google/cloud/location:location_proto"
+        "@com_google_googleapis//google/cloud/location:location_proto",
+        "@com_google_googleapis//google/iam/v1:iam_policy_proto"
```
blakeli0, line 23: So adding this `iam_policy_proto` implicitly enables IAM feature? Is this how we are suppose to enable IAM?

showcase/pom.xml
```diff
@@ -80,6 +80,11 @@
               <!-- TODO(1547): Remove this exclusion once MessagingClientHttpJsonTest#updateBlurbTest is fixed -->
               <excludes>
                 <exclude>**/MessagingClientHttpJsonTest.java</exclude>
+                <exclude>**/ComplianceClientHttpJsonTest.java</exclude>
```
blakeli0, line 83: Are these changes related?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITIam.java
```diff
@@ -0,0 +1,256 @@
...
+
+  @BeforeClass
+  public static void createClients() throws Exception {
+    grpcClient = TestClientInitializer.createGrpcIdentityClient();
+    httpjsonClient = TestClientInitializer.createHttpJsonIdentityClient();
+  }
+
+  @Before
+  public void setupTests() {
+    resourceName = "users/" + UUID.randomUUID().toString().substring(0, 8);
```
blakeli0, line 57: Why the resource name is in this format? I see that it is a wild card resource based on the definition in [googleapis](https://github.com/googleapis/googleapis/blob/bd3fbf4776fe1240f2fddcc525a6b6e7690232ad/google/iam/v1/iam_policy.proto#L106), I guess it is randomly chosen?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITIam.java
```diff
@@ -0,0 +1,256 @@
...
+            .build();
+    Policy policy = httpjsonClient.setIamPolicy(policyRequest);
+    assertThat(policy).isEqualTo(DEFAULT_POLICY);
+  }
+
+  @Test
+  public void testGrpc_setIamPolicy_missingResource() {
+    SetIamPolicyRequest setIamPolicyRequest = SetIamPolicyRequest.newBuilder().build();
+    assertThrows(
+        InvalidArgumentException.class, () -> grpcClient.setIamPolicy(setIamPolicyRequest));
```
blakeli0, line 96: Where is this exception being thrown? Is it in our client libraries? I see that we are adding a lot of error scenarios in this IT, a general comment is that we only want to test the exception if it's being thrown in the IAM code before it reaches backend, or if there is any specific response exception mapping logic in IAM code.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITIam.java
```diff
@@ -0,0 +1,256 @@
...
+        httpjsonClient.getIamPolicy(
+            GetIamPolicyRequest.newBuilder().setResource(policyRequest.getResource()).build());
+    assertThat(policy).isEqualTo(DEFAULT_POLICY);
+  }
+
+  @Test
+  public void testGrpc_getIamPolicy_doesNotExist() {
+    GetIamPolicyRequest getIamPolicyRequest =
+        GetIamPolicyRequest.newBuilder().setResource(resourceName).build();
+    assertThrows(NotFoundException.class, () -> grpcClient.getIamPolicy(getIamPolicyRequest));
```
blakeli0, line 156: I don't think we want to test this scenario, as this exception is based on a backend response. For example, not likely but the backend could return a different status code which will be mapped to a different exception.

showcase/BUILD.bazel
lqiu96: I believe having this in the bazel rule enables it: https://github.com/googleapis/gapic-showcase/blob/b146f892d98ea0d733c45a2315941e8e6ab69c90/schema/google/showcase/v1beta1/showcase_v1beta1.yaml#L29

But this was to generate the Java related Mixin code.

showcase/pom.xml
lqiu96: Yes. It was causing some errors in the ComplienceTest:
```
[ERROR] Errors:
[ERROR]   ComplianceClientHttpJsonTest.getIamPolicyTest » NotFound com.google.api.client.http.HttpResponseException: 404
GET https://localhost:7469/v1beta1/users/%5BUSER%5D/profile/blurbs/legacy/%5BLEGACY_USER%5D~%5BBLURB%5D:getIamPolicy
```

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITIam.java
lqiu96: I was using the IdentityClient to call the IAM RPCs which uses the User resource. I just gave it `user/{random_user_id}`.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITIam.java
lqiu96: This is being thrown in the backend: https://github.com/googleapis/gapic-showcase/blob/b146f892d98ea0d733c45a2315941e8e6ab69c90/server/services/iam_policy_service.go#L27-L29

There is only a few validation checks that the backend iam server does.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITIam.java
lqiu96: This was mapped from the showcase server's error responses: https://github.com/googleapis/gapic-showcase/blob/b146f892d98ea0d733c45a2315941e8e6ab69c90/server/services/iam_policy_service.go#L27-L29

showcase/pom.xml
blakeli0: I see, I think https://github.com/googleapis/sdk-platform-java/issues/1547 at least worth investigating, then we can decide should we fix it or not. It could be that the showcase proto is configured in a way that real protos do not, if not, it could happen when a service team configures their protos in the same way as showcase protos.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITIam.java
blakeli0: If the error is from the backend, I don't think we want to test them, or at most test one of them to make sure exceptions from IAM server can be properly mapped. Test all the permutations is not needed.

lqiu96: Sounds good. I'll keep one test and remove the rest.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITIam.java
blakeli0: Discussed offline that this could be any format.

blakeli0 (approved):

PR #1790, opened 2023-06-21
fix: abstract batch resource and add method to determine if batch should be flushed

mutianf:
Abstract the element and byte counting to a `BatchResource` object so client libraries can add other dimensions to it. Also moved `shouldBeFull` logic to `BatchResource` so libraries can override it and use the other dimensions to determine if the batch should be flushed.

### DIFF
--- a/gax-java/gax/clirr-ignored-differences.xml
+++ b/gax-java/gax/clirr-ignored-differences.xml
@@ -13,4 +13,10 @@
     <method>*setWaitTimeout*</method>
     <to>com.google.api.gax.rpc.ServerStreamingCallSettings$Builder</to>
   </difference>
+  <!-- BatchingDescriptor is an InternalApi intended for google cloud libraries use only -->
+  <difference>
+    <differenceType>7012</differenceType>
+    <className>com/google/api/gax/batching/BatchingDescriptor</className>
+    <method>*</method>
+  </difference>
 </differences>
--- a/gax-java/gax/clirr-ignored-differences.xml
+++ b/gax-java/gax/clirr-ignored-differences.xml
@@ -13,10 +13,4 @@
     <method>*setWaitTimeout*</method>
     <to>com.google.api.gax.rpc.ServerStreamingCallSettings$Builder</to>
   </difference>
-  <!-- BatchingDescriptor is an InternalApi intended for google cloud libraries use only -->
-  <difference>
-    <differenceType>7012</differenceType>
-    <className>com/google/api/gax/batching/BatchingDescriptor</className>
-    <method>*</method>
-  </difference>
 </differences>
--- a/gax-java/gax/src/main/java/com/google/api/gax/batching/BatchResource.java
+++ b/gax-java/gax/src/main/java/com/google/api/gax/batching/BatchResource.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above
+ * copyright notice, this list of conditions and the following disclaimer
+ * in the documentation and/or other materials provided with the
+ * distribution.
+ *     * Neither the name of Google LLC nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.google.api.gax.batching;
+
+import com.google.api.core.InternalApi;
+
+/**
+ * Represent the resource used by a batch including element and byte. It can also be extended to
+ * other things to determine if adding a new element needs to be flow controlled or if the current
+ * batch needs to be flushed.
+ */
+@InternalApi("For google-cloud-java client use only.")
+public interface BatchResource {
+
+  /** Adds the additional resource. */
+  BatchResource add(BatchResource resource);
+
+  /** Returns the element count of this resource. */
+  long getElementCount();
+
+  /** Returns the byte count of this resource. */
+  long getByteCount();
+
+  /** Returns true if the resource is empty. */
+  boolean isEmpty();
+}
--- a/gax-java/gax/src/main/java/com/google/api/gax/batching/BatchResource.java
+++ b/gax-java/gax/src/main/java/com/google/api/gax/batching/BatchResource.java
@@ -47,7 +47,4 @@ public interface BatchResource {
 
   /** Returns the byte count of this resource. */
   long getByteCount();
-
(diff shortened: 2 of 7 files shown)


gax-java/gax/clirr-ignored-differences.xml
```diff
@@ -13,4 +13,9 @@
     <method>*setWaitTimeout*</method>
     <to>com.google.api.gax.rpc.ServerStreamingCallSettings$Builder</to>
   </difference>
+  <difference>
+    <differenceType>7012</differenceType>
```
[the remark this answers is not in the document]
mutianf: 👍  No rush on reviewing yet! It's still a draft :)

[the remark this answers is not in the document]
blakeli0: Yeah thanks for the reminder! I figured so I didn't review the details, left the comment only in case I forget it next time.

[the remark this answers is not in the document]
mutianf: I added it here because `@InternalApi` seems to just be an annotation and the clirr test still failed. I can add a comment mentioning this is intended for only google libraries to use

gax-java/gax/src/main/java/com/google/api/gax/batching/BatcherImpl.java
```diff
@@ -95,6 +95,10 @@ public class BatcherImpl<ElementT, ElementResultT, RequestT, ResponseT>
   private final FlowController flowController;
   private final ApiCallContext callContext;
 
+  private final long elementThreshold;
+
+  private final long bytesThreshold;
+
```
igorbernstein2, line 101: please add comment about how 0 zero is interpreted

gax-java/gax/src/main/java/com/google/api/gax/batching/BatcherImpl.java
```diff
@@ -241,12 +250,15 @@ public ApiFuture<ElementResultT> add(ElementT element) {
 
     SettableApiFuture<ElementResultT> result = SettableApiFuture.create();
     synchronized (elementLock) {
-      currentOpenBatch.add(element, result, throttledTimeMs);
-    }
+      if (!currentOpenBatch.isEmpty()
+          && batchingDescriptor.shouldFlush(
+              currentOpenBatch.resource.add(newResource), elementThreshold, bytesThreshold)) {
+        sendOutstanding();
+      }
```
igorbernstein2, line 257: I dont think the isEmpty check is necessary here. You want to flush sendOutstanding either way...if its empty, then it will be a no-op

gax-java/gax/src/main/java/com/google/api/gax/batching/DefaultBatchResource.java
```diff
@@ -0,0 +1,71 @@
...
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.google.api.gax.batching;
+
+import com.google.common.base.Preconditions;
+
+/**
+ * The default implementation of {@link BatchResource} which tracks the elementCount and byteCount.
```
igorbernstein2, line 35: Please add a note that this requires the Batching Descriptor to use BatchResource

gax-java/gax/src/main/java/com/google/api/gax/batching/DefaultBatchResource.java
```diff
@@ -0,0 +1,71 @@
...
+  DefaultBatchResource(long elementCount, long byteCount) {
+    this.elementCount = elementCount;
+    this.byteCount = byteCount;
+  }
+
+  @Override
+  public BatchResource add(BatchResource resource) {
+    Preconditions.checkArgument(
+        resource instanceof DefaultBatchResource,
+        "BatchResource needs to be an instance of DefaultBatchResource");
```
igorbernstein2, line 51: Please extend the message to include what you actually got using ie
"BatchResource needs to be an instance of DefaultBatchResource, got %s", resource.getClass()

gax-java/gax/src/main/java/com/google/api/gax/batching/DefaultBatchResource.java
```diff
@@ -0,0 +1,71 @@
...
+
+  @Override
+  public long getByteCount() {
+    return byteCount;
+  }
+
+  @Override
+  public boolean isEmpty() {
+    return elementCount == 0;
+  }
```
igorbernstein2, line 70: as per above I dont think this is necessary

igorbernstein2 (approved):
lgtm after comments are addressed

gax-java/gax/src/main/java/com/google/api/gax/batching/DefaultBatchResource.java
```diff
@@ -0,0 +1,71 @@
...
+
+  @Override
+  public BatchResource add(BatchResource resource) {
+    Preconditions.checkArgument(
+        resource instanceof DefaultBatchResource,
+        "BatchResource needs to be an instance of DefaultBatchResource");
+    this.elementCount += ((DefaultBatchResource) resource).elementCount;
+    this.byteCount += ((DefaultBatchResource) resource).byteCount;
+    return this;
```
igorbernstein2, line 54: you should return a new instance

gax-java/gax/clirr-ignored-differences.xml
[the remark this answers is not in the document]
blakeli0: I meant that maybe we can do something like 
```
  <difference>
 <className>com/google/api/gax/batching/BatchingDescriptor</className>
  </difference>
```
if possible. If not, we can ignore it in pom, like [this](https://github.com/googleapis/sdk-platform-java/blob/3b85e00357fa24bc0f8c23103cccafdcf9d71b34/gax-java/gax/pom.xml#L90).

gax-java/gax/src/main/java/com/google/api/gax/batching/DefaultBatchResource.java
```diff
@@ -0,0 +1,61 @@
...
+import com.google.auto.value.AutoValue;
+import com.google.common.base.Preconditions;
+
+/**
+ * The default implementation of {@link BatchResource} which tracks the elementCount and byteCount.
+ */
+@AutoValue
+abstract class DefaultBatchResource implements BatchResource {
+
+  static DefaultBatchResource create(long elementCount, long byteCount) {
```
blakeli0, line 41: Can we use [builders](https://g3doc.corp.google.com/third_party/java_src/auto/value/g3doc/builders.md?cl=head) instead? Builders are easier to maintain than constructor helpers.

gax-java/gax/src/main/java/com/google/api/gax/batching/BatchResource.java
```diff
@@ -0,0 +1,50 @@
...
+
+import com.google.api.core.InternalApi;
+
+/**
+ * Represent the resource used by a batch including element and byte. It can also be extended to
+ * other things to determine if adding a new element needs to be flow controlled or if the current
+ * batch needs to be flushed.
+ */
+@InternalApi("For google-cloud-java client use only.")
+public interface BatchResource {
```
blakeli0, line 40: Do we have to create an interface for it? The default implementation looks like a POJO to me, I would prefer not to have a separate interface if we don't plan to have multiple implementations.

gax-java/gax/src/main/java/com/google/api/gax/batching/BatchingDescriptor.java
```diff
@@ -96,4 +96,24 @@ public interface BatchingDescriptor<ElementT, ElementResultT, RequestT, Response
...
+  /** Create an empty {@link BatchResource}. */
+  default BatchResource createEmptyResource() {
+    return DefaultBatchResource.create(0, 0);
+  }
+
+  /**
+   * Checks if the current {@link BatchResource} should be flushed based on the maxElementThreshold
+   * and maxBytesThreshold.
+   */
+  default boolean shouldFlush(
```
blakeli0, line 114: This method should be moved to `BatchResource`, it does not use anything from this class. Also please add some unit tests just for this method.

gax-java/gax/src/main/java/com/google/api/gax/batching/BatcherImpl.java
```diff
@@ -241,12 +251,14 @@ public ApiFuture<ElementResultT> add(ElementT element) {
 
     SettableApiFuture<ElementResultT> result = SettableApiFuture.create();
     synchronized (elementLock) {
-      currentOpenBatch.add(element, result, throttledTimeMs);
-    }
+      if (batchingDescriptor.shouldFlush(
+          currentOpenBatch.resource.add(newResource), elementThreshold, bytesThreshold)) {
```
blakeli0, line 255: This can be simplified to `resource.add(newResource).shouldFlush(elementThreshold, bytesThreshold))` once the `shouldFlush` method is moved to `BatchResource` or `DefaultBatchResource` if we decide to get rid of the interface.

gax-java/gax/src/main/java/com/google/api/gax/batching/BatchingDescriptor.java
```diff
@@ -96,4 +96,24 @@ public interface BatchingDescriptor<ElementT, ElementResultT, RequestT, Response
 
   /** Returns the size of the passed element object in bytes. */
   long countBytes(ElementT element);
+
+  /** Creates a new {@link BatchResource} with ElementT. */
+  default BatchResource createResource(ElementT element) {
```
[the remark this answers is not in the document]
mutianf: Yes, in bigtable client we want to override the BatchResource implementation, so we don't want to create BatchResource directly in the BatcherImpl code.

blakeli0 (approved):

PR #1804, opened 2023-06-23
chore: do not merge. showcase-clirr test PR

burkedavison:


burkedavison:
First commit tests failed because the showcase source was not updated to reflect the dependency change.

Good. This proves we don't need to perform generation during showcase-clirr.

burkedavison:
With showcase goldens updated, showcase-clirr now failing with binary incompatibilities.

Good.

burkedavison:
Testing complete.

PR #1805, opened 2023-06-23
ci: add showcase-clirr check

burkedavison:
This check compares the Showcase client against the **target branch** for binary incompatible changes.

Previous binary incompatibilities have been caught with the java-iam client, but:
* Showcase covers proto, grpc, and gapic artifacts. The java-iam client covers only proto and grpc.
* Showcase is intended to provide broad feature coverage, thus it will be a more robust binary compatibility check.
* Since this check compares against the target branch, it works with LTS branches.

See https://github.com/googleapis/sdk-platform-java/pull/1804 for an example of this check catching a binary incompatible change.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -229,12 +229,12 @@ jobs:
       - name: Checkout sdk-platform-java @ PR merge commit
         uses: actions/checkout@v3
         # Showcase golden test ensures that src changes are already reflected in the PR.
-      - name: Update Showcase version to 0.0.2-SNAPSHOT
+      - name: Clirr check
         working-directory: showcase
-        run: mvn versions:set -B -ntp -DnewVersion=0.0.2-SNAPSHOT
-      - name: Clirr Check
-        working-directory: showcase
-        run: mvn -B -ntp clirr:check -Dclirr.skip=false -DcomparisonVersion=0.0.1-SNAPSHOT
+        run: |
+          SHOWCASE_CLIENT_VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
+          mvn versions:set -B -ntp -DnewVersion=local
+          mvn clirr:check -B -ntp -Dclirr.skip=false -DcomparisonVersion=$SHOWCASE_CLIENT_VERSION
 
   gapic-generator-java-bom:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -207,6 +207,35 @@ jobs:
...
+          mvn install -B -ntp -T 1C -DskipTests
+      - name: Checkout sdk-platform-java @ PR merge commit
+        uses: actions/checkout@v3
+        # Showcase golden test ensures that src changes are already reflected in the PR.
+      - name: Update Showcase version to 0.0.2-SNAPSHOT
+        working-directory: showcase
+        run: mvn versions:set -B -ntp -DnewVersion=0.0.2-SNAPSHOT
+      - name: Clirr Check
+        working-directory: showcase
+        run: mvn -B -ntp clirr:check -Dclirr.skip=false -DcomparisonVersion=0.0.1-SNAPSHOT
```
meltsufin, line 237: Can we eliminate the assumption about the exact version numbers using something like
`mvn org.apache.maven.plugins:maven-help-plugin:2.1.1:evaluate -Dexpression=project.version`?

burkedavison: Done. Verified in #1804 that new logic continues to catch the binary incompatibility.

.github/workflows/ci.yaml
```diff
@@ -207,6 +207,35 @@ jobs:
...
+          mvn install -B -ntp -T 1C -DskipTests
+      - name: Checkout sdk-platform-java @ PR merge commit
+        uses: actions/checkout@v3
+        # Showcase golden test ensures that src changes are already reflected in the PR.
+      - name: Clirr check
+        working-directory: showcase
+        run: |
+          SHOWCASE_CLIENT_VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
+          mvn versions:set -B -ntp -DnewVersion=local
+          mvn clirr:check -B -ntp -Dclirr.skip=false -DcomparisonVersion=$SHOWCASE_CLIENT_VERSION
```
meltsufin, line 237: I though that `comparisonVersion` is the previous "released" version that we should be comparing against, but here it's the PR version. So, I'm a bit confused.

meltsufin: I guess it works as long as the project version in the PR is the same as the one in the target branch, which should almost always be the case.

burkedavison: Ah. Yes. You're right. I'll make sure it is always true... sec...

meltsufin (approved):

PR #1807, opened 2023-06-26
PoC: Otel metrics

blakeli0:


gax-java/gax/src/main/java/com/google/api/gax/tracing/ApiTracer.java
```diff
@@ -101,6 +101,11 @@ public interface ApiTracer {
   /** Adds an annotation that the attempt succeeded. */
   void attemptSucceeded();
 
+  //This is for libraries to override to intended name
+  default String attemptLatencyName() {
```
mutianf, line 105: If we're going to provide a default implementation, I wonder instead of override the attempt name, maybe we can provide a general class, something like `ApiTracerInstrument` and provide functions like apiTracerInstrument.record(latency, attributes). So client libraries can define their own meter definition. Similar to https://github.com/googleapis/java-bigtable/pull/1796/files#diff-f2ab39ff606949b18fa0ea0a25b6149352064c8b1f775a78cb79eb8acfddde6f `BigtableInsturment` and `BuiltinInstruments`

gax-java/gax/src/main/java/com/google/api/gax/tracing/OpenTelemetryTracerFactory.java
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * A {@link ApiTracerFactory} to build instances of {@link OpencensusTracer}.
+ *
+ * <p>This class wraps the {@link Tracer} provided by Opencensus in {@code Tracing.getTracer()}. It
+ * will be used to create new spans and wrap them in {@link OpencensusTracer} defined in gax.
+ *
+ * <p>This class is thread safe.
+ */
+@InternalApi("For google-cloud-java client use only")
+public class OpenTelemetryTracerFactory extends BaseApiTracerFactory {
```
mutianf, line 46: FYI, we can probably clean up BaseApiTracer and BaseApiTracerFactory, it was created in java7 days for backward compatibility: https://github.com/googleapis/sdk-platform-java/commit/b810cc90249ae92831fc5e9f1fe7ab316be63391

gax-java/gax/src/main/java/com/google/api/gax/tracing/ApiTracer.java
blakeli0: I can see the need of another abstraction layer for customizing some attributes(like extract this method to something like `MetricNames`), however, I think most of things, like `description` and `unit` should be the same for all services, so that we can have a consistent experience for all services. Let me know if I'm missing something.

PR #1812, opened 2023-06-26
fix: generate native image configurations for Gapics

mpeddada1:
Notable Changes:

- This PR leverages the Writer class to parse necessary information such as package name and class name from the collection of GapicClass objects that are returned by the Composer. It then uses that information to generated a `reflect-config.json` file. 
- The drawback of this change is that it only includes a subset of classes, particularly, the ones generated by the gapic generator. We need to also include the classes that are generated by the standard `java_proto_library` bazel rule.

### DIFF
--- a/gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
+++ b/gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
@@ -202,15 +202,6 @@ public class Parser {
       }
     }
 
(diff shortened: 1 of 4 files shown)


burkedavison (changes requested):
Like the other PR, (if we want to continue considering this option) let's please update this PR to:

- [ ] Apply to showcase
- [ ] Invoke the showcase native image tests
- [ ] Remove the RunReachabilityHandlersConcurrently argument
- [ ] Remove the Feature classes using reachability handlers

PR #1831, opened 2023-07-06
feat: generate native image metadata during tar assembly

mpeddada1:
This is an alternate approach to https://github.com/googleapis/sdk-platform-java/pull/1812/. Instead of editing the Writer class, we are generating the reflect-config.json file at a later stage, i.e when the library tar is being assembled. 

In this implementation:

The `_gapic_pkg_tar_impl` implementation package tar.gz files for proto-google-*, gapic-google-* and grpc-google-*. Before the compressed tarballs for each of these directories are created, we implement some logic to 
1) Iterate through all the java files in these directories.
2) Derive the class name and package name for each .java file to build a string of the format, `com.google.cloud.speech.v2.SpeechClient` 
3) Add the string to reflect-config.json along with a set of fixed parameters in the format that is recognizable to the native image compiler:
```
[ 
 {
    "name": "com.google.cloud.speech.v2.SpeechSettings",
    "queryAllMethods": true,
    "queryAllPublicConstructors": true,
    "queryAllDeclaredMethods": true,
    "queryAllPublicMethods": true,
    "allDeclaredClasses": true,
    "allPublicClasses": true
  }
...
]
```

**Local Testing**
- Follow instructions in the [development](https://github.com/googleapis/sdk-platform-java/blob/main/gapic-generator-java/DEVELOPMENT.md#running-the-plugin-under-googleapis-with-local-gapic-generator-java) guide to use the SNAPSHOT version of the gapic-generator-java.
- At the root of the googleapis repo, run `bazel build //google/cloud/speech/v2:google-cloud-speech-v2-java` to generate an compressed tar for the desired. client library. 

** TODOs**
- Add logic to parse the nested classes.


rules_java_gapic/java_gapic_pkg.bzl
```diff
@@ -86,18 +86,34 @@ def _gapic_pkg_tar_impl(ctx):
...
     done
-
     mkdir -p {package_dir_path}
     for dep in {deps}; do
         tar -xzpf $dep -C {package_dir_path}
     done
-    cd {package_dir_path}/{tar_cd_suffix}
 
+    #TODO: Use awk to extract classes inside java class. Such as: awk '/public class/' Classes.java
+    if [[ {package_dir_path} == *proto-google-cloud-* ]] || [[ {package_dir_path} == *gapic-google-cloud-* ]] || [[ {package_dir_path} == *grpc-google-cloud-* ]]; then
```
burkedavison, line 95: Let's update this PR to:

- [ ] Also apply to Showcase client generation
- [ ] Invoke the showcase native image tests (free if rebasing on #1833)
- [ ] Remove the RunReachabilityHandlersConcurrently argument
- [ ] Remove the Feature classes using reachability handlers

mpeddada1:
Closing in favor of https://github.com/googleapis/sdk-platform-java/pull/1841

PR #1841, opened 2023-07-10
chore: convert to static reflect-config.json using client generator

burkedavison:
Work in progress.

### DIFF
--- a/gapic-generator-java/src/main/java/com/google/api/generator/gapic/Generator.java
+++ b/gapic-generator-java/src/main/java/com/google/api/generator/gapic/Generator.java
@@ -14,6 +14,7 @@
 
 package com.google.api.generator.gapic;
 
+import com.google.api.generator.gapic.composer.ClientLibraryReflectConfigComposer.ReflectConfig;
 import com.google.api.generator.gapic.composer.Composer;
 import com.google.api.generator.gapic.model.GapicClass;
 import com.google.api.generator.gapic.model.GapicContext;
@@ -29,7 +30,8 @@ public class Generator {
     GapicContext context = Parser.parse(request);
     List<GapicClass> clazzes = Composer.composeServiceClasses(context);
     GapicPackageInfo packageInfo = Composer.composePackageInfo(context);
+    List<ReflectConfig> reflectConfig = Composer.composeNativeReflectConfig(context, packageInfo);
     String outputFilename = "temp-codegen.srcjar";
-    return Writer.write(context, clazzes, packageInfo, outputFilename);
+    return Writer.write(context, clazzes, packageInfo, reflectConfig, outputFilename);
   }
 }
--- a/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/ClientLibraryReflectConfigComposer.java
+++ b/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/ClientLibraryReflectConfigComposer.java
@@ -1,0 +1,56 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.api.generator.gapic.composer;
+
+import com.google.api.generator.gapic.model.GapicContext;
+import com.google.api.generator.gapic.model.GapicPackageInfo;
+import com.google.api.generator.gapic.model.Message;
+import com.google.common.collect.ImmutableList;
+import java.util.ArrayList;
+import java.util.List;
+
+public class ClientLibraryReflectConfigComposer {
+
+  public static List<ReflectConfig> generateReflectConfig(
+      GapicContext context, GapicPackageInfo packageInfo) {
+    List<String> list = new ArrayList<>();
+    context
+        .messages()
+        .forEach(
+            (String id, Message message) -> {
+              // if (id.startsWith(packageInfo.packageInfo().pakkage())) {
+              list.add(id);
+              if (!message.isEnum()) {
+                list.add(id + "$Builder");
+              }
+              // }
+            });
+    return list.stream().sorted().map(ReflectConfig::new).collect(ImmutableList.toImmutableList());
+  }
+
+  public static class ReflectConfig {
+    String name;
+    boolean queryAllDeclaredConstructors = true;
+    boolean queryAllPublicConstructors = true;
+    boolean queryAllDeclaredMethods = true;
+    boolean queryAllPublicMethods = true;
+    boolean allDeclaredClasses = true;
+    boolean allPublicClasses = true;
+
+    ReflectConfig(String name) {
+      this.name = name;
+    }
+  }
+}
--- a/gax-java/gax-grpc/src/main/resources/META-INF/native-image/com.google.api/gax-grpc/native-image.properties
+++ b/gax-java/gax-grpc/src/main/resources/META-INF/native-image/com.google.api/gax-grpc/native-image.properties
@@ -11,6 +11,4 @@ Args=--add-opens=java.base/java.time=ALL-UNNAMED \
     io.grpc.netty.shaded.io.netty.handler.ssl,\
     io.grpc.internal.RetriableStream,\
     io.grpc.netty.shaded.io.netty \
-  --features=com.google.api.gax.grpc.nativeimage.ProtobufMessageFeature,\
-  com.google.api.gax.grpc.nativeimage.GrpcNettyFeature \
-  -H:-RunReachabilityHandlersConcurrently
+  --features=com.google.api.gax.grpc.nativeimage.GrpcNettyFeature
--- a/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/Composer.java
+++ b/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/Composer.java
@@ -16,6 +16,7 @@ package com.google.api.generator.gapic.composer;
 
 import com.google.api.generator.engine.ast.ClassDefinition;
 import com.google.api.generator.engine.ast.CommentStatement;
+import com.google.api.generator.gapic.composer.ClientLibraryReflectConfigComposer.ReflectConfig;
 import com.google.api.generator.gapic.composer.comment.CommentComposer;
 import com.google.api.generator.gapic.composer.grpc.GrpcServiceCallableFactoryClassComposer;
 import com.google.api.generator.gapic.composer.grpc.GrpcServiceStubClassComposer;
@@ -53,6 +54,11 @@ public class Composer {
 
   public static GapicPackageInfo composePackageInfo(GapicContext context) {
     return addApacheLicense(ClientLibraryPackageInfoComposer.generatePackageInfo(context));
+  }
+
+  public static List<ReflectConfig> composeNativeReflectConfig(
+      GapicContext context, GapicPackageInfo packageInfo) {
+    return ClientLibraryReflectConfigComposer.generateReflectConfig(context, packageInfo);
   }
 
   public static List<GapicClass> generateServiceClasses(GapicContext context) {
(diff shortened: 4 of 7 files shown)


gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -598,9 +600,16 @@ private static Map<String, Message> parseMessages(
         }
       }
     }
+
+    ImmutableList<String> nestedEnums =
+        messageDescriptor.getEnumTypes().stream()
```
blakeli0, line 605: I guess `getEnumTypes()` is empty if a message does not have any nested enums?

burkedavison: Correct. Most messages don't have any nested enums.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/ClientLibraryReflectConfigComposer.java
```diff
@@ -0,0 +1,82 @@
...
+  /** List all classes in the message that should have a reflect-config entry */
+  private static List<String> parseReflectConfig(String id, Message message) {
+    final String name = formatNestedClasses(id);
+
+    List<String> list = new ArrayList<>();
+    list.add(name);
+    for (String nestedEnum : message.nestedEnums()) {
+      list.add(name + "$" + nestedEnum);
+    }
+    if (!message.isEnum()) {
```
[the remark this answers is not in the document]
burkedavison: First-class Enum classes like `CVSSVersion` are represented in the `GapicContext.messages()` as a `Message` instance that [has `enumValues`](https://github.com/googleapis/sdk-platform-java/blob/main/gapic-generator-java/src/main/java/com/google/api/generator/gapic/model/Message.java#L79-L81).

This is discussed in the [Message POJO](https://github.com/googleapis/sdk-platform-java/blob/main/gapic-generator-java/src/main/java/com/google/api/generator/gapic/model/Message.java#L50-L54) and in the [Parser.parseMessages implementation](https://github.com/googleapis/sdk-platform-java/blob/main/gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java#L527).

We can also see this in this PR's reflect-config.json golden file(s), for example: [Severity is an enum](https://github.com/googleapis/sdk-platform-java/blob/main/showcase/proto-gapic-showcase-v1beta1/src/main/java/com/google/showcase/v1beta1/Severity.java), and has a [corresponding reflect-config entry](https://github.com/googleapis/sdk-platform-java/pull/1841/files#diff-5f61fc0dc2f3b6f3bce24db46adeadd86f06793146c8a7a49e1f38f72203677cR3054-R3062).

burkedavison:
Latest commit intended to break the golden tests, and did.
Good. 
Fixing now.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/ClientLibraryReflectConfigComposer.java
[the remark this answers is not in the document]
mpeddada1: Nice! Thank you for the explanation.

gax-java/gax-grpc/src/main/resources/META-INF/native-image/com.google.api/gax-grpc/native-image.properties
```diff
@@ -13,4 +13,4 @@ Args=--add-opens=java.base/java.time=ALL-UNNAMED \
     io.grpc.netty.shaded.io.netty \
   --features=com.google.api.gax.grpc.nativeimage.ProtobufMessageFeature,\
   com.google.api.gax.grpc.nativeimage.GrpcNettyFeature \
-  -H:-RunReachabilityHandlersConcurrently
\ No newline at end of file
+  -H:-RunReachabilityHandlersConcurrently
```
burkedavison: `ProtobufMessageFeature` and `-H:-RunReachabilityHandlersConcurrently` have now been reintroduced.

Implementation plan:
1. Merge and release with this PR's changes. 
2. Wait for Owlbot to use the updated gapic-generator-java, and generate the reflect-config.json for each client libary.
3. In the next release cycle: Remove `ProtobufMessageFeature` and `-H:-RunReachabilityHandlersConcurrently`, perform downstream testing on the client libraries that will then have local reflect-configs.

meltsufin (approved):
Please add an issue to track removal of the feature in the next release cycle, if you haven't already.

burkedavison:
#1844 created and CC'd release managers @JoeWang1127 and @blakeli0 for visibility.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/ClientLibraryReflectConfigComposer.java
```diff
@@ -0,0 +1,92 @@
...
+        .distinct()
+        .sorted()
+        .map(ReflectConfig::new)
+        .collect(ImmutableList.toImmutableList());
+  }
+
+  /** List all classes in the message that should have a reflect-config entry */
+  @VisibleForTesting
+  static List<String> calculateReflectConfigList(String id, Message message) {
+    final String name = formatNestedClasses(id);
```
blakeli0, line 40: Can we add some comment regarding why the format has to be this way? Same thing for the nested enums format below. Maybe a link to a GraalVM doc?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/ClientLibraryReflectConfigComposer.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.List;
+
+public class ClientLibraryReflectConfigComposer {
+  private ClientLibraryReflectConfigComposer() {} // Disallow instance creation
+
+  public static List<ReflectConfig> generateReflectConfig(GapicContext context) {
+    List<String> allConfigs = new ArrayList<>();
+    context.messages().forEach((id, msg) -> allConfigs.addAll(calculateReflectConfigList(id, msg)));
+    return allConfigs.stream()
+        .distinct()
```
blakeli0, line 31: Could there be duplicated configs?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/ClientLibraryReflectConfigComposer.java
```diff
@@ -0,0 +1,92 @@
...
+      }
+      result.append(s);
+      if (Character.isUpperCase(s.charAt(0))) {
+        isNested = true;
+      }
+    }
+    return result.toString();
+  }
+
+  public static class ReflectConfig {
```
blakeli0, line 73: I know this class is not a typical model class and only used for serializing to JSON, but it still is worth its own file in my opinion.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/Generator.java
```diff
@@ -29,7 +30,8 @@ public static CodeGeneratorResponse generateGapic(CodeGeneratorRequest request)
     GapicContext context = Parser.parse(request);
     List<GapicClass> clazzes = Composer.composeServiceClasses(context);
     GapicPackageInfo packageInfo = Composer.composePackageInfo(context);
+    List<ReflectConfig> reflectConfig = Composer.composeNativeReflectConfig(context);
```
blakeli0, line 33: Are we always going to generate the reflect configs for every client library? Do we want to consider a flag to turn it off if needed? Maybe it's a requirement but I'm not sure if every client library need GraalVM support, also the file could be huge for some client libraries.

blakeli0 (approved):

mpeddada1 (approved):
This looks great. 
Pending one minor comment to add the TODO (discussed offline) about the structure of the META-INF/native/image directory. Not an issue rn as we are only expecting one reflect-config.json per library but in the future we may need to enhance this logic a bit to follow convention (https://www.graalvm.org/22.1/reference-manual/native-image/BuildConfiguration/#embedding-a-configuration-file) so that the configs don't get overridden if we have have two files that have the same file path.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/Generator.java
burkedavison: CC @meltsufin and @mpeddada1 for additional opinions and visibility.

> Are we always going to generate the reflect configs for every client library? Do we want to consider a flag to turn it off if needed? Maybe it's a requirement but I'm not sure if every client library need GraalVM support...

Providing the reflect-config as no impact on non-native builds, so the plan is to generate it for every client library. I'd hesitate to create a flag for disabling the generation until we have a use-case for it.

> ...also the file could be huge for some client libraries.

I don't think size should be a concern yet. Compute has one of the largest set of protos (just based off googleapis *.proto file sizes), and its reflect-config.json (in this PR's /test goldens) is 43KB -- and there's so much repeated text that it's going to compress in the jar really well... a simple zip takes it down to 1KB.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/ClientLibraryReflectConfigComposer.java
burkedavison: Yes -- I'll add some comments and make the wording more precise than it is right now.

I've also just verified that this format is required. Without it, native-image provides 50ish warnings that each look like this:
```
Warning: Could not resolve com.google.protobuf.DescriptorProtos.FieldDescriptorProto for reflection configuration. Reason: java.lang.ClassNotFoundException: com.google.protobuf.DescriptorProtos.FieldDescriptorProto.
...
Warning: Could not resolve com.google.showcase.v1beta1.Sequence.Response for reflection configuration. Reason: java.lang.ClassNotFoundException: com.google.showcase.v1beta1.Sequence.Response.
```

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/ClientLibraryReflectConfigComposer.java
burkedavison: I don't know for sure. It depends on the external input(s) we get from protoc; so rather than be concerned about it, I just threw in the `distinct()` to make it a non-issue.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/ClientLibraryReflectConfigComposer.java
burkedavison: Comment added. Unable to find this requirement explicitly in the GraalVM docs, only via behavior.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/model/Message.java
```diff
@@ -74,6 +74,8 @@ public abstract class Message {
   // [Foo, Bar, Car].
   public abstract ImmutableList<String> outerNestedTypes();
 
+  public abstract ImmutableList<String> nestedEnums();
```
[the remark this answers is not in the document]
burkedavison: Clarifying documentation added to all *enum* methods in Message.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/ClientLibraryReflectConfigComposer.java
burkedavison: Done: Extracted class to model package.

burkedavison:
> @mpeddada1 : This looks great. Pending one minor comment...

Done.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/ClientLibraryReflectConfigComposer.java
blakeli0: I might be overthinking about this but my only concern is that we could hide real problems. For example, what does it mean if there are duplicates? Are they legitimate? If yes, is this solution gonna work? If no, is it a proto issue? Maybe we should throw exceptions if there are duplicates found?

blakeli0 (commented):
LGTM.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/ClientLibraryReflectConfigComposer.java
burkedavison: Verified that duplicate entries in the reflect-config.json doesn't cause any issues -- but these are great questions.

On looking at it further:
* We are given as an input the _fully-qualified class names_ (FQN) of the messages and enums that have been parsed from [TypeParser.parseMessageReference](https://github.com/googleapis/sdk-platform-java/blob/main/gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/TypeParser.java#L161-L171) and [TypeParser.parseEnumReference](https://github.com/googleapis/sdk-platform-java/blob/main/gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/TypeParser.java#L216-L233).
* Even if the above logic accidentally produced duplicate names, we'd only be given one of them because of the Map input type.
* So at the point of this `distinct()` call if we have duplicates, it's because the logic invoked from this function caused them.

That should definitely not be possible, and if it happened there'd be at least one class that we were accidentally excluding from the configuration.

I'm adding a runtime check and removing the `distinct()` call.

PR #1854, opened 2023-07-17
Resource name deduplication

diegomarquezp:
Adds deduplication logic for `com.google.api.resourcenames.ResourceName` implementing classes. It is possible that a generated class is also named `ResourceName`, which is not currently handled by the generator.

This approach consists of:
* modifying `TypeStore` to add a special getter for `TypeNode`s with references that will `useFullName`
* modifying `ResourceNameClassComposer` to use the full name of any implemented classes if there is a name collision with the class being generated


gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/resourcename/ResourceNameHelperClassComposer.java
```diff
@@ -160,8 +160,26 @@ private static List<AnnotationNode> createClassAnnotations() {
...
+   * @param implementingClassName class that is implementing the resulting list. Collisions are
+   *     checked
+   * @return
+   */
+  private static List<TypeNode> createImplementsTypes(String implementingClassName) {
+    List<TypeNode> preprocessed = Arrays.asList(FIXED_TYPESTORE.get("ResourceName"));
+    List<TypeNode> result = new ArrayList<>();
+    for (TypeNode typeNode : preprocessed) {
+      if (!typeNode.isPrimitiveType()
+          && typeNode.reference().name().compareTo(implementingClassName) == 0) {
```
blakeli0, line 176: I think `typeNode.reference().name().equals(implementingClassName)` does the same thing?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/store/TypeStore.java
```diff
@@ -33,18 +36,42 @@ public TypeStore(List<Class<?>> concreteClasses) {
...
   public TypeNode get(String typeName) {
     return store.get(typeName);
   }
 
+  public TypeNode getWithFullName(String typeName) {
+    TypeNode stored = store.get(typeName);
+    Preconditions.checkNotNull(stored);
+    Reference ref;
+    if (stored.reference() instanceof ConcreteReference) {
+      ref =
```
blakeli0, line 54: I'm afraid we might be losing information here, I see that `ConcreteReference` has a lot more fields other than `clazz` and `useFullName`, same thing for `VaporReference`.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/resourcename/ResourceNameHelperClassComposer.java
```diff
@@ -160,8 +160,26 @@ private static List<AnnotationNode> createClassAnnotations() {
...
+  /**
+   * Builds a list of types to be implemented. If the type has the same name then we use its full
+   * name
+   *
+   * @param implementingClassName class that is implementing the resulting list. Collisions are
+   *     checked
+   * @return
+   */
+  private static List<TypeNode> createImplementsTypes(String implementingClassName) {
+    List<TypeNode> preprocessed = Arrays.asList(FIXED_TYPESTORE.get("ResourceName"));
```
blakeli0, line 172: `FIXED_TYPESTORE` always returns only one result. Can we check if there is conflict and fix it first, then wrap it as a list in the end?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/resourcename/ResourceNameHelperClassComposer.java
```diff
@@ -160,8 +160,26 @@ private static List<AnnotationNode> createClassAnnotations() {
...
+    List<TypeNode> result = new ArrayList<>();
+    for (TypeNode typeNode : preprocessed) {
+      if (!typeNode.isPrimitiveType()
+          && typeNode.reference().name().compareTo(implementingClassName) == 0) {
+        result.add(FIXED_TYPESTORE.getWithFullName(typeNode.reference().name()));
```
blakeli0, line 177: I know you may want to create a genetic `getWithFullName` method so that it could helpful in the future, but since this is a very specific issue, and we already hardcode the String `ResourceName` in the beginning of the method, I think it might be easier to just fix this issue.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/resourcename/ResourceNameHelperClassComposer.java
diegomarquezp: I switched to this comparison. It is better.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/store/TypeStore.java
diegomarquezp: I moved the logic to `createImplementsTypes` by using each one of the setters in the builder

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/resourcename/ResourceNameHelperClassComposer.java
diegomarquezp: Makes sense. I just moved the deduplication logic into `createImplementsTypes`

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/resourcename/ResourceNameHelperClassComposer.java
diegomarquezp: I used a slightly different but similarly concise way (wrap both possible results in the end). Hope you agree it's good too.

blakeli0 (commented):
Thanks Diego! Looking good to me other than one nit. We should update the description as it does not reflect the latest changes. Also, this PR should probably be a `fix` instead of `feat`.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/resourcename/ResourceNameHelperClassComposer.java
diegomarquezp: Ah I only saw the review comment instead of this one. I'll leave it as a quick follow up

PR #1861, opened 2023-07-18
chore: include src/main/resources in gapic output

burkedavison:


blakeli0 (approved):
LGTM. It would be great if we have a way to test it but it can be a separate PR. For example, we do have a [CI step](https://github.com/googleapis/sdk-platform-java/blob/main/.github/workflows/ci.yaml#L109-L127) to generate the client libraries from googleapis, but it probably need additional configuration to test GraalVM, it may not be feasible either. We should also take this into hermetic build's consideration.

rules_java_gapic/java_gapic_pkg.bzl
```diff
@@ -274,6 +274,14 @@ def _java_gapic_srcs_pkg_impl(ctx):
         unzip -q -o $src -d {package_dir_path}/src/main/java
         rm -r -f {package_dir_path}/src/main/java/META-INF
 
+        # GAPIC libraries generate java/ and resources/ folders, unlike grpc and proto.
+        # If this is a GAPIC library, move its output up one directory.
+        if [ -d {package_dir_path}/src/main/java/java ]; then
```
blakeli0, line 279: `{package_dir_path}/src/main/java/java` does not look right to me. Is it meant to be `{package_dir_path}/src/main/java`?

burkedavison: It is correct - hence the comment trying to clarify what's going on. 

This logic is shared between gapic, grpc, and proto outputs. Due to the one line change in `java-gapic.bzl`, the gapic output is now effectively zipping the 'main' folder, while the grpc and proto outputs (not under our direct control with these rules) continue to effectively zip the 'java' folder.

So to avoid major refactoring, the simplest solution I found was:
> Keep the logic consistent to unzip into the java directory. However, if we detect that what we've unzipped is actually the main folder of a gapic library, then move everything up one directory.

blakeli0 (approved):

suztomo:
@burkedavison I forgot to share useful Bazel debug options yesterday: "--subcommands" and "--sandbox_debug". They are our friends.

burkedavison:
> @burkedavison I forgot to share useful Bazel debug options yesterday: "--subcommands" and "--sandbox_debug". They are our friends.

😄 Yes they are! Thankfully, I have been using them or this might've taken much longer.

PR #1863, opened 2023-07-18
docs: Update RetrySettings javadocs to include polling

### LINKED ISSUE #2214: Gax RetrySettings' documentation should include polling information
Update the existing RetrySettings' javadoc to use references to including polling, default values, and where to look to find custom configurations.

lqiu96:

Fixes #2214


gax-java/gax/src/main/java/com/google/api/gax/retrying/RetrySettings.java
```diff
@@ -74,36 +78,53 @@ public abstract class RetrySettings implements Serializable {
...
 
   /**
-   * InitialRetryDelay controls the delay before the first retry. Subsequent retries will use this
-   * value adjusted according to the RetryDelayMultiplier. The default value is {@code
+   * InitialRetryDelay controls the delay before the first retry/ poll. Subsequent retries and polls
+   * will use this value adjusted according to the RetryDelayMultiplier. The default value is {@code
    * Duration.ZERO}.
+   *
+   * <p>If there are no configurations, Retries have the default initial retry delay value set above
+   * and LROs have a default initial poll delay value of {@code Duration.ofMillis(5000)} (5
```
blakeli0, line 95: Where is the default number coming from? Can we link to where the default number is actually specified? Same comment for all the default numbers below.

gax-java/gax/src/main/java/com/google/api/gax/retrying/RetrySettings.java
```diff
@@ -74,36 +78,53 @@ public abstract class RetrySettings implements Serializable {
...
+   * <p>If there are no configurations, Retries have the default timeout value set above and LROs
+   * have a default total timeout value of {@code Duration.ofMillis(300000)} (5 minutes).
    */
   public abstract Duration getTotalTimeout();
 
   /**
-   * InitialRetryDelay controls the delay before the first retry. Subsequent retries will use this
-   * value adjusted according to the RetryDelayMultiplier. The default value is {@code
+   * InitialRetryDelay controls the delay before the first retry/ poll. Subsequent retries and polls
+   * will use this value adjusted according to the RetryDelayMultiplier. The default value is {@code
```
blakeli0, line 91: Is this comment `The default value is {@code * Duration.ZERO}.` still valid with your new comment regarding default values?

gax-java/gax/src/main/java/com/google/api/gax/retrying/RetrySettings.java
```diff
@@ -166,36 +197,53 @@ public abstract static class Builder {
...
+     *
+     * <p>If there are no configurations, Retries have the default max retry delay value set above
+     * and LROs have a default max poll retry delay value of {@code Duration.ofMillis(45000)} (45
+     * seconds).
      */
     public abstract Builder setMaxRetryDelay(Duration maxDelay);
 
     /**
      * MaxAttempts defines the maximum number of attempts to perform. The default value is {@code
      * 0}. If this value is greater than 0, and the number of attempts reaches this limit, the logic
```
blakeli0, line 242: I remember that both 0 and 1 mean there is no retry from a YAQS. It's kind of already implied as `maxAttempts` should include the first attempt, can you please make it clear in the comment here to reduce similar confusion in the future?

gax-java/gax/src/main/java/com/google/api/gax/retrying/RetrySettings.java
lqiu96: I have a reference here: https://github.com/googleapis/sdk-platform-java/blob/14973463983495c7eb386a0edfabfcac3fcf9359/gax-java/gax/src/main/java/com/google/api/gax/retrying/RetrySettings.java#L70-L72, but it's at the top of the file and may get lost. The default value retry values are populated from the grpc config file and I don't want to link those in here. LROs are preconfigured from the generation.

gax-java/gax/src/main/java/com/google/api/gax/retrying/RetrySettings.java
lqiu96: Updated and hopefully it's clearer what the default values are for retry and polling operations.

gax-java/gax/src/main/java/com/google/api/gax/retrying/RetrySettings.java
lqiu96: Updated. Let me know if this makes more sense.

gax-java/gax/src/main/java/com/google/api/gax/retrying/RetrySettings.java
blakeli0: What are the default numbers if there is no gRPC config file? I guess the numbers are coming from constants in GAX ? The concern I have is if we change the default numbers in the future, we would have to change the numbers here as well, and it's very easy to forget to change here.

gax-java/gax/src/main/java/com/google/api/gax/retrying/RetrySettings.java
blakeli0: Perfect! Now I see why we have two default numbers, one for Retry and one for LRO

gax-java/gax/src/main/java/com/google/api/gax/retrying/RetrySettings.java
blakeli0: `The first attempt will be considered attempt #0.` I think this is true if `maxAttempts` is set to 0, but `The first attempt will be considered attempt #1` if `maxAttempts` is set to 1?

gax-java/gax/src/main/java/com/google/api/gax/retrying/RetrySettings.java
lqiu96: If there is nothing configured inside the grpc config file, the generator will create a RetrySettings with the default values with values listed above (For example: https://github.com/googleapis/google-cloud-java/blob/8452ac3a8eefa8a9e49ff83187b33be26efe6a01/java-batch/google-cloud-batch/src/main/java/com/google/cloud/batch/v1/stub/BatchServiceStubSettings.java#L518). This is what I'll call the RetrySettings default values.

Service teams can modify the grpc config file and that will provide default values for certain callables. These default values is what I'll call the callable default values. This is the RetrySettings which will be configured like: https://github.com/googleapis/google-cloud-java/blob/8452ac3a8eefa8a9e49ff83187b33be26efe6a01/java-batch/google-cloud-batch/src/main/java/com/google/cloud/batch/v1/stub/BatchServiceStubSettings.java#L507-L516

I believe they're both technically default values. The Javadoc I updated only refers the RetrySettings default values since the callable default values differ per service. I can add this explanation in so it's clearer to the readers.

gax-java/gax/src/main/java/com/google/api/gax/retrying/RetrySettings.java
lqiu96: The first attempt will be set as Attempt #0 (https://github.com/googleapis/sdk-platform-java/blob/80f23ec98922bf8f29f3f6960fb20fbeb6231706/gax-java/gax/src/main/java/com/google/api/gax/retrying/ExponentialRetryAlgorithm.java#L104). Every retryAttempt will create the `nextAttempt` class and update the attemptCount: https://github.com/googleapis/sdk-platform-java/blob/80f23ec98922bf8f29f3f6960fb20fbeb6231706/gax-java/gax/src/main/java/com/google/api/gax/retrying/ExponentialRetryAlgorithm.java#L164 and the shouldRetry function inside the algorithm will check if it's allowed. In this case, the first _retry_ attempt will be attempt #1 and that will pass this check: https://github.com/googleapis/sdk-platform-java/blob/80f23ec98922bf8f29f3f6960fb20fbeb6231706/gax-java/gax/src/main/java/com/google/api/gax/retrying/ExponentialRetryAlgorithm.java#L224-L226.

I think it would make sense if I specify that maxAttempts refers to the number of retry attempts instead.

gax-java/gax/src/main/java/com/google/api/gax/retrying/RetrySettings.java
blakeli0: The default numbers in [RetrySettingsComposer](https://github.com/googleapis/sdk-platform-java/blob/main/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/RetrySettingsComposer.java#L69-L73) is what I was referring to. If those numbers change, we would have to update the numbers here as well. But since those numbers are in the generator, I don't think it's a good idea to expose the implementation details to Gax either, I think we can leave them as they are for now.

blakeli0 (approved):

gax-java/gax/src/main/java/com/google/api/gax/retrying/RetrySettings.java
lqiu96: > The default numbers in [RetrySettingsComposer](https://github.com/googleapis/sdk-platform-java/blob/main/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/RetrySettingsComposer.java#L69-L73) is what I was referring to. If those numbers change, we would have to update the numbers here as well. But since those numbers are in the generator, I don't think it's a good idea to expose the implementation details to Gax either, I think we can leave them as they are for now.

Yeah, those values are set in the generator (and only the generator sets defaults for LROs). I think if we make an effort to change those, we'll have to remember to update our docs here manually. I'd like for an automated way to do that, but linking to the generator code doesn't seem ideal.

I'll keep as is for now then.

PR #1866, opened 2023-07-19
chore: gax-java now uses third-party-dependencies' management

burkedavison:
NEEDS_LGTM=@suztomo

### DIFF
--- a/gax-java/pom.xml
+++ b/gax-java/pom.xml
@@ -48,6 +48,13 @@
   <dependencyManagement>
     <dependencies>
       <dependency>
+        <groupId>com.google.cloud</groupId>
+        <artifactId>third-party-dependencies</artifactId>
+        <version>3.13.1-SNAPSHOT</version><!-- {x-version-update:google-cloud-shared-dependencies:current} -->
+        <type>pom</type>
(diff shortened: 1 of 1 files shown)


blakeli0:
So the dependency tree is now
`first-party-dependencies` -> `gapic-generator-java-bom` -> `gax` -> `third-party-dependencies`
but `shared-dependencies-bom` is dependent on both `first-party-dependencies` and `third-party-dependencies` in the mean time, it works but not sure it's the best way. Theoretically we could get rid of `third-party-dependencies` and list all of them in gax.

burkedavison:
@blakeli0 : Correct. Visually: 
image

suztomo:
> Theoretically we could get rid of third-party-dependencies and list all of them in gax.


@burkedavison @blakeli0 The dependency tree you have in mind needs to distinguish BOM and artifacts. Third-party-dependencies is a BOM and gax is not a BOM.


The change looks good to me. No circular dependency. (Even with a circular dependency, Maven may work)

suztomo (approved):

blakeli0 (approved):

PR #1872, opened 2023-07-20
feat: Introduce java.time to Gax-Java

diegomarquezp:
This introduces the use of `java.time` in replacement of `org.threeten` backport classes. 

### Message to library users (release note)
We have made the following backwards-compatible changes in this library:
 * We have changed the method return type and argument type in __internal__ classes from `org.threeten.bp` to `java.time`. This has no effect in the API surface.
 * We have added `java.time` versions for each of the abstract methods involving `org.threeten.bp`. The affected classes are `BatchingSettings`, `RetrySettings`, `TimedAttemptSettings` and `ClientContext`, all of which are `AutoValue` classes, not meant to be extended by our users.
 * All getters and setters using `org.threeten.bp` will be marked final because we will force their behavior to be the same one as the `java.time` counterpart, which has the actual implementation. The affected classes are `BatchingSettings`, `RetrySettings`, `TimedAttemptSettings` and `ClientContext`, all of which are `AutoValue` classes, not meant to be extended by our users.
 * In three `interface`s, we have added `java.time` versions for each of the methods involving `org.threeten.bp`. The affected interfaces are `WatchdogProvider` (public interface), `ApiTracer` (internal) and `ApiCallContext` (internal). The new method in `WatchdogProvider` has a `default` behavior, which is to call the `threeten` counterpart, in order to make this a backwards-compatible change.

## Implementation
It preserves method signatures using `threeten`. The cases are:
### `@AutoValue` clases and classes with public interface involving time objects
Autovalues will have an internal `threeten` variable with a convenience `java.time` wrapper method. For example: 
https://github.com/googleapis/sdk-platform-java/blob/667f2da15d0800d87d0866e36627ac23955c0b01/gax-java/gax/src/main/java/com/google/api/gax/retrying/RetrySettings.java#L78-L89

and 

https://github.com/googleapis/sdk-platform-java/blob/a99d3c1acd4b25a28bd571c8c6f7f7f5d74e1365/gax-java/gax/src/main/java/com/google/api/gax/retrying/RetrySettings.java#L243-L262

### Internal variables
`Duration` or `Instant` private variables are switched to `java.time`. In this case, the wrappers are of `threeten` clasess. For example
https://github.com/googleapis/sdk-platform-java/blob/667f2da15d0800d87d0866e36627ac23955c0b01/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonCallContext.java#L237-L247

## Deprecation warning
[...]

PR #1887, opened 2023-07-27
feat: Introduce @ObsoleteApi

blakeli0:
This PR introduces `@ObsoleteApi`. Deprecating any client side feature including adding `@Deprecated` annotation is considered a breaking change based on the new support policy, hence we need an alternative way to signal us and customers that an API is deprecating. This annotation is not a direct replacement of `@Deprecated`, it's more like a pre-deprecation annotation, it has two major purposes:

1. It provides a signal to us regarding what we are going to deprecate in a major version release. We can easily change all `@ObsoleteApi` methods to `@Deprecated` before a major version release,  and eventually remove it in another major version release.
2. It provides a signal to customers that methods marked with it will be deprecated in a major version release, so they can start using alternative ways as soon as possible. It is not as obvious as `@Deprecated` since IDE would strike through deprecated methods, but customers can see it easily through Javadocs while hovering over on it.

### DIFF
--- a/api-common-java/src/main/java/com/google/api/core/ObsoleteApi.java
+++ b/api-common-java/src/main/java/com/google/api/core/ObsoleteApi.java
@@ -61,5 +61,5 @@ import java.lang.annotation.Target;
 @Documented
 public @interface ObsoleteApi {
   /** Context information such as links to a discussion thread, tracking issue, etc. */
-  String description();
+  String value();
 }


api-common-java/src/main/java/com/google/api/core/ObsoleteApi.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Indicates a public API is obsolete, and will be deprecated in the next major version.
+ *
+ * <p>Usage guidelines:
+ *
+ * <ol>
+ *   <li>This annotation is used only on APIs with non-private visibility. Private interfaces should
+ *       not use it.
+ *   <li>This annotation should only be added if either an alternative API is provided or we plan to
+ *       stop supporting the API in next major version
```
meltsufin, line 47:
```suggestion
 *       stop supporting the API in the next major version.
```

api-common-java/src/main/java/com/google/api/core/ObsoleteApi.java
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * Indicates a public API is obsolete, and will be deprecated in the next major version.
+ *
+ * <p>Usage guidelines:
+ *
+ * <ol>
+ *   <li>This annotation is used only on APIs with non-private visibility. Private interfaces should
+ *       not use it.
+ *   <li>This annotation should only be added if either an alternative API is provided or we plan to
```
meltsufin, line 46:
```suggestion
 *   <li>This annotation should only be added if either an alternative API is provided and/or we plan to
```

meltsufin (approved):

lqiu96:
Maybe we can add a blurb here: https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/README.md#versioning about this new annotation.

blakeli0:
> Maybe we can add a blurb here: https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/README.md#versioning about this new annotation (I believe this annotation is going to be used in Gax?).

Great idea! Added, PTAL.

lqiu96 (approved):

diegomarquezp (approved):
LGTM - I'll start adding this to the threeten PR https://github.com/googleapis/sdk-platform-java/pull/1872

PR #1909, opened 2023-08-07
ci: add downstream compatibility checks

burkedavison:
Adds a new GitHub workflow to perform downstream testing on client libraries.

1. Installs sdk-platform-java to local maven.
2. Checks out the downstream repository at the latest release associated with the main branch, not HEAD.
3. Updates all dependencies on java-shared-dependency to the local version.
4. Compiles and invokes unit tests for the given client libraries. 
 
Does not invoke integration tests. Downstream integration testing is limited to google-cloud-java via the existing GraalVM Kokoro presubmits.

### DIFF
--- a/.kokoro/presubmit/downstream-build.sh
+++ b/.kokoro/presubmit/downstream-build.sh
@@ -15,15 +15,21 @@
 
 set -eo pipefail
 
-## Get the directory of the build script
-scriptDir=$(realpath "$(dirname "${BASH_SOURCE[0]}")")
-## cd to the parent directory, i.e. the root of the git repo
-cd "${scriptDir}/../.."
-
+# Define the google-cloud-java modules to be tested
 if [ -z "${MODULES_UNDER_TEST}" ]; then
   echo "MODULES_UNDER_TEST must be set to run downstream-build.sh"
   exit 1
 fi
+# Define the dedicated client library repos to be tested
+if [ -z "${REPOS_UNDER_TEST}" ]; then
+  echo "REPOS_UNDER_TEST must be set to run downstream-build.sh"
+  exit 1
+fi
+
+## Get the directory of the build script
+scriptDir=$(realpath "$(dirname "${BASH_SOURCE[0]}")")
+cd "${scriptDir}/../.." # git repo root
+source "$scriptDir/common.sh"
 
 # Use GCP Maven Mirror
 mkdir -p "${HOME}/.m2"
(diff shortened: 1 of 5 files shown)


meltsufin (approved):
It looks good, but would it make sense to add some tests for the functions added in `common.sh`?

blakeli0:
> 2. Checks out the downstream repository at the latest release associated with the main branch, not HEAD.

What's the reason to not checkout HEAD on main branch? To avoid possible frictions?

.kokoro/presubmit/downstream-build.sh
```diff
@@ -15,15 +15,17 @@
...
-scriptDir=$(realpath "$(dirname "${BASH_SOURCE[0]}")")
-## cd to the parent directory, i.e. the root of the git repo
-cd "${scriptDir}/../.."
-
 if [ -z "${MODULES_UNDER_TEST}" ]; then
   echo "MODULES_UNDER_TEST must be set to run downstream-build.sh"
   exit 1
 fi
+# Use default value for REPOS_UNDER_TEST if unset. If set to empty string, maintain empty string.
+REPOS_UNDER_TEST=${REPOS_UNDER_TEST-"java-storage"}
```
blakeli0, line 23: Is it intentional to have `java-storage` here?

.kokoro/presubmit/downstream-compatibility.sh
```diff
@@ -0,0 +1,49 @@
...
+  exit 1
+fi
+
+# Get the directory of the build script
+scriptDir=$(realpath "$(dirname "${BASH_SOURCE[0]}")")
+cd "${scriptDir}/../.." # cd to the root of this repo
+source "$scriptDir/common.sh"
+
+echo "Setup maven mirror"
+mkdir -p "${HOME}/.m2"
```
blakeli0, line 30: I might be missing something, I thought `.m2` folders would be created automatically by running `mvn` command?

.kokoro/presubmit/downstream-build.sh
burkedavison: Yes. It's the only current repository with GraalVM tests that don't require cloud project resources.

blakeli0: Thanks! Can we add comments to explain it?

burkedavison:
> What's the reason to not checkout HEAD on main branch? To avoid possible frictions?

* To avoid any issues with main not always being stable. Across all repos, it's possible that main is not always ready to release and may enter into a bad state. OTOH, it's very unlikely that a release commit is in a bad state.
* To make it more difficult to introduce a breaking change by updating sdk-platform-java and the downstream repo at the same time. It's still possible that the downstream repo prepares for the breaking change, releases, then makes the sdk-platform-java breaking change -- but less likely without triggering a conversation.

.kokoro/presubmit/downstream-compatibility.sh
burkedavison: Likely, however we haven't yet invoked mvn at this point.

blakeli0: What's the content in `settings.xml`? If we don't push to anywhere, I think we might be able to just use the default?

burkedavison: https://github.com/googleapis/sdk-platform-java/blob/main/settings.xml

It uses the Google mirror instead of maven central directly.

This logic is not new to this PR. If we remove it from here, we should remove use of the mirror in all the repos that use it. I'd ask this discussion not be considered part of this PR.

https://github.com/googleapis/java-shared-config/blob/763419aefd7efb40b30641641198ca6491b380d2/.kokoro/presubmit/downstream-build.sh#L42-L44

https://github.com/googleapis/sdk-platform-java/blob/33e40ff788beaca80f2158f0d01642d9bc3d3c8f/.kokoro/presubmit/downstream-build.sh#L28-L30

https://github.com/googleapis/google-cloud-java/blob/a082d632ddb3aa413886b731e0b0298f2bec8170/.kokoro/build.sh#L26-L28

blakeli0: > This logic is not new to this PR. If we remove it from here, we should remove use of the mirror in all the repos that use it. I'd ask this discussion not be considered part of this PR.

Agreed. It was mostly for my own understanding, @suztomo what's the benefit of using Google mirror? I guess it's more stable than Maven central when downloading dependencies locally?

blakeli0 (approved):
LGTM. Pending resolution of add more comments.

.kokoro/presubmit/downstream-compatibility.sh
suztomo: It's also faster in Kokoro CI because both are in Google'a infrastructure.

meltsufin (approved):
Thanks for adding the tests!

PR #1916, opened 2023-08-10
feat: add `generate_library.sh` without post processing

### LINKED ISSUE #1922: Implement a shell script to generate a java library hermetically (without post processing)
1. Implement a shell script, `generate_library.sh`, to generate a java library.
2. Compare the result with pure generated code in googleapis-gen, exclude gradle files.

JoeWang1127:
Fixes #1922 ☕️


JoeWang1127:
The differences in the [comparison](https://github.com/googleapis/sdk-platform-java/actions/runs/5868423301/job/15911223019?pr=1916) is because the entries in `gapic_metadata.json` and `package-info.java` are in the same order of proto files in `proto_library` target in BUILD. I sorted the proto files with respect to file name to get a deterministic result (`proto_files=$(find "$proto_path" -type f  -name "*.proto" | sort)`).

However, in a recent submit (cl/536794833), `model_garden_service.proto` is added after `model_service.proto` in BUILD. The `generate_library.sh` still generates library since `model_garden_service.proto` is *before* `model_service.proto` upon sorted and causes the difference.

This is not a source code difference and there are some approaches
1. change the proto files's order in google3
2. ignore the difference in `gapic_metadata.json` and `package-info.java`, though this is risky as real differences, lack of an entry for example, will be ignored.
3. refactor `generate_library.sh` such that list all proto files in proto_path as the same order in BUILD. This will increase the input list of `generate_library.sh` and seems redundant since we have `proto_path` as an input. Of course, here the assumption is all proto files in `proto_path` should be generated, but why put it there if it's not for code generation?

@meltsufin @blakeli0 @suztomo WDYT?

blakeli0:
> in a recent submit (cl/536794833), model_garden_service.proto is added after model_service.proto in BUILD

Can you take a look at the commit history? It might be manually added. 
Either way, if the current generation process is assuming the protos are sorted, then we should add them as input as well, because we can not control the order of the protos, but the result is dependent on them so it makes the generation process not hermetic.

JoeWang1127:
> Can you take a look at the commit history? It might be manually added.

The CL's title suggested that it's auto-generated but I think the proto file is added manually.

> Either way, if the current generation process is assuming the protos are sorted, then we should add them as input as well, because we can not control the order of the protos, but the result is dependent on them so it makes the generation process not hermetic.

I think by sorting the proto files, the result is deterministic and it is also hermetic as given the same proto files and dependencies, the generated code is the same. If we choose to list the proto files as BUILD does, we are tied with BUILD and we need to change the list whenever a new proto files is added (or deleted) in `proto_path` unless I find an example such that a proto file is not generated in `proto_path`.

JoeWang1127:
> In order for us to make a decision, we need to understand the usage of `gapic_metadata.json` and `package-info.java`. If the order of entries in these two files do not matter to customers, then we can change gapic-generator-java to always generated sorted `gapic_metadata.json` and `package-info.java`.

The `package-info.java` has mainly comments and I guess the difference in comments do not matter.

The `gapic_metadata.json` lists metadata of the grpc services. I think typically json file is parsed into map-like structure so the order of the key does not matter.

Of course there maybe other use cases and I'll do some research.

JoeWang1127:
I changed to another library: `google/cloud/confidentialcomputing/v1` and it only has one proto file so it should not have the above problem.

blakeli0:
> I changed to another library: `google/cloud/confidentialcomputing/v1` and it only has one proto file so it should not have the above problem.

This is good for now but we need to have a deterministic order of the protos somewhere, either in the generator or in the script.

.github/workflows/verify_library_generation.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - name: checkout googleapis
+      uses: actions/checkout@v3
+      with:
+        repository: googleapis/googleapis
+        ref: 90a551c2a02613507a2f0f05fd25d133113a1a1c
+        path: googleapis
+    - name: copy protos to proto_path
+      run: |
+        cp -r googleapis/google library_generation
+    - name: generate client library
```
blakeli0, line 33: Not sure if we have an existing pattern to test shell scripts, can we have something like `generate_library_test.sh` that have all the test cases? So that we can test more scenarios and it can be run in local as well? +@meltsufin

library_generation/utilities.sh
```diff
@@ -0,0 +1,110 @@
...
+}
+remove_empty_files() {
+  category=$1
+  find "$destination_path/$category-$folder_name/src/main/java" -type f -size 0 | while read -r f; do rm -f "${f}"; done
+  if [ -d "$destination_path/$category-$folder_name/src/main/java/samples" ]; then
+      mv "$destination_path/$category-$folder_name/src/main/java/samples" "$destination_path/$category-$folder_name"
+  fi
+}
+
+mv_src_files() {
```
blakeli0, line 19: Do we have logic that moves the resource name helper classes from gapic folder to proto folder? Like this [file](https://github.com/googleapis/googleapis-gen/blob/master/google/bigtable/v2/google-cloud-bigtable-v2-java/proto-google-cloud-bigtable-v2-java/src/main/java/com/google/bigtable/v2/TableName.java)

library_generation/utilities.sh
```diff
@@ -0,0 +1,110 @@
...
+
+remove_grpc_version() {
+  find "$destination_path" -type f -name "*Grpc.java" -exec \
+  sed -i 's/value = \"by gRPC proto compiler.*/value = \"by gRPC proto compiler\",/g' {} \;
+}
+
+download_gapic_generator_parent_pom() {
+  gapic_generator_version=$1
+  if [ ! -f parent-pom.xml ]; then
+    curl -LJ -o parent-pom.xml "https://repo1.maven.org/maven2/com/google/api/gapic-generator-java-pom-parent/$gapic_generator_version/gapic-generator-java-pom-parent-$gapic_generator_version.pom"
```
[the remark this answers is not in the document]
JoeWang1127: Sure, I'll look into that.

.github/workflows/verify_library_generation.yaml
```diff
@@ -0,0 +1,56 @@
...
+      shell: bash
+      run: |
+        set -x
+        library_generation/generate_library.sh \
+        -p google/cloud/confidentialcomputing/v1 \
+        -d google-cloud-confidentialcomputing-v1-java \
+        --gapic_generator_version 2.24.0 \
```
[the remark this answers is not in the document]
JoeWang1127: I want to use this check to ensure future change of `generate_library.sh` won't generate "wrong" library, so it's the latter. I'll find a way to dynamically pass all the values.

JoeWang1127:
> This is good for now but we need to have a deterministic order of the protos somewhere, either in the generator or in the script.

In the script, the order is deterministic: `proto_files=$(find "$proto_path" -type f  -name "*.proto" | sort)`

.github/workflows/verify_library_generation.yaml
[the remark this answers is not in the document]
blakeli0: > should I parse the version from googleapis/WORKSAPCE

Yes

> the commit hash should be the latest in master branch of googleapis

Yes, googleapis-gen is always generated from head.

This check is currently checking against googeapis-gen, also keep in mind that we could easily change it to compare with google-cloud-java in the future.

library_generation/utilities.sh
```diff
@@ -0,0 +1,110 @@
...
+
+find_additional_protos_in_yaml() {
+  pattern=$1
+  find_result=$(grep --include=\*.yaml -rw "$proto_path" -e "$pattern")
+  if [ -n "$find_result" ]; then
+    echo "$find_result"
+  fi
+}
+
+search_additional_protos() {
```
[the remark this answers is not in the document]
JoeWang1127: Just like what the Bazel Build generator does [here](https://github.com/googleapis/rules_gapic/blob/main/bazel/src/main/java/com/google/api/codegen/bazel/BazelBuildFileView.java#L78-L86), this function is used to add additional protos like `common_resources.proto`, `iam_policy.proto` and `locations.proto` if they are required in yaml files.

library_generation/utilities.sh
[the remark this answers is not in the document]
blakeli0: I don't think the script should invoke mvn command either, yes we can assume the jar exists in `~/.m2/repository`. We should throw proper error though if the version passed in is invalid(can not find it in maven central or local).

library_generation/generate_library.sh
```diff
@@ -0,0 +1,171 @@
...
+fi
+# pull protoc-gen-grpc-java plugin from maven central
+cd "$working_directory"
+if [ ! -f protoc-gen-grpc-java ]; then
+  curl -LJ -o protoc-gen-grpc-java https://repo1.maven.org/maven2/io/grpc/protoc-gen-grpc-java/"$grpc_version"/protoc-gen-grpc-java-"$grpc_version"-linux-x86_64.exe
+  chmod +x protoc-gen-grpc-java
+fi
+# gapic-generator-java
+if [ ! -f gapic-generator-java.jar ]; then
+  curl -LJ -o gapic-generator-java.jar https://repo1.maven.org/maven2/com/google/api/gapic-generator-java/"$gapic_generator_version"/gapic-generator-java-"$gapic_generator_version".jar
```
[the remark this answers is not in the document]
JoeWang1127: Sure, I'll look into that.

Do you think we need to test SNAPSHOT of `protobuf` and `grpc` as well?

.github/workflows/verify_library_generation.yaml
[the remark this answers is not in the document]
JoeWang1127: @blakeli0 follow up from our meeting, do you still think we need to test the script against master branch of googleapis-gen? As this check will be replaced with a testing script.

.github/workflows/verify_library_generation.yaml
JoeWang1127: As discussed in the meeting, the testing script will be in future PRs.

library_generation/generate_library.sh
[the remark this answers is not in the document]
blakeli0: I don't think so at this moment as we don't develop those two libraries. Did we have this use case in the past @suztomo ?

.github/workflows/verify_library_generation.yaml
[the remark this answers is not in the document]
blakeli0: Yes, I think we still need it. As we discussed last week, we want to get signals as early as possible that if our script is generating client libraries correctly or not. If possible, we even want to test against all libraries in googleapis-gen as @meltsufin mentioned.

I consider testing against googleapis-gen and showcase "Integration tests" of the scripts, while the testing script is "Unit tests", so they serve different purposes. Whether we want to make it part of this PR or how much "Integration tests" we want could be decided separately.

library_generation/utilities.sh
blakeli0: Interesting, do you know how this gapic-generate-java generated [file](https://github.com/googleapis/googleapis-gen/blob/master/google/bigtable/v2/google-cloud-bigtable-v2-java/proto-google-cloud-bigtable-v2-java/src/main/java/com/google/bigtable/v2/TableName.java) is being moved to the proto folder?

library_generation/generate_library.sh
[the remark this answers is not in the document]
JoeWang1127: Now the script can copy SNAPSHOT of gapic-generator-java jar file from maven local repo.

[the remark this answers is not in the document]
suztomo: No point to test protobuf and gRPC.

The point of using SNAPSHOT for gapic-generator-java is that the script uses the same version of gapic-generator-java within the repository. It doesn't rely on Maven Central. It's one step forward to hermeticness.

library_generation/utilities.sh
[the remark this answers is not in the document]
JoeWang1127: Now the script can download SNAPSHOT of `gapic-generator-java-pom-parent/pom.xml`

library_generation/utilities.sh
```diff
@@ -0,0 +1,176 @@
...
+    if [ ! -f "$generator_jar" ]; then
+      echo "Can't copy gapic-generator-java-$gapic_generator_version.jar from maven local repository."
+      exit 1
+    fi
+    cp "$generator_jar" "gapic-generator-java-$gapic_generator_version.jar"
+    echo "Copy gapic-generator-java-$gapic_generator_version.jar from maven local repository."
+  fi
+
+  if [ ! -f "gapic-generator-java-$gapic_generator_version.jar" ]; then
+    curl -LJ -o "gapic-generator-java-$gapic_generator_version.jar" https://repo1.maven.org/maven2/com/google/api/gapic-generator-java/"$gapic_generator_version"/gapic-generator-java-"$gapic_generator_version".jar
```
blakeli0, line 148: Can we have proper error handling for all the downloading script? Either from Maven central or Github, we have a lot of scenarios in the past that these two website could be not responsive. In such case, we need to emit proper error message and exit the script.
Separately, we may want to consider downloading from [Google's Maven mirror](https://github.com/googleapis/sdk-platform-java/blob/main/settings.xml#L28) which could be faster and more consistent per @suztomo

library_generation/utilities.sh
```diff
@@ -0,0 +1,176 @@
...
+}
+
+download_generator() {
+  gapic_generator_version=$1
+  cd "$working_directory"
+  if [[ "$gapic_generator_version" == *"-SNAPSHOT" ]]; then
+    # get SNAPSHOT from maven local repository.
+    generator_jar=$HOME/.m2/repository/com/google/api/gapic-generator-java/$gapic_generator_version/gapic-generator-java-$gapic_generator_version.jar
+    if [ ! -f "$generator_jar" ]; then
+      echo "Can't copy gapic-generator-java-$gapic_generator_version.jar from maven local repository."
```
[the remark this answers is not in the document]
JoeWang1127: I changed the error message.

library_generation/utilities.sh
```diff
@@ -0,0 +1,170 @@
...
+  fi
+}
+
+download_protobuf() {
+  protobuf_version=$1
+  cd "$working_directory"
+  if [ ! -d "protobuf-$protobuf_version" ]; then
+    # pull proto files and protoc from protobuf repository
+    # maven central doesn't have proto files
+    curl -LJ -o "protobuf-$protobuf_version.zip" https://github.com/protocolbuffers/protobuf/releases/download/v"$protobuf_version"/protoc-"$protobuf_version"-linux-x86_64.zip \
```
JoeWang1127, line 144: @blakeli0 I still need to download protobuf from [protobuf releases](https://github.com/protocolbuffers/protobuf/releases) because there're common protos in the .zip file. These are the protos in protobuf.zip. Does `sdk-platform-java` publish these protos so that I can download these files from maven central as well?
```
-rw-r--r--  1 joewa primarygroup  6049 Aug 18 19:50 any.proto
-rw-r--r--  1 joewa primarygroup  7728 Aug 18 19:50 api.proto
drwxr-xr-x  2 joewa primarygroup  4096 Aug 18 19:29 compiler/
-rw-r--r--  1 joewa primarygroup 43479 Aug 18 19:50 descriptor.proto
-rw-r--r--  1 joewa primarygroup  4892 Aug 18 19:50 duration.proto
-rw-r--r--  1 joewa primarygroup  2363 Aug 18 19:50 empty.proto
-rw-r--r--  1 joewa primarygroup  8185 Aug 18 19:50 field_mask.proto
-rw-r--r--  1 joewa primarygroup  2341 Aug 18 19:50 source_context.proto
-rw-r--r--  1 joewa primarygroup  3778 Aug 18 19:50 struct.proto
-rw-r--r--  1 joewa primarygroup  6449 Aug 18 19:50 timestamp.proto
-rw-r--r--  1 joewa primarygroup  6367 Aug 18 19:50 type.proto
-rw-r--r--  1 joewa primarygroup  4044 Aug 18 19:50 wrappers.proto
```

blakeli0: We might be able to download these files from Maven central somewhere but I don't think we can get away from downloading protoc, so this look good to me.

library_generation/utilities.sh
blakeli0: It does make the script simpler but I'm not sure executing `mvn` command is a good idea in this script. This adds one more pre-condition of running the script, and this also means we need to install Maven in anywhere we want to execute this script. Especially for internal presubmit checks, I'm not sure we have Maven installed.  
WDYT @meltsufin?

JoeWang1127: If we don't use `mvn`, how can we download artifacts from [maven mirror repository](https://maven-central.storage-download.googleapis.com/maven2/)? It seems that the url pattern is not the same as maven central.

I can add steps to install `mvn` in the script if it's not in the environment.

library_generation/utilities.sh
JoeWang1127: After unzip `java_gapic_srcjar_raw.srcjar.zip` produced by the generator, [TableName.java](https://github.com/googleapis/googleapis-gen/blob/master/google/bigtable/v2/google-cloud-bigtable-v2-java/proto-google-cloud-bigtable-v2-java/src/main/java/com/google/bigtable/v2/TableName.java) is in `java_gapic_srcjar/proto/src/..`.

The command `mv_src_files "proto" "main"` is used to move this file to `proto-*/src/main/java/`.

I can see [InstanceName.java](https://github.com/googleapis/googleapis-gen/blob/master/google/bigtable/v2/google-cloud-bigtable-v2-java/proto-google-cloud-bigtable-v2-java/src/main/java/com/google/bigtable/v2/InstanceName.java) is also moved to `proto-*/src/main/java/` in the same way.

library_generation/utilities.sh
suztomo: > It seems that the url pattern is not the same as maven central.

It's the same; otherwise Maven users cannot use the mirror.

JoeWang1127: The url to download `com.google.api:gapic-generator-java:2.24.0` is `https://repo1.maven.org/maven2/com/google/api/gapic-generator-java/2.24.0/`.

However, `https://maven-central.storage-download.googleapis.com/maven2/com/google/api/gapic-generator-java/2.24.0/` doesn't return the right page. Is it because I don't have the right permission?

suztomo: > we need to emit proper error message and exit the script


@JoeWang1127 This is the concern Blake has. It's about the script may continue when the download fails. What's your answer?

JoeWang1127: I modified the download command so the script can exit with an error message should the download fail.

suztomo: > The url to download com.google.api:gapic-generator-java:2.24.0 is https://repo1.maven.org/maven2/com/google/api/gapic-generator-java/2.24.0/.

This is not the JAR file.

JoeWang1127: > > The url to download com.google.api:gapic-generator-java:2.24.0 is https://repo1.maven.org/maven2/com/google/api/gapic-generator-java/2.24.0/.
> 
> This is not the JAR file.

I changed the url to `https://maven-central.storage-download.googleapis.com/maven2/com/google/api/gapic-generator-java/2.24.0/gapic-generator-java-2.24.0.jar` now I can download the jar file. Thanks Tomo.

Then it comes back to whether we want to download dependencies using `mvn`.

suztomo: The concern that the curl command might continue when the download fails. Or the curl command appropriately fails with an error message.

How can you address the concern?


(Maven central or Google mirror is a trivial choice. Either one is fine)

JoeWang1127: I changed the curl command to
```
curl -LJ -o "protobuf-$protobuf_version.zip" https://github.com/protocolbuffers/protobuf/releases/download/v"$protobuf_version"/protoc-"$protobuf_version"-linux-x86_64.zip || \
  download_fail "protobuf-$protobuf_version"
```

If the download fails, `download_fail` will print error message and exit the script.

library_generation/utilities.sh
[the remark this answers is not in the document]
JoeWang1127: I added comments in utility functions, PTAL.

library_generation/gapic-generator-java-wrapper
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
```
meltsufin, line 2: Can you add a short comment explaining the purpose of this wrapper?

library_generation/utilities.sh
```diff
@@ -0,0 +1,178 @@
...
+  find_result=$(grep --include=\*.yaml -rw "$proto_path" -e "$pattern")
+  if [ -n "$find_result" ]; then
+    echo "$find_result"
+  fi
+}
+
+# Apart from proto files in proto_path, additional protos are needed in order
+# to generate gapic client libraries.
+# Search additional protos in .yaml files.
+search_additional_protos() {
```
meltsufin, line 62: This whole search for additional protos seems to assume that the protos are always coming from `googleapis`. Can we instead explicitly add a `generate_library.sh` parameter for locations of additional protos?

library_generation/utilities.sh
```diff
@@ -0,0 +1,178 @@
+#!/usr/bin/env bash
+
+set -xeo pipefail
+
+# define utility functions
```
meltsufin, line 5: Can we have "unit tests" for functions in this file?

library_generation/README.md
```diff
@@ -0,0 +1,84 @@
+# Generate Gapic Client Library without post-processing
```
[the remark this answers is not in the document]
JoeWang1127: done

.github/workflows/verify_library_generation.yaml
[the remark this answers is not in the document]
JoeWang1127: Sure, I'll create a script to pull protos from master branch of googleapis, invoke `generate_library.sh` to generate a library and compare the result with master branch of googleapis-gen.

library_generation/gapic-generator-java-wrapper
JoeWang1127: done

.github/workflows/verify_library_generation.yaml
```diff
@@ -0,0 +1,60 @@
...
+    - uses: actions/setup-java@v3
+      with:
+        java-version: ${{ matrix.java }}
+        distribution: temurin
+        cache: maven
+    # add maven central mirror repository
+    - uses: s4u/maven-settings-action@v2.8.0
+      with:
+        mirrors: '[{"id": "google-maven-central", "name": "GCS Maven Central mirror", "mirrorOf": "central", "url": "https://maven-central.storage-download.googleapis.com/maven2/"}]'
+    - name: checkout googleapis
```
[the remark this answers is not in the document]
JoeWang1127: Sure, I'll create a script to test `generate_library.sh` and use Github action to run it.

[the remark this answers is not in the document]
JoeWang1127: I added `generate_library_integration_test.sh` to run tests against googleapis-gen.
I'll resolve this comment now.

.github/workflows/verify_library_generation.yaml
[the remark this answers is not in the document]
JoeWang1127: I added generate_library_integration_test.sh to run tests against googleapis-gen.
I'll resolve this comment now and we can continue discussion in the new script.

library_generation/utilities.sh
JoeWang1127: Can I add unit tests in following PRs?
I have added an integration test in this PR.

library_generation/utilities.sh
JoeWang1127: @suztomo I added `--fail` to curl command so the command return with non-zero exit code when the url is wrong or no file exists in remote server. Thanks for the help to clarify the question.

Regards the `mvn` command, I think it's simple to download artifacts from maven central (it will switch to maven local when downloading a SNAPSHOT version). I also tested that the command will return 1 if the version is wrong or the file doesn't exist. But I can change back to `curl` to remove `mvn` as a dependency. @blakeli0 @meltsufin WDYT?

library_generation/utilities.sh
JoeWang1127: In bazel, all protos are listed as dependencies. 

We can infer that list by downloading `google` folder from `googleapis`. Can I assume all proto files are coming from `google` folder?

In design, we want the script can generate libraries with only the given set of protos, but other protos (e.g., `googleapis/google/api`) are needed to generate a library (showcase is an example), so I think we can't get rid of `googleapis` entirely.

library_generation/utilities.sh
blakeli0: Thanks! Can you make it clear by adding more comments regarding what files the scripts is moving and why?

JoeWang1127: I added comments in `generate_library.sh` to explain which files are moved to which directory. PTAL.

About the reason, I copy these commands from [`java_gapic_pkg.bzl`](https://github.com/googleapis/sdk-platform-java/blob/main/rules_java_gapic/java_gapic_pkg.bzl#L252), I guess this is how GAPIC library is organized?

library_generation/utilities.sh
JoeWang1127: Follow up on meetings and chat, I added instructions on users should pull `google/` directory from `googleapis` before running the script. These additional protos can be found within `google/` directory.

library_generation/README.md
```diff
@@ -0,0 +1,91 @@
...
+The absolute path of `proto_path` is `$cwd/$proto_path`. 
+
+Use `-p` or `--proto_path` to specify the value.
+
+   Note that usually additional proto files are needed to generate a gapic library, 
+   you should also copy `google/` folder from [googleapis](https://github.com/googleapis/googleapis) 
+   into `$cwd`.
+
+### destination_path 
+A directory in `$cwd`. This is the path in which the generated library will reside.
```
suztomo, line 31: The usage is unclear to me. What value should a user specify here as the destination? (I see google-cloud-confidentialcomputing-v1-java example below but this directory does not exist in the java-confidentialcomputing module.)


Can you add how user will use the generated library? "We will use this destination to do ..."

library_generation/README.md
```diff
@@ -0,0 +1,91 @@
...
+You need to run the script with the following parameters.
+
+### proto_path
+A directory in the current working directory (refers as `$cwd`) and copy proto files into it. 
+The absolute path of `proto_path` is `$cwd/$proto_path`. 
+
+Use `-p` or `--proto_path` to specify the value.
+
+   Note that usually additional proto files are needed to generate a gapic library, 
+   you should also copy `google/` folder from [googleapis](https://github.com/googleapis/googleapis)
```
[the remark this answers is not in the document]
JoeWang1127: How about I left the prerequisite and delete this part as I think prerequisite should be done before running the script.

library_generation/README.md
JoeWang1127: The generated library is under `$destination_path`, the directory structure is
```
$destination_path
  |_gapic-*
  |    |_src
  |       |_main
  |          |_java
  |          |_resources
  |       |_test
  |_grpc-*
  |    |_src
  |       |_main
  |          |_java
  |    
  |_proto-*
       |_src
          |_main
             |_java
             |_proto
```

This is how pure generated library looked like in `googleapis-gen`.

This PR will generate the library without post processing, so there's no pom.xml, bom, etc. I think it can't be used by users.

suztomo: The directory structure explanation is great and worth adding here. So users will copy the generated files to the corresponding directories of a library repository.

library_generation/README.md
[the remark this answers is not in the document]
suztomo: That makes sense.

library_generation/README.md
JoeWang1127: done

library_generation/README.md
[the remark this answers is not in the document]
JoeWang1127: done

library_generation/README.md
suztomo: Thank you

library_generation/utilities.sh
meltsufin: Sure, you can do it in a separate PR.

library_generation/utilities.sh
suztomo: @JoeWang1127 In terms of hermetic build, can you make a (small) argument which is better between mvn and curl? Which one is more likely to fail (due to outdated configurations) after 2 years.

library_generation/utilities.sh
diegomarquezp: [Related](https://github.com/googleapis/sdk-platform-java/pull/1935/files#r1301828838)

.github/workflows/verify_library_generation.yaml
JoeWang1127: I changed this action to run `generate_library_integration_test.sh`. The test script can run locally as well.

I'll resolve this comment, please reopen it if necessary.

library_generation/utilities.sh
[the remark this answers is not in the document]
JoeWang1127: I'll resolve this comment, please reopen if necessary.

library_generation/utilities.sh
JoeWang1127: Maven is useful in a java project, but we need to maintain a maven wrapper so that the generation is hermetic, this seems a overkill. 

On the other hand, `curl` is more flexible in shell project, so I decide to use curl to download dependencies.

JoeWang1127: I made the change to use `curl --fail -m 30 --retry 2` to download the dependencies. The function will fail if no file exists or takes long time (> 30s) to finish. Also, use Google maven central mirror instead of maven central.

I'll resolve this comment and please reopen it if necessary.

meltsufin (approved):

library_generation/README.md
```diff
@@ -0,0 +1,112 @@
...
+## Environment
+
+Use Linux environment and install java runtime environment (8 or above).
+
+## Prerequisite
+In order to generate a GAPIC library, you need to pull `google/` from [googleapis](https://github.com/googleapis/googleapis)
+and put it into the directory containing `generate_library.sh` since protos in `google/` 
+are likely referenced by protos from which the library are generated.
+
+In most cases, the `proto_path` is a subdirectory within `google/`.
```
blakeli0, line 14: Is it a requirement? Does the script fail if `proto_path` is not a subdirectory within `google/`?

library_generation/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+The script, `generate_library.sh`, allows you to generate a GAPIC client library from proto files.
+
+## Environment
+
+Use Linux environment and install java runtime environment (8 or above).
+
+## Prerequisite
+In order to generate a GAPIC library, you need to pull `google/` from [googleapis](https://github.com/googleapis/googleapis)
+and put it into the directory containing `generate_library.sh` since protos in `google/`
```
blakeli0, line 11: We have to copy the whole `google/` folder? Or maybe only the common protos?

library_generation/utilities.sh
blakeli0: I think it's OK to have some assumptions at this moment if we make them clear. However it's not very clear to me that where we should put these common protos, and where we should run the script. Can we run the script from anywhere and point `poto_path` to any folder? Left some comments on the README in respective sections.

library_generation/README.md
JoeWang1127: No. Showcase client generation is an example that `proto_path` is outside of `google/`.

library_generation/README.md
JoeWang1127: Some protos reference `google/api`, `google/iam`, `google/type`. I'm not sure whether we have a exhausted list of protos such that only those protos can be referenced. So copy `google` folder is a safe option.

library_generation/README.md
```diff
@@ -0,0 +1,112 @@
...
+This is the path in which the generated library will reside. 
+The absolute path of `destination_path` is `$cwd/$destination_path`. 
+
+Use `-d` or `--destination_path` to specify the value.
+   
+Note that you do not need to create `$detination_path` beforehand.
+
+The directory structure of the generated library is
+```
+$destination_path
```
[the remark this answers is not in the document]
JoeWang1127: done

library_generation/utilities.sh
JoeWang1127: We can run `generate_library.sh` anywhere as long as the script knows how to find all protos (not only protos listed in `proto_path`, but also protos referenced by those protos).

Since we don't know whether we have the permission of an arbitrary directory, it's safe to set `proto_path` as a relative path in the current working directory.

suztomo (approved):

library_generation/README.md
blakeli0: Not part of this PR, but I think we should come up with an exhaustive list before we switch to use it in our generation process. As `google/` folder contains protos from all services, which is a lot more than what we need, and it could easily confuse people in the future.

library_generation/README.md
blakeli0: I think we should remove this sentence from `Prerequisite` section if it is not required.

library_generation/utilities.sh
blakeli0: > as long as the script knows how to find all protos

What does this mean? How do we let the script know how to find protos? Does the folder structure has to be in certain ways?

library_generation/README.md
JoeWang1127: done

library_generation/utilities.sh
JoeWang1127: > What does this mean? How do we let the script know how to find protos? Does the folder structure has to be in certain ways?

By using `proto_path`. 

The folder structure should be the same as the import statements in protos.

For example, we want to generate from `folder1/folder2/protoA`, so `proto_path` should be set to `folder1/folder2`. protoA imports protoB as `folder3/folder4/protoB`, then there should be `folder3/folder4/` (containing `protoB`) in the current working directory (`$cwd`, the directory in which `generate_library.sh` resides).

When generating protoA, protoc will first try to generate protoB (because protoA imports protoB) and it will search for protoB in `folder3/folder4` (notice the relative path). If protoB doesn't in `folder3/folder4`, protoc will exit with error.

library_generation/README.md
JoeWang1127: We can do that, but do we have a guidance that a proto should not reference a proto outside of a list? If not, the list is likely changing and we are back to copy `google`.

library_generation/utilities.sh
blakeli0: Thanks, that's much clearer! Can you please add these explanation to README somewhere?

library_generation/README.md
```diff
@@ -0,0 +1,113 @@
...
+### include_samples (optional)
+Whether generates code samples. The value is either `true` or `false`. 
+The default value is `true`.
+
+Use `--include_samples` to specify the value.
+
+## An example to generate a client library
+```
+library_generation/generate_library.sh \
+-p google/cloud/confidentialcomputing/v1 \
```
blakeli0, line 105: Can you please add comments or more examples regarding `proto_path` that all protos referenced in the path has to be in the current working directory(assuming my understanding is correct from another comment)?

library_generation/README.md
blakeli0: Theoretically, the service team can reference protos from anywhere, but cross service proto depedencies are not allowed per https://google.aip.dev/213, so the service protos should only reference common protos from a few selected folders. I think we may have two services that have cross service proto dependencies but all other services should obey it.

 Now regarding common protos, we don't have a guidance on where they should be and it's possible that new common protos are added to new locations. But if it happens, I think it would be better to explicitly add it, similar to `explicitly declare a new dependency` vs `getting a new dependency transitively from a existing dependency` in Maven's world.

We can discuss more in next project meeting.

library_generation/utilities.sh
JoeWang1127: I added an explanation in prerequisite ([commit](https://github.com/googleapis/sdk-platform-java/pull/1916/commits/1648254a8ab5e10312aa93fc0c0970ccc10a9d6c)).

library_generation/README.md
JoeWang1127: I added an example of generating showcase client.

blakeli0 (approved):

PR #1926, opened 2023-08-17
ci: showcase now uses shared-dependencies

burkedavison:
Primary change:
Showcase now uses shared-config as its parent, and imports shared-deps in its dependency management. This makes showcase more consistent with production client libraries which is useful for testing.

Secondary change:
Shared dependencies (via gapic-generator-java-bom) now includes a declaration for the google-auth-library-oauth2-http testlib.

### DIFF
--- a/gapic-generator-java-bom/pom.xml
+++ b/gapic-generator-java-bom/pom.xml
@@ -30,6 +30,14 @@
         <scope>import</scope>
       </dependency>
       <dependency>
+        <groupId>com.google.auth</groupId>
+        <artifactId>google-auth-library-oauth2-http</artifactId>
+        <version>${google.auth.version}</version>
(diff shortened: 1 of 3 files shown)


gapic-generator-java-bom/pom.xml
```diff
@@ -29,6 +29,14 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
+      <dependency>
```
blakeli0, line 32: Not part of this PR. I think we could add the test jar to the [auth bom](https://github.com/googleapis/google-auth-library-java/blob/main/bom/pom.xml), so that it can be released as part of the above auth bom, then we should be able to remove this declaration. WDYT @suztomo ?

blakeli0 (approved):

PR #1935, opened 2023-08-20
feat: generate showcase without post-processing

diegomarquezp:
Modifies the logic of https://github.com/googleapis/sdk-platform-java/pull/1916 to support generation of `showcase`

**NOTES**
* [Referencing an external repository](https://github.com/googleapis/sdk-platform-java/blob/933cad5e9a6b698e2ec4f755e813f6ddb54de26e/showcase/BUILD.bazel#L30) as the source of the protos when building showcase makes the proto folder to contain [a path that includes the external reference](https://github.com/googleapis/sdk-platform-java/tree/main/showcase/proto-gapic-showcase-v1beta1/src/main/proto). Googleapis-based generated libraries do not contain this difference ([example](https://github.com/googleapis/google-cloud-java/tree/main/java-analyticshub/proto-google-cloud-analyticshub-v1/src/main/proto)) but instead contain the `proto_path` (i.e. path from repository root to proto location). 
  * To fix this, I added an [extra postprocessing step](https://github.com/googleapis/sdk-platform-java/blob/ad59c57d74c3e0c5b4479c9b09edef3e27a31414/showcase/scripts/update.sh#L38-L45) in the current showcase generation
  * I also [tried](https://github.com/googleapis/sdk-platform-java/compare/feat/generate-aiplatform-without-postprocessing...feat/generate-showcase-without-postprocessing-transfer-protos) modifying the bazel BUILD file to copy the protos into showcase but it produced a similar result - a path like `proto/bazel-out/.../schema/...`.
*  The `enable-golden-tests` profile was updated so it compares the current source code of showcase against a recently generated library via the hermetic build scripts.
* The `verify` profile calls the new scripts without merging them into the current showcase source and running a `diff` to confirm nothing has changed

**PENDING**
- [x] use current snapshot of gapic-generator-java to generate showcase (currently uses a published version)


library_generation/generate_showcase_integration_test.sh
```diff
@@ -0,0 +1,89 @@
...
+  git clone https://github.com/googleapis/gapic-showcase
+  pushd gapic-showcase
+  showcase_version=$(get_version_from_WORKSPACE "_showcase_version" $SDK_WORKSPACE)
+  git checkout "v$showcase_version"
+  mv schema ..
+  popd
+  rm -rdf gapic-showcase
+fi
+if [ ! -d google ];then
+  git clone https://github.com/googleapis/googleapis
```
diegomarquezp, line 24: We need some googleapis protos. Showcase protos [reference them](https://github.com/googleapis/gapic-showcase/blob/c6512469703a716c619058249f2bb55d81c5ee37/schema/google/showcase/v1beta1/BUILD.bazel#L33-L40).

library_generation/utilities.sh
```diff
@@ -61,11 +61,11 @@ find_additional_protos_in_yaml() {
 # Search additional protos in .yaml files.
 search_additional_protos() {
   additional_protos="google/cloud/common_resources.proto" # used by every library
-  iam_policy=$(find_additional_protos_in_yaml "name: google.iam.v1.IAMPolicy")
+  iam_policy=$(find_additional_protos_in_yaml "name: '*google.iam.v1.IAMPolicy'*")
```
JoeWang1127, line 64: Why do we need this change?

library_generation/generate_showcase_integration_test.sh
```diff
@@ -0,0 +1,89 @@
...
+ggj_version=$(get_version_from_WORKSPACE _gapic_generator_java_version $SDK_WORKSPACE)
+if [ $(echo $ggj_version | grep 'SNAPSHOT' | wc -l) -gt 0 ]; then
+  echo 'This repo is at a snapshot version. Installing locally...'
+  pushd $SCRIPT_DIR/..
+  mvn clean install -DskipTests -Dclirr.skip
+  popd
+fi
+
+
+grpc_version=$(get_version_from_properties "version.io_grpc" $GAX_PROPERTIES)
```
JoeWang1127, line 41: Why don't we get grpc version from `gapic-generator-java-pom-parent/pom.xml`?

diegomarquezp: Good point. I just saw the [convenience method](https://github.com/googleapis/sdk-platform-java/blob/880c27ba4f4e3efbf87dc11ee9454bd762158197/library_generation/generate_library.sh#L62) to infer from the parent pom. I will rely on this instead.

diegomarquezp: Omitted grpc version in showcase script

diegomarquezp:
Will open this PR to be merged agains `main`

library_generation/generate_showcase_integration_test.sh
```diff
@@ -0,0 +1,87 @@
...
+  rm -rdf googleapis
+fi
+
+GOOGLEAPIS_WORKSPACE=$SCRIPT_DIR/WORKSPACE
+
+ggj_version=$(get_version_from_WORKSPACE _gapic_generator_java_version $SDK_WORKSPACE)
+if [ $(echo $ggj_version | grep 'SNAPSHOT' | wc -l) -gt 0 ]; then
+  echo 'This repo is at a snapshot version. Installing locally...'
+  pushd $SCRIPT_DIR/..
+  mvn clean install -DskipTests -Dclirr.skip
```
blakeli0, line 36: I don't think we want to run `mvn` command in this shell script, we can rely on the behavior of `generate_library.sh` to handle it, just error out if it does not exist.

library_generation/utilities.sh
```diff
@@ -209,3 +209,43 @@ download_fail() {
...
+
+  result="$default"
+  if grep -A 15 "$rule" "$build_file" | grep -q "$pattern"; then
+    result="$if_match"
+  fi
+  echo "$result"
+
+}
+
+get_version_from_properties() {
```
blakeli0, line 246: Is this function used?

.github/workflows/verify_library_generation.yaml
```diff
@@ -5,26 +5,29 @@ on:
...
         cache: maven
-    - name: Run integration tests
+    - name: Run googleapis-gen integration test
       run: |
         set -x
         library_generation/generate_library_integration_test.sh \
         -p google/bigtable/v2 \
         -d google-cloud-bigtable-v2-java \
         --googleapis_gen_url https://cloud-java-bot:${{ secrets.CLOUD_JAVA_BOT_GITHUB_TOKEN }}@github.com/googleapis/googleapis-gen.git
+    - name: Run showcase integration test
```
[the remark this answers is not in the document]
diegomarquezp: @blakeli0 `verify.sh` in the showcase module does [something similar](https://github.com/googleapis/sdk-platform-java/blob/24ae2a3e0237fb2821ea07ccdd874c92ead598a2/showcase/scripts/verify.sh#L68). We discussed with @JoeWang1127 that we will transform this PR into one that modifies the way showcase is built and tested (i.e. change maven exec goals)

library_generation/utilities.sh
diegomarquezp: I removed a few unused functions

library_generation/generate_showcase_integration_test.sh
diegomarquezp: Removed

.github/workflows/verify_library_generation.yaml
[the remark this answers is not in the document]
diegomarquezp: I removed this CI component in favor of diff-based tests being run in the showcase folder

showcase/proto-gapic-showcase-v1beta1/src/main/proto/schema/google/showcase/v1beta1/compliance.proto
blakeli0, line 1: Do we know what caused this permission change?

showcase/scripts/generate_component.sh
blakeli0, line 1: Is the purpose of this script to replace a module(gapic, proto, grpc) of showcase library?

showcase/scripts/generate_showcase.sh
blakeli0, line 1: Can you please add comments on top of each shell script to describe the purpose?

showcase/scripts/showcase_utilities.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -ex
+script_dir=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+
+function generateShowcaseIfMissing {
```
blakeli0, line 5: This function is only used once, I don't think we need another file for it.

showcase/scripts/verify.sh
```diff
@@ -2,69 +2,25 @@
...
-    PROTO_PROJECT_DIR=proto-gapic-showcase-v1beta1
-    PROTO_ARCHIVE_NAME=proto-google-cloud-showcase-v1beta1-java
-    PROTO_ARCHIVE=$(find . -name "$PROTO_ARCHIVE_NAME.tar.gz")
-    create_unpack_dir proto_unpacked
-    PROTO_UNPACK_DIR=$PWD
-
-    tar -xzf "$BAZEL_ROOT/$PROTO_ARCHIVE"
-    delete_unneeded
-    diff -ru "$SHOWCASE_DIR/$PROTO_PROJECT_DIR"/src/main/java "$PROTO_UNPACK_DIR/$PROTO_ARCHIVE_NAME"/src/main/java
+    PROTO_PROJECT_DIR='proto-gapic-showcase-v1beta1'
```
blakeli0, line 10: Glad to see that the process is being simplified a lot!

showcase/gapic-showcase/pom.xml
```diff
@@ -68,11 +67,11 @@
                 </goals>
                 <phase>generate-sources</phase>
                 <configuration>
-                  <executable>bazelisk</executable>
+                  <executable>bash</executable>
                   <arguments>
-                    <argument>run</argument>
-                    <argument>--experimental_proto_descriptor_sets_include_source_info</argument>
-                    <argument>//showcase:update_gapic</argument>
+                    <argument>../scripts/generate_component.sh</argument>
```
blakeli0, line 72: I _think_ the purpose of this `update` profile is to replace all three modules.

blakeli0 (commented):
Overall LGTM. Please make sure all the workflows mentioned in the [README](https://github.com/googleapis/sdk-platform-java/blob/main/showcase/README.md) are still supported. In addition, we should be able to remove all the Bazel files in showcase modules now.

showcase/gapic-showcase/pom.xml
diegomarquezp: I added a small comment in the poms to indicate what the profile does

showcase/scripts/showcase_utilities.sh
diegomarquezp: I removed this file and put the logic in the script that uses it

showcase/scripts/generate_showcase.sh
diegomarquezp: I added comments to the scripts and poms

showcase/scripts/generate_component.sh
diegomarquezp: Replacing the current contents of the showcase folder is optional. I explained this in the newly added comments

showcase/scripts/generate_showcase.sh
```diff
@@ -0,0 +1,60 @@
...
+    rm -rdf googleapis
+  fi
+  sparse_clone https://github.com/googleapis/googleapis.git "WORKSPACE google/api google/rpc google/cloud/common_resources.proto google/longrunning google/iam/v1 google/cloud/location google/type"
+  mv googleapis/google .
+  rm -rdf googleapis
+fi
+
+
+ggj_version=$(get_version_from_versions_txt ../../versions.txt "gapic-generator-java")
+if [ $(echo $ggj_version | grep 'SNAPSHOT' | wc -l) -gt 0 ]; then
```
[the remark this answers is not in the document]
diegomarquezp: Thanks for the catch! I removed it

diegomarquezp:
> we should be able to remove all the Bazel files in showcase modules now.
Indeed. I removed the `BUILD` files as well

library_generation/.gitignore
```diff
@@ -0,0 +1,8 @@
+gapic-generator-java*
+google/
+grpc-java-plugin*
+protobuf*
+output/
+schema/
+out
+protoc*
```
meltsufin, line 8: This seems like a very broad filter that can accidentally ignore needed files. Can we just use a single working directory that would be excluded or do proper cleanup at the end of the generation process?

showcase/scripts/.gitignore
```diff
@@ -0,0 +1,7 @@
+showcase-output/
+gapic-generator-java*
+google/
+out
+protobuf*
+protoc-gen*
+schema/
```
meltsufin, line 7: Similar comment to above. Can we avoid this broad gitignore?

showcase/grpc-gapic-showcase-v1beta1/pom.xml
```diff
@@ -34,11 +34,10 @@
                 </goals>
                 <phase>test</phase>
                 <configuration>
-                  <executable>bazel</executable>
+                  <executable>bash</executable>
                   <arguments>
-                    <argument>run</argument>
-                    <argument>--experimental_proto_descriptor_sets_include_source_info</argument>
-                    <argument>//showcase:verify_grpc</argument>
+                    <argument>../scripts/verify.sh</argument>
```
meltsufin, line 39: Do we have to repeat this plugin configuration in every submodule? Can we just re-generate all of them from the top-level module?

showcase/scripts/generate_component.sh
```diff
@@ -0,0 +1,35 @@
...
+fi
+showcase_folder=$script_dir/..
+case $component in
+    gapic )
+        showcase_component_dir=$showcase_folder/gapic-showcase/src/main ;;
+    proto )
+        showcase_component_dir=$showcase_folder/proto-gapic-showcase-v1beta1/src/main ;;
+    grpc )
+        showcase_component_dir=$showcase_folder/grpc-gapic-showcase-v1beta1/src/main ;;
+esac
```
meltsufin, line 29: Why not just generate all at once?

showcase/scripts/generate_showcase.sh
```diff
@@ -0,0 +1,56 @@
...
+source $lib_gen_scripts_dir/utilities.sh
+
+cd $script_dir
+
+# clone gapic-showcase
+if [ ! -d schema ]; then
+  if [ -d gapic-showcase ]; then
+    rm -rdf gapic-showcase
+  fi
+  showcase_version=$(xmllint --xpath '/*[local-name()="project"]/*[local-name()="properties"]/*[local-name()="gapic-showcase.version"]/text()' $script_dir/../gapic-showcase/pom.xml)
```
meltsufin, line 19: Can you add a comment explaining where this showcase version is being extracted form?

library_generation/.gitignore
diegomarquezp: I removed the .gitignore file and addded a convenience function to perform cleanup

showcase/grpc-gapic-showcase-v1beta1/pom.xml
diegomarquezp: Makes sense. I moved the profile to the parent pom so golden tests and source code update is performed in a single place

showcase/scripts/generate_component.sh
diegomarquezp: I modified this script to be called once

showcase/scripts/generate_showcase.sh
diegomarquezp: Added

showcase/scripts/showcase_utilities.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+function cleanup {
+  script_dir=$1
+  cd "${script_dir}"
+  rm -rdf gapic-generator-java* google schema protobuf-* protoc-gen-grpc-java* showcase-output out
```
meltsufin, line 6: Can we keep all of these folders in a temporary directory because when the script fails midway and cleanup is skipped, it's petty messy? It would be easier if there was a subdirectory under `showcase` like `build` that contained all of this.

library_generation/generate_library_integration_test.sh
blakeli0, line 1: This file is being moved to a subdirectory on main, can we merge the changes on main to this branch?

showcase/scripts/generate_showcase.sh
blakeli0: Thanks!

showcase/scripts/verify.sh
```diff
@@ -1,70 +1,27 @@
...
-case $1 in
-  proto)
-    PROTO_PROJECT_DIR=proto-gapic-showcase-v1beta1
-    PROTO_ARCHIVE_NAME=proto-google-cloud-showcase-v1beta1-java
-    PROTO_ARCHIVE=$(find . -name "$PROTO_ARCHIVE_NAME.tar.gz")
-    create_unpack_dir proto_unpacked
-    PROTO_UNPACK_DIR=$PWD
+# compare with proto library
+PROTO_PROJECT_DIR='proto-gapic-showcase-v1beta1'
+diff -ru "${SHOWCASE_DIR}/${PROTO_PROJECT_DIR}/src" "${SCRIPT_DIR}/showcase-output/proto-showcase-output/src"
```
blakeli0, line 17: Now that we are generating all three modules at the same time, is it possible to have a diff for all three modules?

showcase/scripts/showcase_utilities.sh
blakeli0: +1. Similar to the temp bazel folders like `bazel-out` when we were using bazel.

library_generation/generate_library_integration_test.sh
```diff
@@ -0,0 +1,116 @@
...
+# 4. checkout the master branch googleapis-gen repository and compare the result.
+
+# defaults
+googleapis_gen_url="git@github.com:googleapis/googleapis-gen.git"
+
+while [[ $# -gt 0 ]]
+do
+key="$1"
+case $key in
+    -p|--proto_path)
```
JoeWang1127, line 24: nit: the indention of this line should be 2 whitespace, rather than 4.

.github/workflows/ci.yaml
```diff
@@ -174,6 +174,7 @@ jobs:
           java-version: ${{ matrix.java }}
           distribution: temurin
       - run: mvn -version
+      - run: sudo apt install -y libxml2-utils
```
JoeWang1127, line 177: why do we need to install this tool?

showcase/scripts/generate_showcase.sh
```diff
@@ -0,0 +1,60 @@
...
+
+# clone gapic-showcase
+if [ ! -d schema ]; then
+  if [ -d gapic-showcase ]; then
+    rm -rdf gapic-showcase
+  fi
+  # looks at sdk-platform-java/showcase/gapic-showcase/pom.xml to extract the
+  # version of gapic-showcase
+  # see https://github.com/googleapis/gapic-showcase/releases
+  showcase_version=$(xmllint --xpath '/*[local-name()="project"]/*[local-name()="properties"]/*[local-name()="gapic-showcase.version"]/text()' "${SCRIPT_DIR}/../gapic-showcase/pom.xml")
```
JoeWang1127, line 23: We have to install `xmllint` before running the script.

Do you think retrieving the version using `sed` and `cut` would be better?

.github/workflows/verify_library_generation.yaml
```diff
@@ -5,14 +5,13 @@ on:
...
     - library_generation/**
-
   workflow_dispatch:
 name: verify_library_generation
 jobs:
   integration_tests:
     strategy:
       matrix:
-        java: [ 8 ]
+        java: [ 11 ]
```
JoeWang1127, line 14: I think the generator still supports Java 8 runtime, why increase it to 11?

library_generation/utilities.sh
```diff
@@ -216,3 +214,60 @@ download_fail() {
   >&2 echo "Fail to download ${artifact} from ${repo} repository. Please install ${artifact} first if you want to download a SNAPSHOT."
   exit 1
 }
+
+get_version_from_WORKSPACE() {
```
JoeWang1127, line 218: Could you add unit tests for these utilities functions?

library_generation/utilities.sh
```diff
@@ -63,16 +63,14 @@ find_additional_protos_in_yaml() {
...
+  additional_protos="google/cloud/common_resources.proto" # used by every library
+  iam_policy=$(find_additional_protos_in_yaml "name: '*google.iam.v1.IAMPolicy'*")
+  if [ -n "$iam_policy" ]; then
+    additional_protos="$additional_protos google/iam/v1/iam_policy.proto"
   fi
-  locations=$(find_additional_protos_in_yaml "name: google.cloud.location.Locations")
-  if [ -n "${locations}" ]; then
-    additional_protos="${additional_protos} google/cloud/location/locations.proto"
+  locations=$(find_additional_protos_in_yaml "name: '*google.cloud.location.Locations'*")
+  if [ -n "$locations" ]; then
```
JoeWang1127, line 72:
```suggestion
  if [ -n "${locations}" ]; then
```

library_generation/README.md
```diff
@@ -125,15 +125,3 @@ library_generation/generate_library.sh \
 --include_samples true
 ```
 
-## An example to generate showcase client
```
JoeWang1127, line 128: Since this part is deleted in this PR, do we have suggestions on how to generate/update showcase in this repo?

meltsufin (approved):
You follow up on the outstanding comments in a separate PR.

.github/workflows/verify_library_generation.yaml
diegomarquezp: This was part of the setup to run the showcase script. It is not necessary anymore - removing

library_generation/generate_library_integration_test.sh
diegomarquezp: merged

diegomarquezp: Merged and moved

library_generation/generate_library_integration_test.sh
diegomarquezp: this was the wrong file I brought when merging `main`. The file under `library_generation/test` has the correct indentation

showcase/scripts/generate_showcase.sh
diegomarquezp: I'll try :)

blakeli0 (approved):
LGTM. It would be better if we can add unit tests for new utility functions as @JoeWang1127 mentioned if it's not too much, otherwise it's OK to add it in another PR.

JoeWang1127 (approved):

showcase/scripts/verify.sh
diegomarquezp: Unless there is a better way than preparing two special folders with a similar structure and run `diff` on their roots I think the simplest is to just run `diff` one time for each of the folders (`proto`, `gapic` and `grpc`) in a single call to `verify.sh`. We don't compare all the contents of `gapic-showcase` or `grpc-gapic-showcase-v1beta1` because there are other special files not generated by the scripts (e.g. ITs).

diegomarquezp:
> LGTM. It would be better if we can add unit tests for new utility functions as @JoeWang1127 mentioned if it's not too much, otherwise it's OK to add it in another PR.

Thanks @blakeli0 . I added a couple tests for info extraction functions.

library_generation/utilities.sh
diegomarquezp: added a couple for `BUILD` and `WORKSPACE` files

PR #1948, opened 2023-08-29
chore: add unit tests for `utilities.sh`

### LINKED ISSUE #1947: Write unit tests for utility functions used in library generation
Write unit tests for utility functions defined in [`utilities.sh`](https://github.com/googleapis/sdk-platform-java/blob/main/library_generation/utilities.sh).

JoeWang1127:
Fixes #1947 ☕️

The list of utility functions:
- extract_folder_name*
- remove_empty_files
- mv_src_files
- unzip_src_files
- find_additional_protos_in_yaml*
- search_additional_protos*
- get_gapic_opts*
- remove_grpc_version*
- download_gapic_generator_pom_parent*
- get_grpc_version*
- get_protobuf_version*
- download_tools
- download_generator*
- download_protobuf*
- download_grpc_plugin*
- download_from*
- copy_from
- download_fail

Functions with star (*) are tested in this PR.


library_generation/utilities.sh
```diff
@@ -76,15 +76,15 @@ search_additional_protos() {
 
 # get gapic options from .yaml and .json files from proto_path.
 get_gapic_opts() {
-  gapic_config=$(find "${proto_path}" -type f -name "*gapic.yaml")
+  gapic_config=$(find "$proto_path" -type f -name "*gapic.yaml")
```
suztomo, line 79: prefer "${var}" over "$var"

https://google.github.io/styleguide/shellguide.html

suztomo (approved):

library_generation/utilities.sh
JoeWang1127: Thanks for the reference, I'll modify code accordingly.

library_generation/test/generate_library_unit_tests.sh
```diff
@@ -0,0 +1,222 @@
...
+
+get_grpc_version_test() {
+  pushd "$script_dir"
+  actual_version=$(get_grpc_version "2.24.0")
+  assertEquals "1.56.1" "$actual_version"
+  rm "gapic-generator-java-pom-parent-2.24.0.pom"
+  popd
+}
+
+get_protobuf_version_test() {
```
blakeli0, line 70: Two general suggestions of all the tests in this file:
1. Can we change the test name to something more meaningful? See internal [test naming guide](https://engdoc.corp.google.com/eng/doc/devguide/testing/unit/best-practices.md?cl=head&polyglot=java#naming). For example, this could be `get_protobuf_version_should_succeed_with_valid_generator_version`.
2. Most of the tests are for success scenarios only, can we have some tests for invalid scenarios? For example, maybe `get_protobuf_version_should_faile_with_invalid_generator_version`. In addition, the actual download logic is in [download_gapic_generator_pom_parent](https://github.com/googleapis/sdk-platform-java/blob/main/library_generation/utilities.sh#L100-L115) and I don't see how we handle inalid versions there.

library_generation/test/generate_library_unit_tests.sh
```diff
@@ -0,0 +1,222 @@
...
+
+# Variables used to generate final result
+total_num=0
+succeed_num=0
+failed_num=0
+# Unit tests against ./utilities.sh
+script_dir=$(dirname "$(readlink -f "$0")")
+source "$script_dir"/../utilities.sh
+
+assertEquals() {
```
blakeli0, line 13: We can move these helper methods to the bottom and/or add underscore prefix to indicate they should only be used internally, see this [StackOverflow post](https://stackoverflow.com/questions/2289293/is-there-a-convention-for-naming-private-functions-in-bash).

library_generation/utilities.sh
blakeli0: Thanks Tomo for sharing this! This is really helpful, I see a section regarding [when to use shell](https://google.github.io/styleguide/shellguide.html#when-to-use-shell) which I think it's worth discussing more in our next meeting.

library_generation/test/generate_library_unit_tests.sh
```diff
@@ -0,0 +1,222 @@
...
+  proto_path="$script_dir/resources/monitoring"
+  transport="grpc"
+  rest_numeric_enums="false"
+  gapic_opts="$(get_gapic_opts)"
+  assertEquals \
+  "transport=grpc,grpc-service-config=$proto_path/monitoring_grpc_service_config.json,gapic-config=$proto_path/monitoring_gapic.yaml,api-service-config=$proto_path/monitoring.yaml" \
+  "$gapic_opts"
+}
+
+remove_grpc_version_test() {
```
blakeli0, line 128: Not directly related to the tests, I wonder why we have to remove the gRPC version manually, what made the differences? Separately, I think it might be good to keep the version for troubleshooting/informational purposes.

blakeli0 (commented):
Do you plan to have another PR for the unit testing of `generate_library.sh`?

library_generation/test/generate_library_unit_tests.sh
JoeWang1127: When using `bazel build`, there's a "by gRPC proto compiler" at the beginning of *Grpc.java file. However, the annotation becomes "by gRPC proto compiler xxx" when using shell script, where xxx is the grpc version.

I use this function to remove the version number so that two files are totally identical.

JoeWang1127:
> Do you plan to have another PR for the unit testing of `generate_library.sh`?

I think `generate_library.sh` is tested through the integration tests since it "hooks" everything.

library_generation/test/generate_library_unit_tests.sh
blakeli0: I think `bazel build` eventually calls the same gRPC proto compiler? If not, we need to understand the differences, as there could be more differences when generating other services.

library_generation/test/generate_library_unit_tests.sh
JoeWang1127: > Can we change the test name to something more meaningful? See internal [test naming guide](https://engdoc.corp.google.com/eng/doc/devguide/testing/unit/best-practices.md?cl=head&polyglot=java#naming). For example, this could be get_protobuf_version_should_succeed_with_valid_generator_version.

I renamed the tests.

> Most of the tests are for success scenarios only, can we have some tests for invalid scenarios? For example, maybe get_protobuf_version_should_faile_with_invalid_generator_version. In addition, the actual download logic is in [download_gapic_generator_pom_parent](https://github.com/googleapis/sdk-platform-java/blob/main/library_generation/utilities.sh#L100-L115) and I don't see how we handle inalid versions there.

I added test cases to verify invalid version num. 

The download function will fail and exit if the version is wrong (404 in response) so I thought I couldn't test this scenario in the current shell. Wrapp the function call in `$()` to execute the function in subshell can solve this problem.
```
download_generator_failed_with_invalid_version_test() {
  # The download function will exit the shell
  # if download failed.
  # Use $() to execute the function in subshell so that
  # the other tests can continue executing in the current
  # shell.
  $(download_generator "1.99.0")
  __assertFileDoesNotExist "gapic-generator-java-1.99.0.jar"
}
```

library_generation/test/generate_library_unit_tests.sh
JoeWang1127: I renamed the helper functions.

However, is it a convention to move helper functions to the bottom?

blakeli0:
> > Do you plan to have another PR for the unit testing of `generate_library.sh`?
> 
> I think `generate_library.sh` is tested through the integration tests since it "hooks" everything. The result is verified through `googleapis-gen` (and `google-cloud-java` later).

Unit tests and Integration tests serve different purposes. Integration tests treat the script somewhat like a black box and make sure it works in a real life scenario. Unit tests treat the script as a white box and attempt to test every combinations with test data. 
For example, we have [logics](https://github.com/googleapis/sdk-platform-java/blob/main/library_generation/generate_library.sh#L60-L62C3) to get protobuf version from gapic-generator-java if it is not specified, so we need to have test cases for both protobuf version is specified and is not specified, and technically we need to have two test cases for each `if` condition. 
Now how do we achieve that may not be that simple. For the above case, we could have a test gapic-generator-java pom that has a different protobuf version than the specified protobuf version through the script.

library_generation/test/generate_library_unit_tests.sh
```diff
@@ -0,0 +1,261 @@
...
+succeed_num=0
+failed_num=0
+failed_tests=""
+# Unit tests against ./utilities.sh
+script_dir=$(dirname "$(readlink -f "$0")")
+source "${script_dir}"/../utilities.sh
+
+# Helper functions, they shouldn't be called outside this file.
+__assertEquals() {
+  expected=$1
```
diegomarquezp, line 16: go/shell-style#use-local-variables
Functions should declare their local variables with the `local` keyword. We have several functions (not only in this PR) around that are not following this style. cc: @blakeli0 @suztomo would it make sense to have a "styling up" PR to make our scripts _more_ compliant with the guidelines?

diegomarquezp: go/shell-style#constants-environment-variables-and-readonly-variables
Another example, to have constants with upper case (e.g. `SCRIPT_DIR` instead of `script_dir`)

library_generation/test/generate_library_unit_tests.sh
```diff
@@ -0,0 +1,261 @@
...
+  fi
+
+  rm "${destination_path}/QueryServiceGrpc.java"
+  return "${return_code}"
+}
+
+download_generator_success_with_valid_version_test() {
+  download_generator "2.24.0"
+  __assertFileExists "gapic-generator-java-2.24.0.jar"
+  rm "gapic-generator-java-2.24.0.jar"
```
diegomarquezp, line 158: I think we can have an "`afterEach`" function that cleans up the downloaded jars/tars, even if they don't exist. It could even be useful for local development where I personally find it a bit hard to read dir contents after I call the scripts

library_generation/test/generate_library_unit_tests.sh
diegomarquezp: nvm, I saw at the end of the changes we have https://github.com/googleapis/sdk-platform-java/pull/1948/files#r1311724425 to address this!

library_generation/test/generate_library_unit_tests.sh
JoeWang1127: I think `bazel build` download the gRPC proto compiler source code from github (defined [here](https://github.com/googleapis/sdk-platform-java/blob/1e1ca5d39b4f02b8396b6b846fdec7fcdba6958b/gax-java/repositories.bzl#L62) in gax/repository.bzl) and build from source.

In `generate_library.sh`, it download gPRC compiler executable from maven central. I think both ways will have the same executable.

library_generation/test/generate_library_unit_tests.sh
diegomarquezp: This could be expanded with a `git diff` to confirm we don't have any unexpected untracked files that aren't deleted by such function

library_generation/test/generate_library_unit_tests.sh
JoeWang1127: I plan to propose a PR to format code after merging this one.

JoeWang1127:
> Unit tests and Integration tests serve different purposes. Integration tests treat the script somewhat like a black box and make sure it works in a real life scenario. Unit tests treat the script as a white box and attempt to test every combinations with test data. For example, we have [logics](https://github.com/googleapis/sdk-platform-java/blob/main/library_generation/generate_library.sh#L60-L62C3) to get protobuf version from gapic-generator-java if it is not specified, so we need to have test cases for both protobuf version is specified and is not specified, and technically we need to have two test cases for each `if` condition. Now how do we achieve that may not be that simple. For the above case, we could have a test gapic-generator-java pom that has a different protobuf version than the specified protobuf version through the script. If we can achieve the same results(testing every combinations) through googleapis-gen integration tests, that works too, but I feel it may be more complicated.

Unit test should [test against public APIs](https://engdoc.corp.google.com/eng/doc/devguide/testing/unit/best-practices.md?polyglot=java&cl=head#public-apis) so I think we can test `generate_library.sh` with golden tests: fix the protos and gapic-generator-java version (protobuf and grpc version may or may not provide, but the version should be the same as in gapic-generator-java), verify the script can generate the same source code (compare with golden files). Also, I can write test cases with invalid version of generator, protbuf and grpc to verify `generate_library.sh` exits with non-zero code. WDYT?

library_generation/test/generate_library_unit_tests.sh
blakeli0: > However, is it a convention to move helper functions to the bottom?

I don't think there is any written convention regarding this. From my personal experience and preference, making helpers the first thing in the tests may take away people's attention from real unit tests, especially if you don't have a good naming convention for unit tests. I see that it's much better now with proper names for the helper functions and unit tests, so I'm OK with the way it is.

blakeli0:
> I can write test cases with invalid version of generator, protbuf and grpc to verify `generate_library.sh` exits with non-zero code

SGTM!

> I think we can test generate_library.sh with golden tests: fix the protos and gapic-generator-java version (protobuf and grpc version may or may not provide, but the version should be the same as in gapic-generator-java), verify the script can generate the same source code (compare with golden files)

I like the idea, this is similar to the `golden unit/integration tests` in gapic-generator-java.

library_generation/test/resources/alloydb/alloydb_v1.yaml
blakeli0, line 1: Do the content of these test files matter? If not or only partially matters, can we simplify the content to only what we need? In addition, it would be great if we can name these test files based on what we are testing, and/or add additional comments on each yaml to indicate what this testing yaml is for. e.g. Something like `This yaml is only intended for testing search_additional_protos function` with a yaml name `search_additional_protos_scenario_xyz.yaml`.

blakeli0 (approved):
LGTM pending a question regarding the testing files. There are a few additional follow ups that need to be done(unit tests and golden tests for `generate_library.sh`, style change etc), but they don't have to completed as part of this PR.

library_generation/test/generate_library_unit_tests.sh
blakeli0: LGTM, thanks!

library_generation/test/resources/alloydb/alloydb_v1.yaml
JoeWang1127: This makes sense. 

I'll change the name and contents of yaml files in the reformat PR.

library_generation/test/generate_library_unit_tests.sh
JoeWang1127: Not all tests have clean up actions so it's difficult to write a function and apply it after every test execution.

However, I'll think about it in the reformat PR.

PR #1951, opened 2023-08-30
feat: `generate_library.sh` with postprocessing

diegomarquezp:
This PR adds post-processing to `generate_library.sh`
The changes are intended for pure GAPICs in `google-cloud-java`

### Remarks:
* Although some HW libraries can be generated correctly, we are still at mercy of unmerged PRs ([doc](https://docs.google.com/document/d/19E7Psf8VsfdXP_KyhcnaScpiyFhbREGFtZHEQfJyl_M/edit)) that could affect the contents of its `main` branch. We are skipping testing against HW libraries
* For the case of `java-bigtable` and `dialogflow`, we have owlbot.py files referencing libraries or versions that have not been generated by the current run of `generate_library.sh` ([doc](https://docs.google.com/document/d/1ayW5ySj1-rSzsKgK8Tn4RllJf6Bgq-y3wowjlIG-rZk/edit#heading=h.yny69wxe8z9l)). For this reason, HW libraries are excluded as well as `dialogflow`, which is left as a commented row for future inspection
* We keep `googleapis-gen` tests as a separate option in the integration tests, which can be activated by using `--enable_postprocessing=false`. This is a convenience for local development
* Some libraries will produce `package-info.java` files with incorrect order of variables, although logically equivalent. This is solved by using `LC_COLLATE=C` for byte-level sorting of proto files when passing as input to the gapic generator.
* At some point the script was capable of new-library generation, for which most changes were removed in https://github.com/googleapis/sdk-platform-java/pull/1951/commits/b3169c3a52e32901449583aef79f270933c31fe8 and https://github.com/googleapis/sdk-platform-java/pull/1951/commits/d03aad6ed20cf1531602b1b40655e0d67eee1fe8 for the sake of restricting this PR to consider only pre-existing libraries. It was able to generate some gapic libraries (untouched by humans) that produced no differences even at the `pom.xml` level, whitespace included

### Integration test
1. sparse clone `google-cloud-java` and the specified `repository_path` (e.g. `google-cloud-java/java-asset`)
2. run `generate_library.sh`. 
3. The script removes the `workspace` folder every time it's called
4. Transfer the post-processed files to the `repository_path` in `google-cloud-java`
5. Perform `git diff`
6. Run `compare-poms.py` to compare `pom.xml` files ignoring element order and whitespace
7. If `git diff` or `compare-poms.py` found differences, exit with non-zero and stop the test suite
8. Run for all libraries

### Post-processing workflow:
1. Prepare output from pre-processed library into a folder structure that mimics googleapis-gen (which is the way it's done in other usages, such as [`new-library.py`](https://github.com/googleapis/google-cloud-java/blob/389ac8dca201302a2ceb2299b32eb69fd0d79077/generation/new_client/new-client.py#L266). For example, `java-asset` would be prepared in a folder tree "`google/cloud/asset/v2`"
9. This folder is then sent as a volume/argument to a `owlbot-cli copy-code`
10. The files are arranged as specified in the library's pre-existing `.OwlBot.yaml`. The pre-existing necessary files are specified with `repository_path` (which can be, for example, `google-cloud-java/java-asset`). The files are then sent to an `owl-bot-staging` folder
11. The `owl-bot-staging` folder is sent to a `workspace` folder (root directory the owlbot postprocessor works with)
12. We transfer all pre-existing `pom.xml` files and `owlbot.py` to the `workspace` folder
13. We run the post-processor in the prepared `workspace` folder
14. ~We run `apply_current_versions.sh` (a modified file [borrowed](https://github.com/googleapis/google-cloud-java/blob/389ac8dca201302a2ceb2299b32eb69fd0d79077/generation/apply_current_versions.sh) from google-cloud-java)~


library_generation/run.sh
JoeWang1127, line 1: Should we include this file in the PR?

library_generation/generate_library_integration_test.sh
```diff
@@ -105,24 +111,23 @@ echo "Generating library from $proto_path, to $destination_path..."
 --transport "$transport" \
 --rest_numeric_enums "$rest_numeric_enums" \
 --include_samples "$include_samples" \
+--enable_postprocessing "true" \
+--repo_metadata_json_path $repo_metadata_json_path \
+--owlbot_sha '3a95f1b9b1102865ca551b76be51d2bdb850900c4db2f6d79269e7af81ac8f84' \
```
JoeWang1127, line 116: We should get owlbot sha from .github/.OwlBot.lock.yaml in google-cloud-java

library_generation/run.sh
diegomarquezp: Thanks for the catch. I removed it

library_generation/generate_library_integration_test.sh
diegomarquezp: I added inference logic in the post-processing section

JoeWang1127 (changes requested):
Please modify `library_generation/README.md` to reflect the interface changes.

diegomarquezp:
> @diegomarquezp could you reformat the code in accordance with https://google.github.io/styleguide/shellguide.html?
> 
> The obvious ones are [quoting](https://google.github.io/styleguide/shellguide.html#s5.7-quoting) and [case statement](https://google.github.io/styleguide/shellguide.html#case-statement)

Sure thing, I missed this as a preparation step for review.

blakeli0:
Let's try to merge https://github.com/googleapis/sdk-platform-java/pull/1935 first before merging this. Since the showcase module generation is changed to use the script, the changes in this PR would affect showcase as well, we would probably need more changes to accommodate showcase module.

diegomarquezp:
> @diegomarquezp could you reformat the code in accordance with https://google.github.io/styleguide/shellguide.html?
> 
> The obvious ones are [quoting](https://google.github.io/styleguide/shellguide.html#s5.7-quoting) and [case statement](https://google.github.io/styleguide/shellguide.html#case-statement)

@JoeWang1127 I did most of necessary formatting (i.e. quoting variables with braces and constants) but couldn't quite get the problem with `case` - it looks as suggested by the formatting guide.

(not intended to start a re-review)

library_generation/post-processing/postprocessing_utilities.sh
```diff
@@ -0,0 +1,87 @@
...
+# calls several scripts to perform additional post processing after owlbot is
+# used. If google-cloud-java is downloaded in the root script location, then
+# more processing is performed.
+function other_post_processing_scripts {
+  scripts_root=$1
+  workspace=$2
+  repo_metadata_json_path=$3
+  # postprocessor cleanup
+  bash "${scripts_root}/post-processing/update_owlbot_postprocessor_config.sh" "${workspace}"
+  bash "${scripts_root}/post-processing/delete_non_generated_samples.sh" "${workspace}"
```
meltsufin, line 61: I understand you moved a lot of the code over from new client generation, but do we need all of this functionality here? Should we try to move only the necessary code over?

diegomarquezp: I confirmed a couple scripts were not necessary (generated `java-policysimulator` and `java-alloydb` with differences only in the README (explained in the PR description).

diegomarquezp:
I found out that since the scripts replace the poms, the generation of alloydb disregards the existence of other apis (e.g. generating `alloydb/v1` will remove `v1beta1` from the poms with the current setup).

library_generation/post-processing/postprocessing_utilities.sh
diegomarquezp: I'm afraid I didn't confirm this in the right way. `consolidate_config.sh` and `set_parent_pom.sh` are necessary. 
`consolidate_config` removes things such as `<developer>` section from the generated poms
`set_parent_pom` sets the poms to reference `google-cloud-jar/pom-parent`

library_generation/generate_library.sh
```diff
@@ -78,6 +90,10 @@ if [ -z "${include_samples}" ]; then
   include_samples="true"
 fi
 
+if [ -z "$enable_postprocessing" ]; then
+  enable_postprocessing="false"
```
JoeWang1127, line 94: Is it better to set the default value to `true`?

JoeWang1127:
Could you modify `library_generation/README.md` to reflect the change as well?

JoeWang1127 (changes requested):
Please fix the ci checks.

A quick question: what does it mean by providing a `destination_path` while set `enable_postprossing` to `true`? Will it the generated library reside in `destination_path`?

library_generation/test/generate_library_integration_test.sh
```diff
@@ -74,23 +87,31 @@ echo "Generating library from ${proto_path}, to ${destination_path}..."
...
 pushd "${output_folder}"
-sparse_clone "${googleapis_gen_url}" "${proto_path}"
 
-echo "Compare generation result..."
 RESULT=0
-diff -r "googleapis-gen/${proto_path}/${destination_path}" "${output_folder}/${destination_path}" -x "*gradle*" || RESULT=$?
-
-if [ ${RESULT} == 0 ] ; then
-  echo "SUCCESS: Comparison finished, no difference is found."
+diff -r "google-cloud-java/${monorepo_folder}" "${destination_path}/workspace" \
```
JoeWang1127, line 100: Could you keep the logic of testing against `googleapis-gen` if `enable_postprocessing` is set to `true`?

library_generation/generate_library.sh
diegomarquezp: Modified

diegomarquezp:
https://github.com/googleapis/google-cloud-java/commit/4c9115650a51b522432cfa198d6922311a2568e7 causes different output in poms (artifact will be `apigeeconnect` instead of `apigee-connect`)

diegomarquezp:
https://github.com/googleapis/google-cloud-java/commit/5c5078db41a7c7bcfe107815088df16bdd5a07df#diff-f475a3a9d1237c0f96b9e69f817bf8a42cf682a743f1427463410c6f8f2a66a8 added `accesscontextmanager` which is not currently handled by the script

library_generation/test/compare_poms.py
```diff
@@ -0,0 +1,118 @@
...
+    print_counter(tree2_counter)
+    eprint('tree2')
+    print_counter(tree1_counter)
+  if len(only_in_tree1) > 0 or len(only_in_tree2) > 0:
+    eprint('only in ' + file1)
+    print_counter(only_in_tree1)
+    eprint('only in ' + file2)
+    print_counter(only_in_tree2)
+    return True
+  else:
```
JoeWang1127, line 97: This `else` is redundant.

library_generation/test/compare_poms.py
```diff
@@ -0,0 +1,118 @@
...
+
+  file1 = sys.argv[1]
+  file2 = sys.argv[2]
+  print_whole_trees = sys.argv[3]
+  has_diff = compare_xml(file1, file2, print_whole_trees)
+
+  if has_diff:
+    eprint(f'The poms are different')
+    sys.exit(1)
+  else:
```
JoeWang1127, line 114: This `else` is redundant.

library_generation/README.md
```diff
@@ -150,8 +156,43 @@ Use `--include_samples` to specify the value.
...
+### repo_metadata_json_path (optional)
+Custom `.repo-metadata.json` file to be used in the library generation (this is
+consumed by owlbot)
+If not provided, it will default to the one contained in `repository_path`
+
+### owlbot_sha (optional)
+Owlbot image sha to be used in post-processing
+If not provided, it will default to the one contained in `repository_path`
+
+### repository_path (optional)
```
JoeWang1127, line 173: Do you think it's possible to only have `destination_path`, rather than having two parameters having similar meaning?

After enable post processing, we can create a temp folder to store the pre-processed library. In order to use owlbot cli, the temp folder needs to satisfy a path prefix, which can be inferred from `proto_path`.

meltsufin (commented):
Looks good overall. Just some minor comments.
Also, can you share the performance of the generation process?

library_generation/test/compare_poms.py
diegomarquezp: removed

library_generation/test/compare_poms.py
diegomarquezp: removed

.github/workflows/verify_library_generation.yaml
```diff
@@ -22,11 +23,22 @@ jobs:
         java-version: ${{ matrix.java }}
         distribution: temurin
         cache: maven
+    - uses: actions/setup-python@v4
+      with:
+        python-version: '3.11'
+    - uses: docker-practice/actions-setup-docker@d7be7913d15a6b821cc3ed75b4c03b48e1dde22c
```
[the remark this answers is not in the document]
diegomarquezp: This is docker for ubuntu. We need it since owlbot (cli and postprocessor) runs in docker.

.github/workflows/verify_library_generation.yaml
```diff
@@ -22,11 +23,22 @@ jobs:
...
+      if: matrix.os == 'ubuntu-22.04'
+      timeout-minutes: 12
+    - uses: douglascamata/setup-docker-macos-action@a9f6a5bfeceee5f4bd081e32403859394d2538e1
```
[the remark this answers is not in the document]
diegomarquezp: same answer for macOS. We need this to run owlbot

library_generation/README.md
diegomarquezp: The only interesting situation is with `apigee-connect`. We have the [monorepo-folder](https://github.com/googleapis/google-cloud-java/tree/main/java-apigee-connect) called `apigee-connect`, whereas the [googleapis-gen folder](https://github.com/googleapis/googleapis-gen/tree/master/google/cloud/apigeeconnect/v1/google-cloud-apigeeconnect-v1-java) is `apigeeconnect`. 

This could be solved by inspecting the `repo_short` entry in .repo-metadata.json

library_generation/test/resources/proto_path_list.txt
```diff
@@ -1,19 +1,24 @@
...
-google/cloud/videointelligence/v1p3beta1 google-cloud-videointelligence-v1p3beta1-java
-google/example/library/v1 google-cloud-example-library-v1-java
-google/devtools/containeranalysis/v1 google-cloud-devtools-containeranalysis-v1-java
-google/firestore/bundle google-cloud-firestore-bundle-v1-java
-google/iam/v1 google-iam-v1-java
-google/iam/credentials/v1 google-cloud-iam-credentials-v1-java
-google/logging/v2 google-cloud-logging-v2-java
-google/pubsub/v1 google-cloud-pubsub-v1-java
-google/storage/v2 google-cloud-storage-v2-java
+# proto_path destination_path repository_path is_handwritten
```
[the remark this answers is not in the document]
diegomarquezp: In most cases, yes. I can have the monorepo path as the source of the rest of the parameters (repo metadata, destination path, etc). There is a similar discussion [here](https://github.com/googleapis/sdk-platform-java/pull/1951#discussion_r1368865778)

library_generation/README.md
diegomarquezp: On a second note, it is also possible to get the destination path via `googleapis-gen` following `proto_path`. That will lead us to the `destination_path` already set in `proto_path_list` (e.g. `google-cloud-asset-v1-java`).

blakeli0: I agree with Joe as well, we should try to only use one path. I still don't quite get the need of an additional repo path, can you please explain more about it?

library_generation/post_processing/postprocessing_utilities.sh
```diff
@@ -0,0 +1,132 @@
...
+    --source-repo-commit-hash=none \
+    --source-repo=/pre-processed-libraries \
+    --config-file=.OwlBot.yaml
+
+
+  echo 'running owl-bot post-processor'
+  # run the postprocessor
+  docker run --rm -v "${workspace}:/workspace" --user $(id -u):$(id -g) "${owlbot_postprocessor_image}"
+
+  # get existing versions.txt from downloaded repository
```
blakeli0, line 124: Why do we need to do this in a separate script? I thought it is an existing feature of owlbot that versions in pom are kept as they are?

library_generation/test/compare_poms.py
blakeli0, line 1: This Python script is fine for now for testing purpose. But we should consider a different way of testing before releasing it officially, ideally we should not have a dependency on Python in this repo.

library_generation/README.md
```diff
@@ -150,8 +156,43 @@ Use `--include_samples` to specify the value.
...
+Whether to enable the post-processing steps (usage of owlbot) in the generation
+of this library
+Default is "true".
+
+### repo_metadata_json_path (optional)
+Custom `.repo-metadata.json` file to be used in the library generation (this is
+consumed by owlbot)
+If not provided, it will default to the one contained in `repository_path`
+
+### owlbot_sha (optional)
```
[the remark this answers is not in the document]
diegomarquezp: That's correct, the only case where we send a custom one is for new libraries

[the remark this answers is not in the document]
diegomarquezp: I think we can spare it for now

library_generation/post_processing/postprocessing_utilities.sh
```diff
@@ -0,0 +1,132 @@
...
+
+# Runs the owlbot post-processor docker image.
+# Arguments
+# 1 - workspace: the location of the grpc,proto and gapic libraries to be
+# processed
+# 2 - owlbot_sha: docker image sha that specifies the postprocessor version to
+# be used
+# 3 - repo_metadata_json_path: contains metadata about the library, used by
+# owlbot
+# 4 - include_samples: used to tell if samples are being generated/processed
```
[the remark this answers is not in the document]
diegomarquezp: It was necessary before owlbot-copy. I left them here by mistake. Thanks for the catch

library_generation/README.md
diegomarquezp: I added this logic: we infer `destination_path` from `googleapis-gen` at the specified `proto_path`

library_generation/post_processing/postprocessing_utilities.sh
diegomarquezp: Indeed owlbot has some logic [here](https://github.com/googleapis/synthtool/blob/d52e638b37b091054c869bfa6f5a9fedaba9e0dd/docker/owlbot/java/src/fix-poms.py#L300-L305) - I'm trying to confirm this is actually meant to update existing poms

diegomarquezp:
> @blakeli0 
I agree with Joe as well, we should try to only use one path. I still don't quite get the need of an additional repo path, can you please explain more about it?

I ended up using `repository_path` and infer `destination_path` from `googleapis-gen` + `proto_path`

library_generation/post_processing/postprocessing_utilities.sh
diegomarquezp: I tried the whole `versions.txt` from `google-cloud-java` and the pom ended with versions of other libraries as dependencies. I'm trying to trim the versions we need only and get rid of this script call

.github/workflows/verify_library_generation.yaml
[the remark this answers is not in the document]
meltsufin: The default GitHub Action runner should have docker already installed. See: https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md.

.github/workflows/verify_library_generation.yaml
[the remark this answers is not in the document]
meltsufin: Can you just do `brew install docker`.
I'm a bit concerned about using third-party images with access to secrets.

library_generation/post_processing/postprocessing_utilities.sh
diegomarquezp: Solution: use `versions.txt` at the root of the owlbot image filesystem. I removed this script and its call

diegomarquezp:
> Looks good overall. Just some minor comments. Also, can you share the performance of the generation process?

@meltsufin I added a small time-tracking step and the results were the following

```
+ echo 'generation times in seconds:'
+ cat /home/runner/work/sdk-platform-java/sdk-platform-java/output/generation_times
google-cloud-java/java-apigee-connect 107
google-cloud-java/java-asset 11
google-cloud-java/java-asset 10
google-cloud-java/java-asset 10
google-cloud-java/java-asset 10
google-cloud-java/java-asset 23
google-cloud-java/java-compute 177
google-cloud-java/java-kms 22
google-cloud-java/java-redis 15
google-cloud-java/java-redis 14
google-cloud-java/java-containeranalysis 10
google-cloud-java/java-iam 6
google-cloud-java/java-iamcredentials 10
```

This is only considering the call of `generate_library.sh`, without checking out reposiories for comparison

library_generation/README.md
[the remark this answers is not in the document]
diegomarquezp: I removed them as discussed in our meeting

.github/workflows/verify_library_generation.yaml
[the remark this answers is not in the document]
diegomarquezp: Thanks, I added a manual installation step instead of external actions

.github/workflows/verify_library_generation.yaml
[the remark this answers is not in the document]
diegomarquezp: I added a manual installation step, getting rid of the third party action

library_generation/post_processing/postprocessing_utilities.sh
```diff
@@ -0,0 +1,132 @@
...
+
+  echo 'running owl-bot post-processor'
+  # run the postprocessor
+  docker run --rm -v "${workspace}:/workspace" --user $(id -u):$(id -g) "${owlbot_postprocessor_image}"
+
+  # get existing versions.txt from downloaded repository
+  if [ -d "${output_folder}/google-cloud-java" ];then
+    cp "${output_folder}/google-cloud-java/versions.txt" "${workspace}"
+    pushd "${workspace}"
+    bash "${scripts_root}/post_processing/apply_current_versions.sh"
```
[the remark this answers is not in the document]
diegomarquezp: Makes sense as a separate parameter since it affects the output of the postprocessed library. I implemented this

library_generation/README.md
```diff
@@ -150,8 +156,43 @@ Use `--include_samples` to specify the value.
 Choose the protoc binary type from https://github.com/protocolbuffers/protobuf/releases.
 Default is "linux-x86_64".
 
-## An example to generate a client library
+### enable_postprocessing (optional)
+Whether to enable the post-processing steps (usage of owlbot) in the generation
+of this library
+Default is "true".
+
+### repo_metadata_json_path (optional)
```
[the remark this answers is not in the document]
diegomarquezp: I removed this file in favor of existing libraries already containing this file

meltsufin (approved):

library_generation/generate_library.sh
```diff
@@ -280,5 +295,30 @@ popd # output_folder
...
+# post-processing
+#####################################################
+source "${script_dir}/post_processing/postprocessing_utilities.sh"
+if [ "${enable_postprocessing}" != "true" ];
+then
+  echo "post processing is disabled"
+  exit 0
+fi
+if [ -z "${versions_file}" ];then
+  echo "no versions.txt argument provided. Please provide one in order to enable post-processing"
```
blakeli0, line 309: It's OK to make it required for now since it's only for existing libraries. For new client libraries, I'm assuming we have to generate versions.txt first before calling this script? Is it generated by new client library script currently? cc @JoeWang1127

library_generation/post_processing/postprocessing_utilities.sh
```diff
@@ -0,0 +1,129 @@
...
+}
+
+# Runs the owlbot post-processor docker image.
+# Arguments
+# 1 - workspace: the location of the grpc,proto and gapic libraries to be
+# processed
+# owlbot
+# 4 - scripts_root: location of the generation scripts
+# 5 - destination_path: used to transfer the raw grpc, proto and gapic libraries
+# 6 - repository_path: path from output_folder to the location of the source of
```
blakeli0, line 52: I still don't fully understand the usefulness of `repository_path`? If it's for the poms, they should be already in the output folders, am I missing something?

library_generation/post_processing/postprocessing_utilities.sh
```diff
@@ -0,0 +1,129 @@
...
+# scripts
+
+# returns the metadata json path if given, or defaults to the one found in
+# $repository_path
+# Arguments
+# 1 - repository_path: path from output_folder to the location of the library
+# containing .repo-metadata. It assumes the existence of google-cloud-java in
+# the output folder
+# 2 - output_folder: root for the generated libraries, used in conjunction with
+get_repo_metadata_json() {
```
blakeli0, line 13: Can we have some unit tests for this function and the function below?

library_generation/post_processing/postprocessing_utilities.sh
blakeli0, line 1: This shell is not a utility, and it's actually being used in generate_library.sh, can we call it something like `postprocess_library.sh`? Also it might be subjective, but I don't think it has to be in a sub folder as well.

library_generation/generate_library.sh
diegomarquezp: @blakeli0 the version of a new library [defaults](https://github.com/googleapis/synthtool/blob/247239027cad7678adea8f57e529ebd12600ee93/docker/owlbot/java/src/fix-poms.py#L333) to `0.0.1-SNAPSHOT`.

[example PR](https://github.com/googleapis/google-cloud-java/pull/9993/files)

The new library script [applies current versions](https://github.com/googleapis/google-cloud-java/blob/389ac8dca201302a2ceb2299b32eb69fd0d79077/generation/new_client/new-client.py#L364) but this is for other referenced artifacts such as `google-cloud-java` ([example](https://github.com/googleapis/google-cloud-java/blob/9bf5aae4202c5e6c7c3e23e0ae1d85be849b3a9c/java-redis-cluster/google-cloud-redis-cluster-bom/pom.xml#L12))

I think we still need a versions file for referenced artifacts other than the library being generated when dealing with a new monorepo library

library_generation/post_processing/postprocessing_utilities.sh
diegomarquezp: @blakeli0 the pre-existing poms with their folder structure are indeed necessary for the postprocessor. 

We need `google-cloud-java` to be present beforehand in `output_folder` as well, but the script does not know which folder in the monorepo should obtain the poms, `owlbot.py` and `.OwlBot.yaml` from. I don't think there is a good heuristic to infer the monorepo path from `proto_path` or `destination_path`. For example looking for `.OwlBot.yaml` files that have a `deep-copy` containing `destination_path` may work but we have cases where the owlbot files do not deal with a single gapic library.

To clarify, the purpose of `repository_path` is to give the script a way to find the monorepo folder associated with the library being generated. This folder contains the necessary files for owlbot copy and postprocessor.

library_generation/post_processing/postprocessing_utilities.sh
diegomarquezp: the subfolder made more sense when we have several helper files. Since the number of files has been reduced to 1, it now makes sense to have it in `library_generation`.

I think I can move the helper functions to `utilities.sh`

library_generation/post_processing/postprocessing_utilities.sh
diegomarquezp: added them

library_generation/post_processing/postprocessing_utilities.sh
diegomarquezp: I made a small refactor of the postprocessing file and utilities. Postprocessing went from a function to a script file (called with `bash ...`) and moved the helpers to `utilities.sh`

library_generation/post_processing/postprocessing_utilities.sh
diegomarquezp: As discussed, I removed usage of `destination_path`. The integration test, however, will pass `repository_path` (e.g. java-asset) as `destination_path`. 

NOTE: The usage of owlbot copy works with a yaml file [expecting](https://github.com/googleapis/google-cloud-java/blob/474eb74f8dd6d0c7abf480c5212da44a006272bb/java-asset/.OwlBot.yaml#L28) a certain input folder structure. For example, that yaml has a `deep-copy` section with wildcards in order to build the `owl-bot-staging` folder (i.e. owlbot copy sends the files stored in `deep-copy` to `owlbot-staging` and the owlbot postprocessor runs on them)
I'm taking advantage of the wildcards for owlbot copy to work. The path in this deep-copy section is referring a path in googleapis-gen ([example](https://github.com/googleapis/googleapis-gen/tree/master/google/cloud/asset/v1)) which is could be something like `google-cloud-asset-v1-java` but I just named it `generated-java` (which will be ok for the wildcard)

diegomarquezp: I'm yet to update the readme

library_generation/test/resources/google-cloud-java/java-asset/.repo-metadata.json
blakeli0, line 1: The file name is probably hardcoded in the script so we can not change it. Can we change the folder name `google-cloud-java` and `java-asset`  to something more test specific? Maybe something like `test-monorepo` and `test-service`?

library_generation/post_processing/postprocessing_utilities.sh
blakeli0: Thanks!

library_generation/test/resources/google-cloud-java/java-asset/.repo-metadata.json
diegomarquezp: I changed it to `test-monorepo/test-service`

blakeli0 (approved):

JoeWang1127 (approved):

PR #1963, opened 2023-09-06
feat: Adding vendor and vendor information in header

ddixit14:
This PR adds java vendor information to the "gl-java" header value. Only when the information is present, add it to `javaRuntimeInformation`.

Local testing results for various combinations- 

(1)
the vendor is -> Eclipse Adoptium
the vendor version is -> Temurin-11.0.19+7
**the javaRuntimeInformation is -> 11.0.19__Eclipse-Adoptium__Temurin-11.0.19-7**

(2)
the vendor is ->Amazon.com Inc.
the vendor version is ->Corretto-11.0.19.7.1
**the javaRuntimeInformation is ->11.0.19__Amazon.com-Inc.__Corretto-11.0.19.7.1**

(3)
the vendor is ->Eclipse Adoptium
the vendor version is ->Temurin-11.0.20.1+1
**the javaRuntimeInformation is ->11.0.20.1__Eclipse-Adoptium__Temurin-11.0.20.1-1**

(4) [Case when vendor version is not available]
the vendor is ->Oracle Corporation
the vendor version is ->null
**the javaRuntimeInformation is ->20.0.1__Oracle-Corporation**

### DIFF
--- a/gax-java/gax/src/main/java/com/google/api/gax/core/GaxProperties.java
+++ b/gax-java/gax/src/main/java/com/google/api/gax/core/GaxProperties.java
@@ -29,10 +29,8 @@
  */
 package com.google.api.gax.core;
 
-import static org.graalvm.nativeimage.ImageInfo.PROPERTY_IMAGE_CODE_KEY;
-import static org.graalvm.nativeimage.ImageInfo.PROPERTY_IMAGE_CODE_VALUE_RUNTIME;
-
 import com.google.api.core.InternalApi;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
 import java.io.IOException;
 import java.io.InputStream;
@@ -85,12 +83,6 @@ public class GaxProperties {
 
   /** Returns the version of the running JVM */
   public static String getJavaVersion() {
-    // When running the application as a native image, append `-graalvm` to the
-    // version.
-    String imageCode = System.getProperty(PROPERTY_IMAGE_CODE_KEY);
-    if (imageCode != null && imageCode.equals(PROPERTY_IMAGE_CODE_VALUE_RUNTIME)) {
-      return System.getProperty("java.version") + "-graalvm";
-    }
     return JAVA_VERSION;
   }
 
@@ -99,8 +91,12 @@ public class GaxProperties {
     return GAX_VERSION;
   }
 
-  /** Returns the current runtime version */
-  private static String getRuntimeVersion() {
+  /**
+   * Returns the current runtime version. For GraalVM the values in this method will be fetched at
+   * build time and the values should not differ from runtime (executable)
+   */
+  @VisibleForTesting
+  static String getRuntimeVersion() {
     String javaRuntimeInformation = System.getProperty("java.version");
 
     // append the vendor information to the java-version if vendor is present.
--- a/gax-java/gax/src/main/java/com/google/api/gax/core/GaxProperties.java
+++ b/gax-java/gax/src/main/java/com/google/api/gax/core/GaxProperties.java
@@ -33,6 +33,7 @@ import static org.graalvm.nativeimage.ImageInfo.PROPERTY_IMAGE_CODE_KEY;
 import static org.graalvm.nativeimage.ImageInfo.PROPERTY_IMAGE_CODE_VALUE_RUNTIME;
 
 import com.google.api.core.InternalApi;
+import com.google.common.base.Strings;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.Properties;
@@ -100,23 +101,22 @@ public class GaxProperties {
 
   /** Returns the current runtime version */
   private static String getRuntimeVersion() {
-    String javaVersion = System.getProperty("java.version");
-    String combined = javaVersion;
+    String javaRuntimeInformation = System.getProperty("java.version");
 
     // append the vendor information to the java-version if vendor is present.
     String vendor = System.getProperty("java.vendor");
-    if (vendor != null && !vendor.isEmpty()) {
-      combined = String.format("%s__%s", combined, vendor);
+    if (!Strings.isNullOrEmpty(vendor)) {
+      javaRuntimeInformation = String.format("%s__%s", javaRuntimeInformation, vendor);
     }
 
     // appends the vendor version information to the java-version if vendor version is present.
     String vendorVersion = System.getProperty("java.vendor.version");
-    if (vendorVersion != null && !vendorVersion.isEmpty()) {
-      combined = String.format("%s__%s", combined, vendorVersion);
+    if (!Strings.isNullOrEmpty(vendorVersion)) {
+      javaRuntimeInformation = String.format("%s__%s", javaRuntimeInformation, vendorVersion);
     }
 
     // replacing all characters that are not numbers, letters, underscores, periods, or backslashes
     // with hyphens.
-    return combined.replaceAll("[^0-9a-zA-Z_\\\\.]", "-");
+    return javaRuntimeInformation.replaceAll("[^0-9a-zA-Z_\\\\.]", "-");
   }
 }
--- a/gax-java/gax/src/main/java/com/google/api/gax/core/GaxProperties.java
+++ b/gax-java/gax/src/main/java/com/google/api/gax/core/GaxProperties.java
@@ -93,24 +93,22 @@ public class GaxProperties {
 
   /**
    * Returns the current runtime version. For GraalVM the values in this method will be fetched at
-   * build time and the values should not differ from runtime (executable)
+   * build time and the values should not differ from the runtime (executable)
    */
   @VisibleForTesting
   static String getRuntimeVersion() {
-    String javaRuntimeInformation = System.getProperty("java.version");
+    String javaRuntimeInformation = System.getProperty("java.version", "null");
 
     // append the vendor information to the java-version if vendor is present.
     String vendor = System.getProperty("java.vendor");
     if (!Strings.isNullOrEmpty(vendor)) {
       javaRuntimeInformation = String.format("%s__%s", javaRuntimeInformation, vendor);
+      // appends the vendor version information to the java-version if vendor version is present.
+      String vendorVersion = System.getProperty("java.vendor.version");
+      if (!Strings.isNullOrEmpty(vendorVersion)) {
+        javaRuntimeInformation = String.format("%s__%s", javaRuntimeInformation, vendorVersion);
+      }
     }
-
-    // appends the vendor version information to the java-version if vendor version is present.
-    String vendorVersion = System.getProperty("java.vendor.version");
-    if (!Strings.isNullOrEmpty(vendorVersion)) {
-      javaRuntimeInformation = String.format("%s__%s", javaRuntimeInformation, vendorVersion);
-    }
-
     // replacing all characters that are not numbers, letters, underscores, periods, or backslashes
     // with hyphens.
     return javaRuntimeInformation.replaceAll("[^0-9a-zA-Z_\\\\.]", "-");
--- a/gax-java/gax/src/main/java/com/google/api/gax/core/GaxProperties.java
+++ b/gax-java/gax/src/main/java/com/google/api/gax/core/GaxProperties.java
@@ -100,6 +100,23 @@ public class GaxProperties {
 
   /** Returns the current runtime version */
   private static String getRuntimeVersion() {
-    return System.getProperty("java.version");
+    String javaVersion = System.getProperty("java.version");
+    String combined = javaVersion;
+
+    // append the vendor information to the java-version if vendor is present.
+    String vendor = System.getProperty("java.vendor");
+    if (vendor != null && !vendor.isEmpty()) {
+      combined = String.format("%s__%s", combined, vendor);
+    }
+
+    // appends the vendor version information to the java-version if vendor version is present.
+    String vendorVersion = System.getProperty("java.vendor.version");
+    if (vendorVersion != null && !vendorVersion.isEmpty()) {
+      combined = String.format("%s__%s", combined, vendorVersion);
+    }
+
+    // replacing all characters that are not numbers, letters, underscores, periods, or backslashes
+    // with hyphens.
+    return combined.replaceAll("[^0-9a-zA-Z_\\\\.]", "-");
   }
 }
--- a/gax-java/gax/src/test/java/com/google/api/gax/core/GaxPropertiesTest.java
+++ b/gax-java/gax/src/test/java/com/google/api/gax/core/GaxPropertiesTest.java
@@ -84,7 +84,7 @@ public class GaxPropertiesTest {
   }
 
   @Test
-  public void testGetJavaRuntimeInfoCaseGraalVM() {
+  public void testGetJavaRuntimeInfo_graalVM() {
 
     System.setProperty("java.version", "17.0.3");
     System.setProperty("java.vendor", "GraalVM Community");
@@ -95,7 +95,7 @@ public class GaxPropertiesTest {
   }
 
   @Test
-  public void testGetJavaRuntimeInfoCaseTemurin() {
+  public void testGetJavaRuntimeInfo_temurin() {
     System.setProperty("java.version", "11.0.19");
     System.setProperty("java.vendor", "Eclipse Adoptium");
     System.setProperty("java.vendor.version", "Temurin-11.0.19+7");
@@ -105,7 +105,7 @@ public class GaxPropertiesTest {
   }
 
   @Test
-  public void testGetJavaRuntimeInfoCaseCoretto() {
+  public void testGetJavaRuntimeInfo_coretto() {
     System.setProperty("java.version", "11.0.19");
     System.setProperty("java.vendor", "Amazon.com Inc.");
     System.setProperty("java.vendor.version", "Corretto-11.0.19.7.1");
@@ -115,7 +115,7 @@ public class GaxPropertiesTest {
   }
 
   @Test
-  public void testGetJavaRuntimeInfoCaseOracle() {
+  public void testGetJavaRuntimeInfo_oracle() {
     System.setProperty("java.version", "20.0.1");
     System.setProperty("java.vendor", "Oracle Corporation");
     // case where java.vendor.version is null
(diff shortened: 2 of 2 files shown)


gax-java/gax/src/main/java/com/google/api/gax/core/GaxProperties.java
```diff
@@ -100,6 +100,23 @@ public static String getGaxVersion() {
 
   /** Returns the current runtime version */
   private static String getRuntimeVersion() {
```
blakeli0, line 102: This is pretty complicated logic, can we add some unit tests for it? Since this is a private method, we can test through `getJavaVersion()`.

gax-java/gax/src/main/java/com/google/api/gax/core/GaxProperties.java
```diff
@@ -100,6 +100,23 @@ public static String getGaxVersion() {
...
-    return System.getProperty("java.version");
+    String javaVersion = System.getProperty("java.version");
```
JoeWang1127, line 103: `javaVersion` is only used once, why not using `combined` instead?

gax-java/gax/src/main/java/com/google/api/gax/core/GaxProperties.java
```diff
@@ -100,6 +100,23 @@ public static String getGaxVersion() {
...
+    String combined = javaVersion;
+
+    // append the vendor information to the java-version if vendor is present.
+    String vendor = System.getProperty("java.vendor");
+    if (vendor != null && !vendor.isEmpty()) {
```
blakeli0, line 108: This can be simplified by Guava's `Strings.isNullOrEmpty`, similar to [here](https://github.com/googleapis/sdk-platform-java/blob/6e9ef3f157957a57d5e1c06cee7e755054db6af0/gapic-generator-java/src/main/java/com/google/api/generator/gapic/model/Service.java#L52)

blakeli0:
Do we need to append this info for [graalvm](https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/gax/src/main/java/com/google/api/gax/core/GaxProperties.java#L91) as well?

gax-java/gax/src/main/java/com/google/api/gax/core/GaxProperties.java
meltsufin: +1 Can we have some tests? Also, some integration tests would be nice. If you want to get fancy, you can do a shell script that installs a couple of versions of Java and tests against those.

ddixit14:
to-do : Manually confirm graalvm-behaviour (change to graalvm, print the result of gaxProperties getJavaVersion().

suztomo:
ci / build(8) except for gapic-generator-java (pull_request) Failing after 1m

gax-java/gax/src/test/java/com/google/api/gax/core/GaxPropertiesTest.java
```diff
@@ -57,10 +57,85 @@ public void testGaxVersion() {
     }
   }
 
+  private static String originalJavaVersion = System.getProperty("java.version");
+  private static String originalJavaVendor = System.getProperty("java.vendor");
+  private static String originalJavaVendorVersion = System.getProperty("java.vendor.version");
+
+  public static void cleanup() {
```
suztomo, line 64: Use `@After` https://junit.org/junit4/javadoc/4.13/org/junit/After.html to avoid calling `cleanup()`.

gax-java/gax/src/test/java/com/google/api/gax/core/GaxPropertiesTest.java
```diff
@@ -57,10 +57,85 @@ public void testGaxVersion() {
...
+    assertEquals("17.0.3__GraalVM-Community__GraalVM-CE-22.1.0", runtimeInfo);
+    cleanup();
+  }
+
   @Test
-  public void testGetVersion_nativeImage() {
-    System.setProperty(PROPERTY_IMAGE_CODE_KEY, PROPERTY_IMAGE_CODE_VALUE_RUNTIME);
-    String javaVersion = GaxProperties.getJavaVersion();
-    assertTrue(javaVersion.endsWith("-graalvm"));
+  public void testGetJavaRuntimeInfoCaseOne() {
```
suztomo, line 97: Avoid nameing "CaseOne". It doesn't tell what it is testing.

gax-java/gax/src/main/java/com/google/api/gax/core/GaxProperties.java
```diff
@@ -98,8 +91,28 @@ public static String getGaxVersion() {
     return GAX_VERSION;
   }
 
-  /** Returns the current runtime version */
-  private static String getRuntimeVersion() {
-    return System.getProperty("java.version");
+  /**
+   * Returns the current runtime version. For GraalVM the values in this method will be fetched at
+   * build time and the values should not differ from runtime (executable)
```
suztomo, line 96:
```suggestion
   * build time and the values should not differ from the runtime (executable).
```

gax-java/gax/src/main/java/com/google/api/gax/core/GaxProperties.java
```diff
@@ -98,8 +91,28 @@ public static String getGaxVersion() {
...
+   */
+  @VisibleForTesting
+  static String getRuntimeVersion() {
+    String javaRuntimeInformation = System.getProperty("java.version");
```
meltsufin, line 100: Can `java.version` be `null`?

ddixit14: I think there is a possibility of this happening. In such a case, would we still want to capture the vendor information, since the vendor version might help us deduce java-version?

ddixit14:
> Do we need to append this info for [graalvm](https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/gax/src/main/java/com/google/api/gax/core/GaxProperties.java#L91) as well?

Handled graalvm case as well.

gax-java/gax/src/main/java/com/google/api/gax/core/GaxProperties.java
suztomo: In such a case, I want this method not to fail with NullPointerException. @ddixit14  Can you add a test case for it?

gax-java/gax/src/test/java/com/google/api/gax/core/GaxPropertiesTest.java
```diff
@@ -57,10 +58,81 @@ public void testGaxVersion() {
...
+    System.setProperty("java.vendor", "Oracle Corporation");
+    // case where java.vendor.version is null
+    System.clearProperty("java.vendor.version");
+
+    String runtimeInfo = GaxProperties.getRuntimeVersion();
+    assertEquals("20.0.1__Oracle-Corporation", runtimeInfo);
+  }
+
+  @Test
+  public void testGetJavaRuntimeInfoCaseNullValues() {
```
suztomo, line 129:
```suggestion
  public void testGetJavaRuntimeInfoCaseNullVendorVersion() {
```

gax-java/gax/src/test/java/com/google/api/gax/core/GaxPropertiesTest.java
blakeli0: +1. We have a pretty good [internal guide](https://engdoc.corp.google.com/eng/doc/devguide/testing/unit/best-practices.md?cl=head&polyglot=java#naming) for unit test naming conventions. This could be something like `testGetRuntimeVersion_shouldIncludeVendorVersionIfItExists` or `testGetRuntimeVersion_vendorVersionExist_returnsWithVendorVersion`.

gax-java/gax/src/main/java/com/google/api/gax/core/GaxProperties.java
```diff
@@ -98,8 +91,28 @@ public static String getGaxVersion() {
...
+
+    // append the vendor information to the java-version if vendor is present.
+    String vendor = System.getProperty("java.vendor");
+    if (!Strings.isNullOrEmpty(vendor)) {
+      javaRuntimeInformation = String.format("%s__%s", javaRuntimeInformation, vendor);
+    }
+
+    // appends the vendor version information to the java-version if vendor version is present.
+    String vendorVersion = System.getProperty("java.vendor.version");
```
blakeli0, line 109: Is there a case `java.vendor` does not exist but `java.vendor.version` does not? I feel we should only append it if both exist.

suztomo: > we should only append it if both exist.

Right, that's the spec in the document. @ddixit14 Can you take care of this?

suztomo (approved):
Can you check any usage of `getJavaVersion` in java-storage, java-bigtable, java-pubsub, and google-cloud-java (my random picked repositories)?

gax-java/gax/src/main/java/com/google/api/gax/core/GaxProperties.java
ddixit14: The change is updated and pushed.

gax-java/gax/src/test/java/com/google/api/gax/core/GaxPropertiesTest.java
```diff
@@ -57,10 +58,94 @@ public void testGaxVersion() {
...
+
+    if (Strings.isNullOrEmpty(originalJavaVendorVersion)) {
+      System.clearProperty("java.vendor.version");
+    } else {
+      System.setProperty("java.vendor.version", originalJavaVendorVersion);
+    }
+  }
+
+  @Test
+  public void testGetJavaRuntimeInfo_graalVM() {
```
blakeli0, line 87: The three test cases below are testing the same scenario from what I can see, we should only keep one.

gax-java/gax/src/test/java/com/google/api/gax/core/GaxPropertiesTest.java
```diff
@@ -57,10 +58,94 @@ public void testGaxVersion() {
...
+  public void testGetJavaRuntimeInfo_nullJavaVersion() {
+    // We don't expect this case to happen, however we don't want the method to fail when it really
+    // happens.
+    System.clearProperty("java.version");
+    // case where java.vendor and java.vendor.version is null
+    System.setProperty("java.vendor", "oracle");
+    System.setProperty("java.vendor.version", "20.0.1");
+
+    String runtimeInfo = GaxProperties.getRuntimeVersion();
+    assertEquals("null__oracle__20.0.1", runtimeInfo);
```
blakeli0, line 149: Is this expected to have a String `null` in the version info?

ddixit14: yes. If the system property does not exist, we are assigning it a null value.

gax-java/gax/src/test/java/com/google/api/gax/core/GaxPropertiesTest.java
ddixit14: All the three test cases are actual system property values for the various vendors, so just wanted to make sure we are capturing them fine.

blakeli0: It's OK to use real life examples, but we also don't want to duplicate the testing effort. Is the purpose of this unit test to test all real life vendors? Or test the functionality of the method? Someone else should be able to understand/maintain/test the functionality of the method without knowing the real life examples. 
From unit tests' perspective, what are the different scenarios between them? From what I can see, they are all testing the same scenario that vendor and vendor version exist, and replacing spaces with underscores.
In addition, we should really mention the test scenario instead of a real life example in the test name.

gax-java/gax/src/test/java/com/google/api/gax/core/GaxPropertiesTest.java
suztomo: We are assigning it a "null" as String.

gax-java/gax/src/test/java/com/google/api/gax/core/GaxPropertiesTest.java
ddixit14: These are just a few "major" vendors, not all real life vendors. All the tests are not testing the same scenario (vendor and vendor version exist, and replacing spaces with underscores). Test for graalvm takes care of cases where only spaces exist, whereas test for temurin takes care of spaces along with unsupported characters, also because we officially support it. I have removed the coretto one. I have also updated the test name for each.

blakeli0: Thanks! LGTM

blakeli0 (approved):

gax-java/gax/src/test/java/com/google/api/gax/core/GaxPropertiesTest.java
```diff
@@ -57,10 +58,85 @@ public void testGaxVersion() {
...
+
+    if (Strings.isNullOrEmpty(originalJavaVendorVersion)) {
+      System.clearProperty("java.vendor.version");
+    } else {
+      System.setProperty("java.vendor.version", originalJavaVendorVersion);
+    }
+  }
+
+  @Test
+  public void testGetJavaRuntimeInfo_spaces() {
```
suztomo, line 87: Can you rename it back to the Java vendor name? For those that represent real vendor value, the name should have the vendor name.

blakeli0: Is the purpose of this unit test to test real vendor name or test the functionality of the method?

suztomo: @ddixit14 The test name should answer Blake'a question.

ddixit14: I have tried to capture both.

suztomo:
Test case for this?

the vendor is ->Amazon.com Inc.
the vendor version is ->Corretto-11.0.19.7.1
the javaRuntimeInformation is ->11.0.19__Amazon.com-Inc.__Corretto-11.0.19.7.1

gax-java/gax/src/test/java/com/google/api/gax/core/GaxPropertiesTest.java
```diff
@@ -57,10 +58,85 @@ public void testGaxVersion() {
...
+
+    if (Strings.isNullOrEmpty(originalJavaVendorVersion)) {
+      System.clearProperty("java.vendor.version");
+    } else {
+      System.setProperty("java.vendor.version", originalJavaVendorVersion);
+    }
+  }
+
+  @Test
+  public void testGetJavaRuntimeInfo_graalVM_withSpaces() {
```
suztomo, line 87: The name is ambiguous why it's added.

ddixit14: modified it along with adding comments.

ddixit14:
> Test case for this?
> 
> the vendor is ->Amazon.com Inc. the vendor version is ->Corretto-11.0.19.7.1 the javaRuntimeInformation is ->11.0.19__Amazon.com-Inc.__Corretto-11.0.19.7.1

Added this back.

gax-java/gax/src/test/java/com/google/api/gax/core/GaxPropertiesTest.java
```diff
@@ -57,10 +58,108 @@ public void testGaxVersion() {
...
+    String runtimeInfo = GaxProperties.getRuntimeVersion();
+    assertEquals("11.0.19__Amazon.com-Inc.__Corretto-11.0.19.7.1", runtimeInfo);
+  }
+
   @Test
-  public void testGetVersion_nativeImage() {
-    System.setProperty(PROPERTY_IMAGE_CODE_KEY, PROPERTY_IMAGE_CODE_VALUE_RUNTIME);
-    String javaVersion = GaxProperties.getJavaVersion();
-    assertTrue(javaVersion.endsWith("-graalvm"));
+  public void testGetJavaRuntimeInfo_specialCharacters() {
```
suztomo, line 120: The intention is clear. Good test name.

suztomo (approved):

gax-java/gax/src/test/java/com/google/api/gax/core/GaxPropertiesTest.java
blakeli0: I really think that unit tests should only care about the functionality/behavior of the method and [test names](https://g3doc.corp.google.com/eng/doc/devguide/testing/unit/best-practices.md?cl=head&polyglot=java#naming) should include the behavior being tested and expected outcomes. If we only care about the behavior, the scope is limited and each test is easy to understand. If we start to test real vendor names, what's the scope of it? Do we try to test all the available vendors? If not, why do we pick the selected vendors? 
That being said, if you feel strongly about testing with real vendor names, please add comments explaining the why these real vendors are selected and differences between each vendor.

ddixit14: I have separated the real world tests and functionality tests, and given appropriate names and comments. Can you please review them once, and there's still any issue then we can continue the discussion?

ddixit14: As for why we are using real vendor in testing is because this change is being done to capture the customer usage (which version/vendor/vendor.version the customer is using). If we make sure that we are able to capture real vendors information correctly, our usage data will be more accurate. As for the other functionality tests, I have changed the name and commented what is being tested. As for why these 3 real vendors are chosen. (1) graalvm - we support this and we have a specific dashboard for its usage, so for more accuracy. also because I changed its implementation (2) temurin - we officially support this.

meltsufin (approved):

gax-java/gax/src/test/java/com/google/api/gax/core/GaxPropertiesTest.java
blakeli0: Thanks Deepankar! LGTM.

suztomo: I also want real world values used in unit tests. For future code readers, it works as a document of what kind of values the function deals with in real world.

PR #1964, opened 2023-09-06
chore: add unit tests for `generate_library.sh`

JoeWang1127:
In this PR:
- extract utility functions used in testing to `test_utilities.sh`.
- add unit tests against `generate_library.sh`.

A follow up of #1948.


library_generation/test/generate_library_unit_tests.sh
```diff
@@ -33,6 +34,36 @@ __assertFileDoesNotExist() {
   return 1
 }
 
+# Convert OS type to OS architecture, the default OS architecture is
```
meltsufin, line 37: Can you detect the OS in some easy way?

library_generation/test/generate_library_unit_tests.sh
```diff
@@ -219,6 +250,133 @@ download_grpc_plugin_failed_with_invalid_arch_test() {
...
+  local os_architecture
+  local destination="google-cloud-alloydb-v1-java"
+  os_architecture=$(__get_os_architecture "${os_type}")
+  cd "${script_dir}/resources/protos"
+  "${script_dir}"/../generate_library.sh \
+    -p google/cloud/alloydb/v1 \
+    -d ../"${destination}" \
+    --gapic_generator_version 2.24.0 \
+    --protobuf_version 23.2 \
+    --grpc_version 1.55.1 \
```
meltsufin, line 263: Is the idea to keep these fixed forever or to update as new versions are released? If we intend to update, we probably need renovate-bot support here.

meltsufin (commented):
I see a bunch of goldens being added. Is that intentional?

When we were talking about unit tests, I mostly had in mind testing individual functions in `utilities.sh` but I don't really see those tests in this PR.

library_generation/test/generate_library_unit_tests.sh
JoeWang1127: I think the version should be fixed.

JoeWang1127:
> I see a bunch of goldens being added. Is that intentional?

Yes.

> When we were talking about unit tests, I mostly had in mind testing individual functions in `utilities.sh` but I don't really see those tests in this PR.

Unit tests against `utilities.sh` is in #1948.

library_generation/test/generate_library_unit_tests.sh
```diff
@@ -2,12 +2,13 @@
...
 set -xeo pipefail
 
+os_type=$1
 # Variables used to generate final result
 total_num=0
 succeed_num=0
 failed_num=0
 failed_tests=""
-# Unit tests against ./utilities.sh
+# Unit tests against utilities.sh and generate_library.sh
```
blakeli0, line 11: I think it's better to differentiate the "true unit tests" for `utilities.sh` and "golden unit tests" for `generate_library.sh`. Maybe have a separate file for `generate_library_golden_unit_test`? Also mention that what are the scenarios we couldn't test just through the "true unit tests".

library_generation/test/generate_library_unit_tests.sh
blakeli0: I think it depends on what we want to achieve here. In the context of unit test for `generate_library.sh`, I think the versions should be fixed as well because we are testing the logic of the shell script not the generator. 
For integration tests against googleapis-gen or google-cloud-java in the future though, I think we should probably use release-please to update these versions to the latest.

meltsufin (commented):
`generate_library_unit_tests.sh` runs the entire `generate_library.sh` even for testing a small thing like skipping the `protobuf_version` argument. Maybe, it will be easier to create true unit tests by breaking down the code in `generate_library.sh` into smaller functions.

Also, can we do the "golden" tests in a separate PR? I think we should make a distinction between golden tests and unit tests.

library_generation/test/generate_library_unit_tests.sh
JoeWang1127: I didn't use testing framework here so if we create another file for golden unit test for `generate_library.sh` I have to rewrite all test utility functions, like `assertEquals`.

How about I extract the test utilities to a separate file?

library_generation/test/generate_library_unit_tests.sh
```diff
@@ -219,6 +250,133 @@ download_grpc_plugin_failed_with_invalid_arch_test() {
   __assertEquals 1 $((res))
 }
 
+generate_library_success_with_valid_versions() {
+  local os_architecture
+  local destination="google-cloud-alloydb-v1-java"
+  os_architecture=$(__get_os_architecture "${os_type}")
+  cd "${script_dir}/resources/protos"
+  "${script_dir}"/../generate_library.sh \
+    -p google/cloud/alloydb/v1 \
```
blakeli0, line 259: There are over 300 files added as part of this PR, since the protos under test does not matter, can we use a service that has very few messages? Or even better if we can have a minimum test proto?

JoeWang1127: Sure, do you have a better candidate in mind?

blakeli0: Similar to this [test proto](https://github.com/googleapis/sdk-platform-java/blob/main/gapic-generator-java/src/test/proto/wicked.proto) in gapic-generator-java, I guess we'll need test yaml and json as well but they can have minimum content as well.

library_generation/test/generate_library_unit_tests.sh
blakeli0: > How about I extract the test utilities to a separate file?

Perfect idea!

JoeWang1127:
The change in the generator is not propagated to `googleapis-gen`. I'll rerun integration tests later.
```
+ diff -r googleapis-gen/google/bigtable/v2/google-cloud-bigtable-v2-java google-cloud-bigtable-v2-java -x '*gradle*'
diff -r -x '*gradle*' googleapis-gen/google/bigtable/v2/google-cloud-bigtable-v2-java/proto-google-cloud-bigtable-v2-java/src/main/java/com/google/bigtable/v2/InstanceName.java google-cloud-bigtable-v2-java/proto-google-cloud-bigtable-v2-java/src/main/java/com/google/bigtable/v2/InstanceName.java
140c140
<     if (o != null || getClass() == o.getClass()) {
---
>     if (o != null && getClass() == o.getClass()) {
diff -r -x '*gradle*' googleapis-gen/google/bigtable/v2/google-cloud-bigtable-v2-java/proto-google-cloud-bigtable-v2-java/src/main/java/com/google/bigtable/v2/TableName.java google-cloud-bigtable-v2-java/proto-google-cloud-bigtable-v2-java/src/main/java/com/google/bigtable/v2/TableName.java
157c157
<     if (o != null || getClass() == o.getClass()) {
---
>     if (o != null && getClass() == o.getClass()) {
```

library_generation/test/generate_library_unit_tests.sh
JoeWang1127: I use `uname -sm` to get os information.

JoeWang1127:
> `generate_library_unit_tests.sh` runs the entire `generate_library.sh` even for testing a small thing like skipping the `protobuf_version` argument. Maybe, it will be easier to create true unit tests by breaking down the code in `generate_library.sh` into smaller functions.
> 
> Also, can we do the "golden" tests in a separate PR? I think we should make a distinction between golden tests and unit tests.

Unit tests against `utilities.sh` are added through #1948.

This PR is intended to add golden unit tests against `generate_library.sh`. It also has some test cases with invalid versions (to verify the script can exit early).

I can create a new script to run golden unit tests as @blakeli0 suggested.

library_generation/test/generate_library_unit_tests.sh
JoeWang1127: I can copy some proto/yaml/json from [showcase repo](https://github.com/googleapis/gapic-showcase) and only keep the minimum content. WDYT?

JoeWang1127:
@blakeli0 @meltsufin Since we have concerns that vast lines of golden file is polluting the repo, I plan to use commit hash in `googleapis-gen` to remove golden files. WDYT?

blakeli0:
> @blakeli0 @meltsufin Since we have concerns that vast lines of golden file is polluting the repo, I plan to use commit hash in `googleapis-gen` to remove golden files. WDYT?

I see a few options:
1. As @meltsufin suggested in another [comment](https://github.com/googleapis/sdk-platform-java/pull/1964#pullrequestreview-1618187174), try to split the script more so that we don't have to run the whole script to test a few edge cases.
2. If it's too hard or impossible, try to use a set of mock configs(proto, yaml, json). For example, just one RPC with one request message and one response message, that should reduce the generated golden files to minimum(maybe only 5 files?).
3. If neither of the above options works, then we can use the commit hash in `googleapis-gen`, but keep in mind that we have to keep the versions up to date since `googleapis-gen` is always on main, and we will have to change it to  `google-cloud-java` later very soon as we are adding `owlbot` to the script.

JoeWang1127:
> If neither of the above options works, then we can use the commit hash in googleapis-gen, but keep in mind that we have to keep the versions up to date since googleapis-gen is always on main, and we will have to change it to google-cloud-java later very soon as we are adding owlbot to the script.

I can pin the commit hash in googleapis (protos) and googleapis-gen (generated files) so that we don't need to update versions.

JoeWang1127:
> As @meltsufin suggested in another https://github.com/googleapis/sdk-platform-java/pull/1964#pullrequestreview-1618187174, try to split the script more so that we don't have to run the whole script to test a few edge cases.

Currently the script can't do this. 

However, I can add a `dry_run` option to verify the dependencies versions and return without actually generating the library.

meltsufin:
> > As @meltsufin suggested in another [#1964 (review)](https://github.com/googleapis/sdk-platform-java/pull/1964#pullrequestreview-1618187174), try to split the script more so that we don't have to run the whole script to test a few edge cases.
> 
> Currently the script can't do this.

Can you modify it though so it's more a sequence of function calls? Think of each major block of code in the script as a function.

> However, I can add a `dry_run` option to verify the dependencies versions and return without actually generating the library.

This wouldn't solve the problem of breaking the script up into testable units.

> If we decide to implement this, I suggest to implement it in another PR and remove all golden tests in this PR and only keep the unit tests (cases with invalid versions).

Sounds good. Let's tackle golden tests in a separate PR.

library_generation/test/generate_library_unit_tests.sh
JoeWang1127: done.

JoeWang1127:
@blakeli0 @meltsufin I remove all golden tests in this PR and will create another one for them.

blakeli0 (approved):

meltsufin (approved):

PR #1961, opened 2023-09-05
chore: shell code format

JoeWang1127:
Proposed changes in this PR:
- Change `generate_library_integration_test.sh` to `library_generation/test`.
- Change testing resource name to reflect test cases ([context](https://github.com/googleapis/sdk-platform-java/pull/1948#discussion_r1316149961)).
- Reformat shell script in accordance with [Shell Style Guide](https://google.github.io/styleguide/shellguide.html)

A follow up of #1948.

PR #1968, opened 2023-09-07
chore: add a lint for shell script

JoeWang1127:
A follow up of #1961.

### DIFF
--- a/.github/workflows/verify_library_generation.yaml
+++ b/.github/workflows/verify_library_generation.yaml
@@ -47,5 +47,7 @@ jobs:
       with:
         scandir: 'library_generation'
         format: tty
+        severity: error
       env:
+        # exclude SC2317 (info): Command appears to be unreachable.
         SHELLCHECK_OPTS: -e SC2317
--- a/.github/workflows/verify_library_generation.yaml
+++ b/.github/workflows/verify_library_generation.yaml
@@ -44,7 +44,8 @@ jobs:
     - uses: actions/checkout@v3
     - name: Run ShellCheck
       uses: ludeeus/action-shellcheck@2.0.0
-      scandir: 'library_generation'
-      format: tty
+      with:
+        scandir: 'library_generation'
+        format: tty
       env:
         SHELLCHECK_OPTS: -e SC2317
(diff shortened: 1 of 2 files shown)


JoeWang1127:
The lint for shell script is only triggered when there's change in `library_generation/` directory. 

To test the ci, I added a trivial change and the log is [here](https://github.com/googleapis/sdk-platform-java/actions/runs/6113885418/job/16594280379).

suztomo:
1. Do you find anything that is different from Google shell script style? https://google.github.io/styleguide/shellguide.html (such as indent size)? If not, that's good. (No need to put effort to find discrepancy)
2. Can you add a commit in this pull request that touch some file in the "library_generation/" directory? This pull request should be merged only when the lint passes.

suztomo (approved):

.github/workflows/verify_library_generation.yaml
```diff
@@ -38,3 +38,13 @@ jobs:
       run: |
         set -x
         library_generation/test/generate_library_unit_tests.sh
+  lint:
+    runs-on: ubuntu-22.04
+    steps:
+    - uses: actions/checkout@v3
+    - name: Run ShellCheck
+      uses: ludeeus/action-shellcheck@2.0.0
```
blakeli0, line 46: @suztomo Do we have any guidelines or best practices regarding using third_party Github actions? This one has 200+ stars so there are some usages, but do we think there is any risks in using it?

suztomo: I don't think of any risk using this because it doesn't add any code to the repository.

JoeWang1127: The underlying [ShellCheck](https://github.com/koalaman/shellcheck) is open-sourced and recommended by Google.

If we think third-party action is unreliable, I can write a commend to invoke `shellcheck` directly. WDYT?

JoeWang1127: I think we can use the action for now and if it's broken in the future, I can always convert it to `shellcheck` command.

meltsufin: It would be nice to be able to run it locally too.

meltsufin (approved):

blakeli0 (approved):

PR #1988, opened 2023-09-13
feat: hermetic build OS detection

diegomarquezp:
Adds OS detection to `generate_library.sh`

Manual input from workflow matrix is removed in favor of automatic detection

### DIFF
--- a/library_generation/utilities.sh
+++ b/library_generation/utilities.sh
@@ -275,22 +275,19 @@ get_version_from_versions_txt() {
 }
 
 detect_os_architecture() {
-  case "${OSTYPE}" in
-    "linux-gnu"*|"freebsd"*)
-      os_architecture="linux-$(uname -m)"
+  local os_type
+  local os_architecture
+  os_type=$(uname -sm)
+  case "${os_type}" in
+    *"Linux x86_64"*)
+      os_architecture="linux-x86_64"
       ;;
-    "darwin"*)
-      os_architecture="osx-$(uname -m)"
-      ;;
-    "cygwin"|"msys")
-      os_architecture="win32"
+    *"Darwin x86_64"*)
+      os_architecture="osx-x86_64"
       ;;
     *)
-      >&2 echo 'Could not detect OS. Please specify it with --os_architecture'
-      >&2 echo 'Also, see https://github.com/protocolbuffers/protobuf/releases for a list of available OS (e.g. linux-aarch_64)'
-      exit 1
+      os_architecture="osx-aarch_64"
       ;;
   esac
-  >&2 echo "Detected OS architecture: ${os_architecture}" >2
   echo "${os_architecture}"
 }
--- a/library_generation/utilities.sh
+++ b/library_generation/utilities.sh
@@ -276,9 +276,9 @@ get_version_from_versions_txt() {
 
 detect_OS() {
   if [[ "${OSTYPE}" == "linux-gnu"* ]] || [[ "${OSTYPE}" == "freebsd"* ]]; then
-    os_architecture="linux-x86_64"
+    os_architecture="linux-$(uname -m)"
   elif [[ "${OSTYPE}" == "darwin"* ]]; then
-    os_architecture="osx-x86_64"
+    os_architecture="osx-$(uname -m)"
   elif [[ "${OSTYPE}" == "cygwin" ]] || [[ "${OSTYPE}" == "msys" ]]; then
     os_architecture="win32"
   else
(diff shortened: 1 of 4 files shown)


diegomarquezp:
[Mismatch in `diff`](https://github.com/googleapis/sdk-platform-java/actions/runs/6175923393/job/16763775744?pr=1988) due to `googleapis` recently pushing a [commit](https://github.com/googleapis/googleapis/commit/e8a136feaca2547dd5566ef79841d28f76a80eb5) not found in `googleapis-gen` ([link](https://github.com/googleapis/googleapis-gen/tree/b19bf82fcb471fe7a53d107528dbf5b78eee26e4/google/bigtable/v2))

library_generation/utilities.sh
```diff
@@ -273,3 +273,19 @@ get_version_from_versions_txt() {
   version=$(grep "$key:" "${versions}" | cut -d: -f3) # 3rd field is snapshot
   echo "${version}"
 }
+
+detect_OS() {
```
JoeWang1127, line 277: Judging by what the function does, the name should be `detect_os_architecture`

library_generation/utilities.sh
```diff
@@ -273,3 +273,19 @@ get_version_from_versions_txt() {
...
+  if [[ "${OSTYPE}" == "linux-gnu"* ]] || [[ "${OSTYPE}" == "freebsd"* ]]; then
+    os_architecture="linux-x86_64"
+  elif [[ "${OSTYPE}" == "darwin"* ]]; then
```
JoeWang1127, line 280: What about `osx-aarm_64`?

I think you can use `uname -sm` to detect OS type and CPU architecture, but I'm not sure whether windows can run this command.

library_generation/utilities.sh
```diff
@@ -273,3 +273,19 @@ get_version_from_versions_txt() {
   version=$(grep "$key:" "${versions}" | cut -d: -f3) # 3rd field is snapshot
   echo "${version}"
 }
+
+detect_OS() {
+  if [[ "${OSTYPE}" == "linux-gnu"* ]] || [[ "${OSTYPE}" == "freebsd"* ]]; then
```
JoeWang1127, line 278: Maybe replace `if` with `case`?

JoeWang1127 (changes requested):

library_generation/utilities.sh
diegomarquezp: makes more sense. Changed

library_generation/utilities.sh
diegomarquezp: looks better, changed

library_generation/utilities.sh
diegomarquezp: I defaulted to `win32` regardless of architecture and changed it to `uname -m` to detect the architecture only, while keep reliance on `${OSTYPE}`

library_generation/utilities.sh
```diff
@@ -273,3 +273,24 @@ get_version_from_versions_txt() {
   version=$(grep "$key:" "${versions}" | cut -d: -f3) # 3rd field is snapshot
   echo "${version}"
 }
+
+detect_os_architecture() {
+  case "${OSTYPE}" in
```
JoeWang1127, line 278: where does `OSTYPE` come from?

library_generation/utilities.sh
```diff
@@ -273,3 +273,24 @@ get_version_from_versions_txt() {
...
+    "linux-gnu"*|"freebsd"*)
+      os_architecture="linux-$(uname -m)"
+      ;;
+    "darwin"*)
+      os_architecture="osx-$(uname -m)"
```
JoeWang1127, line 283: When I run `uname -m` on my machine, the output is `arm64`.
However, the right os architecture should be osx-aarch_64.
You can reference the os architecture [here](https://github.com/protocolbuffers/protobuf/releases)

I also implemented a os architecture detection function:
```
get_os_architecture() {
  local os_type
  local os_architecture
  os_type=$(uname -sm)
  case "${os_type}" in
    *"Linux x86_64"*)
      os_architecture="linux-x86_64"
      ;;
    *"Darwin x86_64"*)
      os_architecture="osx-x86_64"
      ;;
    *)
      os_architecture="osx-aarch_64"
      ;;
  esac
  echo "${os_architecture}"
}
```

diegomarquezp: ah, thanks for the function :) I couldn't confirm for the `aarch64` case, but I switched the detection function to this one

JoeWang1127 (approved):

PR #1991, opened 2023-09-14
chore: Change `generate_library_integration_test.sh `to test on multiple libraries in `googleapis-gen`

JoeWang1127:
In this PR:
- Change `generate_library_integration_test.sh` to test on multiple libraries in `googleapis-gen`. The libraries are chosen from [here](https://github.com/googleapis/sdk-platform-java/tree/main/test/integration/goldens).

### DIFF
--- a/library_generation/test/generate_library_integration_test.sh
+++ b/library_generation/test/generate_library_integration_test.sh
@@ -7,9 +7,9 @@ set -xeo pipefail
 # Specifically, this script will do
 # 1. checkout the master branch of googleapis/google and WORKSPACE
 # 2. parse version of gapic-generator-java, protobuf and grpc from WORKSPACE
-# 3. generate a library with given proto_path and destination_path by invoking
-#    `generate_library.sh`. GAPIC options to generate a library will be parsed
-#    from proto_path/BUILD.bazel.
+# 3. generate a library with proto_path and destination_path in a proto_path
+#    list by invoking `generate_library.sh`. GAPIC options to generate a library
+#    will be parsed from proto_path/BUILD.bazel.
 # 4. checkout the master branch googleapis-gen repository and compare the result.
 
 # defaults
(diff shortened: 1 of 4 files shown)


library_generation/test/resources/proto_path_list.txt
```diff
@@ -0,0 +1,15 @@
+# This file is used in integration test against `generate_library.sh`.
+# Format:
+# proto_path destination_path
+google/bigtable/v2 google-cloud-bigtable-v2-java
+google/cloud/apigeeconnect/v1 google-cloud-apigeeconnect-v1-java
+google/cloud/asset/v1 google-cloud-asset-v1-java
+google/cloud/compute/v1 google-cloud-compute-v1-java
+google/cloud/kms/v1 google-cloud-kms-v1-java
+google/example/library/v1 google-cloud-example-library-v1-java
+# google/iam/v1 google-iam-v1-java
```
JoeWang1127, line 10: Temporarily exclude to generate `iam` due to #1992, will address the issue in another PR.

library_generation/test/generate_library_integration_test.sh
```diff
@@ -56,42 +58,49 @@ protobuf_version=$(get_version_from_WORKSPACE "protobuf-" WORKSPACE "-")
...
+  # include gapic_metadata.json and package-info.java after
+  # resolving https://github.com/googleapis/sdk-platform-java/issues/1986
+  diff -r "googleapis-gen/${proto_path}/${destination_path}" "${output_folder}/${destination_path}" -x "*gradle*" -x "gapic_metadata.json" -x "package-info.java" || RESULT=$?
 
-popd # output_folder
+  if [ ${RESULT} == 0 ] ; then
+    echo "SUCCESS: Comparison finished, no difference is found."
+  else
+    echo "FAILURE: Differences found in proto path: ${proto_path}."
+    exit "${RESULT}"
```
diegomarquezp, line 101: Would it make sense to build a list of failed libraries and then exit if the list is non emptyto have better insight on potential problems?

JoeWang1127: I think the integration tests here is to mimic what happens in the real world, so we need to modify `generate_library.sh` if any difference exists.

We do have unit tests to verify failed cases (e.g., invalid versions).

blakeli0 (approved):
LGTM. Good to see that we found one more edge case. Please still try to test the script against all libraries at least once as we discussed in the project status meeting.

meltsufin (approved):

PR #2006, opened 2023-09-18
chore: Fix flaky testCancelOuterFutureAfterStart test

lqiu96:

Fixes https://github.com/googleapis/sdk-platform-java/issues/1498

---
Reproduced:
Had some difficulty reproducing this with 20k invocations. Lowered the InitialRetryDelay value to allow less variance for the Randomized Retry Delay and was able to reproduce this fairly easily.

Setting the InitialRetryDelay to 100ms and got:
```
Random Delay: 3
Random Delay: 7

java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:87)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.junit.Assert.assertTrue(Assert.java:53)
	at com.google.api.gax.retrying.ScheduledRetryingExecutorTest.testCancelOuterFutureAfterStart(ScheduledRetryingExecutorTest.java:273)
```

Ran this with 20k invocations with 0 flakes with jitter off.

I believe the reason this test hasn't flaked as much is due to the extremely low (but non-zero) odds of from four RRD values.


gax-java/gax/src/test/java/com/google/api/gax/retrying/ScheduledRetryingExecutorTest.java
```diff
@@ -250,31 +251,47 @@ public void testCancelGetAttempt() throws Exception {
...
           getRetryingExecutor(getAlgorithm(retrySettings, 0, null), localExecutor);
       RetryingFuture<String> future =
           executor.createFuture(callable, FakeCallContext.createDefault().withTracer(tracer));
       callable.setExternalFuture(future);
       future.setAttemptFuture(executor.submit(future));
 
-      Thread.sleep(30L);
+      // The test sleeps a duration long enough to ensure that the future has been submitted for
+      // execution
+      Thread.sleep(50L);
```
blakeli0, line 286: So the sleep time should be long enough to ensure the future has been submitted, but also short enough so the future is not completed(<275ms in the example)? If that's the case, can we make the retry delay longer and make this sleep time longer? So that we have higher chance of the future being submitted?

gax-java/gax/src/test/java/com/google/api/gax/retrying/ScheduledRetryingExecutorTest.java
```diff
@@ -250,31 +251,47 @@ public void testCancelGetAttempt() throws Exception {
...
+            .toBuilder()
+            // These params were selected to ensure that future tries to run and fail (at least
+            // once) but does not complete before it is cancelled. Assuming no computation time,
+            // it would take 25 + 50 + 100 + 100 = 275ms for the future to complete, which should
+            // be more than enough time to cancel the future.
+            .setInitialRetryDelay(Duration.ofMillis(25L))
+            .setMaxRetryDelay(Duration.ofMillis(100L))
+            .setRetryDelayMultiplier(2.0)
+            .setTotalTimeout(Duration.ofMillis(1000L))
+            // Set this test to not use jitter as the randomized retry delay (RRD) may introduce
```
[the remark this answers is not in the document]
lqiu96: Jitter is just calculated as a random number between [0, Retry_Delay): https://github.com/googleapis/sdk-platform-java/blob/aacfb531eed26c88958721e222e5f58682cad63f/gax-java/gax/src/main/java/com/google/api/gax/retrying/ExponentialRetryAlgorithm.java#L259.

I'm not sure how to quantify how much randomness. It is possible that every RRD value is calculated to be 1ms

gax-java/gax/src/test/java/com/google/api/gax/retrying/ScheduledRetryingExecutorTest.java
lqiu96: It is possible to make the retry delay longer, but that doesn't eliminate the possibility that RRD is calculated as multiple consecutive low values (only makes it less likely).

I believe removing Jitter (only for testing) would make the RRD values consistent and guarantee no flakes.

gax-java/gax/src/test/java/com/google/api/gax/retrying/ScheduledRetryingExecutorTest.java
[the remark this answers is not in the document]
blakeli0: Thanks, that's good enough for me.

gax-java/gax/src/test/java/com/google/api/gax/retrying/ScheduledRetryingExecutorTest.java
blakeli0: Yes the removal of jitter should make it much more stable, but I _think_ there is still a chance that the future is not submitted after 50 ms if the CPU is overwhelmed. 
If I'm understanding correctly, the first retry is at 25ms and the last retry is at 275ms, so the sleep time should between 25 and 275. However. there is no guarantee that the first retry is at 25ms since it's a different thread, so there is a chance that after 50ms, future is not started yet. Let me know if it makes sense.

lqiu96: I see. I can a pick a number closer to 175ms (right before the 4th retry occurs) ... say 150ms? Obviously, still no guarantee that it would occur, but would be extremely unlikely the CPU is stuck that long.

I can add in another check the ensure that the future runs at least once.

blakeli0: >  I can a pick a number closer to 175ms (right before the 4th retry occurs) ... say 150ms?

If it's too close to the end, then we may have a problem on the other side, that the sleep may take too long so the future is already completed. That's why I was suggesting maybe we can have the retry delay to 1000ms, so that the sleep time 150ms is long enough for 25 ms, and short enough for 1000ms.   

> I can add in another check the ensure that the future runs at least once.

Actually I get this from your [comment](https://github.com/googleapis/sdk-platform-java/pull/2006/files#diff-253797ac36ea1cf440da370eab649ec38de9cf6db2a086857b187a5581d2a0bdR258-R259), I'm not sure if it's worth it to add one. I wouldn't add it if it would make the test more flaky and does not add much value.

lqiu96: Ok yeah agreed that 150ms might cut it too close to the other end. I can change the retry multiplier so that the max retry delay will hit 1000ms and buy us enough time.

The reason for the check to ensure it runs once is to check that the future from `executor.submit()` was run. The first attempt is always scheduled to "run immediately" with 0ms delay (should be be able to be scheduled within say a sleep time of 150ms).

blakeli0 (approved):

PR #2023, opened 2023-09-21
fix(hermetic-build): obtain gapic-generator-java locally on release branch

diegomarquezp:
This prevents the [failures](https://github.com/googleapis/sdk-platform-java/actions/runs/6263228150/job/17007141447?pr=1978) from showcase tests on the release-please branch since it tries to fetch a non-published version (`2.26.0`). The fix is to check whether we are on a release please branch as an additional condition to fetch the generator jar and pom locally.

### DIFF
--- a/library_generation/utilities.sh
+++ b/library_generation/utilities.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
 set -xeo pipefail
+util_script_dir=$(dirname "${BASH_SOURCE}")
 
 # Utility functions used in `generate_library.sh` and showcase generation.
 extract_folder_name() {
@@ -76,18 +77,15 @@ remove_grpc_version() {
 download_gapic_generator_pom_parent() {
   local gapic_generator_version=$1
   if [ ! -f "gapic-generator-java-pom-parent-${gapic_generator_version}.pom" ]; then
-    # release please branches will point to a non-snapshot that hasn't been
-    # published. We want to get the generator pom locally in this case
-    if [[ "${gapic_generator_version}" == *"-SNAPSHOT" ]] || [[ "$(is_release_branch)" == "true" ]]; then
-      # copy a SNAPSHOT version from maven local repository.
-      copy_from "$HOME/.m2/repository/com/google/api/gapic-generator-java-pom-parent/${gapic_generator_version}/gapic-generator-java-pom-parent-${gapic_generator_version}.pom" \
+    # first, try to fetch the generator pom locally
+    bash -c "source ${util_script_dir}/utilities.sh && copy_from \"$HOME/.m2/repository/com/google/api/gapic-generator-java-pom-parent/${gapic_generator_version}/gapic-generator-java-pom-parent-${gapic_generator_version}.pom\" \
+    \"gapic-generator-java-pom-parent-${gapic_generator_version}.pom\"" || not_found_locally="true"
+    if [[ "${not_found_locally}" == "true" ]];then 
+      # download gapic-generator-java-pom-parent from Google maven central mirror.
+      download_from \
+      "https://maven-central.storage-download.googleapis.com/maven2/com/google/api/gapic-generator-java-pom-parent/${gapic_generator_version}/gapic-generator-java-pom-parent-${gapic_generator_version}.pom" \
       "gapic-generator-java-pom-parent-${gapic_generator_version}.pom"
-      return
     fi
-    # download gapic-generator-java-pom-parent from Google maven central mirror.
-    download_from \
-    "https://maven-central.storage-download.googleapis.com/maven2/com/google/api/gapic-generator-java-pom-parent/${gapic_generator_version}/gapic-generator-java-pom-parent-${gapic_generator_version}.pom" \
-    "gapic-generator-java-pom-parent-${gapic_generator_version}.pom"
   fi
   # file exists, do not need to download again.
 }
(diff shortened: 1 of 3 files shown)


JoeWang1127 (approved):

library_generation/utilities.sh
```diff
@@ -76,7 +76,9 @@ remove_grpc_version() {
 download_gapic_generator_pom_parent() {
   local gapic_generator_version=$1
   if [ ! -f "gapic-generator-java-pom-parent-${gapic_generator_version}.pom" ]; then
-    if [[ "${gapic_generator_version}" == *"-SNAPSHOT" ]]; then
+    # release please branches will point to a non-snapshot that hasn't been
+    # published. We want to get the generator pom locally in this case
+    if [[ "${gapic_generator_version}" == *"-SNAPSHOT" ]] || [[ "$(is_release_branch)" == "true" ]]; then
```
blakeli0, line 81: Can we always try to download from local maven repo first regardless of the name, and fall back to maven central if not exist? We are going to have multiple release branches in the future and this would become hard to maintain. 
Similarly, we would have the same problem in the [main generation script](https://github.com/googleapis/sdk-platform-java/blob/main/library_generation/utilities.sh#L123-L137), can we try to fix it there as well @JoeWang1127 ?

diegomarquezp: thanks for the quick review! 
I modified the generator download (pom & jar) to default to a local fetch before trying the external source

library_generation/utilities.sh
```diff
@@ -76,16 +77,15 @@ remove_grpc_version() {
 download_gapic_generator_pom_parent() {
   local gapic_generator_version=$1
   if [ ! -f "gapic-generator-java-pom-parent-${gapic_generator_version}.pom" ]; then
-    if [[ "${gapic_generator_version}" == *"-SNAPSHOT" ]]; then
-      # copy a SNAPSHOT version from maven local repository.
-      copy_from "$HOME/.m2/repository/com/google/api/gapic-generator-java-pom-parent/${gapic_generator_version}/gapic-generator-java-pom-parent-${gapic_generator_version}.pom" \
+    # first, try to fetch the generator pom locally
+    bash -c "source ${util_script_dir}/utilities.sh && copy_from \"$HOME/.m2/repository/com/google/api/gapic-generator-java-pom-parent/${gapic_generator_version}/gapic-generator-java-pom-parent-${gapic_generator_version}.pom\" \
```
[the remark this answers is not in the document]
diegomarquezp: I added this to load the functions into the secondary shell. This is all with the intent to capture the [exit code](https://github.com/googleapis/sdk-platform-java/blob/e2bf12e85777b3c78e1e01829f2018bb8fb22484/library_generation/utilities.sh#L177-L189) in case the local search doesn't work.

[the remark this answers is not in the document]
JoeWang1127: Since we always try to copy files locally before downloading from maven central, I think we can modify `copy_from` not failing but return a value to indicate whether copy is success or not.

Also, we need to modify the error message in `download_fail` to indicate the file is not found neither locally or in maven central should the download fail.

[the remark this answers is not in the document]
diegomarquezp: Thanks for pointing this out, makes sense. I modified `copy_from` to return (`echo`) a value instead. I added a message to indicate an external repo attempt will be made in case it wasn't found locally. I also created a more generic function for the pom parent and the jar.

library_generation/utilities.sh
```diff
@@ -174,17 +165,18 @@ download_from() {
...
 }
 
+# copies the specified file in $1 to $2
+# will print "true" if $1 (local_repo) points to a non-existent file
 copy_from() {
   local local_repo=$1
   local save_as=$2
-  cp "${local_repo}" "${save_as}" || \
-    download_fail "${save_as}" "maven local"
+  cp "${local_repo}" "${save_as}" && echo "false" || echo "true"
```
JoeWang1127, line 173: what does this line mean?

It seems to echo `true` if `cp` failed.

diegomarquezp: I added a comment on top of the function. I will modify the comment for it to have a better explanation

diegomarquezp: also added a variable to hold the result in order to improve readability

library_generation/utilities.sh
```diff
@@ -174,17 +165,20 @@ download_from() {
   curl -LJ -o "${save_as}" --fail -m 30 --retry 2 "$url" || download_fail "${save_as}" "${repo}"
 }
 
+# copies the specified file in $1 to $2
+# will return "false" if the copy was successful, otherwise true (to indicate
```
JoeWang1127, line 169: It maybe unintuitive to indicate a successful copy as `false`, though it's a minor issue.

diegomarquezp: I inverted the logic and changed variable names accordingly

library_generation/utilities.sh
```diff
@@ -114,25 +102,28 @@ download_tools() {
...
-      copy_from "$HOME/.m2/repository/com/google/api/gapic-generator-java/${gapic_generator_version}/gapic-generator-java-${gapic_generator_version}.jar" \
-      "gapic-generator-java-${gapic_generator_version}.jar"
-      return
+    # first, try to fetch the generator locally
+    local local_fetch_successful=$(copy_from "$HOME/.m2/repository/com/google/api/${project}/${gapic_generator_version}/${artifact}" \
+      "${artifact}")
+    if [[ "${local_fetch_successful}" == "false" ]];then 
+      # download gapic-generator-java artifact from Google maven central mirror if not
+      # found locally
+      >&2 echo "${artifact} not found locally. Attempting a download from Maven Central"
```
blakeli0, line 122: Can we add logs for successful local fetch or Maven central download? Otherwise it may be misleading that this is the only log we see when we are downloading from Maven Central.

diegomarquezp: thanks, I added a couple success messages for both local and remote fetch

blakeli0 (approved):

PR #2038, opened 2023-09-26
chore: use grpc version defined in gax-java, rather than `googleapis/WORKSPACE`

JoeWang1127:


blakeli0 (approved):

JoeWang1127:
> I think [this](https://github.com/googleapis/sdk-platform-java/blob/main/library_generation/utilities.sh#L86) is where we get the gRPC compiler version implicitly?

Yes, from parent pom.

PR #2040, opened 2023-09-27
chore: download the parent pom to output folder

JoeWang1127:
In this PR:
- download the parent pom to output folder

### DIFF
--- a/library_generation/utilities.sh
+++ b/library_generation/utilities.sh
@@ -81,27 +81,31 @@ download_gapic_generator_pom_parent() {
 get_grpc_version() {
   local gapic_generator_version=$1
   local grpc_version
+  cd "${output_folder}"
   # get grpc version from gapic-generator-java-pom-parent/pom.xml
   download_gapic_generator_pom_parent "${gapic_generator_version}"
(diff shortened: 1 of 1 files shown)


library_generation/utilities.sh
```diff
@@ -81,27 +81,31 @@ download_gapic_generator_pom_parent() {
 get_grpc_version() {
   local gapic_generator_version=$1
   local grpc_version
+  cd "${output_folder}"
   # get grpc version from gapic-generator-java-pom-parent/pom.xml
   download_gapic_generator_pom_parent "${gapic_generator_version}"
   grpc_version=$(grep grpc.version "gapic-generator-java-pom-parent-${gapic_generator_version}.pom" | sed 's/<grpc\.version>\(.*\)<\/grpc\.version>/\1/' | sed 's/^[[:space:]]*//;s/[[:space:]]*$//')
-  echo "$grpc_version"
+  cd ..
```
JoeWang1127, line 88: If I use `popd` here, the value of directory stack somehow gets into `grpc_version`.
The grpc_version becomes `sdk-platform-java sdkplatform-java/output 1.58.0`.

diegomarquezp: I think it's because the `stdout` of the function includes the output of `pushd` and `popd`.  I tried  `pushd $folder > /dev/null` and the units worked (and I could replicate the incorrect grpc version you mentioned if I don't use `> /dev/null`)

JoeWang1127: cool, I'll try that.

JoeWang1127: It works, thanks.

diegomarquezp (approved):

PR #2046, opened 2023-09-27
feat: generate proto-only library

### LINKED ISSUE #2047: Generate proto-only libraries
The `generate_library.sh` should be able to generate libraries which don't have a `java_gapic_library` bazel target.

JoeWang1127:
Fix #2047.

### DIFF
--- a/library_generation/README.md
+++ b/library_generation/README.md
@@ -91,6 +91,17 @@ Use `--grpc_version` to specify the value.
 
 Note that if specified, the version should be compatible with gapic-generator-java.
 
+### proto_only (optional)
+Whether this is a proto-only library (no `gapic-*` directory in the generated 
+library).
+The default value is `false`.
+
+When set to `true`, the GAPIC generator will not be invoked.
+Therefore, GAPIC options (`transport`, `rest_numeric_enums`) and 
+`gapic_additional_protos` will be ignored.
+
+Use `--proto_only` to specify the value.
+
 ### gapic_additional_protos (optional)
 Additional protos that pass to the generator.
 The default value is `google/cloud/common_resources.proto`.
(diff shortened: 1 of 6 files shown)


JoeWang1127:
I've test on three proto-only libraries:
```
google/cloud/alloydb/connectors/v1alpha google-cloud-alloydb-connectors-v1alpha-java
google/firestore/bundle google-cloud-firestore-bundle-v1-java
google/iam/v1 google-iam-v1-java
```
1. Library generated from `google/firestore/bndle` is the same as in `googleapis-gen`.
2. The difference of library generated from `google/cloud/alloydb/connectors/v1alpha` is:
```
Only in /Users/joewa/Documents/workspace/sdk-platform-java/output/google-cloud-alloydb-connectors-v1alpha-java/proto-google-cloud-alloydb-connectors-v1alpha-java/src/main: proto
```
It turns out that allydb connectors don't copy proto files to library.
3. The difference of library generated from `google/iam/v1` is:
```
Only in /Users/joewa/Documents/workspace/sdk-platform-java/output/google-iam-v1-java/proto-google-iam-v1-java/src/main/java/com/google/iam/v1: logging
Only in /Users/joewa/Documents/workspace/sdk-platform-java/output/google-iam-v1-java/proto-google-iam-v1-java/src/main: proto
```
It has the same issue with 2 and I need to exclude `google/iam/v1/logging`.

diegomarquezp (commented):
do we want to add at least one of the proto-only libraries you mentioned in `proto_path_list.txt`?

JoeWang1127:
> do we want to add at least one of the proto-only libraries you mentioned in `proto_path_list.txt`?

Good idea. I added `google/firestore/bundle` to the list.

diegomarquezp (approved):

Issue #1938, opened 2023-08-23
api-common using deprecated javax.annotation-api


sydney-munro:
Google Cloud storage recieved https://github.com/googleapis/java-storage/issues/2176 due to a transient dependency on javax.annotaion-api. This should be updated to the new jakarta namespace.

suztomo:
> javax package is not compliant with Jakarta platform. Please release google-cloud-storage:jar which is Jakarta platform compliant.

Digging further.

suztomo:
@abhishekshan03 Would you share details about the noncompliance? Do you get any error message when you use certain tools from Jakarta EE?

I found this article https://jakarta.ee/blogs/javax-jakartaee-namespace-ecosystem-progress/ but it didn't have any specifics about `javax.annotation-api`.

abhishekshan03:
Hi @suztomo 
Please pardon my limited knowledge. As far as I know packages javax.* are not part of Jakarta EE platform and hence are deprecated and will be replaced by [jakarta annotations](https://jakarta.ee/specifications/annotations/)
As an example please refer to below issue.
https://github.com/jruby/jruby/issues/6514

MyDogIsMyPersonality:
Is there any idea on when this might be released? It is blocking a major upgrade for us.

suztomo:
@abhishekshan03 @MyDogIsMyPersonality Are you suffering from some errors with certain tools when they're used with GCP Java libraries? I think it's easy to swap but I'd like to ensure to verify the fix.

MyDogIsMyPersonality:
There is more information in issue 2176, but Jakarta won't upgrade due to the transient dependency in javax.

suztomo:
@MyDogIsMyPersonality https://github.com/googleapis/java-storage/issues/2176 does not have any errors.

> Jakarta won't upgrade

Would you elaborate a bit more? Are you referencing a certain version of one of the tools listed in https://jakarta.ee/compatibility/ ?

abhishekshan03:
> @abhishekshan03 @MyDogIsMyPersonality Are you suffering from some errors with certain tools when they're used with GCP Java libraries? I think it's easy to swap but I'd like to ensure to verify the fix.
> 
> Thank you for sharing [jruby/jruby#6514](https://github.com/jruby/jruby/issues/6514). I see it's fixed there.

@suztomo  
[...]

PR #2060, opened 2023-10-03
chore: [Not for review] removed GuardeBy

suztomo:
As part of https://github.com/googleapis/sdk-platform-java/issues/1938. This pull request is an experiment to see the effect of the GuardedBy annotation.


suztomo:
The Bazel build caught the bug!

https://github.com/googleapis/sdk-platform-java/actions/runs/6397756942/job/17366399303?pr=2060:

```
external/com_google_api_gax_java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonClientCallImpl.java:287: error: [GuardedBy] This access should be guarded by 'this.lock', which is not currently held
        if (inDelivery && newActiveDeliveryLoop) {
            ^
    (see https://errorprone.info/bugpattern/GuardedBy)
external/com_google_api_gax_java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonClientCallImpl.java:292: error: [GuardedBy] This access should be guarded by 'this.lock', which is not currently held
        inDelivery = true;
        ^
    (see https://errorprone.info/bugpattern/GuardedBy)
```

PR #2064, opened 2023-10-04
fix: ensure that channel pool ref count never go negative

igorbernstein2:
I'm not sure the root cause that might cause this, but this guard should tone down the noise that is created and help us identify the root cause


igorbernstein2:
This PR is broken, will open a new shortly

PR #2065, opened 2023-10-04
fix: ensure that channel pool ref count never goes negative (take2)

igorbernstein2:
I'm not sure the root cause that might cause this, but this guard should tone down the noise that is created and help us identify the root cause

### DIFF
--- a/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
+++ b/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
@@ -68,8 +68,7 @@ import org.threeten.bp.Duration;
  * <p>Package-private for internal use.
  */
 class ChannelPool extends ManagedChannel {
-  @VisibleForTesting
-  static final Logger LOG = Logger.getLogger(ChannelPool.class.getName());
+  @VisibleForTesting static final Logger LOG = Logger.getLogger(ChannelPool.class.getName());
   private static final Duration REFRESH_PERIOD = Duration.ofMinutes(50);
 
   private final ChannelPoolSettings settings;
@@ -427,7 +426,7 @@ class ChannelPool extends ManagedChannel {
 
     void release() {
       if (!wasReleased.compareAndSet(false, true)) {
-        Exception e =  new IllegalStateException("Entry was already released");
+        Exception e = new IllegalStateException("Entry was already released");
         LOG.log(Level.WARNING, e.getMessage(), e);
         return;
       }
@@ -537,7 +536,8 @@ class ChannelPool extends ManagedChannel {
 
       RetainedEntry entry = getRetainedEntry(affinity);
 
-      return new ReleasingClientCall<>(entry.getChannel().newCall(methodDescriptor, callOptions), entry);
+      return new ReleasingClientCall<>(
+          entry.getChannel().newCall(methodDescriptor, callOptions), entry);
     }
   }
 
--- a/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
+++ b/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
@@ -415,6 +415,10 @@ class ChannelPool extends ManagedChannel {
     return localEntries.get(index);
   }
 
+  /**
+   * This represents the reserved refcount of a single RPC using a channel. It the responsibility of
+   * that RPC to call release exactly once when it completes to release the Channel.
+   */
   private static class RetainedEntry {
     private final Entry entry;
     private final AtomicBoolean wasReleased;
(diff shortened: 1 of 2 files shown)


gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
```diff
@@ -415,10 +415,33 @@ private Entry getEntry(int affinity) {
     return localEntries.get(index);
   }
 
+  private static class RetainedEntry {
```
blakeli0, line 418: Can this extends `Entry` so all the existing references to `Entry` can stay as they are?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
```diff
@@ -415,10 +415,33 @@ private Entry getEntry(int affinity) {
...
+    private final AtomicBoolean wasReleased;
+
+    public RetainedEntry(Entry entry) {
+      this.entry = entry;
+      wasReleased = new AtomicBoolean(false);
+    }
+
+    void release() {
+      if (!wasReleased.compareAndSet(false, true)) {
+        Exception e = new IllegalStateException("Entry was already released");
```
blakeli0, line 429: This is another stacktrace we don't want to log, can we just log the WARNING without a stacktrace?
Or if we think this is a true error, then we can throw the Exception?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
```diff
@@ -415,10 +415,33 @@ private Entry getEntry(int affinity) {
...
+        LOG.log(Level.WARNING, e.getMessage(), e);
+        return;
+      }
+      entry.release();
```
blakeli0, line 433: We can probably add a FINE log on success release?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
igorbernstein2: no, Entry represents a shared count, while this represents a single lease from the shared count

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
igorbernstein2: We definitely need a stacktrace here, it will tell us how the refcount got negative. Otherwise we are no wiser than before.....we are just hiding the problem under the rug. And in this case alert firing woud be correct

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
igorbernstein2: we would be generating a log message per RPC then, which would 10s of thousands per sec....it would just be noise

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
blakeli0: To me `Entry was already released` is a true error, why are we just logging instead of throwing an exception?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
blakeli0: Thanks! Didn't notice `release()` is called on every RPC. Can we add some comments on `release()` method?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
blakeli0: I see, can we add some comments on this class?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
igorbernstein2: Added some javadoc to the class

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
igorbernstein2: Because throwing will bubble up to the enduser, who had nothing to do with the error. Instead I want to keep it localized to the double release and protect the rest of the application

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
igorbernstein2: done

meltsufin:
@igorbernstein2 @blakeli0 Can you please create a corresponding issue for better tracking?

blakeli0:
Closing in favor of #2185

PR #2084, opened 2023-10-10
chore: include/exclude certain protos when generating libraries

JoeWang1127:
In this PR: add exceptions to include/exclude certain protos from `proto_path`.
- google/cloud/filestore: include `google/cloud/common/operation_metadata.proto`
- google/cloud/oslogin: include `google/cloud/oslogin/common/common.proto`
- google/devtools/containeranalysis/v1beta1: exclude `google/devtools/containeranalysis/v1beta1/cvss/cvss.proto` in java compiler
- google/cloud/aiplatform/v1beta1: 
  - exclude `google/cloud/aiplatform/v1beta1/schema/io_format.proto`
  - exclude `google/cloud/aiplatform/v1beta1/schema/*.proto` in java compiler

Follow up PR will add exceptions to exclude subdirectories within `proto_path`.

### DIFF
--- a/library_generation/generate_library.sh
+++ b/library_generation/generate_library.sh
@@ -108,6 +108,9 @@ mkdir -p "${output_folder}/${destination_path}"
 folder_name=$(extract_folder_name "${destination_path}")
 pushd "${output_folder}"
 proto_files=$(find "${proto_path}" -type f  -name "*.proto" | sort)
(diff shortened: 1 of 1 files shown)


diegomarquezp (approved):
Since it's a short-term solution and the logic is simple, I think it's fine to hardcode the edge cases

blakeli0 (approved):

PR #2090, opened 2023-10-11
chore: add exceptions to search in `proto_path`, but not in its subdirectories

JoeWang1127:
In this PR, add special cases in which only search protos in `proto_path`, but not in its subdirectories. The cases are:
- google/api

### DIFF
--- a/library_generation/generate_library.sh
+++ b/library_generation/generate_library.sh
@@ -107,7 +107,13 @@ mkdir -p "${output_folder}/${destination_path}"
 # get a fixed order.
 folder_name=$(extract_folder_name "${destination_path}")
 pushd "${output_folder}"
-proto_files=$(find "${proto_path}" -type f  -name "*.proto" | sort)
+depth=""
+case "${proto_path}" in
+  "google/api")
+    depth="-maxdepth 1"
+    ;;
+esac
+proto_files=$(find "${proto_path}" ${depth} -type f  -name "*.proto" | sort)
 # include or exclude certain protos in grpc plugin and gapic generator java.
 case "${proto_path}" in
   "google/cloud/aiplatform/v1beta1"*)


library_generation/generate_library.sh
```diff
@@ -107,9 +107,20 @@ mkdir -p "${output_folder}/${destination_path}"
...
+find_depth=""
+case "${proto_path}" in
+  "google/api" | "google/cloud" | "google/iam/v1" | "google/rpc")
+    find_depth="-maxdepth 1"
+    ;;
+esac
+proto_files=$(find "${proto_path}" ${find_depth} -type f  -name "*.proto" | sort)
 # include or exclude certain protos in grpc plugin and gapic generator java.
 case "${proto_path}" in
+  "google/cloud")
```
blakeli0, line 119: Is this folder generated into java-common-protos [here](https://github.com/googleapis/sdk-platform-java/blob/main/java-common-protos/proto-google-common-protos/src/main/proto/google/cloud/extended_operations.proto) eventually?

library_generation/generate_library.sh
```diff
@@ -107,9 +107,20 @@ mkdir -p "${output_folder}/${destination_path}"
 # get a fixed order.
 folder_name=$(extract_folder_name "${destination_path}")
 pushd "${output_folder}"
-proto_files=$(find "${proto_path}" -type f  -name "*.proto" | sort)
+find_depth=""
+case "${proto_path}" in
```
blakeli0, line 111: How many services we need to search for subdirectories? If there is only a few, maybe we should make searching for current folder the default?Because in some cases it could be wrong to generate for subdirectories.

For example, [fleetengine/v1](https://github.com/googleapis/googleapis/tree/master/google/maps/fleetengine/v1) and [fleetengine/delivery/v1](https://github.com/googleapis/googleapis/tree/master/google/maps/fleetengine/delivery/v1) are two separate apis.

library_generation/generate_library.sh
JoeWang1127: Yes.

library_generation/generate_library.sh
JoeWang1127: > How many services we need to search for subdirectories? 

Less than 5 libraries, aiplatform, for example.

Though in this PR, all special cases are common protos and I'm not sure whether subdirectory protos are included in service clients.

> For example, [fleetengine/v1](https://github.com/googleapis/googleapis/tree/master/google/maps/fleetengine/v1) and [fleetengine/delivery/v1](https://github.com/googleapis/googleapis/tree/master/google/maps/fleetengine/delivery/v1) are two separate apis.

I didn't test this case because they are self-service client.

diegomarquezp (approved):

blakeli0 (approved):

PR #2105, opened 2023-10-11
feat: log DirectPath misconfiguration

mohanli-ml:
We met issues where customers wanted to use DirectPath, but DirectPath was misconfigured and it took 2 weeks for debugging. So we want to add WARNING logs if DirectPath flag/option is set but DirectPath is not used.

Issue: https://github.com/googleapis/sdk-platform-java/issues/2164.

@blakeli0 @frankyn


mohanli-ml:
@blakeli0 This log will happen in every channel creation. I am worried this PR will make too many logs, especially if the customer has a large channel pool size. Does Java open source has something similar to LOG_EVERY_N_SEC as C++ absl?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -266,6 +270,40 @@ boolean isDirectPathXdsEnabled() {
...
+    if (isDirectPathOptionSet || isDirectPathXdsEnvSet || isDirectPathXdsOptionSet) {
+      // Case 1: use DirectPath with gRPCLB
+      if (isDirectPathOptionSet && !(isDirectPathXdsEnvSet || isDirectPathXdsOptionSet)) {
+        // TODO: Add the warning once we move traffic out from gRPCLB
+      }
+
+      // Case 2: just enable DirectPath xDS
+      if (!isDirectPathOptionSet && (isDirectPathXdsEnvSet || isDirectPathXdsOptionSet)) {
+        LOG.log(
+            Level.WARNING, "DirectPath is misconfigured. Please set the attemptDirectPath option.");
```
frankyn, line 287: @singhravidutt this PR is w.r.t R.1 for observability.

Would hadoop connector customers be able to see these warnings emitted when running the connector or would this require a change?

frankyn (commented):
Pulling in a colleague from Dataproc team to weigh in; Thanks for tackling this @mohanli-ml;

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
singhravidutt: Yes, we should be able to surface these logs.

singhravidutt, line 287: should we add values of `isDirectPathOptionSet`, `isDirectPathXdsEnvSet` and `isDirectPathXdsOptionSet` as part of log statement?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -266,6 +270,40 @@ boolean isDirectPathXdsEnabled() {
     return false;
   }
 
+  private void logDirectPathMisconfig() {
```
singhravidutt, line 273: IIUC, this logic is added specifically for log statements. How do we make sure it is in parity with directPath decision making?

frankyn: Are you referring to the bit that actually handles connecting to DirectPath in gRPC?

mohanli-ml: The DirectPath decision making is L363:
```
if (isDirectPathEnabled() && isNonDefaultServiceAccountAllowed() && isOnComputeEngine()) {
```
And I came up with the 4 cases based on it. WDYT?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
mohanli-ml: In this case user enables DirectPathXds either by env or option, but DirectPath itself is not enabled. I think the log is already clear that the user needs to set `attemptDirectPath` and there is no need to log the values for judging the case. WDYT?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
singhravidutt: What I meant was that this logic will be invoked only when directPath is misconfigured. But this is not the logic which decides if directPath is misconfigured. This code snippet could become stale if there is change in logic which is actually deciding when directpath is misconfigured.

Just as an e.g. this is the [code snippet](https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java?#L325) which decides directPath misconfiguration 
```
  if (isDirectPathEnabled() && isNonDefaultServiceAccountAllowed() && isOnComputeEngine()) {
```

Here we check for `isNonDefaultServiceAccountAllowed`. Hypothetically in future if `isNonDefaultServiceAccountAllowed` check is removed there is no easy way to know that `fun:logDirectPathMisconfig` needs to be updated as well.

I guess there is no easy way to enforce it and this is fine for now.

mohanli-ml: Thanks for your inputs! `isNonDefaultServiceAccountAllowed()` and `isOnComputeEngine()` should be kept even after DirectPath rolls out, and then we will need to cleanup env/options, and this warning logs should also be cleaned.

blakeli0:
> @blakeli0 This log will happen in every channel creation. I am worried this PR will make too many logs, especially if the customer has a large channel pool size. Does Java open source has something similar to LOG_EVERY_N_SEC as C++ absl?

It's a valid concern, I would suggest refactor the code so that we don't have to log it on channel creation in `createSingleChannel()`. From what I can see, we can move pretty much all the logging logic to the constructor/builder of `InstantiatingGrpcChannelProvider`.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -266,6 +270,43 @@ boolean isDirectPathXdsEnabled() {
...
+      }
+
+      // Case 2: just enable DirectPath xDS
+      if (!isDirectPathOptionSet && (isDirectPathXdsEnvSet || isDirectPathXdsOptionSet)) {
+        LOG.log(
+            Level.WARNING, "DirectPath is misconfigured. Please set the attemptDirectPath option.");
+      }
+
+      // The following WARNING logs are GCS only.
+      if (serviceAddress.contains("storage.googleapis.com")
```
blakeli0, line 291: This is absolutely not allowed. Gax must not be aware of anything related to a specific service. Looks like we need two logging strategies for opt-in and opt-out, can we make it a flag and Storage can pass the flag into gax?

blakeli0:
@mohanli-ml Thanks for opening this PR! The logic to determine different logs seems pretty complicated and people have different opinions. Can you please create an issue and/or an internal doc, list all the scenarios there, and make sure we all agree on the approach first?
cc: @meltsufin

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
frankyn: Isn't this generally useful though? This log helps call out that attempting directpath wasn't configured correctly to developers.

frankyn: Missed your comment above, about adding this into Constructor which seems like a good path for this; just calling out this is generally useful.

blakeli0: Yes I agree this log is generally useful. But hardcoding the storage specific endpoint `storage.googleapis.com` in Gax is not allowed.

mohanli-ml: I have a new idea. Let’s only make the WARNING log if DirectPath xDS env/option is set. DirectPath xDS is opt-in for GCS/CBT/CS, so we do not need to have two strategies. WDYT?

mohanli-ml:
> > @blakeli0 This log will happen in every channel creation. I am worried this PR will make too many logs, especially if the customer has a large channel pool size. Does Java open source has something similar to LOG_EVERY_N_SEC as C++ absl?
> 
> It's a valid concern, I would suggest refactor the code so that we don't have to log it on channel creation in `createSingleChannel()`. From what I can see, we can move pretty much all the logging logic to the constructor/builder of `InstantiatingGrpcChannelProvider`.

Yes I moved the method to the channel pool creation, so no matter how many channels customers want, the log should only be logged once. Can you double check if this is the right place?

mohanli-ml:
> @mohanli-ml Thanks for opening this PR! The logic to determine different logs seems pretty complicated and people have different opinions. Can you please create an issue and/or an internal doc, list all the scenarios there, and make sure we all agree on the approach first? cc: @meltsufin

Before we do this, can you take a look at the latest code?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
frankyn: I think these warnings should be emitted for both cases; a user is attempting DirectPath and it would be helpful to know they're doing something wrong in configuring it. Otherwise, it's a blackbox and users are left to weeks of debugging what went wrong. The goal of this is to make it easier to understand a user did something wrong.

mohanli-ml: I see. Your concern is that what if a customer set `attemptDirectPath` but not `attemptDirectPathXds` option/env, there will be no warnings. 

Let me provide some context here. `attemptDirectPath` is originally created for DirectPath gRPCLB (gRPCLB is the name resolver). Then DirectPath xDS is developed (TD is the name resolver) and we decide to migrate DirectPath gRPCLB to DirectPath xDS. To differentiate the two DirectPath infrastructures, `attemptDirectPathXds` option/env are created:
* If `attemptDirectPath` is not set, CloudPath will be used;
* If `attemptDirectPath` is set AND `attemptDirectPathXds` option/env is NOT set, DirectPath gRPCLB is used;
* If `attemptDirectPath` is set AND `attemptDirectPathXds` option/env is set, DirectPath xDS is used.

Currently the migration is still ongoing as CBT still has DirectPath gRPCLB traffic (CBT's timeline is Q4). After that, I will cleanup DirectPath gRPCLB code in the gax library. For example, `attemptDirectPath` will be removed and we just need to keep `attemptDirectPathXds` option/env.

With that being said, I can change the DirectPath enabling logic to:
* **If `attemptDirectPath` is not set AND If `attemptDirectPathXds` option/env is not set, CloudPath will be used**;
* If `attemptDirectPath` is set AND `attemptDirectPathXds` option/env is not set, DirectPath gRPCLB is used;
* **If `attemptDirectPathXds` option/env is set, DirectPath xDS is used**.

In this way, GCS customers just need to set `attemptDirectPathXds` option/env, and they do not need to care about `attemptDirectPathXds`. Then the current log logic should be good. WDYT?

frankyn: Changing the way the options work sounds like a reasonable thing to do after migration is complete.
In java-storage, the attemptDirectPath and attemptDirectPathXds issue has been addressed by enabling both, but these warnings would have helped save the team time. We don't expose configuration of each flag attemptDirectPath and attemptDirectPathXds directly to users instead only expose 1 flag that enables both.

My original goal is to add warnings to let GCS users know they have misconfigured the client to use DirectPath. This also lends to supporting future warnings if new conditions are added in the future.

The three cases surfaced in this logic:
1) attemptDirectPath vs. attemptDirectPathXds is one way; 
2) non GCE credentials; this is likely to bite a customer
3) Not using GCE; not as much a concern, but we don't document DirectPath in public documentation so a user could do something silly and use the our clients outside GCE without understanding that DP is not available.

I think the warnings would be generally useful outside of GCS.

mohanli-ml: > Moreover, we don't expose configuration of each flag attemptDirectPath and attemptDirectPathXds directly to users instead only expose 1 flag that enables both.

Sorry I am confused. What are the two cases in the previous comment `I think these warnings should be emitted for both cases`?

frankyn: Maybe we are speaking past each other? 

> I have a new idea. Let’s only make the WARNING log if DirectPath xDS env/option is set. DirectPath xDS is opt-in for GCS/CBT/CS, so we do not need to have two strategies. WDYT?

The point i was driving at is that there's nothing telling a user they configured DirectPath incorrectly when they attempt DP; and selecting the cases only when attemptDirectPathXds is set to emit these warnings misses this point.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -266,6 +271,33 @@ boolean isDirectPathXdsEnabled() {
...
+  private void logDirectPathMisconfig() {
+    boolean isDirectPathOptionSet = Boolean.TRUE.equals(attemptDirectPath);
+    boolean isDirectPathXdsEnvSet =
+        Boolean.parseBoolean(envProvider.getenv(DIRECT_PATH_ENV_ENABLE_XDS));
+    boolean isDirectPathXdsOptionSet = Boolean.TRUE.equals(attemptDirectPathXds);
+    if (isDirectPathXdsEnvSet || isDirectPathXdsOptionSet) {
+      // Case 1: does not enable DirectPath
+      if (!isDirectPathOptionSet) {
+        LOG.log(
+            Level.WARNING, "DirectPath is misconfigured. Please set the attemptDirectPath option.");
```
frankyn, line 283: Recommend:
`Please enable the attemptDirectPath option along with the attemptDirectPathXds option`

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -266,6 +271,33 @@ boolean isDirectPathXdsEnabled() {
...
+        if (!isNonDefaultServiceAccountAllowed()) {
+          LOG.log(
+              Level.WARNING,
+              "DirectPath is misconfigured. Please make sure the credential is an instance of"
+                  + " ComputeEngineCredentials.");
+        }
+        // Case 3: not running on GCE
+        if (!isOnComputeEngine()) {
+          LOG.log(
+              Level.WARNING, "DirectPath is misconfigured. Please run in the GCE environment. ");
```
frankyn, line 295: Suggestion: 
```
DirectPath is misconfigured. DirectPath is only available in a GCE environment.
```

You also have whitespace at the end.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
mohanli-ml: Done.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -266,6 +271,33 @@ boolean isDirectPathXdsEnabled() {
...
+      if (!isDirectPathOptionSet) {
+        LOG.log(
+            Level.WARNING, "DirectPath is misconfigured. Please set the attemptDirectPath option.");
+      } else {
+        // Case 2: credential is not correctly set
+        if (!isNonDefaultServiceAccountAllowed()) {
+          LOG.log(
+              Level.WARNING,
+              "DirectPath is misconfigured. Please make sure the credential is an instance of"
+                  + " ComputeEngineCredentials.");
```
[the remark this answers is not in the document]
mohanli-ml: Done. 

GKE workload identity is compatible with DirectPath.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
mohanli-ml: Done.

frankyn (approved):

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
blakeli0: I have the same concern as @singhravidutt, the logic in `logDirectPathMisconfig()` could evolve to be different than `(isDirectPathEnabled() && isNonDefaultServiceAccountAllowed() && isOnComputeEngine())`. Ideally, we should check if we should use DirectPath or not during the initialization of this class, set the value to a field `useDirectPath`, and use it to check if we should create a DirectPath channel afterwards.
See the following code that demonstrates my idea: 
```
public InstantiatingGrpcChannelProvider build() {
      if (shouldUseDirectPath()) {
          setUseDirectPath(true);
      }
      return new InstantiatingGrpcChannelProvider(this);
}

  private boolean shouldUseDirectPath() {
    if (isDirectPathEnabled()) {
      return isOnComputeWithValidCredential();
    } else {
      if (isDirectPathXdsEnabled()) {
        //log
        isOnComputeWithValidCredential();
      }
    }
    return false;
  }

  private boolean isOnComputeWithValidCredential() {
    if (!isNonDefaultServiceAccountAllowed()) {
      //log
      return false;
    }
    if (!isOnComputeEngine()) {
      //log
      return false;
    }
    return true;
  }

```

This also makes the new code much more testable as [SonarCheck](https://sonarcloud.io/component_measures?id=googleapis_gapic-generator-java_integration_tests&pullRequest=2105&metric=new_coverage&view=list) is failing due to lacking test coverage of the new code.  

@mohanli-ml Let me know if it makes sense, if you think it might be too much at this moment, our team can take it as a follow up PR.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -266,6 +271,37 @@ boolean isDirectPathXdsEnabled() {
...
+    if (isDirectPathXdsEnvSet || isDirectPathXdsOptionSet) {
+      // Case 1: does not enable DirectPath
+      if (!isDirectPathOptionSet) {
+        LOG.log(
+            Level.WARNING,
+            "DirectPath is misconfigured. Please set the attemptDirectPath option along with the"
+                + " attemptDirectPathXds option.");
+      } else {
+        // Case 2: credential is not correctly set
+        if (!isNonDefaultServiceAccountAllowed()) {
```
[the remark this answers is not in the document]
mohanli-ml: Yes it is possible, this is the original fix for this problem, https://github.com/googleapis/gax-java/pull/1250.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -233,6 +237,7 @@ public TransportChannel getTransportChannel() throws IOException {
   }
 
   private TransportChannel createChannel() throws IOException {
+    logDirectPathMisconfig();
```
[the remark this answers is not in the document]
mohanli-ml: Done.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -266,6 +271,37 @@ boolean isDirectPathXdsEnabled() {
     return false;
   }
 
+  private void logDirectPathMisconfig() {
+    boolean isDirectPathOptionSet = Boolean.TRUE.equals(attemptDirectPath);
+    boolean isDirectPathXdsEnvSet =
+        Boolean.parseBoolean(envProvider.getenv(DIRECT_PATH_ENV_ENABLE_XDS));
+    boolean isDirectPathXdsOptionSet = Boolean.TRUE.equals(attemptDirectPathXds);
+    if (isDirectPathXdsEnvSet || isDirectPathXdsOptionSet) {
```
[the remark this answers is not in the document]
mohanli-ml: Done.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -266,6 +271,37 @@ boolean isDirectPathXdsEnabled() {
...
+      // Case 1: does not enable DirectPath
+      if (!isDirectPathOptionSet) {
```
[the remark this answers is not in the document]
mohanli-ml: Done.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
mohanli-ml: Thanks for the suggestion! Moving the DirectPath enabling logic to the builder instead of checking it every time when construct a channel sounds good to me. However, pay attention to the logic here:
```
private boolean shouldUseDirectPath() {
    if (isDirectPathEnabled()) {       <- This should be `if (isDirectPathEnabled() && !isDirectPathXdsEnabled())`
      return isOnComputeWithValidCredential();
    } else {
      if (isDirectPathXdsEnabled()) {
        //log
        isOnComputeWithValidCredential();
      }
    }
    return false;
  }
```
In short I feel the logic is already complex and I would suggest to do this with the cleanup of removing DirectPath gRPCLB, which is blocked by CBT and is expected to be done in Q4.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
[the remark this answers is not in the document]
blakeli0: If we move `logDirectPathMisconfig()` into the Builder class then it should be fine, ideally we should do it but then we would need to move all related logics into it as well. The quickest solution would be add it to the [constructor](https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L142).

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
blakeli0: > In short I feel the logic is already complex and I would suggest to do this with the cleanup of removing DirectPath gRPCLB, which is blocked by CBT and is expected to be done in Q4.

I think that's reasonable. For now, can you please add some unit tests for these logging statement? So that the test coverage for new code meets our least standard and gives us more confidence when we refactor it. For example, one way to test it is to create a [FakeLogger](https://github.com/googleapis/sdk-platform-java/pull/2065/files#diff-303a4fc40681b63e18e938fb84ba5753e81efb2ae40aeffde3813b58304ad37dR723-R739) and assert the log message captured in the FakeLogger.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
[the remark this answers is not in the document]
mohanli-ml: Done.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
mohanli-ml: Done. Thanks!

blakeli0 (approved):

mohanli-ml:
@blakeli0 Hi Blake, I just realized that this PR could be misleading. For example, a customer's application may contains a GCS client and a Bigquery client. The customer use `setAttemptDirectPath` to enable DirectPath in its GCS client, and set `GOOGLE_CLOUD_ENABLE_DIRECT_PATH_XDS` to enable DirectPath xDS. In this case, this [warning log](https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L276-L282) will be printed in the customer's Bigquery client. I think this is a P2 or P3 issue, and could you find someone to fix it (as you mentioned your team want to take over DirectPath)?

cc @frankyn

frankyn:
Warnings started being emitted in recent version of java-storage w.r.t ["incorrect" GCE credentials](https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L285-L290); however requests are still using GCE ADC, was there a change recently to credentials that could trigger this incorrectly?

Issue #2081, opened 2023-10-09
The count of outstanding RPCs in ChannelPool could go negative.


blakeli0:
The count of outstanding RPCs in ChannelPool could go negative in some cases, and an [exception](https://github.com/googleapis/sdk-platform-java/blob/726d02d60181dc56ffa0911757ec5516b2b5e237/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java#L495) would be throw. In such case, there is no way a customer could recover from such error. See [internal issue](https://b.corp.google.com/issues/298510633) for details.

We need to prevent the count of outstanding RPCs to go negative and also figure out why this happened.

meltsufin:
Do we understand how this could happen and are we able to reproduce the error?
cc: @igorbernstein2

blakeli0:
> Do we understand how this could happen and are we able to reproduce the error? cc: @igorbernstein2

I don't think we understand how this could happen, @igorbernstein2 tried to add the [stacktrace](https://github.com/googleapis/sdk-platform-java/pull/2065) so we might be able to figure it out next time it happens. This is not reproducible either by customers, we might be able to reproduce it by digging really deep into the codebase.

meltsufin:
In the internal bug I see some ideas. Can we try to reproduce them?

PR #2185, opened 2023-10-19
fix: Make sure outstanding RPCs count in ChannelPool can not go negative

blakeli0:
fixes: #2081 

Add two flags `wasClosed` and `wasReleased` to `ReleasingClientCall` to check various scenarios. The combination of these two flags can make sure the count of outstanding RPCs can never go negative, and help us identify what exactly goes wrong next time it happens. 

### More background for the purpose of `outstandingRPCs`

The primary purpose of keeping a count for outstanding RPCs is to track when a channel is
safe to close. In grpc, initialization & starting of rpcs is split between 2 methods:
Channel#newCall() and ClientCall#start. gRPC already has a mechanism to safely close channels
that have rpcs that have been started. However, it does not protect calls that have been
created but not started. In the sequence: Channel#newCall() Channel#shutdown()
ClientCall#Start(), gRpc will error out the call telling the caller that the channel is
shutdown.
Hence, the increment of outstanding RPCs has to happen when the ClientCall is initialized,
as part of Channel#newCall(), not after the ClientCall is started. The decrement of
outstanding RPCs has to happen when the ClientCall is closed or the ClientCall failed to
start.

### DIFF
--- a/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
+++ b/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
@@ -470,7 +470,7 @@ class ChannelPool extends ManagedChannel {
     private void release() {
       int newCount = outstandingRpcs.decrementAndGet();
       if (newCount < 0) {
-        throw new IllegalStateException("Bug: reference count is negative!: " + newCount);
+        LOG.log(Level.WARNING, "Bug: Reference count is negative!: " + newCount);
       }
 
       // Must check outstandingRpcs after shutdownRequested (in reverse order of retain()) to ensure
@@ -526,6 +526,8 @@ class ChannelPool extends ManagedChannel {
   static class ReleasingClientCall<ReqT, RespT> extends SimpleForwardingClientCall<ReqT, RespT> {
     @Nullable private CancellationException cancellationException;
     final Entry entry;
+    private final AtomicBoolean wasClosed = new AtomicBoolean();
+    private final AtomicBoolean wasReleased = new AtomicBoolean();
 
     public ReleasingClientCall(ClientCall<ReqT, RespT> delegate, Entry entry) {
       super(delegate);
@@ -542,17 +544,32 @@ class ChannelPool extends ManagedChannel {
             new SimpleForwardingClientCallListener<RespT>(responseListener) {
               @Override
               public void onClose(Status status, Metadata trailers) {
-                try {
+                if (wasClosed.compareAndSet(false, true)) {
                   super.onClose(status, trailers);
-                } finally {
-                  entry.release();
+                  if (wasReleased.compareAndSet(false, true)) {
+                    entry.release();
+                  } else {
+                    LOG.log(
+                        Level.WARNING,
+                        "The entry is already released, this indicates that there is an exception on start of the call");
+                  }
+                } else {
+                  LOG.log(
+                      Level.WARNING,
+                      "onClose() has already being called, please make sure onClose() is not being manually called, otherwise this may indicate a bug in gRPC library");
                 }
               }
             },
             headers);
       } catch (Exception e) {
         // In case start failed, make sure to release
-        entry.release();
+        if (wasReleased.compareAndSet(false, true)) {
+          entry.release();
+        } else {
+          LOG.log(
+              Level.WARNING,
+              "The entry is already released, this indicates that onClose() has already being called previously");
+        }
         throw e;
       }
     }
--- a/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
+++ b/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
@@ -561,13 +561,16 @@ class ChannelPool extends ManagedChannel {
               @Override
               public void onClose(Status status, Metadata trailers) {
                 if (wasClosed.compareAndSet(false, true)) {
-                  super.onClose(status, trailers);
-                  if (wasReleased.compareAndSet(false, true)) {
-                    entry.release();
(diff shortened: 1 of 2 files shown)


gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
```diff
@@ -542,17 +544,32 @@ public void start(Listener<RespT> responseListener, Metadata headers) {
...
             headers);
       } catch (Exception e) {
         // In case start failed, make sure to release
-        entry.release();
+        if (wasReleased.compareAndSet(false, true)) {
+          entry.release();
+        } else {
+          LOG.log(
+              Level.WARNING,
+              "The entry is already released, this indicates that onClose() has already being called previously");
```
meltsufin, line 571:
```suggestion
              "The entry is already released. This indicates that onClose() has already been called previously");
```

meltsufin (approved):
I'm not sure if all of these logs should be warnings, but I think the safeguards make sense.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
```diff
@@ -542,17 +560,32 @@ public void start(Listener<RespT> responseListener, Metadata headers) {
             new SimpleForwardingClientCallListener<RespT>(responseListener) {
               @Override
               public void onClose(Status status, Metadata trailers) {
-                try {
+                if (wasClosed.compareAndSet(false, true)) {
                   super.onClose(status, trailers);
-                } finally {
-                  entry.release();
+                  if (wasReleased.compareAndSet(false, true)) {
+                    entry.release();
```
igorbernstein2, line 566: This needs to be in a finally block so that ref count can be decremented even if the caller's responseListener throws an error. In other words it doesnt matter if the caller's handler has a bug, from the channel perspective, the call is done and the ref count should be decremented

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
```diff
@@ -542,17 +560,32 @@ public void start(Listener<RespT> responseListener, Metadata headers) {
...
+                  } else {
+                    LOG.log(
+                        Level.WARNING,
+                        "Call was not closed but entry was released. This may be due to an exception on start of the call.");
```
igorbernstein2, line 570: I would reword this to "channel refcount was released before the rpc closed. This may be due to ..."

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
```diff
@@ -526,6 +542,8 @@ public <RequestT, ResponseT> ClientCall<RequestT, ResponseT> newCall(
   static class ReleasingClientCall<ReqT, RespT> extends SimpleForwardingClientCall<ReqT, RespT> {
     @Nullable private CancellationException cancellationException;
     final Entry entry;
+    private final AtomicBoolean wasClosed = new AtomicBoolean();
+    private final AtomicBoolean wasReleased = new AtomicBoolean();
```
igorbernstein2, line 546: Why both bools?

blakeli0: So that we can identify different scenarios next time it happens
1. Multiple calls to onClose()
2. onClose() being called first, try-catch of `start()` being called second
3. try-catch of `start()` being called first, onClose() being called second.

With only one flag, I don't think we can identify these different scenarios.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
blakeli0: Because `channel refcount` is easier to understand than `entry`? Or put more focus on `released before closed`?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
blakeli0: Updated

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
```diff
@@ -542,17 +560,35 @@ public void start(Listener<RespT> responseListener, Metadata headers) {
...
+                      LOG.log(
+                          Level.WARNING,
+                          "Entry was released before the call is closed. This may be due to an exception on start of the call.");
+                    }
+                  }
+                } else {
+                  LOG.log(
+                      Level.WARNING,
+                      "Call is being closed more than once. Please make sure that onClose() is not being manually called.");
                 }
```
igorbernstein2, line 579: The nesting makes it hard to read, I think early returns would make it easier:

```java
if (wasClosed.compareAndSet(false, true)) {
   LOG.log(
                      Level.WARNING,
                      "Call is being closed more than once. Please make sure that onClose() is not being manually called.");
   return;
}

if (!wasReleased.compareAndSet(false, true)) {
    LOG.log(
                      Level.WARNING,
                      "Call is being closed more than once. Please make sure that onClose() is not being manually called.");
    return;
}
entry.release();
```

blakeli0: Thanks for the suggestion! Yeah with try-finally being added, it's getting harder to read. I extracted the checking of `wasClosed` to earlier to remove one indent. Kept the checking of `wasRelease` in finally because I think it's logically easier to understand even though has more indents.

PR #2194, opened 2023-10-23
feat: add mixin protos

### LINKED ISSUE #2085: The generator should generate code and tests for Mixins if Minxins are defined in service yaml
When there's Mixins in service yaml, generator should generate corresponding test code.

For example: 
In order to generate tests for the Mixins, we need to pass `google/iam/iam_policy.proto` and `google/cloud/location/locations.proto` to the generator explicitly even though Mixins defined in [baremetalsolution_v2.yaml](https://github.com/googleapis/googleapis/blob/master/google/cloud/baremetalsolution/v2/baremetalsolution_v2.yaml#L6C2-L10).

The code difference with and without Mixin protos: https://paste.googleplex.com/6122778548764672

JoeWang1127:
Fixes #2085 ☕️


JoeWang1127:
A test to verify the PR can generate the identical library with and without the Mixin protos in the `proto_library_with_info` target. I'll use google/cloud/baremetalsolution/v2 as an example.
1. Build the GAPIC library using main branch as a baseline. In googleapis repository root:
```
bazelisk build google/cloud/baremetalsolution/v2:google-cloud-baremetalsolution-v2-java
```
2. Pull this PR to your local repository, install it locally:
```
mvn clean install -DskipTests
```
3. In googleapis repository, change the WORKSPACE to use a local gapic-generator-java ([guidance](https://github.com/googleapis/sdk-platform-java/blob/392ad578313526e25bbd25421af3bd1e3339716e/gapic-generator-java/DEVELOPMENT.md#running-the-plugin-under-googleapis-with-local-gapic-generator-java)). Change `_gapic_generator_java_version` to `2.28.1-SNAPSHOT` (the current local version).
4. In googleapis repository, build the library:
```
bazelisk build google/cloud/baremetalsolution/v2:google-cloud-baremetalsolution-v2-java
```
5. Compare the library built in step 1 and 4, the difference are versions in `build.gradle`:
```
joewa-macbookpro:googleapis joewa$ diff -r ~/Downloads/main/ ~/Downloads/branch_with_2_protos/
diff -r /Users/joewa/Downloads/main/gapic-google-cloud-baremetalsolution-v2-java/build.gradle /Users/joewa/Downloads/branch_with_2_protos/gapic-google-cloud-baremetalsolution-v2-java/build.gradle
9,14c9,14
<   api 'com.google.api:gax:2.36.0'
<   testImplementation 'com.google.api:gax:2.36.0:testlib'
<   api 'com.google.api:gax-grpc:2.36.0'
<   testImplementation 'com.google.api:gax-grpc:2.36.0:testlib'
<   api 'com.google.api:gax-httpjson:2.36.0'
<   testImplementation 'com.google.api:gax-httpjson:2.36.0:testlib'
---
>   api 'com.google.api:gax:2.36.1-SNAPSHOT'
>   testImplementation 'com.google.api:gax:2.36.1-SNAPSHOT:testlib'
>   api 'com.google.api:gax-grpc:2.36.1-SNAPSHOT'
>   testImplementation 'com.google.api:gax-grpc:2.36.1-SNAPSHOT:testlib'
>   api 'com.google.api:gax-httpjson:2.36.1-SNAPSHOT'
>   testImplementation 'com.google.api:gax-httpjson:2.36.1-SNAPSHOT:testlib'
```
6. In googleapis, remove [two lines](https://github.com/googleapis/googleapis/blob/master/google/cloud/baremetalsolution/v2/BUILD.bazel#L53-L54) in `google/cloud/baremetalsolution/v2/BUILD.bazel`.
7. Build the library with `--subcommands` option:
```
bazelisk build --subcommands google/cloud/baremetalsolution/v2:google-cloud-baremetalsolution-v2-java
```
Note that in the protoc command to generate library with gapic-generator-java (search for keyword `protoc-gen-java_gapic` in the output), there's no Mixin protos in the command line.
8. Compare the library built in step 4 and 7, there's no difference.
```
joewa-macbookpro:googleapis joewa$ diff -r ~/Downloads/branch_with_2_protos/ ~/Downloads/branch_without_2_protos/
joewa-macbookpro:googleapis joewa$ 
```

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -1059,13 +1064,50 @@ private static Map<String, FileDescriptor> getFilesToGenerate(CodeGeneratorReque
 
       fileDescriptors.put(fileDescriptor.getName(), fileDescriptor);
     }
+    extraMixins.forEach(fileDescriptor -> parseDependencies(fileDescriptor, fileDescriptors));
```
blakeli0, line 1067: I _think_ we can just add the three mixin FileDescriptor to `fileDescriptors` here. Can you please give it a try?

JoeWang1127: done

JoeWang1127: It seems that if we always add Mixin protos, in `reflect-config.json`, all message types associated with Mixins are generated (difference: https://paste.googleplex.com/6265925991923712).

Is this OK?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -91,15 +95,17 @@ public class Parser {
...
   // Mirrors the sanitizer allowlist.
-  private static final Set<String> MIXIN_ALLOWLIST =
-      ImmutableSet.of(
-          "google.iam.v1.IAMPolicy",
-          "google.longrunning.Operations",
-          "google.cloud.location.Locations");
+  private static final Map<String, FileDescriptor> MIXIN_ALLOWLIST =
+      ImmutableMap.of(
+          "google.iam.v1.IAMPolicy", IamPolicyProto.getDescriptor(),
+          "google.longrunning.Operations", LocationsProto.getDescriptor(),
```
lqiu96, line 101: Should this be `OperationsProto.getDescriptor()`?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -271,12 +277,14 @@ public static List<Service> parseServices(
...
-                .map(a -> a.getName())
-                .collect(Collectors.toSet());
+        serviceYamlProtoOpt
+            .map(
+                value ->
+                    value.getApisList().stream()
+                        .map(Api::getName)
+                        .filter(MIXIN_ALLOWLIST::containsKey)
+                        .collect(Collectors.toSet()))
+            .orElse(Collections.emptySet());
```
lqiu96, line 287: I think if we set the ServiceYaml Mixins as the source of truth, then we probably don't need the checks we're adding in L265-L273 and that we should validate that the ServiceYaml Mixins are correct.

I'm thinking:
1. L265-L273 could be updated to just filter to match for the Mixins in the allowlist. If the new filtered list is the same size as the Services list in L242, then we block generation
2. Since mixedInApis is the source of truth, we should validate mixidInApis matches the Mixins found in the filtered list (they should all match)

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
JoeWang1127: Yes, thanks for catching this.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -232,9 +239,8 @@ public static List<Service> parseServices(
       List<Service> outputMixinServices,
       Transport transport) {
     Map<String, FileDescriptor> fileDescriptors = getFilesToGenerate(request);
-
     List<Service> services = new ArrayList<>();
-    for (String fileToGenerate : request.getFileToGenerateList()) {
+    for (String fileToGenerate : mergeExtraMixinsWith(request.getFileToGenerateList())) {
```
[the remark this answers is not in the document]
JoeWang1127: Sure, that makes sense.

gapic-generator-java/src/test/resources/echo_without_mixins_v1beta1.yaml
```diff
@@ -0,0 +1,93 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+type: google.api.Service
+config_version: 3
+name: showcase.googleapis.com
+title: Client Libraries Showcase API
+
+apis:
```
blakeli0, line 20: Per https://google.aip.dev/client-libraries/4234, mixins must be configured in `apis` section.
```
To be generated into a client library, a mixin service must be declared under [apis](https://github.com/googleapis/googleapis/blob/master/google/api/service.proto#L96) in a [google.api.Service](https://github.com/googleapis/googleapis/blob/master/google/api/service.proto). 
Furthermore, only the RPCs with [google.api.http](https://github.com/googleapis/googleapis/blob/master/google/api/http.proto) bindings declared in the [http](https://github.com/googleapis/googleapis/blob/master/google/api/service.proto#L124) configuration of the same [google.api.Service](https://github.com/googleapis/googleapis/blob/master/google/api/service.proto) can be generated (more details on these bindings in [AIP-127](https://google.aip.dev/127)). If a mixin service RPC does not have a [google.api.http](https://github.com/googleapis/googleapis/blob/master/google/api/http.proto) rule declared in the [google.api.Service](https://github.com/googleapis/googleapis/blob/master/google/api/service.proto), it must not be generated.
```

I think we should have at least 4 test cases:
1. Mixins should be generated if they are configured in both `apis` and `http` sections.
2. Mixins should not be generated if they are configured in `apis` but not in `http` sections.
3. Mixins should not be generated if they are configured in `http` but not in `apis` sections. This is probably a misconfiguration.
4. Mixins should not be generated if they are not configured in `apis` or `http` sections.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -318,9 +326,9 @@ public static List<Service> parseServices(
       apiDefinedRpcs.addAll(
           service.methods().stream().map(m -> m.name()).collect(Collectors.toSet()));
     }
-    // Mix-in APIs only if the protos are present and they're defined in the service.yaml file.
+    // Mix-in APIs if they're defined in the service.yaml file.
     Set<Service> outputMixinServiceSet = new HashSet<>();
-    if (servicesContainBlocklistedApi && !mixedInApis.isEmpty()) {
+    if (servicesContainBlocklistedApi) {
```
blakeli0, line 331: `!mixedInApis.isEmpty()` check is still required, see my [comments](https://github.com/googleapis/sdk-platform-java/pull/2194/files#r1376566678) in the tests.

gapic-generator-java/src/test/resources/echo_without_mixins_v1beta1.yaml
JoeWang1127: Thanks for pointing it out.

I'll add two tests for each scenario, one with Mixin protos and one without.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
JoeWang1127: done.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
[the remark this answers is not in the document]
JoeWang1127: done.

JoeWang1127:
After the change, the methods in longrunning Operations are generated in showcase since service yaml has configured Mixins in both apis and http bindings.

blakeli0:
> After the change, the methods in longrunning Operations are generated in showcase since service yaml has configured Mixins in both apis and http bindings.

I think this is fine as they are supposed to be generated, cc @lqiu96  to confirm.

lqiu96:
> > After the change, the methods in longrunning Operations are generated in showcase since service yaml has configured Mixins in both apis and http bindings.
> 
> I think this is fine as they are supposed to be generated, cc @lqiu96 to confirm.

Yeah, this looks like the correct behavior as it's both listed in the [http section](https://github.com/googleapis/gapic-showcase/blob/c6512469703a716c619058249f2bb55d81c5ee37/schema/google/showcase/v1beta1/showcase_v1beta1.yaml#L78-L85) and is listed in the [apis section](https://github.com/googleapis/gapic-showcase/blob/c6512469703a716c619058249f2bb55d81c5ee37/schema/google/showcase/v1beta1/showcase_v1beta1.yaml#L30) of the service yaml file. Both conditions hit and should be generated.

I need to do some more digging because I think Operations is the weirdest mixin in terms of behavior. It's the oldest mixin and has its own client. Plus we've had to make some handwritten changes to the Operations client in Gax so I have no idea how well these new stubs play (or don't play) with the Operations Client and custom http bindings.

At first thought, this seems like the more cohesive and intended behavior for GAPICs. No longer are IAM and Location callables spliced into the ServiceClient and Operations in a separate client. But also having duplicated Operation callables (1 in the ServiceClient and 1 in the OperationsClient) also seems odd.

JoeWang1127:
> But also having duplicated Operation callables (1 in the ServiceClient and 1 in the OperationsClient) also seems odd.

If this is not right, should we address it in this PR? To me, this sounds like something happens in composer.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
JoeWang1127: If we *always* add all Mixins, then `reflect-config.json` will contain message types associated with all Mixins. Previously, we only add Mixins defined in service yaml so certain message types are excluded.

I changed the logic to only add fileDescriptor of Mixins, but not fileDescriptors of their dependencies, but the result is the same (all message types associated with Mixins are in `reflect-config.json`).

> Ideally, we can keep all other logics and this is all we need to change in this PR

We need to change the logic of getting filesToGenerate as well; otherwise Mixin methods won't be generated. I confirmed this with unit tests.
https://github.com/googleapis/sdk-platform-java/blob/1b861ec78a7d48549140224cc15aa793e3d890ec/gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java#L245

JoeWang1127:
I updated the golden files in `showcase/` and `test/integration`
However, the build and integration tests failed (they passed in my local environment).

lqiu96:
> > But also having duplicated Operation callables (1 in the ServiceClient and 1 in the OperationsClient) also seems odd.
> 
> If this is not right, should we address it in this PR? To me, this sounds like something happens in composer.

I'll need to take another look at this, but I'm not entirely sure how much we address/ fix in the Composer right now. Removing the OperationsClient (imo seems like the better long term option) is a breaking change and breaks LRO functionality. We might need to remove the callables from the ServiceClient if we only want one set of Operation callables. Maybe it could make sense where both exist and the use of the OperationsClient is tucked away.

JoeWang1127:
> I updated the golden files in `showcase/` and `test/integration` However, the build and integration tests failed (they passed in my local environment).

Run `mvn clean install` in the repository root and update the golden files solved this issue.

Under the hook, both bazel and `generate_library.sh` don't re-install the SNAPSHOT version of the generator, so you have to re-install it manually to reflect latest changes in the generator.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
JoeWang1127: > We need to change the logic of getting filesToGenerate as well; otherwise Mixin methods won't be generated. I confirmed this with unit tests.

This part is gone.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -252,31 +255,29 @@ public static List<Service> parseServices(
...
-        !serviceYamlProtoOpt.isPresent()
-            ? Collections.emptySet()
-            : serviceYamlProtoOpt.get().getApisList().stream()
-                .filter(a -> MIXIN_ALLOWLIST.contains(a.getName()))
-                .map(a -> a.getName())
-                .collect(Collectors.toSet());
+    // Mixin services that are stated in the service yaml file.
+    List<Service> mixinServices =
+        serviceYamlProtoOpt
+            .map(
```
blakeli0, line 261: This logic is getting hard to read with a lot of nested indents, can we refactor it to make it easier to read?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -252,31 +255,29 @@ public static List<Service> parseServices(
...
+                                parseService(
+                                    MIXIN_ALLOWLIST.get(mixinApi),
+                                    messageTypes,
+                                    resourceNames,
+                                    serviceYamlProtoOpt,
+                                    serviceConfigOpt,
+                                    outputArgResourceNames,
+                                    transport)
+                                    .stream())
+                        .sorted((s1, s2) -> s2.name().compareTo(s1.name()))
```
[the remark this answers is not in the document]
JoeWang1127: In the main branch, the Mixin services are sorted, so I copied the logic.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
JoeWang1127: Sure, will do.

test/integration/goldens/storage/src/com/google/storage/v2/StorageClient.java
```diff
@@ -848,8 +848,7 @@ public final Bucket lockBucketRetentionPolicy(LockBucketRetentionPolicyRequest r
    * }
    * }</pre>
    *
-   * @param resource REQUIRED: The resource for which the policy is being requested. See the
-   *     operation documentation for the appropriate value for this field.
+   * @param resource
```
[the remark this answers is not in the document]
JoeWang1127: No, I'm not familiar with how comments are generated.

@lqiu96 do you know where I should look into?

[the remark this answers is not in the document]
JoeWang1127: After a quick investigation, this might have something related to `SourceCodeInfo` doesn't present in FileDescriptor, though I don't know why.

lqiu96:
I used this generator change to generate a new Speech v1 library as it has custom http bindings for Mixins. I believe having the operation callables generated in the client is fine as the HttpJsonTransportStub generates the ApiMethodDescriptors with the custom bindings. It looks like only the RPC listed in the http section of the service yaml files are generated as well and this is the behavior that we want.

I used the newly generated callables to make the Operation calls and it looks to make the calls properly. Sample code: gpaste/6136898236973056

Should have no impact to customers except that we're exposing a few new methods per client. The actual LRO still uses the GAX OperationsClient so logic is untouched. Eventually we could migrate away from using the GAX's OperationsClient and use the callables in each client, but that is out of the scope of this PR.

test/integration/goldens/storage/src/com/google/storage/v2/StorageClient.java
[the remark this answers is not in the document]
lqiu96: I'm not the most familiar with how comments are generated either. I *believe* most of the logic is inside ServceClientCommentComposer and these lines are relevant: https://github.com/googleapis/sdk-platform-java/blob/80f23ec98922bf8f29f3f6960fb20fbeb6231706/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/comment/ServiceClientCommentComposer.java#L191-L201

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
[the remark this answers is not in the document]
JoeWang1127: If they are not sorted, the mixin methods will have random orders in the output java files.

test/integration/goldens/kms/src/com/google/cloud/kms/v1/KeyManagementServiceClient.java
```diff
@@ -3685,9 +3685,7 @@ public final UnaryCallable<GetIamPolicyRequest, Policy> getIamPolicyCallable() {
 
   // AUTO-GENERATED DOCUMENTATION AND METHOD.
   /**
-   * Lists information about the supported locations for this service.
-   *
-   * <p>Sample code:
+   * Sample code:
```
JoeWang1127, line 3688: This is the last file with comment issue.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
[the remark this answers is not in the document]
burkedavison: Very-optional syntax sugar if you want to make the inversion order more visually explicit:
```java
.sorted(Comparator.comparing(Service::getName).reversed())
```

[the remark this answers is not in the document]
JoeWang1127: Thanks for the suggestion @burkedavison.

test/integration/goldens/kms/src/com/google/cloud/kms/v1/KeyManagementServiceClient.java
JoeWang1127: After some investigation, I think this is because `LocationsProto.getDescriptor()` doesn't contain `sourceCodeInfo_`

JoeWang1127: Now this comment is restored.

test/integration/goldens/storage/src/com/google/storage/v2/StorageClient.java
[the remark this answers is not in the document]
JoeWang1127: Thanks @lqiu96, this is the place to add method description. I think for this part to work, the fileDescriptor has to have SourceCodeInfo.

test/integration/goldens/kms/src/com/google/cloud/kms/v1/KeyManagementServiceClient.java
blakeli0: What was the fix?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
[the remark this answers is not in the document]
blakeli0: I see that we are doing sorting in multiple places, is the order here actually used? For example, there is a sorting for [updatedMixinMethods](https://github.com/googleapis/sdk-platform-java/pull/2194/files#diff-648bf4fdb398cabe3694d2e383a4b273f690c2279af3247ede8c4d0e584a624cR371-R374), which kind of implies that it is used for the order of mixing methods. Can you please confirm?

test/integration/goldens/kms/src/com/google/cloud/kms/v1/KeyManagementServiceClient.java
JoeWang1127: If the protos are dependent on mixins or mixins are passed through command line, then we don't add mixin file descriptors to `fileDescriptors` because the latter don't have SourceCodeInfo.

https://github.com/googleapis/sdk-platform-java/blob/ab6f4944df4ffc6a650a1a1e53d7302a64524933/gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java#L1062-L1063

Then use file descriptor in `fileDescriptors`, rather than `MIXIN_ALLOWLIST`.

https://github.com/googleapis/sdk-platform-java/blob/ab6f4944df4ffc6a650a1a1e53d7302a64524933/gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java#L275

Though, in the future, if the protos are not dependent on mixins and no mixins are provided, the mixin method description (not the sample code) might lose. The main issue is I don't know how to get `SourceCodeInfo` from `getFileDescriptor` method.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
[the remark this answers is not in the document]
JoeWang1127: This sort ensures the Mixin methods in are grouped by Mixin services, for example, methods from operations.proto are always before methods from locations.proto.

https://github.com/googleapis/sdk-platform-java/blob/ab6f4944df4ffc6a650a1a1e53d7302a64524933/gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java#L283

These two sorts seems no impact on output file, judging by the comments, they are needed to determinist test result.

https://github.com/googleapis/sdk-platform-java/blob/ab6f4944df4ffc6a650a1a1e53d7302a64524933/gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java#L370-L376

https://github.com/googleapis/sdk-platform-java/blob/ab6f4944df4ffc6a650a1a1e53d7302a64524933/gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java#L388-L392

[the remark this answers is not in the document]
blakeli0: Makes sense, thanks!

JoeWang1127:
The main issue of this PR is that I can't get `sourceCodeInfo` from generated Java code and I' not sure whether this is by design (see [comment](https://github.com/protocolbuffers/protobuf/issues/8728#issuecomment-860740530) in a relevant protobuf issue) or not.

The impact of the above issue is the missing of method description in Mixin protos as these descriptions only exist in `sourceCodeInfo` of a descriptor file.

We need to think another way to address #2085.

PR #2233, opened 2023-11-09
fix: use correct path as input to owlbot-copy

diegomarquezp:
Currently, the library generation integration test for `google/iam/v1` succeeds although no java files were transferred to the final folder. A check was added in the IT to prevent this situation from passing.

This fix is intended to adapt to `.OwlBot.yaml` files that specify the full path in googleapis-gen to the desired source library. The tradeoff is an extra argument to `generate_library.sh`

This [example](https://github.com/googleapis/google-cloud-java/blob/aa81513e6f6081bb7d8f1dd24cb498fd71fd8913/java-iam/.OwlBot.yaml#L27) does not use wildcards:
```yaml
deep-copy-regex:
  - source: "/google/iam/v2beta/google-cloud-iam-v2beta-java/gapic-google-cloud-iam-v2beta-java/src"
    dest: "/owl-bot-staging/java-iam/v2beta/google-iam-policy/src"
  - source: "/google/iam/v2/google-cloud-iam-v2-java/gapic-google-cloud-iam-v2-java/src"
    dest: "/owl-bot-staging/java-iam/v2/google-iam-policy/src"
```

[Our current logic](https://github.com/googleapis/sdk-platform-java/blob/445477f7fec3a6d4be9df4c9cf3f01c007307645/library_generation/postprocess_library.sh#L61-L70) uses a `generated-java` folder that will not produce a match for owlbot copy using the yaml above.
This logic assumed all `.OwlBot.yaml` files looked like [this one](https://github.com/googleapis/google-cloud-java/blob/aa81513e6f6081bb7d8f1dd24cb498fd71fd8913/java-alloydb/.OwlBot.yaml#L26)

```yaml
deep-copy-regex:
- source: "/google/cloud/alloydb/(v.*)/.*-java/proto-google-.*/src"
  dest: "/owl-bot-staging/java-alloydb/$1/proto-google-cloud-alloydb-$1/src"
```

### DIFF
--- a/library_generation/generate_library.sh
+++ b/library_generation/generate_library.sh
@@ -72,6 +72,10 @@ case $key in
     versions_file="$2"
     shift
     ;;
+  --googleapis_gen_folder_name)
+    googleapis_gen_folder_name="$2"
+    shift
+    ;;
   *)
     echo "Invalid option: [$1]"
     exit 1
@@ -300,10 +304,13 @@ then
   echo "post processing is disabled"
   exit 0
 fi
-if [ -z "${versions_file}" ];then
-  echo "no versions.txt argument provided. Please provide one in order to enable post-processing"
-  exit 1
-fi
+for necessary_argument in versions_file googleapis_gen_folder_name; do
+  if [ -z "${necessary_argument}" ];then
+    echo "no ${necessary_argument} argument provided. Please provide one in order to enable post-processing"
+    exit 1
+  fi
+done
+
 workspace="${output_folder}/workspace"
 if [ -d "${workspace}" ]; then
   rm -rdf "${workspace}"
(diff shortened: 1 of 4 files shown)


JoeWang1127 (approved):

blakeli0:
> This fix is intended to adapt to `.OwlBot.yaml` files that specify the full path in googleapis-gen to the desired source library. The tradeoff is an extra argument to `generate_library.sh`

I would rather us fixing the non-standard owlbot configuration rather than accommodating it in our scripts, especially at the cost of one more parameter. Is there any harm in fixing the owlbot config for iam?

diegomarquezp:
> > This fix is intended to adapt to `.OwlBot.yaml` files that specify the full path in googleapis-gen to the desired source library. The tradeoff is an extra argument to `generate_library.sh`
> 
> I would rather us fixing the non-standard owlbot configuration rather than accommodating it in our scripts, especially at the cost of one more parameter. Is there any harm in fixing the owlbot config for iam?

@blakeli0 Sounds good. I created https://github.com/googleapis/google-cloud-java/pull/10050 to address this.

diegomarquezp:
This approach is obsolete. https://github.com/googleapis/google-cloud-java/pull/10050 fixed this

PR #2234, opened 2023-11-10
chore: update showcase version

JoeWang1127:
In this PR:
- Update showcase version to `0.29.0`.
- Update showcase client.


JoeWang1127:
Should I worry about ci/showcase-clirr check?

burkedavison:
No, we don't care about clirr in this case since we are deliberately changing the showcase source and there are no downstream compatibility concerns with this. Clirr check can be ignored here.

burkedavison (approved):

burkedavison (approved):

PR #2235, opened 2023-11-10
fix: confirm owlbot-copy succeeeded to transfer java files

diegomarquezp:
Follow up of https://github.com/googleapis/google-cloud-java/pull/10050
* Confirms java files have been transferred by owl-bot-copy
* Adds `google/iam/v2` in the integration test list. 
* `google/iam/v1` is blacklisted from postprocessing list because it is not listed in its [owlbot yaml ](https://github.com/googleapis/google-cloud-java/blob/18020600f38e9cf966d3b4c12035954a7124344a/java-iam/.OwlBot.yaml#L25-L28)

### DIFF
--- a/library_generation/test/generate_library_integration_test.sh
+++ b/library_generation/test/generate_library_integration_test.sh
@@ -85,7 +85,7 @@ fi
 grep -v '^ *#' < "${proto_path_list}" | while IFS= read -r line; do
   proto_path=$(echo "$line" | cut -d " " -f 1)
   repository_path=$(echo "$line" | cut -d " " -f 2)
-  is_handwritten=$(echo "$line" | cut -d " " -f 3)
+  skip_postprocessing=$(echo "$line" | cut -d " " -f 3)
   # parse destination_path
   pushd "${output_folder}"
   echo "Checking out googleapis-gen repository..."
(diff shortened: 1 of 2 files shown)


library_generation/test/resources/proto_path_list.txt
```diff
@@ -17,7 +17,7 @@ google/cloud/redis/v1 java-redis false
 google/cloud/redis/v1beta1 java-redis false
 # google/example/library/v1 google-cloud-example-library-v1-java null false
 google/devtools/containeranalysis/v1 java-containeranalysis false
-google/iam/v1 java-iam false
+google/iam/v2 java-iam false
```
JoeWang1127, line 20: Could you add another line for `google/iam/v2` because I use `google/iam/v1` to test proto-only library.

diegomarquezp: I'll disable `v1` for the post-processing IT and add `v2` separately. The `iam` [owlbot yaml](https://github.com/googleapis/google-cloud-java/blob/18020600f38e9cf966d3b4c12035954a7124344a/java-iam/.OwlBot.yaml#L25-L28) only supports `v2` and `v2beta1` so using `v1` will fail to move any java files into the workspace

library_generation/test/generate_library_integration_test.sh
```diff
@@ -173,6 +173,11 @@ grep -v '^ *#' < "${proto_path_list}" | while IFS= read -r line; do
   echo "Generate library finished."
   echo "Compare generation result..."
   if [ $enable_postprocessing == "true" ]; then
```
JoeWang1127, line 175: nit: `"${enable_postprocessing}"`

JoeWang1127 (approved):

library_generation/test/generate_library_integration_test.sh
diegomarquezp: thanks, I added this and a couple other format fixes

blakeli0 (approved):

Issue #1858, opened 2023-07-18
Improve LRO's Error Message on Exception


lqiu96:
We should provide customers a more detailed error message when there is a timeout for LROs. Customers are seeing `CancellationException` and have no idea how to proceed. It may be perceived as an error on their end when they instead need to update the Retry configurations.

Possible fix:
Update the `CancellationException()` to throw a more detailed error message (https://github.com/googleapis/sdk-platform-java/blob/aeea0d1b22647cbdc344e9916dcc595f1df945db/gax-java/gax/src/main/java/com/google/api/gax/longrunning/OperationTimedPollAlgorithm.java#L80)

(Stretch)
Update the google-cloud-java README docs with more information and link it from the CancellationException.

diegomarquezp:
Finding: `The task was cancelled.` comes from [guava](https://github.com/google/guava/blob/9b8035d654cd4e36e79e1b8329f4457e0f3087f8/guava/src/com/google/common/util/concurrent/AbstractFuture.java#L590). Apparently the `OperationTimedPollAlgorithm` exception would bubble up if it failed instead of getting cancelled.

PR #2236, opened 2023-11-10
fix: improve information on CancellationException for LROs

diegomarquezp:
Fixes https://github.com/googleapis/sdk-platform-java/issues/1858
When a LRO runs out of retries/time, it throws a `CancellationException` that is then swallowed by guava's `AbstractFuture`, meaning there is no simple way to bubble up a custom exception with message.

This approach uses a logger with a message that points to LRO documentation to ensure the user is informed on what they can do about this.

### DIFF
--- a/gax-java/gax/src/main/java/com/google/api/gax/longrunning/OperationTimedPollAlgorithm.java
+++ b/gax-java/gax/src/main/java/com/google/api/gax/longrunning/OperationTimedPollAlgorithm.java
@@ -77,7 +77,9 @@ public class OperationTimedPollAlgorithm extends ExponentialRetryAlgorithm {
     if (super.shouldRetry(nextAttemptSettings)) {
       return true;
     }
-    throw new CancellationException();
+    throw new CancellationException(
+        "The task has been cancelled. Please refer to "
+            + "https://github.com/googleapis/google-cloud-java#lro-timeouts for more information");
   }
 
   // Note: if the potential time spent is exactly equal to the totalTimeout,
--- a/gax-java/gax/src/main/java/com/google/api/gax/longrunning/OperationTimedPollAlgorithm.java
+++ b/gax-java/gax/src/main/java/com/google/api/gax/longrunning/OperationTimedPollAlgorithm.java
@@ -35,7 +35,10 @@ import com.google.api.core.NanoClock;
 import com.google.api.gax.retrying.ExponentialRetryAlgorithm;
 import com.google.api.gax.retrying.RetrySettings;
 import com.google.api.gax.retrying.TimedAttemptSettings;
+import com.google.common.annotations.VisibleForTesting;
 import java.util.concurrent.CancellationException;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * Operation timed polling algorithm, which uses exponential backoff factor for determining when the
@@ -44,6 +47,8 @@ import java.util.concurrent.CancellationException;
  */
 public class OperationTimedPollAlgorithm extends ExponentialRetryAlgorithm {
 
+  @VisibleForTesting
+  public static Logger LOGGER = Logger.getLogger(OperationTimedPollAlgorithm.class.getName());
   public static final String LRO_TROUBLESHOOTING_LINK =
       "https://github.com/googleapis/google-cloud-java#lro-timeouts";
 
@@ -81,9 +86,11 @@ public class OperationTimedPollAlgorithm extends ExponentialRetryAlgorithm {
     if (super.shouldRetry(nextAttemptSettings)) {
       return true;
     }
-    throw new CancellationException(
-        "The task has been cancelled. Please refer to "
-            + LRO_TROUBLESHOOTING_LINK + " for more information");
+    if (LOGGER.isLoggable(Level.SEVERE)) {
+      LOGGER.log(Level.SEVERE, "The task has been cancelled. Please refer to "
+          + LRO_TROUBLESHOOTING_LINK + " for more information");
+    }
+    throw new CancellationException();
   }
 
   // Note: if the potential time spent is exactly equal to the totalTimeout,
--- a/gax-java/gax/src/main/java/com/google/api/gax/longrunning/OperationTimedPollAlgorithm.java
+++ b/gax-java/gax/src/main/java/com/google/api/gax/longrunning/OperationTimedPollAlgorithm.java
@@ -43,6 +43,10 @@ import java.util.concurrent.CancellationException;
  * algorithm cancels polling.
  */
 public class OperationTimedPollAlgorithm extends ExponentialRetryAlgorithm {
+
+  public static final String LRO_TROUBLESHOOTING_LINK =
+      "https://github.com/googleapis/google-cloud-java#lro-timeouts";
+
   /**
    * Creates the polling algorithm, using the default {@code NanoClock} for time computations.
    *
@@ -79,7 +83,7 @@ public class OperationTimedPollAlgorithm extends ExponentialRetryAlgorithm {
     }
     throw new CancellationException(
         "The task has been cancelled. Please refer to "
-            + "https://github.com/googleapis/google-cloud-java#lro-timeouts for more information");
+            + LRO_TROUBLESHOOTING_LINK + " for more information");
   }
 
   // Note: if the potential time spent is exactly equal to the totalTimeout,
--- a/gax-java/gax/src/test/java/com/google/api/gax/util/FakeLogHandler.java
+++ b/gax-java/gax/src/test/java/com/google/api/gax/util/FakeLogHandler.java
@@ -1,0 +1,26 @@
+package com.google.api.gax.util;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.logging.Handler;
+import java.util.logging.LogRecord;
+import java.util.stream.Collectors;
+
+public class FakeLogHandler extends Handler {
+  List<LogRecord> records = new ArrayList<>();
+
+  @Override
+  public void publish(LogRecord record) {
+    records.add(record);
+  }
+
+  @Override
+  public void flush() {}
+
+  @Override
+  public void close() throws SecurityException {}
+
+  public List<String> getAllMessages() {
+    return records.stream().map(LogRecord::getMessage).collect(Collectors.toList());
+  }
+}
--- a/gax-java/gax/src/test/java/com/google/api/gax/util/FakeLogHandler.java
+++ b/gax-java/gax/src/test/java/com/google/api/gax/util/FakeLogHandler.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package com.google.api.gax.util;
 
 import java.util.ArrayList;
@@ -6,6 +21,10 @@ import java.util.logging.Handler;
 import java.util.logging.LogRecord;
 import java.util.stream.Collectors;
 
+/*
+ * Convenience class that stores the log entries produced by any logger
+ * It can then be inspected - its entries are a list log records
+ */
 public class FakeLogHandler extends Handler {
   List<LogRecord> records = new ArrayList<>();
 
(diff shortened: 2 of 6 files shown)


gax-java/gax/src/main/java/com/google/api/gax/longrunning/OperationTimedPollAlgorithm.java
```diff
@@ -77,6 +87,13 @@ public boolean shouldRetry(TimedAttemptSettings nextAttemptSettings)
     if (super.shouldRetry(nextAttemptSettings)) {
       return true;
     }
+    if (LOGGER.isLoggable(Level.SEVERE)) {
```
lqiu96, line 90: QQ, is there a reason `SEVERE` logging level was picked? It seems like that is meant for more fatal application issues (and less for an RPC timeout), but I also am not sure if there is a better level (Info? Warning?).

diegomarquezp: That makes sense. My reasoning was more like "exception=severe", but in the context of a timeout on a request I agree something like `WARNING` is better

diegomarquezp: I decreased it to WARNING

gax-java/gax/src/main/java/com/google/api/gax/longrunning/OperationTimedPollAlgorithm.java
```diff
@@ -35,14 +35,24 @@
...
+import java.util.logging.Logger;
 
 /**
  * Operation timed polling algorithm, which uses exponential backoff factor for determining when the
  * next polling operation should be executed. If the polling exceeds the total timeout this
  * algorithm cancels polling.
  */
 public class OperationTimedPollAlgorithm extends ExponentialRetryAlgorithm {
+
+  @VisibleForTesting
```
lqiu96, line 50: nit: I don't think the logger needs a VisibleForTesting annotation.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITLongRunningOperation.java
```diff
@@ -121,6 +124,8 @@ public void testHttpJson_LROSuccessfulResponse_doesNotExceedTotalTimeout() throw
   @Test
   public void testGRPC_LROUnsuccessfulResponse_exceedsTotalTimeout_throwsDeadlineExceededException()
       throws Exception {
+    FakeLogHandler logHandler = new FakeLogHandler();
```
lqiu96, line 127: nit: Might be better to create field for FakeLogHandler that all tests can use, if possible

setUp() -> Creates it
clean() -> clears all the message so the next test can run

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITLongRunningOperation.java
```diff
@@ -160,6 +168,8 @@ public void testGRPC_LROUnsuccessfulResponse_exceedsTotalTimeout_throwsDeadlineE
   public void
       testHttpJson_LROUnsuccessfulResponse_exceedsTotalTimeout_throwsDeadlineExceededException()
           throws Exception {
+    FakeLogHandler logHandler = new FakeLogHandler();
+    OperationTimedPollAlgorithm.LOGGER.addHandler(logHandler);
```
lqiu96, line 172: Same here

lqiu96 (approved):
LGTM, added a few nits. Can you also have @burkedavison or @JoeWang1127 take a quick look? Thanks for this @diegomarquezp

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITLongRunningOperation.java
diegomarquezp: Makes sense, thanks! I only added a `setUp()` since creating a new handler wipes out the old handler with its maessages

gax-java/gax/src/main/java/com/google/api/gax/longrunning/OperationTimedPollAlgorithm.java
```diff
@@ -35,14 +35,23 @@
...
+import java.util.logging.Logger;
 
 /**
  * Operation timed polling algorithm, which uses exponential backoff factor for determining when the
  * next polling operation should be executed. If the polling exceeds the total timeout this
  * algorithm cancels polling.
  */
 public class OperationTimedPollAlgorithm extends ExponentialRetryAlgorithm {
+
+  public static Logger LOGGER = Logger.getLogger(OperationTimedPollAlgorithm.class.getName());
```
JoeWang1127, line 50: Why the logger has to be `public`?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITLongRunningOperation.java
```diff
@@ -149,6 +161,9 @@ public void testGRPC_LROUnsuccessfulResponse_exceedsTotalTimeout_throwsDeadlineE
       assertThrows(CancellationException.class, operationFuture::get);
       int attemptCount = operationFuture.getPollingFuture().getAttemptSettings().getAttemptCount();
       assertThat(attemptCount).isGreaterThan(1);
+      assertEquals(1, logHandler.getAllMessages().size());
+      assertThat(logHandler.getAllMessages().get(0))
```
JoeWang1127, line 165: Is there a guarantee that the message is the first element of the list?

gax-java/gax/src/test/java/com/google/api/gax/util/FakeLogHandler.java
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright 2017 Google LLC
```
JoeWang1127, line 2: Please change year to 2023.

gax-java/gax/src/main/java/com/google/api/gax/longrunning/OperationTimedPollAlgorithm.java
diegomarquezp: This is so it can be accessed by the test in order to add a log handler. @lqiu96, maybe we can fall back to `@VisibleForTesting` because it is... visible only for testing?

gax-java/gax/src/test/java/com/google/api/gax/util/FakeLogHandler.java
diegomarquezp: Changed

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITLongRunningOperation.java
diegomarquezp: So far the implementation has only one case where the logger is used. I think the current assertions are a safe heuristic. Maybe I can assert that at least one message contains the entry I'm looking for?

gax-java/gax/src/main/java/com/google/api/gax/longrunning/OperationTimedPollAlgorithm.java
lqiu96: Ah, right. Let's try the the default package scope, which I believe should still be accessible. If not, then we can add the annotation.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITLongRunningOperation.java
JoeWang1127: > Maybe I can assert that at least one message contains the entry I'm looking for?

I think this maybe better.

gax-java/gax/src/main/java/com/google/api/gax/longrunning/OperationTimedPollAlgorithm.java
JoeWang1127: There're some suggestions made by Quality Gate: https://sonarcloud.io/project/issues?id=googleapis_gapic-generator-java&pullRequest=2236&resolved=false&types=CODE_SMELL

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITLongRunningOperation.java
```diff
@@ -39,6 +43,13 @@
  */
 public class ITLongRunningOperation {
```
burkedavison, line 44: Are we sure that this already-complex integration test the best place to add these new assertions? Rephrased: Is the showcase server needed to verify the client's logging behavior?

The changes we're making are in `com.google.api.gax.longrunning.OperationTimedPollAlgorithm`, but we don't yet have any unit tests for this class. Is it possible for us to write a direct unit test that focuses purely on the logging behavior of `OperationTimedPollAlgorithm.shouldRetry()`?

Without looking at this PR and only the logic we're trying to test, I would have expected to see a set of unit tests that were roughly:
```
// WARNING: Complete pseudocode below...
public void testTroubleshootingLinkProvided() {
  LOGGER.addHandler(handler);
  assertThrows(..., algorithm.shouldRetry(NEXT_ATTEMPT_SHOULD_NOT_RETRY));
  assertThat(handler.getLogMessage().exists())
  assertThat(handler.getLogMessage().contains(LRO_TROUBLESHOOTING_LINK));
}
public void testTroubleshootingLinkNotProvided() {
  LOGGER.addHandler(handler);
  assertThat(algorithm.shouldRetry(NEXT_ATTEMPT_SHOULD_RETRY), is(true));
  assertThat(handler.getLogMessage().doesNotExist())
}
```

Then the integration test can be kept clear of client-side-only implementation details, allowing the person debugging a problem to focus only on the relevant interactions.

diegomarquezp: @burkedavison thank you. At some point writing the test in this file helped me find that the exception thrown in `OperationTimedPollAlgorithm` does not bubble up to the client (hence the use of a logger), but maybe that's just part of finding out about the behavior of LROs. I will move the test case to a unit test in gax

gax-java/gax/src/test/java/com/google/api/gax/longrunning/OperationTimedPollAlgorithmTest.java
```diff
@@ -0,0 +1,122 @@
...
+        OperationTimedPollAlgorithm.create(FAST_RETRY_SETTINGS, clock);
+    TimedAttemptSettings settings =
+        TimedAttemptSettings.newBuilder()
+            .setGlobalSettings(FAST_RETRY_SETTINGS)
+            .setRetryDelay(Duration.ofMillis(1l))
+            .setRpcTimeout(Duration.ofMillis(1l))
+            .setRandomizedRetryDelay(Duration.ofMillis(1l))
+            .setAttemptCount(0)
+            .setFirstAttemptStartTimeNanos(clock.nanoTime())
+            .build();
```
burkedavison, line 114: Thank you for the improvement by moving this to a unit test!

At this point, I'm just nit-picking: you could choose to extract all three of these variables to class fields, and further reduce both of these tests to their essential checks.

gax-java/gax/src/test/java/com/google/api/gax/longrunning/OperationTimedPollAlgorithmTest.java
```diff
@@ -0,0 +1,122 @@
...
+    clock.incrementNanoTime(1 * 1000 * 1000 * 1000);
```
burkedavison, line 115: Consider a quick comment here to clarify what part of the logic this is triggering to force a no-retry.
For example (if this is accurate):
```java
clock.incrementNanoTime(1 * 1000 * 1000 * 1000); // Force Rpc Timeout
```

burkedavison (approved):

gax-java/gax/src/test/java/com/google/api/gax/longrunning/OperationTimedPollAlgorithmTest.java
diegomarquezp: good point. I added this comment

gax-java/gax/src/test/java/com/google/api/gax/longrunning/OperationTimedPollAlgorithmTest.java
diegomarquezp: added

gax-java/gax/src/test/java/com/google/api/gax/longrunning/OperationTimedPollAlgorithmTest.java
```diff
@@ -0,0 +1,114 @@
...
+          .setMaxRetryDelay(Duration.ofMillis(1L))
+          .setInitialRpcTimeout(Duration.ofMillis(1L))
+          .setMaxAttempts(0)
+          .setJittered(false)
+          .setRpcTimeoutMultiplier(1)
+          .setMaxRpcTimeout(Duration.ofMillis(1L))
+          .setTotalTimeout(Duration.ofMillis(5L))
+          .build();
+  private static TimedAttemptSettings timedAttemptSettings;
+  private static FakeApiClock clock;
```
burkedavison, line 61: By making these static and initializing them in an instance method (`@Before`), the test now becomes flaky if multithreading is ever enabled. One test may overwrite the reference during a different test.

```suggestion
  private TimedAttemptSettings timedAttemptSettings;
  private FakeApiClock clock;
```

Issue #2122, opened 2023-10-14
Support a AST Node to auto-generate an interface


arpan14:
I was exploring to use the java auto-generator for generating some of Cloud Spanner Java Client Library. For the same I was exploring how to auto-gen an interface. While I found a few classes to generate a `Class` (`ClassDefinition`) and `Method` (`MethodDefinition`) but could not see an AST node for interface.

Can we support this in the framework?

suztomo:
Would you add the URL of your requirements doc as a "go" link in this ticket too?

suztomo:
Got the document.

PR #2249, opened 2023-11-17
chore: auto-generate Spanner RPC Options

arpan14:


gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/spanner/SpannerOptionsStubClassComposer.java
```diff
@@ -0,0 +1,376 @@
...
+    List<MethodDefinition> setterMethods = new ArrayList<>();
+    List<MethodDefinition> hasMethods = new ArrayList<>();
+    List<MethodDefinition> getterMethods = new ArrayList<>();
+    List<MethodDefinition> allMethods = new LinkedList<>();
+
+    for(Map.Entry<Field, String> field: fields.entrySet()) {
+
+      // generate public method to set option value
+      /**
+       * public static Options.TransactionOption maxBatchingDelayMs(int maxBatchingDelayMs) {
```
arpan14, line 96: @blakeli0 There is an error over here due to a mismatch between the desired return type and the return expression. Refer - https://github.com/googleapis/sdk-platform-java/issues/2122

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/spanner/SpannerOptionsStubClassComposer.java
```diff
@@ -0,0 +1,376 @@
...
+            .setMethods(Arrays.asList(constructor, appendToOptions))
+            .build();
+    return classDefinition;
+  }
+
+  private static TypeNode getOptionsGroupTypeNode(final String value) {
+    // TODO add other mapping groups to the below switch case
+    switch (value) {
+      case "TransactionOption" :
+        return TypeNode.withReference(TRANSACTION_OPTION);
```
arpan14, line 312: @blakeli0 I had a doubt on this. This method returns a `TypeNode` that is used to specify the interfaces to implement. For ex - `implements Options.TransactionOption` . Now `TransactionOption` is supposed to be an inner interface with the `Options` class. How can we do such chaining of classes while creating a type node?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/spanner/SpannerOptionsConfigParser.java
blakeli0, line 1: I would put this file into `parse` package, instead of `composer`. Otherwise looks good.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/spanner/SpannerOptionsStubClassComposer.java
```diff
@@ -0,0 +1,394 @@
...
+/**
+ * Generate an Option class which represents an Option parameter.
+ */
+public class SpannerOptionsStubClassComposer implements ClassComposer {
+  private static final SpannerOptionsStubClassComposer INSTANCE =
+      new SpannerOptionsStubClassComposer();
+  private static final TypeStore FIXED_TYPESTORE = createStaticTypes();
+
+  private static final VaporReference READ_AND_QUERY_OPTION =
+      createVaporReference("com.google.cloud.spanner", "ReadAndQueryOption", "Options");
```
blakeli0, line 63: Using concrete reference is usually preferred, because it reduces the chance of generating non-compilable code. It may not be possible in this case since it's a Spanner class, but once the code is migrated to Spanner repo, we should have reference to the actual classes.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/spanner/SpannerOptionsStubClassComposer.java
```diff
@@ -0,0 +1,394 @@
...
+      default:
+        throw new RuntimeException(String.format("Invalid value => %s provided.", value));
+    }
+  }
+
+  private static List<AnnotationNode> createClassAnnotations() {
+    return Arrays.asList(
+        AnnotationNode.builder()
+            .setType(FIXED_TYPESTORE.get("Generated"))
+            .setDescription("by gapic-generator-java")
```
blakeli0, line 337: This probably can be something like `spanner-generator`

PR #2272, opened 2023-11-29
build: do not declare javadoc plugin version

alicejli:
The maven-javadoc-plugin version is defined in the shared config pom.xml.
https://github.com/googleapis/java-shared-config/blob/778a547a09de71dbf9e5a42b155f12d15c319864/pom.xml#L472


alicejli:
Pretty sure the `showcase-clirr` check failure is unrelated:
```
Error:  PROTOC FAILED: gapic_metadata.proto:1:1: Invalid control characters encountered in text.
```
as it is mentioned in: https://github.com/googleapis/sdk-platform-java/pull/2270

burkedavison (approved):

PR #2275, opened 2023-11-29
refactor: Add EndpointContext

lqiu96:

## Description
EndpointContext is a new class created to resolve the correct endpoint. Client library users are able customize the endpoint in various ways and this class is the source of truth to resolve all the customizations.

Currently supports:
- ClientSettings/ StubSettings endpoint value
- mTLS endpoint value
- switchToMtls flag

The logic to resolve the endpoint was moved from the ClientContext into the EndpointContext. In this PR, no additional logic was added.

### DIFF
--- a/gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
+++ b/gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
@@ -206,7 +206,7 @@ public abstract class ClientContext {
             .setMtlsEndpoint(settings.getMtlsEndpoint())
             .setSwitchToMtlsEndpointAllowed(settings.getSwitchToMtlsEndpointAllowed())
             .build();
-    String endpoint = endpointContext.resolveEndpoint();
+    String endpoint = endpointContext.getResolvedEndpoint();
     if (transportChannelProvider.needsEndpoint()) {
       transportChannelProvider = transportChannelProvider.withEndpoint(endpoint);
     }
--- a/gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
+++ b/gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
@@ -224,12 +224,8 @@ public abstract class ClientContext {
     if (transportChannelProvider.needsCredentials() && credentials != null) {
       transportChannelProvider = transportChannelProvider.withCredentials(credentials);
     }
-    String endpoint =
-        getEndpoint(
-            settings.getEndpoint(),
-            settings.getMtlsEndpoint(),
-            settings.getSwitchToMtlsEndpointAllowed(),
-            new MtlsProvider());
+    EndpointContext endpointContext = settings.getEndpointContext();
+    String endpoint = endpointContext.resolveEndpoint();
     if (transportChannelProvider.needsEndpoint()) {
       transportChannelProvider = transportChannelProvider.withEndpoint(endpoint);
     }
--- a/gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
+++ b/gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
@@ -200,7 +200,12 @@ public abstract class ClientContext {
     if (transportChannelProvider.needsCredentials() && credentials != null) {
       transportChannelProvider = transportChannelProvider.withCredentials(credentials);
     }
-    EndpointContext endpointContext = settings.getEndpointContext();
+    EndpointContext endpointContext =
+        EndpointContext.newBuilder()
+            .setClientSettingsEndpoint(settings.getEndpoint())
+            .setMtlsEndpoint(settings.getMtlsEndpoint())
+            .setSwitchToMtlsEndpointAllowed(settings.getSwitchToMtlsEndpointAllowed())
+            .build();
     String endpoint = endpointContext.resolveEndpoint();
     if (transportChannelProvider.needsEndpoint()) {
       transportChannelProvider = transportChannelProvider.withEndpoint(endpoint);
--- a/gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
+++ b/gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
@@ -37,7 +37,6 @@ import com.google.api.gax.core.BackgroundResource;
 import com.google.api.gax.core.ExecutorAsBackgroundResource;
 import com.google.api.gax.core.ExecutorProvider;
 import com.google.api.gax.rpc.internal.QuotaProjectIdHidingCredentials;
-import com.google.api.gax.rpc.mtls.MtlsProvider;
 import com.google.api.gax.tracing.ApiTracerFactory;
 import com.google.api.gax.tracing.BaseApiTracerFactory;
 import com.google.auth.Credentials;
@@ -141,29 +140,6 @@ public abstract class ClientContext {
    */
   public static ClientContext create(ClientSettings settings) throws IOException {
     return create(settings.getStubSettings());
-  }
-
-  /** Returns the endpoint that should be used. See https://google.aip.dev/auth/4114. */
-  static String getEndpoint(
-      String endpoint,
-      String mtlsEndpoint,
-      boolean switchToMtlsEndpointAllowed,
-      MtlsProvider mtlsProvider)
-      throws IOException {
-    if (switchToMtlsEndpointAllowed) {
-      switch (mtlsProvider.getMtlsEndpointUsagePolicy()) {
-        case ALWAYS:
-          return mtlsEndpoint;
-        case NEVER:
-          return endpoint;
-        default:
-          if (mtlsProvider.useMtlsClientCertificate() && mtlsProvider.getKeyStore() != null) {
-            return mtlsEndpoint;
-          }
-          return endpoint;
-      }
-    }
-    return endpoint;
   }
 
   /**
--- a/gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
+++ b/gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
@@ -1,0 +1,128 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above
+ * copyright notice, this list of conditions and the following disclaimer
+ * in the documentation and/or other materials provided with the
+ * distribution.
+ *     * Neither the name of Google LLC nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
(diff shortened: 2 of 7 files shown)


blakeli0:
Can we integrate EndpointContext into the existing code and make sure all functionalities are not affected? It may not be a good practice to have a `chore` PR introduces new classes without any reference to it. Because it does not provide  additional functionality, and it's not easy for reviewers to understand the purpose of the new class. It would be better if we can make this PR a proper `refactor`.

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -0,0 +1,129 @@
...
+  public abstract static class Builder {
+    public abstract Builder setClientSettingsEndpoint(String clientSettingsEndpoint);
+
+    public abstract Builder setTransportChannelEndpoint(String transportChannelEndpoint);
+
+    public abstract Builder setMtlsEndpoint(String mtlsEndpoint);
+
+    public abstract Builder setSwitchToMtlsEndpointAllowed(boolean switchToMtlsEndpointAllowed);
+
+    @VisibleForTesting
```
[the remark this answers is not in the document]
lqiu96: Yep. Removed.

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -0,0 +1,126 @@
...
+
+/** Contains the fields required to resolve the endpoint */
+@InternalApi
+@AutoValue
+public abstract class EndpointContext {
+  @Nullable
+  public abstract String clientSettingsEndpoint();
+
+  @Nullable
+  public abstract String transportChannelEndpoint();
```
blakeli0, line 47: Will we ever need to pass in `transportChannelEndpoint`? 
I see that there is no getters for the endpoint set to transport channel, maybe we want to expose one? So that we can set all the info to `EndpointContext`?
It's also OK to leave `transportChannelEndpoint` out of `EndpointContext` for now, if that's the case, then it's better to remove this field at this moment.

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -0,0 +1,126 @@
...
+          if (mtlsProvider.useMtlsClientCertificate() && mtlsProvider.getKeyStore() != null) {
+            return mtlsEndpoint;
+          }
+          return endpoint;
+      }
+    }
+    return endpoint;
+  }
+
+  public String resolveEndpoint() throws IOException {
```
blakeli0, line 101: Can we try to resolve the endpoint in the `build()` method and just expose a getter for the resolved endpoint?

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -0,0 +1,126 @@
...
+import com.google.common.annotations.VisibleForTesting;
+import java.io.IOException;
+import javax.annotation.Nullable;
+
+/** Contains the fields required to resolve the endpoint */
+@InternalApi
+@AutoValue
+public abstract class EndpointContext {
+  @Nullable
+  public abstract String clientSettingsEndpoint();
```
blakeli0, line 44: I don't think this can be null as it is coming from [StubSettings.endpoint](https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/gax/src/main/java/com/google/api/gax/rpc/StubSettings.java#L73)?

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -0,0 +1,126 @@
...
+
+  public static Builder newBuilder() {
+    return new AutoValue_EndpointContext.Builder().setSwitchToMtlsEndpointAllowed(false);
+  }
+
+  @VisibleForTesting
+  void determineEndpoint() throws IOException {
+    MtlsProvider mtlsProvider = mtlsProvider() == null ? new MtlsProvider() : mtlsProvider();
+    String customEndpoint =
+        transportChannelEndpoint() != null ? transportChannelEndpoint() : clientSettingsEndpoint();
```
blakeli0, line 69: I don't see `transportChannelEndpoint` being set in the ClientContext. Was it intentional?

gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
```diff
@@ -224,12 +200,13 @@ public static ClientContext create(StubSettings settings) throws IOException {
...
-            settings.getMtlsEndpoint(),
-            settings.getSwitchToMtlsEndpointAllowed(),
-            new MtlsProvider());
+    EndpointContext endpointContext =
+        EndpointContext.newBuilder()
+            .setClientSettingsEndpoint(settings.getEndpoint())
+            .setMtlsEndpoint(settings.getMtlsEndpoint())
+            .setSwitchToMtlsEndpointAllowed(settings.getSwitchToMtlsEndpointAllowed())
+            .build();
+    String endpoint = endpointContext.resolveEndpoint();
```
blakeli0, line 209: It maybe outside the scope of this PR, do we plan to pass the whole `EndpointContext` to `ApiCallContext` so that we can do the validation later?

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
lqiu96: I believe for most non-modified GAPICs, this is the case. It should be set via the {Service}StubSettings like: https://github.com/googleapis/google-cloud-java/blob/6227504db7c511b379cad2d67960d63f36eed39b/java-grafeas/src/main/java/io/grafeas/v1/stub/GrafeasStubSettings.java#L635 with a non-null value (as the generator will fail if it can't find it).

I'm accounting for the Grafeas edge case: https://github.com/googleapis/google-cloud-java/blob/6227504db7c511b379cad2d67960d63f36eed39b/java-grafeas/src/main/java/io/grafeas/v1/stub/GrafeasStubSettings.java#L403-L405 which has post-processed changes.

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
lqiu96: > Will we ever need to pass in transportChannelEndpoint?

I believe so to keep with the existing behavior where transportChannelEndpoint overrides the clientSettingsEndpoint if set. Right now, the ClientContext doesn't set it inside the EndpointContext but the code to use the endpoint exists here: https://github.com/googleapis/sdk-platform-java/blob/da6607b17130ab045640618d505fda915ddb8e49/gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java#L233-L235

The reason I haven't added it in yet is because I _most likely_ will need to expose an additional `getEndpoint()` method to the TransportChannelProvider: https://github.com/googleapis/sdk-platform-java/blob/da6607b17130ab045640618d505fda915ddb8e49/gax-java/gax/src/main/java/com/google/api/gax/rpc/TransportChannelProvider.java#L144 (Note: grpc and httpjson's transportchannel already have a `getEndpoint()` method). I was planning on doing this on the next PR as this one only introduces the EndpointContext.

I see your point about including it then, so I'll drop the transportChannelEndpoint values from this PR and add it in the next PR.

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
lqiu96: I believe this should be answered above. I'll remove this logic from this PR and add it in the next.

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
lqiu96: I'm using Autovalue's Builder which I believe generates the `build()` implementation for us. I could create my own Builder class if that's preferable.

gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
lqiu96: Yes, it'll be in a future PR.

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
blakeli0: I see, this is good for now then. Let's keep this in mind and come back for it later if we have time. I'm not sure why we removed the default endpoint for grafeas, and how does grafeas create channels without a default endpoint. But ideally this should not be nullable, otherwise gRPC channel creation could fail.

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
blakeli0: SGTM

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
blakeli0: Yeah, it's a common practice to validate or normalize a field value in the `build()` method, see internal docs [here](go/autovalue/builders-howto#normalize). We have [examples](https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPoolSettings.java#L163-L181) in our repo as well.

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -0,0 +1,122 @@
...
+
+    abstract EndpointContext autoBuild();
+
+    public EndpointContext build() {
+      try {
+        EndpointContext endpointContext = autoBuild();
+        endpointContext.determineEndpoint();
+        return endpointContext;
+      } catch (IOException e) {
+        throw new RuntimeException("Unable to determine the endpoint: " + e.getMessage());
```
lqiu96, line 118: throw IOExcpetion

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -0,0 +1,122 @@
...
+    return endpoint;
+  }
+
+  public String getResolvedEndpoint() {
+    return resolvedEndpoint;
+  }
+
+  @AutoValue.Builder
+  public abstract static class Builder {
+    public abstract Builder setClientSettingsEndpoint(String clientSettingsEndpoint);
```
lqiu96, line 102: Add Javadoc for this field (in builder and setter)

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -0,0 +1,122 @@
...
+          }
+          return endpoint;
+      }
+    }
+    return endpoint;
+  }
+
+  public String getResolvedEndpoint() {
+    return resolvedEndpoint;
+  }
```
lqiu96, line 98: Add Javadoc here

blakeli0 (approved):
LGTM. Before merging, can you please add more info to the description to explain the purpose of `EndpointContext`? 
I would reword the title as well, I think this is more like a `refactor` instead of `feat`. Because we added a new class to encapsulate existing features but didn't introduce any new features.

lqiu96:
> LGTM. Before merging, can you please add more info to the description to explain the purpose of `EndpointContext`? I would reword the title as well, I think this is more like a `refactor` instead of `feat`. Because we added a new class to encapsulate existing features but didn't introduce any new features.

Added.

PR #2278, opened 2023-11-30
feat: Structs mapper utility

kolea2:
Originally implemented in https://github.com/googleapis/java-logging/blob/d3f897c26b5b55436d2212a0d7452940f15c413f/google-cloud-logging/src/main/java/com/google/cloud/logging/Structs.java, this is a common utility that other projects can benefit from.

Test is from: https://github.com/googleapis/java-logging/blob/d3f897c26b5b55436d2212a0d7452940f15c413f/google-cloud-logging/src/test/java/com/google/cloud/logging/StructsTest.java

The only difference between the java-logging implementation and this is that builder and class methods in Structs.java have been made public.

### DIFF
--- a/java-core/google-cloud-core/src/main/java/com/google/cloud/Structs.java
+++ b/java-core/google-cloud-core/src/main/java/com/google/cloud/Structs.java
@@ -1,0 +1,149 @@
+/*
+ * Copyright 2016 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud;
+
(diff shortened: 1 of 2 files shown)


blakeli0:
@kolea2 Can you please create an issue for tracking purpose?

java-core/google-cloud-core/src/main/java/com/google/cloud/Structs.java
```diff
@@ -0,0 +1,149 @@
...
+import java.util.AbstractSet;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * This class contains static utility methods that operate on or return protobuf's {@code Struct}
+ * objects.
+ */
+public final class Structs {
```
blakeli0, line 39: We can add an `@InternalApi` to prevent this class being used by customers directly.

java-core/google-cloud-core/src/main/java/com/google/cloud/Structs.java
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    @Override
+    public Set<Entry<String, Object>> entrySet() {
+      return entrySet;
+    }
+  }
+
+  /** Returns an unmodifiable map view of the {@link Struct} parameter. */
+  @InternalApi
```
blakeli0, line 89: Adding `@InternalApi` to the class already implies that all methods in the class are internal APIs, we don't have to add them again to the public methods. Same thing for the `newStruct` method below.

kolea2: fixed, thanks!

blakeli0 (approved):

PR #2114, opened 2023-10-12
feat: update javadocs for Client classes to include table of methods

alicejli:
- This updates the javadoc for generated `Client` classes to include a table of methods and method variants
- This replaces the generic "API Surfaces" comments
- As it is currently implemented, this would affect all generated `Client` classes (e.g. handwritten libraries would not be able to avoid this change)

Refer to cl/572240700 for staging links.

Fixes #2069  ☕️

PR #2280, opened 2023-12-01
fix: format proto comments in Client Overview

alicejli:
Follow up PR to https://github.com/googleapis/sdk-platform-java/pull/2114 which breaks generation for java-bigqueryreservation due to errant formatting of the proto comments.

cl/587006892 contains the mock for bigqueryreservation's ReservationServiceClent.

I think it could make sense to add an integration test for this as the original integration tests didn't catch this bug, but I also understand that there are already a lot of integration tests. Thoughts?

### DIFF
--- a/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/comment/ServiceClientCommentComposer.java
+++ b/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/comment/ServiceClientCommentComposer.java
@@ -242,7 +242,7 @@ public class ServiceClientCommentComposer {
           .append(method.method)
           .append("</td>\n")
           .append("     <td>")
-          .append("<p>" + method.description + "</p>")
+          .append(CommentFormatter.formatAsJavaDocComment(method.description, null))
           .append("</td>\n")
           .append("     <td>\n");
       generateUnorderedListMethodVariants(
--- a/gapic-generator-java/src/test/java/com/google/api/generator/engine/writer/JavaWriterVisitorTest.java
+++ b/gapic-generator-java/src/test/java/com/google/api/generator/engine/writer/JavaWriterVisitorTest.java
@@ -71,6 +71,7 @@ import com.google.api.generator.engine.ast.VaporReference;
 import com.google.api.generator.engine.ast.Variable;
 import com.google.api.generator.engine.ast.VariableExpr;
 import com.google.api.generator.engine.ast.WhileStatement;
+import com.google.api.generator.engine.writer.JavaFormatter.FormatException;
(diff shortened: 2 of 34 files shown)


gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/comment/ServiceClientCommentComposer.java
```diff
@@ -242,7 +242,7 @@ private static String createTableOfMethods(List<MethodAndVariants> methodAndVari
           .append(method.method)
           .append("</td>\n")
           .append("     <td>")
-          .append("<p>" + method.description + "</p>")
+          .append(CommentFormatter.formatAsJavaDocComment(method.description, null))
```
blakeli0, line 245: Can we make the whole table a String and format with `CommentFormatter`? This fix should be good for now, but in the future, someone else may add another line with dynamic content that could result in the same bug.

blakeli0:
> I think it could make sense to add an integration test for this as the original integration tests didn't catch this bug, but I also understand that there are already a lot of integration tests. Thoughts?

Can we mock this scenario with showcase? By adding the problematic comment to [one of the showcase protos](https://github.com/googleapis/gapic-showcase/tree/3188c8a6089597d937ed6ab5b529716ab92d5ff3/schema/google/showcase/v1beta1).

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/comment/ServiceClientCommentComposer.java
alicejli: The challenge with formatting the whole table with `CommentFormatter` is that it replaces *all* of the html elements for the table with the HTML-escaped character sequence for them, so the output doesn't stay formatted as a table. You can see the example output here: cl/572240700

burkedavison (approved):

alicejli:
> > I think it could make sense to add an integration test for this as the original integration tests didn't catch this bug, but I also understand that there are already a lot of integration tests. Thoughts?
> 
> Can we mock this scenario with showcase? By adding the problematic comment to [one of the showcase protos](https://github.com/googleapis/gapic-showcase/tree/3188c8a6089597d937ed6ab5b529716ab92d5ff3/schema/google/showcase/v1beta1).

Good idea - PR opened: https://github.com/googleapis/gapic-showcase/pull/1391

alicejli:
@blakeli0 I added a failing test for JavaWriter per our offline discussion. Let me know any additional thoughts, thanks!

blakeli0 (approved):

PR #2285, opened 2023-12-04
test: BatcherImplTest.testThrottlingBlocking to check actual wait time

### LINKED ISSUE #1931: gax-java(mac-os): BatcherImplTest.testThrottlingBlocking- adding elements to batcher should be blocked by FlowControlled
The build has failed : https://github.com/googleapis/gapic-generator-java/actions/runs/5897706299

suztomo:
Fixes #1931 . BatcherImplTest.testThrottlingBlocking to check actual wait time, because I observe that `Future.wait` does not always throw TimeoutException in a timely manner. With the additional logging, we observed the following failure:

```
[INFO] Running com.google.api.gax.batching.BatcherImplTest
2023 1:24:02.786 com.google.api.gax.batching.BatcherImplTest testThrottlingBlocking INFO: Calling future.get(10 ms)
2023 1:24:02.841 com.google.api.gax.batching.BatcherImplTest lambda$testThrottlingBlocking$0 INFO: Calling flowController.release
2023 1:24:02.841 com.google.api.gax.batching.BatcherImplTest lambda$testThrottlingBlocking$0 INFO: Called flowController.release
2023 1:24:02.843 com.google.api.gax.batching.BatcherImplTest testThrottlingBlocking INFO: future.get(10 ms) unexpectedly returned.
...
[ERROR] Failures: 
[ERROR]   BatcherImplTest.testThrottlingBlocking:905 adding elements to batcher should be blocked by FlowControlled
```

(See the `future.get(10 ms)` returned after 57 ms.)

It's probably because the main thread does not always get the CPU resource. This pull request fixes the flaky test failure by checking the actual wait time in the main thread. If it's taking more than 10 milliseconds, then it means the method `future.get(10 ms)` does not return within 10 milliseconds (Good).

### DIFF
--- a/gax-java/gax/pom.xml
+++ b/gax-java/gax/pom.xml
@@ -93,6 +93,14 @@
           </excludes>
         </configuration>
       </plugin>
+      <plugin>
+        <!-- Troubleshooting a flaky test in https://github.com/googleapis/sdk-platform-java/issues/1931 -->
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <argLine>-Djava.util.logging.SimpleFormatter.format="%1$tY %1$tl:%1$tM:%1$tS.%1$tL %2$s %4$s: %5$s%6$s%n"</argLine>
+        </configuration>
+      </plugin>
     </plugins>
   </build>
 </project>
--- a/gax-java/gax/src/test/java/com/google/api/gax/batching/BatcherImplTest.java
+++ b/gax-java/gax/src/test/java/com/google/api/gax/batching/BatcherImplTest.java
@@ -85,6 +85,8 @@ import org.threeten.bp.Duration;
 @RunWith(JUnit4.class)
 public class BatcherImplTest {
 
+  private static final Logger logger = Logger.getLogger(BatcherImplTest.class.getName());
+
   private static final ScheduledExecutorService EXECUTOR =
       Executors.newSingleThreadScheduledExecutor();
 
@@ -889,15 +891,20 @@ public class BatcherImplTest {
           () -> {
             try {
               Thread.sleep(throttledTime);
+              logger.info("Calling flowController.release");
               flowController.release(1, 1);
+              logger.info("Called flowController.release");
             } catch (InterruptedException e) {
             }
           });
 
       try {
+        logger.info("Calling future.get(10 ms)");
         future.get(10, TimeUnit.MILLISECONDS);
+        logger.info("future.get(10 ms) unexpectedly returned.");
         assertWithMessage("adding elements to batcher should be blocked by FlowControlled").fail();
       } catch (TimeoutException e) {
+        logger.info("future.get(10 ms) timed out expectedly.");
         // expected
       }
 
--- a/gax-java/gax/src/test/java/com/google/api/gax/batching/BatcherImplTest.java
+++ b/gax-java/gax/src/test/java/com/google/api/gax/batching/BatcherImplTest.java
@@ -873,6 +873,7 @@ public class BatcherImplTest {
                 public void run() {
                   batcherAddThreadHolder.add(Thread.currentThread());
                   batcher.add(1);
+                  logger.info("Called batcher.add(1)");
                 }
               });
 
@@ -887,6 +888,7 @@ public class BatcherImplTest {
       } while (batcherAddThreadHolder.isEmpty()
           || batcherAddThreadHolder.get(0).getState() != Thread.State.WAITING);
 
+      long beforeGetCall = System.currentTimeMillis();
       executor.submit(
           () -> {
             try {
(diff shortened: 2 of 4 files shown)


gax-java/gax/pom.xml
```diff
@@ -93,6 +93,14 @@
           </excludes>
         </configuration>
       </plugin>
+      <plugin>
+        <!-- Troubleshooting a flaky test in https://github.com/googleapis/sdk-platform-java/issues/1931 -->
```
suztomo, line 97: I keep this just in case the problem occurs and we have the logging. This surefire (unit test) configuration does not affect the released libraries to users.

gax-java/gax/src/test/java/com/google/api/gax/batching/BatcherImplTest.java
```diff
@@ -885,20 +888,35 @@ public void run() {
       } while (batcherAddThreadHolder.isEmpty()
           || batcherAddThreadHolder.get(0).getState() != Thread.State.WAITING);
 
+      long beforeGetCall = System.currentTimeMillis();
       executor.submit(
           () -> {
             try {
               Thread.sleep(throttledTime);
+              logger.info("Calling flowController.release");
```
blakeli0, line 896: Can we lower the logging level so that the logs are not generated by default?

suztomo: I want these logs in case the problem recurs.

suztomo: I will remove these logs in February, if the flaky test does not recur.

gax-java/gax/src/test/java/com/google/api/gax/batching/BatcherImplTest.java
```diff
@@ -885,20 +888,35 @@ public void run() {
...
               flowController.release(1, 1);
+              logger.info("Called flowController.release");
             } catch (InterruptedException e) {
             }
           });
 
       try {
+        logger.info("Calling future.get(10 ms)");
```
lqiu96, line 904: I'm not too familiar with Batcher or this test, but what is the significance of 10ms here? If a TimeoutException is expected, why can't we lower the future.get() block time?

gax-java/gax/src/test/java/com/google/api/gax/batching/BatcherImplTest.java
```diff
@@ -885,20 +888,35 @@ public void run() {
...
         future.get(10, TimeUnit.MILLISECONDS);
-        assertWithMessage("adding elements to batcher should be blocked by FlowControlled").fail();
+        long afterGetCall = System.currentTimeMillis();
+        long actualWaitTimeMs = afterGetCall - beforeGetCall;
+
+        // In a flaky test troubleshooting, we observed that "future.get" method might not throw
+        // TimeoutException in a timely manner. It's because the main thread is not always having
+        // CPU resource. As long as the "future.get" does not return within the timeout, this test
```
lqiu96, line 911: qq: What do you mean by "main thread is not always having CPU resource"?

gax-java/gax/src/test/java/com/google/api/gax/batching/BatcherImplTest.java
blakeli0: SGTM.

gax-java/gax/src/test/java/com/google/api/gax/batching/BatcherImplTest.java
suztomo: From the log (in the pull request description), the problem is that the TimeoutException isn't always thrown. Lowering the 10 ms to smaller number won't help.

suztomo: > Is this just potentially a scheduling issue on a machine that has a single CPU/ few CPUs?

Yes.

lqiu96: Ok, from my understanding it would be either case:
- future.get() takes more than 10ms to return -> TimeoutException
- future.get() returns a result within 10ms -> Check that the actual waitTime was at least 10ms

I think that should suffice due to the variance in scheduling. CC: @blakeli0 Do you know of another option to handle this flakiness?

lqiu96 (approved):

gax-java/gax/src/test/java/com/google/api/gax/batching/BatcherImplTest.java
lqiu96: Ah, updated the comment above. The docs for `Future.get()` seem to suggest that it will throw the exception once 10ms passes, but I'm guessing that is just best effort then.

gax-java/gax/src/test/java/com/google/api/gax/batching/BatcherImplTest.java
suztomo: https://github.com/googleapis/sdk-platform-java/pull/2519

PR #2289, opened 2023-12-05
chore: modify instructions on how to update goldens

JoeWang1127:
In this PR:
- Modify instructions on how to update goldens in showcase and integration tests.

### DIFF
--- a/showcase/README.md
+++ b/showcase/README.md
@@ -17,9 +17,10 @@ update to a compatible client version in `./WORKSPACE`.
 
 ```shell
 # Install the showcase server version defined in gapic-showcase/pom.xml
-showcase/ $ go install github.com/googleapis/gapic-showcase/cmd/gapic-showcase@v"$(cd gapic-showcase;mvn help:evaluate -Dexpression=gapic-showcase.version -q -DforceStdout)"
-showcase/ $ PATH=$PATH:`go env GOPATH`/bin
-showcase/ $ gapic-showcase --help
+cd showcase
+go install github.com/googleapis/gapic-showcase/cmd/gapic-showcase@v"$(cd gapic-showcase;mvn help:evaluate -Dexpression=gapic-showcase.version -q -DforceStdout)"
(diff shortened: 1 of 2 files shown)


showcase/README.md
```diff
@@ -78,6 +79,10 @@ Note:
 Open a new terminal window in the root project directory.
 
 ```shell
-$ cd showcase
-$ mvn compile -P update
+# Likely you need to reinstall gapic-generator-java 
+# before updating the showcase.
+# In repository's root directory
+mvn clean install -DskipTests
```
blakeli0, line 85: It should be outside of the scope of this PR, but just want to bring up ideas for possible enhancement. I _think_ if we make showcase a submodule of the repo (compare to a standalone module right now), then we should be able to run showcase tests/update goldens without re-installing the generator, but it may have other implications. @suztomo may have more ideas about it.

test/integration/goldens/README.md
```diff
@@ -22,6 +22,10 @@ update the goldens using source files. Run the command below to overwrite the go
 in `redis` folder.
 
 ```sh
+# Likely you need to reinstall gapic-generator-java 
+# before updating Goldens.
```
lqiu96, line 26: Same as above

showcase/README.md
```diff
@@ -78,6 +79,10 @@ Note:
 Open a new terminal window in the root project directory.
 
 ```shell
-$ cd showcase
-$ mvn compile -P update
+# Likely you need to reinstall gapic-generator-java 
+# before updating the showcase.
```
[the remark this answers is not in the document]
JoeWang1127: Makes sense, I'll remove the comment.

lqiu96 (approved):

showcase/README.md
suztomo: I think that idea will work. We just have to be careful that the showcase artifact won't be released to Maven Central.

PR #2293, opened 2023-12-08
feat: Add Getters/Setters for Universe Domain

lqiu96:
Introduce new Getter/Setter for Universe Domain inside the StubSettings and ClientContext.

TransportChannelProvider adds a new method `getEndpoint()` to expose the custom set endpoint value.

### DIFF
--- a/gax-java/gax/src/main/java/com/google/api/gax/rpc/TransportChannelProvider.java
+++ b/gax-java/gax/src/main/java/com/google/api/gax/rpc/TransportChannelProvider.java
@@ -91,6 +91,15 @@ public interface TransportChannelProvider {
   boolean needsEndpoint();
 
   /**
+   * TransportChannelProvider allows setting a custom endpoint which may not be the resolved
+   * endpoint. Resolving the endpoint is always required for gRPC and HttpJson Transports and is not
+   * dependent if the user set a custom endpoint.
+   */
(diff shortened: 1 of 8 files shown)


gax-java/gax/src/main/java/com/google/api/gax/rpc/TransportChannelProvider.java
```diff
@@ -90,6 +90,15 @@ public interface TransportChannelProvider {
...
 
+  /**
+   * TransportChannelProvider allows setting a custom endpoint which may not be the resolved
+   * endpoint. Resolving the endpoint is _always_ required for gRPC and HttpJson Transports.
+   * Unlike {@link #needsEndpoint()}, it is not dependent if the user set a custom endpoint.
+   */
+  default boolean needsResolvedEndpoint() {
+    return false;
+  }
+
```
lqiu96, line 101: Alternatively, a possible option would be that ClientContext always overrides the endpoint in the TransportChannelProvider (ignoring the `needsEndpoint()` call). This would require changing the behavior of `withEndpoint(...)` for two subclasses: LocalChannelProvider and FixedChannelProvider as they throw an exception on the `withEndpoint(...)` call.

Adding a this new method would prevent requiring changing the behavior of the public classes.

Issue #2296, opened 2023-12-08
TransportChannelProvider's endpoint value should use the resolved endpoint from EndpointContext


lqiu96:
TransportChannelProvider should create a valid TransportChannel using the endpoint result from the EndpointContext. 

Currently, the behavior is:
1. Use the user configured endpoint set from the TransportChannelProvider's `setEndpoint()` method
2. If there is no user configured endpoint in the TransportChannelProvider, use the clientSettings endpoint

The resolved endpoint may not be the the user configured endpoint in the TransportChannelProvider.

A few options to resolve this:
1. Expose a new method (i.e. `needsResolvedEndpoint()`) which is similar to `needsEndpoint()`. This will always be set as true for gRPC and HttpJson and is not determined by checking if the TransportChannelProvider has user configuration. It should be possible to add the `@ObsoleteApi` annotation to `needsEndpoint()`.
2. Always override the TransportChannelProvider's endpoint (regardless of what was set in before). This works because the user set endpoint in TransportChannelProvider is stored before being overwritten. This would require changes behavior changes to `withEndpoint(...)` in FixedTransportProvider and LocalTransportProvider as they currently throw an exception if it is called.

Option 1 would add a new public method to the public interface, but would avoid behavior breaking changes.

PR #2297, opened 2023-12-08
refactor: Use EndpointContext's resolved endpoint for TransportChannelProvider

lqiu96:
Fixes https://github.com/googleapis/sdk-platform-java/issues/2296

Changing from `needsEndpoint()` -> `needsResolvedEndpoint()` would not be a breaking behavior change as the EndpointContext would use the TransportChannelProvider's endpoint over the ClientSettings endpoint if provided.


lqiu96:
Actually, after reviewing this again, I don't think this is needed. I'll close it for now and and revisit again if this is needed.

PR #2300, opened 2023-12-11
feat: Parse Host Service Name

lqiu96:
Changes:
Generator parses the Default Host value to get the Host Service Name. The parsed value is passed to the {Service}StubSettings and then passed to the EndpointContext.

HostServiceName is either set to be package-private scope or marked with `@InternalApi` as it's meant for usage by GAPIC clients.

### DIFF
--- a/gapic-generator-java/src/main/java/com/google/api/generator/gapic/model/Service.java
+++ b/gapic-generator-java/src/main/java/com/google/api/generator/gapic/model/Service.java
@@ -53,7 +53,7 @@ public abstract class Service {
   }
 
   public String hostServiceName() {
-    // Default Host is guaranteed to exist and be non-null and non-empty
+    // Host Service Name is guaranteed to exist and be non-null and non-empty
     // Parser will fail if the default host is not supplied
     return parseHostServiceName(defaultHost());
   }
--- a/gapic-generator-java/src/main/java/com/google/api/generator/gapic/model/Service.java
+++ b/gapic-generator-java/src/main/java/com/google/api/generator/gapic/model/Service.java
@@ -52,6 +52,12 @@ public abstract class Service {
     return !Strings.isNullOrEmpty(description());
   }
 
+  public String hostServiceName() {
+    // Default Host is guaranteed to exist and be non-null and non-empty
+    // Parser will fail if the default host is not supplied
+    return parseHostServiceName(defaultHost());
+  }
+
   public String apiShortName() {
     if (!Strings.isNullOrEmpty(defaultHost())) {
       return parseApiShortName(defaultHost());
@@ -182,6 +188,17 @@ public abstract class Service {
       apiVersion = "";
     }
     return apiVersion;
+  }
+
+  // Parse the service name from the default host configured in the protos
+  // or service yaml file. For Google Cloud Services, the default host value
+  // is expected to contain `.googleapis.com`. Exceptions may exist (i.e. localhost),
+  // in which case we will return an empty string.
+  private static String parseHostServiceName(String defaultHost) {
+    if (defaultHost.contains(".googleapis.com")) {
+      return Iterables.getFirst(Splitter.on(".").split(defaultHost), defaultHost);
+    }
+    return "";
   }
 
   // Parse defaultHost for apiShortName for the RegionTag. Need to account for regional default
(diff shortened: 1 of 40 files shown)


gax-java/gax/src/main/java/com/google/api/gax/rpc/StubSettings.java
```diff
@@ -414,6 +426,14 @@ public B setClock(ApiClock clock) {
       return self();
     }
 
+    // This meant to be used internally by GAPIC clients to configure an endpoint.
+    // Users should avoid settings this as it could change their endpoint.
+    @InternalApi
+    public B setHostServiceName(String hostServiceName) {
```
blakeli0, line 432: Do we have to expose a setter for it? Since the host service name is a fixed String, I think we should implement it in a similar way as the to-be-deprecated default endpoint, that we generate a getter that returns the default host service name.

gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
```diff
@@ -100,6 +100,10 @@ public abstract class ClientContext {
   @Nonnull
   public abstract Duration getStreamWatchdogCheckInterval();
 
+  // Package-Private scope for internal use only. Shared between StubSettings and ClientContext
+  @Nullable
+  abstract String getHostServiceName();
```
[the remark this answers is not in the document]
lqiu96: Sure, I'll rename to `serviceName`.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/model/Service.java
```diff
@@ -52,6 +52,12 @@ public boolean hasDescription() {
     return !Strings.isNullOrEmpty(description());
   }
 
+  public String hostServiceName() {
+    // Default Host is guaranteed to exist and be non-null and non-empty
```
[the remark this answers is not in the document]
lqiu96: Ah, comment is misleading. I meant to write that the `hostServiceName is guaranteed` as the generator won't generate a library if it doesn't exist.

> I remember we discussed that the service name could come from the proto or service yaml, is that still true?

It is still true. Checks the proto's default_host value first, then the service yaml file, and fails otherwise

gax-java/gax/src/main/java/com/google/api/gax/rpc/StubSettings.java
lqiu96: I see, I think that makes sense. I'll update it.

gax-java/gax/src/main/java/com/google/api/gax/rpc/StubSettings.java
```diff
@@ -137,6 +140,13 @@ public final ApiClock getClock() {
     return clock;
   }
 
+  // Intended for Internal Use and Overriden by generated ServiceStubSettings classes.
+  // Meant to be shared between StubSettings and ClientContext.
+  @InternalApi
+  public String getServiceName() {
+    return "";
+  }
```
lqiu96, line 148: `ClientContext.create()` takes in the root StubSettings (not the subclasses). Create a default method that generated subclasses can override.

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/DeprecatedServiceStubSettings.golden
```diff
@@ -102,6 +102,12 @@ public class DeprecatedServiceStubSettings extends StubSettings<DeprecatedServic
             "Transport not supported: %s", getTransportChannelProvider().getTransportName()));
   }
 
+  /** Returns the default service name. */
+  @Override
+  public String getServiceName() {
+    return "";
+  }
```
lqiu96, line 109: Probably can exclude this from being generated if it's going to return `""`. The parent StubSettings already returns `""`

gapic-generator-java/src/main/java/com/google/api/generator/gapic/model/Service.java
```diff
@@ -184,6 +190,17 @@ private static String parseApiVersion(String protoPackage) {
     return apiVersion;
   }
 
+  // Parse the service name from the default host configured in the protos
+  // or service yaml file. For Google Cloud Services, the default host value
+  // is expected to contain `.googleapis.com`. Exceptions may exist (i.e. localhost),
+  // in which case we will return an empty string.
+  private static String parseHostServiceName(String defaultHost) {
```
blakeli0, line 197: I know we are already testing this method indirectly through golden tests. Since this is in Service not composers, and we can easily mock the input, it would be great if we can have unit tests testing it directly as well.

lqiu96: Added some tests for parsing, thanks!

PR #2291, opened 2023-12-07
test: increasing timeout in testOnResponseError

suztomo:
Fixes #1975 

(This test is multithread because RetryingServerStreamingCallable uses an executor.)

This change increases the timeout in HttpJsonDirectServerStreamingCallableTest.testOnResponseError to avoid flakiness. There's no guarantee that this multithread test executes the CountDownLatch.countDown() within 2 seconds.

Will this 60-second timeout make the test execution slower? => No, it won't. `CountDownLatch.await(timeout)` returns immediately when the count becomes zero, without waiting for timeout.

image

In b/281936326, Lawrence tried to reproduce the issue in Windows, but it didn't. I think the best way is to increase timeout and observe any recurrence.

PR #2308, opened 2023-12-15
test: increasing timeout in testServerStreamingStart

suztomo:
Fixes https://github.com/googleapis/sdk-platform-java/issues/1678

The assertion error tells `latch.await(2, TimeUnit.SECONDS)` did not return true.

https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CountDownLatch.html#await-long-java.util.concurrent.TimeUnit-

> Returns:
> true if the count reached zero and false if the waiting time elapsed before the count reached zero

Under a multithreading test, there's no guarantee that `latch.await(2, TimeUnit.SECONDS)` finishes within 2 second.
Increasing the value to 60 second to (drastically) reduce the flakiness. (This is the same tactics as https://github.com/googleapis/sdk-platform-java/pull/2291)

Will this 60-second timeout make the test execution slower? => No, it won't. `CountDownLatch.await(timeout)` returns immediately when the count becomes zero, without waiting for timeout.


JoeWang1127 (approved):
Let's merge this now.

PR #2312, opened 2023-12-18
feat: add parsing of autopopulated fields from serviceyaml

alicejli:
See b/309351289 for more details.

This PR adds parsing of the service_config.yaml file for autopopulated fields and adds them to the `Field` and `Method` classes. 

Follow-up: Add logic to `AbstractTransportServiceStubClassComposer.java` to auto-populate the field (similar to how dynamic routing headers is added). We may go with a different design in the end, but that should be one viable approach.

Note that I updated `echo.proto` for testing; once this feature is implemented in gapic-showcase, I can look to remove those manual updates.


gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -683,9 +689,28 @@ static List<Method> parseMethods(
...
+      Optional<com.google.api.Service> serviceYamlProtoOpt,
       Set<ResourceName> outputArgResourceNames,
       Transport transport) {
     List<Method> methods = new ArrayList<>();
+
+    // Parse the serviceYaml for autopopulated methods and fields once and put into a map
+    ImmutableMap.Builder<String, List<String>> autoPopulatedMethodsWithFieldsBuilder =
+        ImmutableMap.builder();
+    if (serviceYamlProtoOpt.isPresent()) {
+      if (serviceYamlProtoOpt.get().getPublishing().getMethodSettingsList().size() > 0) {
```
JoeWang1127, line 701: Should we worry about `getPublishing()` return `null`?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -1000,6 +1034,21 @@ private static Field parseField(
...
+    // TODO: Autopopulation requires fieldInfo.formatValue = UUID4; when more types are supported,
+    // this logic will need to be updated
+    if (fieldOptions.hasExtension(FieldInfoProto.fieldInfo)) {
+      fieldInfoFormat = fieldOptions.getExtension(FieldInfoProto.fieldInfo).getFormat();
+      if (fieldOptions.getExtension(FieldInfoProto.fieldInfo).getFormatValue() == 1) {
+        isEligibleToBeAutoPopulated = true;
+      }
+    }
+    if (fieldOptions.getExtensionCount(FieldBehaviorProto.fieldBehavior) > 0) {
+      if (fieldOptions.getExtension(FieldBehaviorProto.fieldBehavior).contains(2)) ;
```
JoeWang1127, line 1048: What does "2" mean in this line?

gapic-generator-java/src/test/java/com/google/api/generator/gapic/protoparser/ServiceYamlParserTest.java
```diff
@@ -37,4 +39,22 @@ public void parseServiceYaml_basic() {
...
+    Path yamlPath = Paths.get(YAML_DIRECTORY, yamlFilename);
+    Optional<com.google.api.Service> serviceYamlProtoOpt =
+        ServiceYamlParser.parse(yamlPath.toString());
+    assertTrue(serviceYamlProtoOpt.isPresent());
+
+    com.google.api.Service serviceYamlProto = serviceYamlProtoOpt.get();
+    assertEquals("showcase.googleapis.com", serviceYamlProto.getName());
+
+    Publishing publishingSettings = serviceYamlProto.getPublishing();
+    java.util.List<MethodSettings> methodSettings = publishingSettings.getMethodSettingsList();
```
JoeWang1127, line 55: nit: can you add  `import java.util.List;`?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/model/Method.java
```diff
@@ -69,6 +69,19 @@ public boolean isPaged() {
   // [["content", "error"], ["content", "error", "info"]].
   public abstract ImmutableList<List<MethodArgument>> methodSignatures();
 
+  @Nullable
+  public abstract List<String> autoPopulatedFields();
```
blakeli0, line 73: We can set the default value to an empty list so that it is not nullable.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -683,9 +689,28 @@ static List<Method> parseMethods(
...
       Optional<GapicServiceConfig> serviceConfigOpt,
+      Optional<com.google.api.Service> serviceYamlProtoOpt,
       Set<ResourceName> outputArgResourceNames,
       Transport transport) {
     List<Method> methods = new ArrayList<>();
+
+    // Parse the serviceYaml for autopopulated methods and fields once and put into a map
+    ImmutableMap.Builder<String, List<String>> autoPopulatedMethodsWithFieldsBuilder =
+        ImmutableMap.builder();
+    if (serviceYamlProtoOpt.isPresent()) {
```
blakeli0, line 700: There are three nested indentations, can we extract this logic to a separate method so it's more maintainable and testable?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/model/Method.java
```diff
@@ -69,6 +69,19 @@ public boolean isPaged() {
...
+
+  /**
+   * If a service's service_config.yaml file contains method_settings.auto_populated_fields for this
+   * method, and the method is a Unary-type, then this is true
+   */
+  public boolean hasAutoPopulatedFields() {
```
blakeli0, line 79: Can we add some unit tests for this method?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
alicejli: Refactored

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -1000,6 +1034,21 @@ private static Field parseField(
       }
     }
 
+    // If field is annotated with fieldInfo.format = UUID4 and fieldBehavior != REQUIRED, then
+    // autopopulate the field
+    // TODO: Autopopulation requires fieldInfo.formatValue = UUID4; when more types are supported,
+    // this logic will need to be updated
+    if (fieldOptions.hasExtension(FieldInfoProto.fieldInfo)) {
+      fieldInfoFormat = fieldOptions.getExtension(FieldInfoProto.fieldInfo).getFormat();
+      if (fieldOptions.getExtension(FieldInfoProto.fieldInfo).getFormatValue() == 1) {
```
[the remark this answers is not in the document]
alicejli: have refactored it slightly such that now it just sets `Field.isRequired` and `Field.fieldInfoFormat`. I think the logic of auto-population should exist on the `AbstractTransportServiceStubClassComposer.java` (e.g. it checks that the field name is listed in `Method.autoPopulatedFields()`, thefield is not required, and the field format is UUID4 before auto-generating it). 

`FieldTest.java` doesn't currently exist - I'm not sure if it makes sense to add given that the unit tests for `Field` would be relatively trivial. But thought I'd ask given our earlier convo on tests :)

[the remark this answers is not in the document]
blakeli0: > FieldTest.java doesn't currently exist - I'm not sure if it makes sense to add given that the unit tests for Field would be relatively trivial

I agree, if it's just getters/setters, we don't have to create unit tests for it. There are different opinions around it, like this [discussion](https://stackoverflow.com/questions/6197370/should-unit-tests-be-written-for-getter-and-setters), but I don't think we should do it. 

However, I see that there are changes in hashcode/equals as well, which usually need to be tested if they are handwritten. And I'm not sure why we [added them manually](https://github.com/googleapis/sdk-platform-java/pull/622) in the first place instead of using the hashcode/equals provided by [Autovalue](https://yaqs.corp.google.com/eng/q/4616341951873024). Let's discuss it offline tomorrow.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/model/Field.java
```diff
@@ -32,6 +33,12 @@ public abstract class Field {
 
   public abstract TypeNode type();
 
+  // If the field is annotated with google.api.field_behavior = REQUIRED, then this is true.
```
blakeli0, line 36: Can we add more comments mentioning that this is used for checking if a field should be auto-populated only?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/model/Field.java
```diff
@@ -72,6 +79,8 @@ public boolean equals(Object o) {
     return name().equals(other.name())
         && originalName().equals(other.originalName())
         && type().equals(other.type())
+        && isRequired() == other.isRequired()
+        && fieldInfoFormat() == other.fieldInfoFormat()
```
blakeli0, line 83: We need to add these two to the hashcode method below as well if we continue to use these manually written hashcode/equals methods

alicejli: Ah I see - I'm not clear how the integers are chosen to create the hashing (if we do decide to use these manual methods). We can discuss offline.

alicejli: Added the fields to the `hashCode()` method below as discussed yesterday.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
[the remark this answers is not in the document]
alicejli: added them to hashcode() per our offline discussion yesterday and added additional unit tests to ParserTest

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -1124,4 +1158,27 @@ static String parseNestedProtoTypeName(String fullyQualifiedName) {
...
+        .collect(
+            Collectors.toMap(
+                MethodSettings::getSelector, MethodSettings::getAutoPopulatedFieldsList));
+  }
+
+  @VisibleForTesting
+  static boolean hasMethodSettings(Optional<com.google.api.Service> serviceYamlProtoOpt) {
+    return serviceYamlProtoOpt.isPresent()
+        && serviceYamlProtoOpt.get().hasPublishing()
+        && !serviceYamlProtoOpt.get().getPublishing().getMethodSettingsList().isEmpty();
```
blakeli0, line 1182: Could `getMethodSettingsList()` return null? If it is, I think we can change this to a null check. We don't have to check if it's empty because `Collectors.toMap()` should map it to an empty map if `getMethodSettingsList()` is empty.

gapic-generator-java/src/test/proto/echo.proto
```diff
@@ -217,6 +223,12 @@ message ExpandRequest {
   google.rpc.Status error = 2;
 
   string info = 3;
+
+  // TODO: @alicejli to remove this one auto-population of request-id is added officially to gapic-showcase. This is for testing purposes only.
```
blakeli0, line 227: No need to remove it as this is for unit tests.

gapic-generator-java/src/test/resources/echo_v1beta1.yaml
```diff
@@ -94,4 +94,9 @@ http:
...
       additional_bindings:
         - post: '/v1beta2/{name=operations/**}:cancel'
-        - post: '/v1beta3/{name=operations/**}:cancel'
\ No newline at end of file
+        - post: '/v1beta3/{name=operations/**}:cancel'
+publishing:
+    method_settings:
+    - selector: google.showcase.v1beta1.Echo.Echo
+      auto_populated_fields:
+      - request_id
```
blakeli0, line 102: Just a note that we need to add more test cases for scenarios like fields do not exist, fields are not of String type etc. After we completed the minimum requirements for Storage.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -1000,6 +1022,16 @@ private static Field parseField(
       }
     }
 
+    if (fieldOptions.hasExtension(FieldInfoProto.fieldInfo)) {
+      fieldInfoFormat = fieldOptions.getExtension(FieldInfoProto.fieldInfo).getFormat();
+    }
+    if (fieldOptions.getExtensionCount(FieldBehaviorProto.fieldBehavior) > 0) {
```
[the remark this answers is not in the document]
alicejli: Unfortunately that doesn't work as `hasExtension` can't be called on repeated fields, and `FieldBehaviorProto` can be set on repeated fields, so I had to check it via `getExtensionCount` instead.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
alicejli: I think it will never be null because the default is `java.util.Collections.emptyList()`.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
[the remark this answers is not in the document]
blakeli0: I see, that's good to know.

gapic-generator-java/src/test/resources/echo_v1beta1.yaml
alicejli: Added a TODO here for that. Thanks!

blakeli0 (approved):

PR #2313, opened 2023-12-19
feat: Full Endpoint Resolution from EndpointContext

lqiu96:
Features:
- Introduces the concept of Universe Domain
- Endpoint + Universe Domain resolution is determined solely by the EndpointContext
- Support and compatibility for GDC-H, DirectPath, and mTLS

### DIFF
--- a/gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
+++ b/gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
@@ -172,8 +172,8 @@ public abstract class ClientContext {
             .setSwitchToMtlsEndpointAllowed(settings.getSwitchToMtlsEndpointAllowed())
             .setUsingGDCH(usingGDCH)
             .build();
-    String endpoint = endpointContext.getResolvedEndpoint();
-    String universeDomain = endpointContext.getResolvedUniverseDomain();
+    String endpoint = endpointContext.resolvedEndpoint();
+    String universeDomain = endpointContext.resolvedUniverseDomain();
 
     String settingsGdchApiAudience = settings.getGdchApiAudience();
     if (usingGDCH) {
--- a/gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
+++ b/gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
@@ -173,7 +173,6 @@ public abstract class ClientContext {
             .setUsingGDCH(usingGDCH)
             .build();
     String endpoint = endpointContext.resolvedEndpoint();
-    String universeDomain = endpointContext.resolvedUniverseDomain();
 
     String settingsGdchApiAudience = settings.getGdchApiAudience();
     if (usingGDCH) {
@@ -271,7 +270,7 @@ public abstract class ClientContext {
         .setClock(clock)
         .setDefaultCallContext(defaultCallContext)
         .setServiceName(settings.getServiceName())
-        .setUniverseDomain(universeDomain)
+        .setUniverseDomain(settings.getUniverseDomain())
         .setEndpoint(settings.getEndpoint())
         .setQuotaProjectId(settings.getQuotaProjectId())
         .setStreamWatchdog(watchdog)
--- a/gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
+++ b/gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
@@ -105,6 +105,9 @@ public abstract class ClientContext {
   abstract String getServiceName();
 
   @Nullable
+  public abstract String getUniverseDomain();
+
+  @Nullable
   public abstract String getEndpoint();
 
   @Nullable
@@ -157,15 +160,29 @@ public abstract class ClientContext {
     final ScheduledExecutorService backgroundExecutor = backgroundExecutorProvider.getExecutor();
 
     Credentials credentials = settings.getCredentialsProvider().getCredentials();
+    boolean usingGDCH = credentials instanceof GdchCredentials;
+    EndpointContext endpointContext =
+        EndpointContext.newBuilder()
+            .setServiceName(settings.getServiceName())
+            .setUniverseDomain(settings.getUniverseDomain())
+            .setClientSettingsEndpoint(settings.getEndpoint())
+            .setTransportChannelProviderEndpoint(
+                settings.getTransportChannelProvider().getEndpoint())
+            .setMtlsEndpoint(settings.getMtlsEndpoint())
+            .setSwitchToMtlsEndpointAllowed(settings.getSwitchToMtlsEndpointAllowed())
+            .setUsingGDCH(usingGDCH)
+            .build();
+    String endpoint = endpointContext.getResolvedEndpoint();
+    String universeDomain = endpointContext.getResolvedUniverseDomain();
 
     String settingsGdchApiAudience = settings.getGdchApiAudience();
-    if (credentials instanceof GdchCredentials) {
+    if (usingGDCH) {
       // We recompute the GdchCredentials with the audience
       String audienceString;
       if (!Strings.isNullOrEmpty(settingsGdchApiAudience)) {
         audienceString = settingsGdchApiAudience;
-      } else if (!Strings.isNullOrEmpty(settings.getEndpoint())) {
-        audienceString = settings.getEndpoint();
+      } else if (!Strings.isNullOrEmpty(endpoint)) {
+        audienceString = endpoint;
       } else {
         throw new IllegalArgumentException("Could not infer GDCH api audience from settings");
       }
@@ -204,16 +221,6 @@ public abstract class ClientContext {
     if (transportChannelProvider.needsCredentials() && credentials != null) {
       transportChannelProvider = transportChannelProvider.withCredentials(credentials);
     }
-    EndpointContext endpointContext =
-        EndpointContext.newBuilder()
-            .setServiceName(settings.getServiceName())
-            .setClientSettingsEndpoint(settings.getEndpoint())
-            .setTransportChannelProviderEndpoint(
-                settings.getTransportChannelProvider().getEndpoint())
-            .setMtlsEndpoint(settings.getMtlsEndpoint())
-            .setSwitchToMtlsEndpointAllowed(settings.getSwitchToMtlsEndpointAllowed())
-            .build();
-    String endpoint = endpointContext.getResolvedEndpoint();
     if (transportChannelProvider.needsEndpoint()) {
       transportChannelProvider = transportChannelProvider.withEndpoint(endpoint);
     }
@@ -264,6 +271,7 @@ public abstract class ClientContext {
         .setClock(clock)
         .setDefaultCallContext(defaultCallContext)
         .setServiceName(settings.getServiceName())
+        .setUniverseDomain(universeDomain)
         .setEndpoint(settings.getEndpoint())
         .setQuotaProjectId(settings.getQuotaProjectId())
         .setStreamWatchdog(watchdog)
@@ -331,6 +339,8 @@ public abstract class ClientContext {
 
     // Package-Private scope for internal use only. Shared between StubSettings and ClientContext
     abstract Builder setServiceName(String serviceName);
+
+    public abstract Builder setUniverseDomain(String universeDomain);
 
     public abstract Builder setEndpoint(String endpoint);
 
--- a/gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
+++ b/gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
@@ -81,7 +81,7 @@ public abstract class EndpointContext {
 
   private String resolvedEndpoint;
   // Default to the GDU
-  private String resolvedUniverseDomain = GOOGLE_DEFAULT_UNIVERSE;
+  private String resolvedUniverseDomain;
 
   public static Builder newBuilder() {
     return new AutoValue_EndpointContext.Builder()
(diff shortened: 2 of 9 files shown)


gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
```diff
@@ -157,15 +160,29 @@ public static ClientContext create(StubSettings settings) throws IOException {
...
-    if (credentials instanceof GdchCredentials) {
+    if (usingGDCH) {
       // We recompute the GdchCredentials with the audience
       String audienceString;
       if (!Strings.isNullOrEmpty(settingsGdchApiAudience)) {
         audienceString = settingsGdchApiAudience;
-      } else if (!Strings.isNullOrEmpty(settings.getEndpoint())) {
-        audienceString = settings.getEndpoint();
+      } else if (!Strings.isNullOrEmpty(endpoint)) {
+        audienceString = endpoint;
```
lqiu96, line 185: Confirm this behavior is correct

lqiu96: The GDCH audienceString's endpoint should match with endpoint set inside the TransportChannelProvider. Use the EndpointContext's resolved endpoint for both.

lqiu96: The else block is kept as the custom endpoint could be set as "" (empty string). `URI.create("")` does not throw an exception, so it is caught in the else block.

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -69,28 +77,76 @@ public abstract class EndpointContext {
...
+    // Universe Domain defaults to the GDU if it's not provided by the user.
+    resolvedUniverseDomain = universeDomain() != null ? universeDomain() : GOOGLE_DEFAULT_UNIVERSE;
+
+    if (Strings.isNullOrEmpty(customEndpoint)) {
+      customEndpoint = buildEndpointTemplate(serviceName(), resolvedUniverseDomain);
+      resolvedEndpoint =
+          mtlsEndpointResolver(
+              customEndpoint, mtlsEndpoint(), switchToMtlsEndpointAllowed(), mtlsProvider);
+      // Check if mTLS is configured with non-GDU
+      if (resolvedEndpoint.equals(mtlsEndpoint())
```
blakeli0, line 121: This validation works for now but there some potential concerns. Ideally we should not set `resolvedEndpoint` until the validation is passed. Here we set the resolved endpoints first, then use the set endpoint for validation. In case there are some changes to the validation logic, e.g. log warnings instead of throwing exception, it would be error prone.

A fix would be to either create a temp variable to hold the value and set it as the last step, or preferably we change this method to return a resolved endpoint, and set it separately. So that it separates the concerns of calculating and setting resolved endpoints.

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -69,28 +77,76 @@ public abstract class EndpointContext {
...
+        throw new IllegalArgumentException(
+            "mTLS is not supported in any universe other than googleapis.com");
+      }
+    } else {
+      resolvedEndpoint = customEndpoint;
+    }
+  }
+
+  // GDC-H has no concept of Universe Domain. Do not set the resolvedUniverseDomain value
+  private void determineGDCHEndpoint(String customEndpoint) {
```
blakeli0, line 132: We can change this method to return a resolved GDCH endpoint as well, instead of resolving and setting in the same place, if you plan to change the `determineEndpoint()` to return a resolved endpoint.

gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
```diff
@@ -104,6 +104,9 @@ public abstract class ClientContext {
   @Nullable
   abstract String getServiceName();
 
+  @Nullable
+  public abstract String getUniverseDomain();
```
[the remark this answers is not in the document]
lqiu96: This is for the for logic that connects ClientContext <-> StubSettings. I believe it was added a while back for reasons and this getter is (ideally) only for StubSettings usage.

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -33,13 +33,18 @@
...
 import javax.annotation.Nullable;
 
-/** Contains the fields required to resolve the endpoint */
+/** Contains the fields required to resolve the endpoint and Universe Domain */
 @InternalApi
 @AutoValue
 public abstract class EndpointContext {
+  // Default to port 443 for HTTPS. Using HTTP requires explicitly setting the endpoint
+  private static final String ENDPOINT_TEMPLATE = "SERVICE_NAME.UNIVERSE_DOMAIN:443";
+  static final String GOOGLE_DEFAULT_UNIVERSE = "googleapis.com";
```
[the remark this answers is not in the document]
lqiu96: Yeah. We can migrate to that constant once it's released.

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -69,28 +77,76 @@ public abstract class EndpointContext {
...
+    } else if (customEndpoint == null) {
+      resolvedEndpoint = buildEndpointTemplate(serviceName(), GOOGLE_DEFAULT_UNIVERSE);
+    } else {
+      resolvedEndpoint = customEndpoint;
+    }
+  }
+
+  // Construct the endpoint with the template
+  private String buildEndpointTemplate(String serviceName, String resolvedUniverseDomain) {
+    return ENDPOINT_TEMPLATE
```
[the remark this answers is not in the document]
lqiu96: Sure, sounds good. I'll change.

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -69,28 +77,76 @@ public abstract class EndpointContext {
...
+    if (usingGDCH()) {
+      determineGDCHEndpoint(customEndpoint);
+      return;
+    }
+    // Check for "" (empty string)
+    if (universeDomain() != null && universeDomain().isEmpty()) {
+      throw new IllegalArgumentException("The universe domain value cannot be empty.");
+    }
+    // Universe Domain defaults to the GDU if it's not provided by the user.
+    resolvedUniverseDomain = universeDomain() != null ? universeDomain() : GOOGLE_DEFAULT_UNIVERSE;
```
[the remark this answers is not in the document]
lqiu96: Sounds good, will do.

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
lqiu96: Makes sense. I'll update.

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
lqiu96: Sure, will update.

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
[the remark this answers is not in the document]
blakeli0: > Can we default the universe domain to GDU instead of introducing a new field resolvedUniverseDomain?

Do we still need `resolvedUniverseDomain`?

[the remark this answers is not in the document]
lqiu96: I planned to remove it and use GDU by default, but I think there is a small edge case. Found it when running the IT GDCH tests, specifically regarding: https://github.com/googleapis/sdk-platform-java/blob/3d885bab4e3fd71f682e3d477c4cb07337629681/showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java#L189-L192.

L190's ClientContext.create(...) creates an EndpointContext. Nothing is passed in via the universeDomain getter as it's a GDC-H flow, but the universe domain would be set to GDU by default.

L192's createStub() will create the Stub and it takes the params from the ClientContext (which has the GDU as the universe domain) and it will fail the GDC-H logic block.

I'm not sure the way that the IT GDC-H tests are set is makes sense (or if I'm missing something), specifically regarding explicitly creating the ClientSettings -> ClientContext -> StubSettings steps. But I think it shows that doing this is possible by the user.

I learned towards keeping the resolvedUniverseDomain so that the user can differentiate between what the user configuration and what the resolved universe domain is (null, GDU, user configured). This would possibly be done in a future enhancement as a getter.

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -149,12 +118,124 @@ public abstract static class Builder {
...
+    abstract boolean usingGDCH();
+
+    abstract String resolvedUniverseDomain();
+
     abstract EndpointContext autoBuild();
 
+    private String determineUniverseDomain() {
+      // Do not set the universe domain for GDC-H
+      if (usingGDCH()) {
+        return null;
```
blakeli0, line 150: Can we return GDU here instead of null? And make `resolvedUniverseDomain` not nullable? I see that `resolvedUniverseDomain` is used multiple places later, it is probably good now but may easily introduce null pointer exceptions later.

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -149,12 +118,124 @@ public abstract static class Builder {
...
+      return endpoint;
+    }
+
+    // GDC-H has no concept of Universe Domain. Do not set the resolvedUniverseDomain value
+    private String determineGDCHEndpoint(String customEndpoint) {
+      if (universeDomain() != null) {
+        throw new IllegalArgumentException(
+            "Universe domain configuration is incompatible with GDC-H");
+      } else if (customEndpoint == null) {
+        return buildEndpointTemplate(serviceName(), GOOGLE_DEFAULT_UNIVERSE);
```
blakeli0, line 199: Can we use `resolvedUniverseDomain` if we default it to GDU?

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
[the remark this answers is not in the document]
blakeli0: > I learned towards keeping the resolvedUniverseDomain so that the user can differentiate between what the user configuration and what the resolved universe domain is (null, GDU, user configured)

Makes sense. Please see my new comments regarding `resolvedUniverseDomain`, I think it's better to make it not nullable, but maybe there is a edge case preventing us from doing so.

gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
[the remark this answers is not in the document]
blakeli0: Yeah, for every new `ClientSettings`, it seems that we have to duplicate it in `StubSettings` as well.

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -69,58 +75,19 @@ public abstract class EndpointContext {
...
-  public abstract Builder toBuilder();
+  public abstract boolean usingGDCH();
 
-  private String resolvedEndpoint;
-
-  public static Builder newBuilder() {
-    return new AutoValue_EndpointContext.Builder().setSwitchToMtlsEndpointAllowed(false);
-  }
+  @Nullable
+  public abstract String resolvedUniverseDomain();
```
blakeli0, line 81: Do we expect `resolvedUniverseDomain` to be used by anything outside of this class? If not, I think we can make the this method and the setter in the builder package private (I don't think we can make it completed private though due to AutoValue limitations).

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
lqiu96: The resolvedUniverseDomain is what is stored in the ClientContext and StubSettings (not the user configured universe domain). Following the GDC-H case above, it's possible to manually construct the StubSettings and ClientContext and each would construct the EndpointContext. If we set the resolvedUniverseDomain to the GDU by default, then it would fail every case as the Universe Domain can never be set for GDC-H (must be null).

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
lqiu96: Actually on second thought, I think you're right. I've made it package private. I originally used it as the value I set in the ClientContext, but I can just follow what we are doing with endpoint and pass the user set values for that. This way the resolvedUniverseDomain can be non-null (default to GDU).

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
lqiu96: Updated. Thanks!

blakeli0 (approved):
LGTM. Please add Javadocs before merging.

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -48,6 +51,9 @@ public abstract class EndpointContext {
   @Nullable
   public abstract String serviceName();
 
+  @Nullable
```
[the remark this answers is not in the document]
lqiu96: Done. Thanks

Issue #1966, opened 2023-09-06
Encode any Enums in Routing Header as an Int


lqiu96:
See [go/clientlibs-routine-headers-numeric-enums](http://goto.google.com/clientlibs-routine-headers-numeric-enums) for more information.

Enums in every implicit/ explicit routing header should be encoded as an Int. 

The Routing Headers seem to be encoded here via a ParamsExtractor.
Example: https://github.com/googleapis/sdk-platform-java/blob/44701e88de88428fb8a94692ffcc85ac815cc203/showcase/gapic-showcase/src/main/java/com/google/showcase/v1beta1/stub/HttpJsonComplianceStub.java#L850-L855
This should be modified to get the Int value instead.

Generator changes:
Affect the `AbstractTransportServiceStubClassComposer`. Specifically around: https://github.com/googleapis/sdk-platform-java/blob/44701e88de88428fb8a94692ffcc85ac815cc203/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java#L1252-L1258

Both methods should be modified so that the changes are applied for both implicit and explicit routing headers.

ddixit14:
Created a new issue for implicit routing headers case -  https://github.com/googleapis/sdk-platform-java/issues/2368 .

ddixit14:
FYI, only implicit routing headers can be fixed currently because explicit routing headers are implemented as strings by design. See this [doc](https://docs.google.com/document/d/1FtOMCBaTH7HRgBaTaWHBS8gk9phQhrPuuLoThFjGsH8/edit?usp=sharing&resourcekey=0-_dpciY4eZCQwyh5-4sIPOA) for more details.

lqiu96:
Implicit Routing Headers were implemented. Discussion is pending explicit routing headers and how to properly implement it.

blakeli0:
Closing this issue as there was no progress for how to implement it for explicit routing headers, which was designed to be String in the first place. Will re-open if this becomes a priority.

PR #2328, opened 2024-01-02
chore: Numeric enums in routing headers

ddixit14:


ddixit14:
Code changes description:
createRequestFieldGetterExpr() is being called in both cases (routing_header rule present v/s absent) to map the parameters.

In first case (routingHeaderRule() == null, createRequestParamsExtractorBodyForHttpBindings() is called), createRequestFieldGetterExpr() is modified to accept an extra boolean parameter, isFieldEnum(). And if the field is enum, I am adding an extra “Value” to the fieldMethod name (https://github.com/googleapis/sdk-platform-java/pull/2328/files#diff-13cc1d69966fca5c433b2c243aa4e8414520cf05e295f02ed52f2fe5097cac21R1474) during the last iteration of the descendant fields.

In the 2nd case when routingHeaderRule()!=null, I am sending false in isFieldEnum() because `routingHeaderParam` does not have anything to check if it is an enum or not (or is there? I am not able to find it).

gapic-generator-java/src/test/proto/explicit_dynamic_routing_header_testing.proto
```diff
@@ -265,9 +266,58 @@ service ExplicitDynamicRoutingHeaderTesting {
       }
     };
   }
+
+  // This method echoes the ComplianceData request. This method exercises
+  // including a non-proto3-optional enum field in the URL path while sending the
+  // entire request object in the REST body.
+  rpc RepeatDataPathEnum(RepeatRequest) returns (RepeatResponse) {
+    option (google.api.http) = {
+      post: "/v1beta1/repeat/kingdom/{info.f_kingdom}:pathenum"
```
blakeli0, line 275: This is for implicit routing header, can you please add another test case for explicit routing header?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
```diff
@@ -1361,7 +1361,7 @@ private void createRequestParamsExtractorBodyForRoutingHeaders(
     for (int i = 0; i < routingHeaderParams.size(); i++) {
       RoutingHeaderRule.RoutingHeaderParam routingHeaderParam = routingHeaderParams.get(i);
       MethodInvocationExpr requestFieldGetterExpr =
-          createRequestFieldGetterExpr(requestVarExpr, routingHeaderParam.fieldName());
+          createRequestFieldGetterExpr(requestVarExpr, routingHeaderParam.fieldName(),false);
```
blakeli0, line 1364: Why we want to pass `false` here? I believe this line is for explicit routing header, which we should cover as well.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/model/GapicContext.java
```diff
@@ -18,13 +18,10 @@
...
 import com.google.gapic.metadata.GapicMetadata;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.TreeMap;
-import java.util.stream.Collectors;
+
 import javax.annotation.Nullable;
+import java.util.*;
```
blakeli0, line 23: Wildcard importing is [not allowed](https://engdoc.corp.google.com/eng/doc/devguide/java/style/index.md?cl=head#s3.3.1-wildcard-imports). You can change your Intellij setting to not automatically do it.

gapic-generator-java/src/test/proto/explicit_dynamic_routing_header_testing.proto
blakeli0, line 1: Can you please simplify the changes in this file to minimum?

blakeli0 (commented):
Since this only solves the implicit routing headers, can you please create a separate issue to track it and update the description of this PR accordingly? Leaving a comment at https://github.com/googleapis/sdk-platform-java/issues/1966 would be great as well.

ddixit14:
> Since this only solves the implicit routing headers, can you please create a separate issue to track it and update the description of this PR accordingly? Leaving a comment at #1966 would be great as well.

Done. Here is the issue created - https://github.com/googleapis/sdk-platform-java/issues/2368

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
```diff
@@ -1361,7 +1364,7 @@ private void createRequestParamsExtractorBodyForRoutingHeaders(
     for (int i = 0; i < routingHeaderParams.size(); i++) {
       RoutingHeaderRule.RoutingHeaderParam routingHeaderParam = routingHeaderParams.get(i);
       MethodInvocationExpr requestFieldGetterExpr =
-          createRequestFieldGetterExpr(requestVarExpr, routingHeaderParam.fieldName());
+          createRequestFieldGetterExpr(requestVarExpr, routingHeaderParam.fieldName(), false);
```
lqiu96, line 1367: Could you a quick comment above to explain why this is always set to false?

lqiu96:
FYI @ddixit14, we have Showcase ITs for Dynamic Routing Headers at https://github.com/googleapis/sdk-platform-java/blob/762c125abadb5e56682224c9f1587c71e5c6e653/showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITDynamicRoutingHeaders.java

I think the [compliance.proto](https://github.com/googleapis/gapic-showcase/blob/330a3a25eb8c4d0948860f1898fe7dcc05dc4ea5/schema/google/showcase/v1beta1/compliance.proto#L212) has a enum field that you can possibly test against. It may be possible to add a test to intercept the the enum value is sent as an int value in the headers.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
ddixit14: Because currently explicit routing header parameters are implemented as strings by design. Hence passing false for them, they are not enums.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
```diff
@@ -1462,7 +1465,7 @@ private Expr fieldValuesNotNullConditionExpr(
   }
 
   private MethodInvocationExpr createRequestFieldGetterExpr(
-      VariableExpr requestVarExpr, String fieldName) {
+      VariableExpr requestVarExpr, String fieldName, Boolean isFieldEnum) {
```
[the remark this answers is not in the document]
ddixit14: good catch. no specific reason. I've used boolean now, to be consistent with other such variables in the class. Thanks!!

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
```diff
@@ -1472,6 +1475,9 @@ private MethodInvocationExpr createRequestFieldGetterExpr(
       String currFieldName = descendantFields.get(i);
       String bindingFieldMethodName =
           String.format("get%s", JavaStyle.toUpperCamelCase(currFieldName));
+      if (i == descendantFields.size() - 1 && isFieldEnum) {
```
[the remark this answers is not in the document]
ddixit14: added as a comment.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
ddixit14: Added as a comment.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITDynamicRoutingHeaders.java
```diff
@@ -114,16 +121,36 @@ public void onClose(int statusCode, HttpJsonMetadata trailers) {
...
+        ComplianceSettings.newBuilder()
+            .setCredentialsProvider(NoCredentialsProvider.create())
+            .setTransportChannelProvider(
+                ComplianceSettings.defaultGrpcTransportProviderBuilder()
+                    .setChannelConfigurator(ManagedChannelBuilder::usePlaintext)
+                    .setInterceptorProvider(() -> interceptorList)
+                    .build())
+            .build();
+    return ComplianceClient.create(grpcComplianceSettings);
+  }
```
lqiu96, line 142: Can you move this to `TestClientInitializer` and invoke it via something like `TestClientInitializer.createGrpcComplianceClient(...)`?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITDynamicRoutingHeaders.java
```diff
@@ -175,6 +203,22 @@ public void testGrpc_matchesHeaderName() {
...
+    RepeatRequest request =
+        RepeatRequest.newBuilder().setInfo(ComplianceData.newBuilder().setFKingdomValue(5)).build();
+    RepeatResponse actualResponse = grpcComplianceClient.repeatDataSimplePath(request);
+    String headerValue = grpcComplianceInterceptor.metadata.get(REQUEST_PARAMS_HEADER_KEY);
+    assertThat(headerValue).isNotNull();
+    List<String> requestHeaders =
+        Arrays.stream(headerValue.split(SPLIT_TOKEN)).collect(Collectors.toList());
+    List<String> expectedHeaders =
+        ImmutableList.of(
+            "info.f_bool=false", "info.f_double=0.0", "info.f_int32=0", "info.f_kingdom=5");
```
lqiu96, line 218: Can you add a comment above the expextedHeaders explaining that the fields besides `info.f_kingdom` are expected to be the default values (false, 0.0, 0) since the request's info message doesn't set the values?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITDynamicRoutingHeaders.java
ddixit14: Done.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITDynamicRoutingHeaders.java
ddixit14: Done.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITDynamicRoutingHeaders.java
```diff
@@ -175,6 +189,24 @@ public void testGrpc_matchesHeaderName() {
     assertThat(requestHeaders).containsExactlyElementsIn(expectedHeaders);
   }
 
+  @Test
+  public void testGrpc_implicitHeadersEnum() {
```
lqiu96, line 193: nit: Rename this to be a bit more descriptive. Something like `testGrpc_implicitHeaders_enumsEncodedasInt()` to describe that this test covers specifically enums -> int.

lqiu96: Also, can you add an httpjson test case as well. I think it affects both transports. Ignore this comment if it only affects gRPC.

lqiu96:
Also, we talked earlier about possibly enhancing echo.proto to also test for some explicit + implicit dynamic routing enum use cases. I took another look into echo.proto and I think it makes sense since a bulk of the routing header test cases are there: https://github.com/googleapis/gapic-showcase/blob/330a3a25eb8c4d0948860f1898fe7dcc05dc4ea5/schema/google/showcase/v1beta1/echo.proto#L48-L86

I think you can create an issue in the gapic-showcase repo and someone can pick it up in the future.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITDynamicRoutingHeaders.java
```diff
@@ -114,16 +121,22 @@ public void onClose(int statusCode, HttpJsonMetadata trailers) {
...
 
   @Before
   public void createClients() throws Exception {
     // Create gRPC Interceptor and Client
     grpcInterceptor = new GrpcCapturingClientInterceptor();
     grpcClient = TestClientInitializer.createGrpcEchoClient(ImmutableList.of(grpcInterceptor));
 
+    // Create gRPC ComplianceClient and Interceptor
+    grpcComplianceInterceptor = new GrpcCapturingClientInterceptor();
+    grpcComplianceClient = TestClientInitializer.createGrpcComplianceClient(ImmutableList.of(grpcComplianceInterceptor));
```
lqiu96, line 138: nit: Add a small comment above that we include compliance client in this IT because the EchoClient doesn't have an implicit enum routing test case. Possibly link the ticket in gapic-showcase to here.

lqiu96 (approved):
I added a few more nits but overall LGTM to me. Please resolve the nits and let @blakeli0 to take another look before merging. Thanks for the PR!

blakeli0:
> Also, we talked earlier about possibly enhancing echo.proto to also test for some explicit + implicit dynamic routing enum use cases. I took another look into echo.proto and I think it makes sense since a bulk of the routing header test cases are there: https://github.com/googleapis/gapic-showcase/blob/330a3a25eb8c4d0948860f1898fe7dcc05dc4ea5/schema/google/showcase/v1beta1/echo.proto#L48-L86
> 
> I think you can create an issue in the gapic-showcase repo and someone can pick it up in the future.

If we already have a test case in `compliance.proto`, what is the advantage of adding one to `echo.proto`? Is it mostly for simplifying our showcase tests, so that we don't have to create a separate ComplianceClient? If it's for latter, I don't think we have to add one. 
Ideally the showcase protos should be grouped into features, so that routing header test cases can be in a dedicated proto. Before that happens, I think having implicit routing header test case in a separate proto is OK.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITDynamicRoutingHeaders.java
blakeli0: > nit: Rename this to be a bit more descriptive. Something like `testGrpc_implicitHeaders_enumsEncodedasInt()` to describe that this test covers specifically enums -> int.

+1000 to this. See [unit tests naming](https://engdoc.corp.google.com/eng/doc/devguide/testing/unit/best-practices.md?cl=head&polyglot=java#naming) for details.

lqiu96:
> > Also, we talked earlier about possibly enhancing echo.proto to also test for some explicit + implicit dynamic routing enum use cases. I took another look into echo.proto and I think it makes sense since a bulk of the routing header test cases are there: https://github.com/googleapis/gapic-showcase/blob/330a3a25eb8c4d0948860f1898fe7dcc05dc4ea5/schema/google/showcase/v1beta1/echo.proto#L48-L86
> > I think you can create an issue in the gapic-showcase repo and someone can pick it up in the future.
> 
> If we already have a test case in `compliance.proto`, what is the advantage of adding one to `echo.proto`? Is it mostly for simplifying our showcase tests, so that we don't have to create a separate ComplianceClient? If it's for latter, I don't think we have to add one. Ideally the showcase protos should be grouped into features, so that routing header test cases can be in a dedicated proto. Before that happens, I think having implicit routing header test case in a separate proto is OK.

For now, it's mainly for the latter reason. Ideally we would have a separate showcase proto for routing headers, but it looks like the explicit headers are mostly contained in the echo proto. It's not urgent to add an enum to the echo proto, but a smaller task to enable showcase tests for more routing headers use cases is to add it there (especially for any future explicit enhancements/ improvements. In the future/whenever someone has the time, we can create a new proto specifically for routing headers.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITDynamicRoutingHeaders.java
ddixit14: Done.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITDynamicRoutingHeaders.java
ddixit14: renamed the tests and added httpjson case as well. thank you.

blakeli0 (approved):

PR #2329, opened 2024-01-02
feat: Add Universe Domain to Java-Core

lqiu96:
Google Auth Library v1.21.0 has been released. 

ServiceOptions exposes Getters and Setters for Universe Domain. There is a helper method to resolve the endpoint and to validate the universe domain.

### DIFF
--- a/gapic-generator-java-pom-parent/pom.xml
+++ b/gapic-generator-java-pom-parent/pom.xml
@@ -27,7 +27,7 @@
         consistent across modules in this repository -->
     <javax.annotation-api.version>1.3.2</javax.annotation-api.version>
     <grpc.version>1.60.0</grpc.version>
-    <google.auth.version>1.20.0</google.auth.version>
+    <google.auth.version>1.21.0</google.auth.version>
     <google.http-client.version>1.43.3</google.http-client.version>
     <gson.version>2.10.1</gson.version>
     <guava.version>32.1.3-jre</guava.version>
--- a/gapic-generator-java-pom-parent/pom.xml
+++ b/gapic-generator-java-pom-parent/pom.xml
@@ -27,7 +27,7 @@
         consistent across modules in this repository -->
     <javax.annotation-api.version>1.3.2</javax.annotation-api.version>
     <grpc.version>1.60.0</grpc.version>
-    <google.auth.version>1.21.0</google.auth.version>
+    <google.auth.version>1.20.0</google.auth.version>
     <google.http-client.version>1.43.3</google.http-client.version>
     <gson.version>2.10.1</gson.version>
     <guava.version>32.1.3-jre</guava.version>
--- a/gapic-generator-java-pom-parent/pom.xml
+++ b/gapic-generator-java-pom-parent/pom.xml
@@ -27,7 +27,7 @@
         consistent across modules in this repository -->
     <javax.annotation-api.version>1.3.2</javax.annotation-api.version>
     <grpc.version>1.59.1</grpc.version>
-    <google.auth.version>1.20.0</google.auth.version>
+    <google.auth.version>1.21.0</google.auth.version>
     <google.http-client.version>1.43.3</google.http-client.version>
     <gson.version>2.10.1</gson.version>
     <guava.version>32.1.3-jre</guava.version>
--- a/gapic-generator-java-pom-parent/pom.xml
+++ b/gapic-generator-java-pom-parent/pom.xml
@@ -26,7 +26,7 @@
     <!-- External dependencies, especially gRPC and Protobuf version, should be
         consistent across modules in this repository -->
     <javax.annotation-api.version>1.3.2</javax.annotation-api.version>
-    <grpc.version>1.59.1</grpc.version>
+    <grpc.version>1.60.0</grpc.version>
     <google.auth.version>1.21.0</google.auth.version>
     <google.http-client.version>1.43.3</google.http-client.version>
     <gson.version>2.10.1</gson.version>
--- a/java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
+++ b/java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
@@ -805,7 +805,7 @@ public abstract class ServiceOptions<
       throw new IllegalArgumentException("Universe Domain cannot be empty");
     } else {
       if (host != null) {
-        return normalizeEndpoint(host);
+        return normalizeEndpoint();
       }
       return formatEndpoint(serviceName, getUniverseDomain());
     }
@@ -815,8 +815,9 @@ public abstract class ServiceOptions<
     return serviceName + "." + universeDomain + ":443";
   }
 
-  // Best effort endpoint normalization to ensure it results in {host}:{port} format
-  private String normalizeEndpoint(String host) {
+  // Best effort endpoint normalization to ensure it results in {domain}:{port} format
+  // for gRPC-Java. `host` is expected to be in format of http(s)://{domain}
+  String normalizeEndpoint() {
     URI uri = URI.create(host);
     String scheme = uri.getScheme();
     int port = uri.getPort();
@@ -826,14 +827,9 @@ public abstract class ServiceOptions<
       return String.format("%s:%s", uri.getHost(), port > 0 ? port : 80);
     } else if (scheme.equals("https")) {
       return String.format("%s:%s", uri.getHost(), port > 0 ? port : 443);
+    } else {
+      throw new RuntimeException("Invalid host: " + host + ". Expecting http(s)://{domain}");
     }
-    // If scheme does not match above, then host does not have a valid scheme. The
-    // host param value is probably in format {host}:{port}.
-    if (host.contains(":")) {
-      return host;
-    }
-    // If the scheme and port are not provided, default to 443
-    return host + ":443";
   }
 
   /**
@@ -845,7 +841,7 @@ public abstract class ServiceOptions<
   public String getResolvedApiaryHost(String serviceName) {
     String resolvedUniverseDomain =
         getUniverseDomain() != null ? getUniverseDomain() : Credentials.GOOGLE_DEFAULT_UNIVERSE;
-    return "https://" + serviceName + "." + resolvedUniverseDomain + "/";
+    return "https://www." + serviceName + "." + resolvedUniverseDomain + "/";
   }
 
   /** Validates that Credentials' Universe Domain matches the user configured Universe Domain. */
--- a/java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
+++ b/java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
@@ -848,8 +848,8 @@ public abstract class ServiceOptions<
    * Validates that Credentials' Universe Domain matches the resolved Universe Domain. Currently,
    * this is only intended for BigQuery and Storage Apiary Wrapped Libraries.
    *
-   * <p>This validation call should be made prior to Apiary RPCs invocations. If the checks find
-   * that there is an invalid universe domain, the call should not be made.
+   * <p>This validation call should be made prior to any RPC invocation. This call is used to gate
+   * the RPC invocation if there is no valid universe domain.
    */
   @InternalApi
   public boolean hasValidUniverseDomain() throws IOException {
--- a/java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
+++ b/java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
@@ -813,29 +813,40 @@ public abstract class ServiceOptions<
     }
   }
 
-  /** Host parameter is expected in the format of {scheme}:{host} */
+  private String formatEndpoint(String serviceName, String universeDomain) {
+    return serviceName + "." + universeDomain + ":443";
+  }
+
   private String normalizeEndpoint(String host) {
     URI uri = URI.create(host);
     String scheme = uri.getScheme();
     int port = uri.getPort();
 
-    // http:// defaults to 80, https:// or no scheme default to 443
-    if ("http".equals(scheme)) {
-      port = port > 0 ? port : 80;
-    } else if ("https".equals(scheme)) {
-      port = port > 0 ? port : 443;
-    } else {
-      port = 443;
+    // If no port is provided, http:// defaults to 80 and https:// defaults to 443
+    if (scheme.equals("http")) {
+      return String.format("%s:%s", uri.getHost(), port > 0 ? port : 80);
+    } else if (scheme.equals("https")) {
+      return String.format("%s:%s", uri.getHost(), port > 0 ? port : 443);
     }
-    return String.format("%s:%s", uri.getHost(), port);
+    // If scheme does not match above, then host does not have a valid scheme. The
+    // host param value is probably in format {host}:{port}.
+    if (host.contains(":")) {
+      return host;
+    }
+    // If the scheme and port are not provided, default to 443
+    return host + ":443";
   }
 
-  // Temporarily used for Apiary Wrapped Libraries. To be removed in the future.
+  /**
+   * Temporarily used for BigQuery and Storage Apiary Wrapped Libraries. To be removed in the
+   * future. Returns the host to be used for the rootUrl and output is in format of:
+   * "https://serviceName.universeDomain/"
+   */
   @InternalApi
-  public String getResolvedApiaryEndpoint(String serviceName) {
+  public String getResolvedApiaryHost(String serviceName) {
     String resolvedUniverseDomain =
         getUniverseDomain() != null ? getUniverseDomain() : GOOGLE_DEFAULT_UNIVERSE;
-    return formatEndpoint(serviceName, resolvedUniverseDomain);
+    return "https://" + serviceName + "." + resolvedUniverseDomain + "/";
   }
 
   /** Validates that Credentials' Universe Domain and user configured Universe Domain matches. */
@@ -846,9 +857,5 @@ public abstract class ServiceOptions<
         universeDomain != null ? universeDomain : GOOGLE_DEFAULT_UNIVERSE;
     return true;
     //    return credentials.getUniverseDomain() != resolvedUniverseDomain;
-  }
-
-  private String formatEndpoint(String serviceName, String universeDomain) {
-    return serviceName + "." + universeDomain + ":443";
   }
 }
--- a/java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
+++ b/java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
@@ -815,4 +815,14 @@ public abstract class ServiceOptions<
         .replace("SERVICE_NAME", serviceName)
         .replace("UNIVERSE_DOMAIN", universeDomain);
   }
+
+  public boolean isValidUniverseDomain() throws IOException {
+    Credentials credentials = getCredentials();
+    String universeDomain = getUniverseDomain();
+    if (universeDomain == null) {
+      universeDomain = ServiceOptions.GOOGLE_DEFAULT_UNIVERSE;
+    }
+    return true;
+    //    return credentials.getUniverseDomain() != universeDomain;
+  }
 }
--- a/java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
+++ b/java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
@@ -202,13 +202,18 @@ public abstract class ServiceOptions<
     }
 
     /**
-     * Universe Domain is the domain for Google Cloud Services. It follows the format of
-     * `{ServiceName}.{UniverseDomain}`. For example, speech.googleapis.com would have a Universe
-     * Domain value of `googleapis.com` and cloudasset.test.com would have a Universe Domain of
-     * `test.com`. If this value is not set, this will default to `googleapis.com`.
+     * Universe Domain is the domain for Google Cloud Services. A Google Cloud endpoint follows the
+     * format of `{ServiceName}.{UniverseDomain}`. For example, speech.googleapis.com would have a
+     * Universe Domain value of `googleapis.com` and cloudasset.test.com would have a Universe
+     * Domain of `test.com`.
+     *
+     * <p>If this value is not set, the resolved UniverseDomain will default to `googleapis.com`.
+     *
+     * @throws NullPointerException if {@code universeDomain} is {@code null}. The resolved
+     *     universeDomain will be `googleapis.com` if this value is not set.
      */
     public B setUniverseDomain(String universeDomain) {
-      this.universeDomain = universeDomain;
+      this.universeDomain = checkNotNull(universeDomain);
       return self();
     }
 
@@ -597,10 +602,13 @@ public abstract class ServiceOptions<
   }
 
   /**
-   * Universe Domain is the domain for Google Cloud Services. It follows the format of
-   * `{ServiceName}.{UniverseDomain}`. For example, speech.googleapis.com would have a Universe
-   * Domain value of `googleapis.com` and cloudasset.test.com would have a Universe Domain of
-   * `test.com`. If this value is not set, this will default to `googleapis.com`.
+   * Universe Domain is the domain for Google Cloud Services. A Google Cloud endpoint follows the
+   * format of `{ServiceName}.{UniverseDomain}`. For example, speech.googleapis.com would have a
+   * Universe Domain value of `googleapis.com` and cloudasset.test.com would have a Universe Domain
+   * of `test.com`.
+   *
+   * @return The universe domain value set in the Builder's setter. Does not return the resolved
+   *     Universe Domain
    */
   public String getUniverseDomain() {
     return universeDomain;
@@ -793,9 +801,10 @@ public abstract class ServiceOptions<
   }
 
   /**
-   * Returns the resolved endpoint for the Service to connect to Google Cloud
+   * Returns the resolved host for the Service to connect to Google Cloud
    *
-   * <p>The resolved endpoint is always in `host:port` format
+   * <p>The resolved host will be in `https://www.{serviceName}.{resolvedUniverseDomain}/` format.
+   * The resolvedUniverseDomain will be set to `googleapis.com` if universeDomain is null.
    */
   @InternalApi
   public String getResolvedHost(String serviceName) {
@@ -816,23 +825,23 @@ public abstract class ServiceOptions<
   /**
    * Temporarily used for BigQuery and Storage Apiary Wrapped Libraries. To be removed in the
    * future. Returns the host to be used for the rootUrl and output is in the format of:
-   * "https://serviceName.universeDomain/"
+   * "https://www.serviceName.universeDomain/"
    */
   @InternalApi
   public String getResolvedApiaryHost(String serviceName) {
     String resolvedUniverseDomain =
-        getUniverseDomain() != null ? getUniverseDomain() : Credentials.GOOGLE_DEFAULT_UNIVERSE;
+        universeDomain != null ? universeDomain : Credentials.GOOGLE_DEFAULT_UNIVERSE;
     return "https://www." + serviceName + "." + resolvedUniverseDomain + "/";
   }
 
   /**
-   * Validates that Credentials' Universe Domain matches the user configured Universe Domain.
-   * Currently, this is intended for BigQuery and Storage Apiary Wrapped Libraries
+   * Validates that Credentials' Universe Domain matches the resolved Universe Domain. Currently,
+   * this is intended for BigQuery and Storage Apiary Wrapped Libraries
    */
   @InternalApi
-  public boolean isValidUniverseDomain() throws IOException {
+  public boolean hasValidUniverseDomain() throws IOException {
     String resolvedUniverseDomain =
-        getUniverseDomain() != null ? getUniverseDomain() : Credentials.GOOGLE_DEFAULT_UNIVERSE;
+        universeDomain != null ? universeDomain : Credentials.GOOGLE_DEFAULT_UNIVERSE;
     return resolvedUniverseDomain.equals(getCredentials().getUniverseDomain());
   }
 }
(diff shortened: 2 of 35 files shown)


java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -767,4 +794,40 @@ public String getClientLibToken() {
...
+  public String getResolvedEndpoint(String serviceName) {
+    if (universeDomain == null) {
+      return formatEndpoint(serviceName, GOOGLE_DEFAULT_UNIVERSE);
+    } else if (Strings.isNullOrEmpty(universeDomain)) {
+      throw new IllegalArgumentException("Universe Domain cannot be empty");
+    } else {
+      if (host == null) {
+        return formatEndpoint(serviceName, getUniverseDomain());
+      }
+      return host;
```
blakeli0, line 808: Can we refactor the logic in this method to something like
```
if (host != null) {
  return host;
}
//check universeDomain
``` 
This would be clearer to comprehend that `host` should override `universeDomain` if specified, and reduce the nested indentations.

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -767,4 +794,40 @@ public String getClientLibToken() {
   public String getQuotaProjectId() {
     return quotaProjectId;
   }
+
+  /** Returns the resolved endpoint for the Service to connect to Google Cloud */
+  public String getResolvedEndpoint(String serviceName) {
```
blakeli0, line 799: I think `host` in java-core is basically `endpoint` in Gax without the port, based on the value of [DEFAULT_HOST](https://github.com/googleapis/sdk-platform-java/blob/main/java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java#L85). So we should either rename this to `getResolvedHost` and remove the port concatenation in `formatEndpoint`, or return `host:port` if `host` is specified.
Can you please verify the format of `host`?

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -81,6 +82,7 @@ public abstract class ServiceOptions<
     implements Serializable {
 
   public static final String CREDENTIAL_ENV_NAME = "GOOGLE_APPLICATION_CREDENTIALS";
+  public static final String GOOGLE_DEFAULT_UNIVERSE = "googleapis.com";
```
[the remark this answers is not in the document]
lqiu96: Yep, this PR is dependent on the Auth release. I was going to migrate to using the Auth constant.

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -767,4 +794,40 @@ public String getClientLibToken() {
...
+      if (host == null) {
+        return formatEndpoint(serviceName, getUniverseDomain());
+      }
+      return host;
+    }
+  }
+
+  // Temporarily used for Apiary Wrapped Libraries. To be removed in the future.
+  @InternalApi
+  public String getResolvedApiaryEndpoint(String serviceName) {
```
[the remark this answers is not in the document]
lqiu96: Yep, onyl for Bigquery and Storage for now

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
lqiu96: Sure, will update!

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
lqiu96: I think the format of host is intended to be `scheme://host:port`. For gRPC, the channel is expecting something in format of `host:port` and httpjson is open to anything.

I see that the handwritten libraries handle these inside their own codebase:
1. Storage gRPC (strips the scheme and adds a port if missing): https://github.com/googleapis/java-storage/blob/2b5b27eac3279db815b36b252830d0905ade0665/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java#L163-L170
2. Storage HttpJson (sets it directly): https://github.com/googleapis/java-storage/blob/2b5b27eac3279db815b36b252830d0905ade0665/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java#L143
3. Spanner (enforces that a port exists): https://github.com/googleapis/java-spanner/blob/21f5eba7d46c9011803accf0aaf87e6072066419/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java#L1469-L1471
4. Bigquery (sets it directly): https://github.com/googleapis/java-bigquery/blob/7a1bbd2d33e71644d83c42b71c67bac099882a4a/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/spi/v2/HttpBigQueryRpc.java#L108

I'll update this to return in `host:port` format since that's the minimum required from both transports. The handwritten libraries already check to add in the port if it's missing.

gapic-generator-java-pom-parent/pom.xml
```diff
@@ -27,7 +27,7 @@
         consistent across modules in this repository -->
     <javax.annotation-api.version>1.3.2</javax.annotation-api.version>
     <grpc.version>1.60.0</grpc.version>
-    <google.auth.version>1.20.0</google.auth.version>
+    <google.auth.version>1.21.0</google.auth.version>
```
lqiu96, line 30: Manually update as renovate bot has not picked up the changes yet.

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -767,4 +794,40 @@ public String getClientLibToken() {
...
+    String resolvedUniverseDomain =
+        getUniverseDomain() != null ? getUniverseDomain() : GOOGLE_DEFAULT_UNIVERSE;
+    return formatEndpoint(serviceName, resolvedUniverseDomain);
+  }
+
+  /** Validates that Credentials' Universe Domain and user configured Universe Domain matches. */
+  public boolean isValidUniverseDomain() throws IOException {
```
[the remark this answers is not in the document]
lqiu96: Google-Auth library has released v1.21.0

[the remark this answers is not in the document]
blakeli0: I still don't see this method being called? Do we expect this method to be called from handwritten libraries?

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -767,4 +792,62 @@ public String getClientLibToken() {
...
+   *
+   * <p>The resolved endpoint is always in `host:port` format
+   */
+  public String getResolvedEndpoint(String serviceName) {
+    if (universeDomain == null) {
+      return formatEndpoint(serviceName, Credentials.GOOGLE_DEFAULT_UNIVERSE);
+    } else if (universeDomain.isEmpty()) {
+      throw new IllegalArgumentException("Universe Domain cannot be empty");
+    } else {
+      if (host != null) {
```
blakeli0, line 807: If host is set and universeDomain is null, I think we should use `normalizeEndpoint()` instead of `formatEndpoint(serviceName, Credentials.GOOGLE_DEFAULT_UNIVERSE)`?

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -767,4 +792,62 @@ public String getClientLibToken() {
...
+      }
+      return formatEndpoint(serviceName, getUniverseDomain());
+    }
+  }
+
+  private String formatEndpoint(String serviceName, String universeDomain) {
+    return serviceName + "." + universeDomain + ":443";
+  }
+
+  // Best effort endpoint normalization to ensure it results in {domain}:{port} format
```
blakeli0, line 818: Is `domain` of format `serviceName.universeDomain`?

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -199,6 +201,17 @@ public B setHost(String host) {
       return self();
     }
 
+    /**
+     * Universe Domain is the domain for Google Cloud Services. It follows the format of
```
suztomo, line 205: It's unclear What "It" is that has the format in next line.

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -767,4 +791,48 @@ public String getClientLibToken() {
   public String getQuotaProjectId() {
     return quotaProjectId;
   }
+
+  /**
+   * Returns the resolved endpoint for the Service to connect to Google Cloud
```
suztomo, line 796: I felt the term "host" in method name and "endpoint" in Javadoc are confusing.

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -767,4 +791,48 @@ public String getClientLibToken() {
...
+   *
+   * <p>The resolved endpoint is always in `host:port` format
```
suztomo, line 798: I see this method returning "https://...".

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -199,6 +201,17 @@ public B setHost(String host) {
...
     }
 
+    /**
+     * Universe Domain is the domain for Google Cloud Services. It follows the format of
+     * `{ServiceName}.{UniverseDomain}`. For example, speech.googleapis.com would have a Universe
+     * Domain value of `googleapis.com` and cloudasset.test.com would have a Universe Domain of
+     * `test.com`. If this value is not set, this will default to `googleapis.com`.
+     */
+    public B setUniverseDomain(String universeDomain) {
+      this.universeDomain = universeDomain;
```
suztomo, line 211: Can you add `checkNotNull(universeDomain)` if universeDomain cannot be null?

If universeDomain is ok to be null, would you add a paragraph explaining when a caller want to set this field as null?

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -767,4 +791,48 @@ public String getClientLibToken() {
...
+
+  /**
+   * Returns the resolved endpoint for the Service to connect to Google Cloud
+   *
+   * <p>The resolved endpoint is always in `host:port` format
+   */
+  @InternalApi
+  public String getResolvedHost(String serviceName) {
+    if (universeDomain != null && universeDomain.isEmpty()) {
+      throw new IllegalArgumentException("Universe Domain cannot be empty");
```
suztomo, line 803: I felt it's wrong a method that "Returns the resolved endpoint for the Service to connect to Google Cloud" works only when `universeDomain` is set.

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -767,4 +791,48 @@ public String getClientLibToken() {
...
+        getUniverseDomain() != null ? getUniverseDomain() : Credentials.GOOGLE_DEFAULT_UNIVERSE;
+    return "https://www." + serviceName + "." + resolvedUniverseDomain + "/";
+  }
+
+  /**
+   * Validates that Credentials' Universe Domain matches the user configured Universe Domain.
+   * Currently, this is intended for BigQuery and Storage Apiary Wrapped Libraries
+   */
+  @InternalApi
+  public boolean isValidUniverseDomain() throws IOException {
```
suztomo, line 833:
```suggestion
  public boolean hasValidUniverseDomain() throws IOException {
```

java-core/google-cloud-core/src/test/java/com/google/cloud/ServiceOptionsTest.java
```diff
@@ -140,13 +144,44 @@ public class ServiceOptionsTest {
           + "  \"client_email\": \"someclientid@developer.gserviceaccount.com\",\n"
           + "  \"client_id\": \"someclientid.apps.googleusercontent.com\",\n"
           + "  \"type\": \"service_account\",\n"
-          + "  \"quota_project_id\": \"some-quota-project-id\"\n"
+          + "  \"quota_project_id\": \"some-quota-project-id\",\n"
+          + "  \"universe_domain\": \"googleapis.com\"\n"
+          + "}";
+
+  private static final String JSON_KEY_NON_GDU =
```
suztomo, line 151: Would you add source code comment how you generated this value?

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -767,4 +791,48 @@ public String getClientLibToken() {
...
+    String resolvedUniverseDomain =
+        getUniverseDomain() != null ? getUniverseDomain() : Credentials.GOOGLE_DEFAULT_UNIVERSE;
```
suztomo, line 835: It's itchy that getUniverseDomain is called twice. Can you save the returned value in a variable?

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -767,4 +791,48 @@ public String getClientLibToken() {
...
+    // The DEFAULT_HOST value is not a valid host for handwritten libraries.
+    if (!host.equals(DEFAULT_HOST)) {
+      return host;
+    }
+    return "https://www." + serviceName + "." + resolvedUniverseDomain + "/";
+  }
+
+  /**
+   * Temporarily used for BigQuery and Storage Apiary Wrapped Libraries. To be removed in the
+   * future. Returns the host to be used for the rootUrl and output is in the format of:
```
suztomo, line 818:
```suggestion
   * future when XXX. Returns the host to be used for the rootUrl and output is in the format of:
```

Would you fill "when XXX" to clarify what we're waiting for before removing this method.

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -582,6 +596,16 @@ public String getProjectId() {
     return projectId;
   }
 
+  /**
+   * Universe Domain is the domain for Google Cloud Services. It follows the format of
+   * `{ServiceName}.{UniverseDomain}`. For example, speech.googleapis.com would have a Universe
+   * Domain value of `googleapis.com` and cloudasset.test.com would have a Universe Domain of
+   * `test.com`. If this value is not set, this will default to `googleapis.com`.
```
suztomo, line 603: > If this value is not set, this will default to `googleapis.com`.

I felt it's unclear what is the second "this" in "this will default to `googleapis.com`". When universe domain is not set, this is null. I think you want to add "when" as in "this will default to `googleapis.com when ...`.

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -199,6 +201,17 @@ public B setHost(String host) {
       return self();
     }
 
+    /**
+     * Universe Domain is the domain for Google Cloud Services. It follows the format of
+     * `{ServiceName}.{UniverseDomain}`. For example, speech.googleapis.com would have a Universe
+     * Domain value of `googleapis.com` and cloudasset.test.com would have a Universe Domain of
+     * `test.com`. If this value is not set, this will default to `googleapis.com`.
```
suztomo, line 208: Would you add a paragraph that explains the validity of the argument in terms of `isValidUniverseDomain`?

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
suztomo, line 796: Would you add a paragraph that explains the specification of the resolution in this Javadoc?

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -767,4 +791,48 @@ public String getClientLibToken() {
...
+  @InternalApi
+  public String getResolvedApiaryHost(String serviceName) {
+    String resolvedUniverseDomain =
+        getUniverseDomain() != null ? getUniverseDomain() : Credentials.GOOGLE_DEFAULT_UNIVERSE;
+    return "https://www." + serviceName + "." + resolvedUniverseDomain + "/";
+  }
+
+  /**
+   * Validates that Credentials' Universe Domain matches the user configured Universe Domain.
+   * Currently, this is intended for BigQuery and Storage Apiary Wrapped Libraries
```
suztomo, line 830: Can you add a paragraph when these libraries should call this method? Or is it optional to validate universe domain?

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
suztomo: It reads:

> A is B. It has the format "xxx.{A}".

A's format uses A in it.

lqiu96: Sounds good, will update.

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
lqiu96: universeDomain _can_ be null. The resolved universeDomain will be `googleapis.com`. I'll add a sentence (there should be no reason to manually set this field as null).

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
lqiu96: I updated the method, but forgot to update the javadocs. Thanks!

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
lqiu96: This is a required check. If universedomain is set as `""`, we must error out.

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
lqiu96: I do think the above is the correct behavior. I want to keep consistency with the current GAPICs behavior where the value put in the setter is what is returned by the getter (this is a future enhancement to be able to return the resolved values).

The resolved universeDomain is going to be `googleapis.com` when the inputted universeDomain is null. I'll update the docs to be clearer.

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
lqiu96: I pivoted to use host last night. I'll update the docs to reference host instead of endpoint. Host is used for Java-Core for backwards compatibility even though GAPICs use the endpoint terminology.

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
lqiu96: Sounds good! I think this makes sense.

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
lqiu96: Sounds good, will update.

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
lqiu96: I think I misunderstood. The inputted universeDomain can never be null, but the universeDomain that is used get resolve the final universeDomain can be (as it will set the resolved universeDomain as `googleapis.com` if the user does not touch the setter). I'll update.

java-core/google-cloud-core/src/test/java/com/google/cloud/ServiceOptionsTest.java
lqiu96: I didn't generate this value. I copied it from above and added a new json entry for `universe_domain`. This `universe_domain` addition to Credentials was part of Auth v1.21.0

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
suztomo: Ensure this won't cause the discrepancy in user-facing interfaces.

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -767,4 +799,52 @@ public String getClientLibToken() {
...
+  /**
+   * Returns the resolved host for the Service to connect to Google Cloud
+   *
+   * <p>The resolved host will be in `https://www.{serviceName}.{resolvedUniverseDomain}/` format.
+   * The resolvedUniverseDomain will be set to `googleapis.com` if universeDomain is null.
+   */
+  @InternalApi
+  public String getResolvedHost(String serviceName) {
+    if (universeDomain != null && universeDomain.isEmpty()) {
+      throw new IllegalArgumentException("Universe Domain cannot be empty");
```
suztomo, line 812:
```suggestion
      throw new IllegalArgumentException("Universe Domain cannot be empty string");
```

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
suztomo: I misunderstood the meaning of the code. Added suggested edits.

java-core/google-cloud-core/src/test/java/com/google/cloud/ServiceOptionsTest.java
suztomo: Can you add that "copied it from above and added a new json entry for universe_domain"?

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -767,4 +799,52 @@ public String getClientLibToken() {
...
+    // overriden to include the serviceName.
+    if (!host.equals(DEFAULT_HOST)) {
+      return host;
+    }
+    return "https://www." + serviceName + "." + resolvedUniverseDomain + "/";
+  }
+
+  /**
+   * Temporarily used for BigQuery and Storage Apiary Wrapped Libraries. To be removed in the
+   * future. Returns the host to be used as the rootUrl.
```
suztomo, line 828: Would you fill "when XXX" to clarify what we're waiting for before removing this method.

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -767,4 +799,52 @@ public String getClientLibToken() {
...
+  @InternalApi
+  public String getResolvedApiaryHost(String serviceName) {
+    String resolvedUniverseDomain =
+        universeDomain != null ? universeDomain : Credentials.GOOGLE_DEFAULT_UNIVERSE;
+    return "https://www." + serviceName + "." + resolvedUniverseDomain + "/";
+  }
+
+  /**
+   * Validates that Credentials' Universe Domain matches the resolved Universe Domain. Currently,
+   * this is only intended for BigQuery and Storage Apiary Wrapped Libraries
```
suztomo, line 842: Can you add a paragraph when these libraries should call this method? Or is it optional to validate universe domain?

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
suztomo, line 828: I believe rootUrl is not a generic term but specific to our code. Can you add a link to `rootUrl`?

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
lqiu96: Guidance is to use `The universe domain cannot be empty`. I'll update that for consistency among languages.

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
suztomo: Then how about setting the requirement in `setUniverseDomain`?

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
lqiu96: Sure, I'll something.

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
lqiu96: Sounds good, will add.

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -767,4 +799,52 @@ public String getClientLibToken() {
   public String getQuotaProjectId() {
     return quotaProjectId;
   }
+
+  /**
+   * Returns the resolved host for the Service to connect to Google Cloud
+   *
+   * <p>The resolved host will be in `https://www.{serviceName}.{resolvedUniverseDomain}/` format.
```
suztomo, line 806: Would you add a reference about adding "www."? I checked `www.speech.googleapis.com` but it (the DNS entry) does not exist.

```
suztomo@suztomo2:~$ ping www.speech.googleapis.com
ping: www.speech.googleapis.com: Name or service not known
```

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -767,4 +799,52 @@ public String getClientLibToken() {
...
+    if (universeDomain != null && universeDomain.isEmpty()) {
+      throw new IllegalArgumentException("Universe Domain cannot be empty");
+    }
+    String resolvedUniverseDomain =
+        universeDomain != null ? universeDomain : Credentials.GOOGLE_DEFAULT_UNIVERSE;
+    // The host value set to DEFAULT_HOST if the user didn't configure a host. If the
+    // user set a host the library uses that value, otherwise, construct the host for the user.
+    // The DEFAULT_HOST value is not a valid host for handwritten libraries and should be
+    // overriden to include the serviceName.
+    if (!host.equals(DEFAULT_HOST)) {
```
blakeli0, line 820: This could throw null pointer exception if host is null. I see that host is set to DEFAULT_HOST in the constructor, but I would change the code to the following to prevent any possible NullPointerException in the future.
```suggestion
    if (!DEFAULT_HOST.equals(host)) {
```

gapic-generator-java-pom-parent/pom.xml
blakeli0: Let's create a separate PR to update this dependency, so it can properly show up in release notes.

blakeli0 (approved):
LGTM. Please resolve outstanding comments before merging.

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
lqiu96: Added a link to DEFAULT_HOST

java-core/google-cloud-core/src/test/java/com/google/cloud/ServiceOptionsTest.java
lqiu96: Added

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
lqiu96: This works for `cloudkms.googleapis.com` not sure why there is an issue for `speech.googleapis.com`

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
suztomo: You could say there's specification for the exception message.

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
lqiu96: I think that makes sense. Guidance for temporary Apiary is to use whatever is passed in as the universe domain. The empty check is for handwritten and gapics. I'll ask about the empty string check, but it shouldn't be blocking for this.

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
lqiu96: Added

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -767,4 +799,62 @@ public String getClientLibToken() {
...
+   *
+   * <p>The resolved host will be in `https://{serviceName}.{resolvedUniverseDomain}` format. The
+   * resolvedUniverseDomain will be set to `googleapis.com` if universeDomain is null. The format is
+   * similar to the DEFAULT_HOST value in java-core.
+   *
+   * @see <a
+   *     href="https://github.com/googleapis/sdk-platform-java/blob/097964f24fa1989bc74b4807a253f0be4e9dd1ea/java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java#L85">DEFAULT_HOST</a>
+   */
+  @InternalApi
+  public String getResolvedHost(String serviceName) {
```
suztomo, line 814: Calling it as host is itchy. Fine but ensure to explain this to library owners.
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/net/URL.html#getHost()

suztomo (approved):

PR #2330, opened 2024-01-02
feat: Validate the Universe Domain

lqiu96:
Universe Domain validation ensures that the user configured value matches the one set in the Credentials. If validation does not succeed, the call should not proceed. Validating the unvierse domain should occur prior to each RPC invocation.

Behavior shared between transports:
- Try to validate the credentials. If universe domain matches, proceed with the call. Otherwise, return an Unauthorized exception back to the user.

Transport Specific Behavior:
- gRPC: If the Credentials is unable to get the universe domain due to an Auth Exception, mark the exception as Unavailable and retry the call following the retry bounds defined by the RPC. Otherwise, return the exception back to the user.
- HttpJson: If the Credentials is unable to get the universe domain due to an Auth Exception, return the exception back to the user.

## CallContext's `merge()`
Callables are created with the TransportCallableFactory using the defaultCallContext. The defaultCallContext is merged with the call's callcontext in a default invocation: https://github.com/googleapis/sdk-platform-java/blob/97ae228a262738e09ddf7b4dab95bb81a4f8860a/gax-java/gax/src/main/java/com/google/api/gax/rpc/UnaryCallable.java#L126

GAPIC clients expose callables which invoke the `call()` with a null callcontext: https://github.com/googleapis/google-cloud-java/blob/c12ae4aacc597161211d53f15c60d6ab904d19d2/java-accessapproval/google-cloud-accessapproval/src/main/java/com/google/cloud/accessapproval/v1/AccessApprovalAdminClient.java#L701-L703. 
Merge is not invoked and does not result in creating the EndpointContext over and over again.

### DIFF
--- a/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
+++ b/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
@@ -465,9 +465,7 @@ public final class GrpcCallContext implements ApiCallContext {
     }
 
     // The EndpointContext is not updated as there should be no reason for a user
-    // to update this. EndpointContext is an internal class used by the client library
-    // to resolve the endpoint. It is created once the library is initialized and
-    // should not be updated.
+    // to update this.
     return new GrpcCallContext(
         newChannel,
         newCredentials,
--- a/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
+++ b/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
@@ -612,8 +612,8 @@ public final class GrpcCallContext implements ApiCallContext {
   @Override
   public void validateUniverseDomain() {
     EndpointContext endpointContext = getEndpointContext();
+    Credentials credentials = getCredentials();
     try {
-      Credentials credentials = getCredentials();
       if (!endpointContext.hasValidUniverseDomain(credentials)) {
         throw ApiExceptionFactory.createException(
             new Throwable(
@@ -627,11 +627,12 @@ public final class GrpcCallContext implements ApiCallContext {
     } catch (IOException e) {
       // Check if it is an Auth Exception
       if (e instanceof Retryable) {
+        Retryable retryable = (Retryable) e;
         // Keep the behavior the same as gRPC-Java. Mark as Auth Exceptions as Unavailable
         throw ApiExceptionFactory.createException(
             new Throwable("Unable to retrieve the Universe Domain from the Credentials."),
             GrpcStatusCode.of(Status.Code.UNAVAILABLE),
-            ((Retryable) e).isRetryable());
+            retryable.isRetryable());
       }
       // Return the exception back to the user
       throw new RuntimeException(e);
--- a/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
+++ b/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
@@ -37,6 +37,8 @@ import com.google.api.gax.rpc.ApiExceptionFactory;
 import com.google.api.gax.rpc.EndpointContext;
 import com.google.api.gax.rpc.StatusCode;
 import com.google.api.gax.rpc.TransportChannel;
+import com.google.api.gax.rpc.UnauthenticatedException;
+import com.google.api.gax.rpc.UnavailableException;
 import com.google.api.gax.rpc.internal.ApiCallContextOptions;
 import com.google.api.gax.rpc.internal.Headers;
 import com.google.api.gax.tracing.ApiTracer;
@@ -72,6 +74,9 @@ import org.threeten.bp.Duration;
  */
 @BetaApi("Reference ApiCallContext instead - this class is likely to experience breaking changes")
 public final class GrpcCallContext implements ApiCallContext {
+  private static final GrpcStatusCode UNAUTHENTICATED_STATUS_CODE =
+      GrpcStatusCode.of(Status.Code.UNAUTHENTICATED);
+
   static final CallOptions.Key<ApiTracer> TRACER_KEY = CallOptions.Key.create("gax.tracer");
 
   private final Channel channel;
@@ -101,7 +106,7 @@ public final class GrpcCallContext implements ApiCallContext {
         ApiCallContextOptions.getDefaultOptions(),
         null,
         null,
-        EndpointContext.getDefault());
+        null);
   }
 
   /** Returns an instance with the given channel and {@link CallOptions}. */
@@ -118,7 +123,7 @@ public final class GrpcCallContext implements ApiCallContext {
         ApiCallContextOptions.getDefaultOptions(),
         null,
         null,
-        EndpointContext.getDefault());
+        null);
   }
 
   private GrpcCallContext(
@@ -459,8 +464,10 @@ public final class GrpcCallContext implements ApiCallContext {
       newCallOptions = newCallOptions.withOption(TRACER_KEY, newTracer);
     }
 
-    EndpointContext newEndpointContext = endpointContext.merge(grpcCallContext.endpointContext);
-
+    // The EndpointContext is not updated as there should be no reason for a user
+    // to update this. EndpointContext is an internal class used by the client library
+    // to resolve the endpoint. It is created once the library is initialized and
+    // should not be updated.
     return new GrpcCallContext(
         newChannel,
         newCredentials,
@@ -473,7 +480,7 @@ public final class GrpcCallContext implements ApiCallContext {
         newOptions,
         newRetrySettings,
         newRetryableCodes,
-        newEndpointContext);
+        endpointContext);
   }
 
   /** The {@link Channel} set on this context. */
@@ -605,34 +612,42 @@ public final class GrpcCallContext implements ApiCallContext {
     return options.getOption(key);
   }
 
-  /** {@inheritDoc} */
-  @Override
+  /**
+   * Validate the Universe Domain to ensure that the user configured Universe Domain and the
+   * Credentials' Universe Domain match. An exception will be raised if there are any issues when
+   * trying to validate (i.e. unable to access the universe domain).
+   *
+   * @throws UnauthenticatedException Thrown if the universe domain that the user configured does
+   *     not match the Credential's universe domain.
+   * @throws UnavailableException If client library is unable to retrieve the universe domain from
+   *     the Credentials and the RPC is configured to retry Unavailable exceptions, the client
+   *     library will attempt to retry with the RPC's defined retry bounds. If the retry bounds have
+   *     been exceeded and the library is still unable to retrieve the universe domain, the
+   *     exception will be thrown back to the user.
+   */
+  @InternalApi
   public void validateUniverseDomain() {
     EndpointContext endpointContext = getEndpointContext();
-    Credentials credentials = getCredentials();
     try {
-      if (!endpointContext.hasValidUniverseDomain(credentials)) {
-        throw ApiExceptionFactory.createException(
-            new Throwable(
-                String.format(
-                    EndpointContext.INVALID_UNIVERSE_DOMAIN_ERROR_TEMPLATE,
-                    endpointContext.resolvedUniverseDomain(),
-                    credentials.getUniverseDomain())),
-            GrpcStatusCode.of(Status.Code.PERMISSION_DENIED),
-            false);
-      }
+      endpointContext.validateUniverseDomain(getCredentials(), UNAUTHENTICATED_STATUS_CODE);
     } catch (IOException e) {
-      // Check if it is an Auth Exception
+      // Check if it is an Auth Exception (All instances of IOException from endpointContext's
+      // `validateUniverseDomain()` call should be an Auth Exception).
       if (e instanceof Retryable) {
         Retryable retryable = (Retryable) e;
         // Keep the behavior the same as gRPC-Java. Mark as Auth Exceptions as Unavailable
         throw ApiExceptionFactory.createException(
-            new Throwable("Unable to retrieve the Universe Domain from the Credentials."),
+            "Unable to retrieve the Universe Domain from the Credentials. Retrying the attempt.",
+            e,
             GrpcStatusCode.of(Status.Code.UNAVAILABLE),
             retryable.isRetryable());
       }
-      // Return the exception back to the user
-      throw new RuntimeException(e);
+      // This exception below should never be raised as all IOExceptions should be caught above.
+      throw ApiExceptionFactory.createException(
+          EndpointContext.UNABLE_TO_RETRIEVE_CREDENTIALS_ERROR_MESSAGE,
+          e,
+          UNAUTHENTICATED_STATUS_CODE,
+          false);
     }
   }
 
--- a/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
+++ b/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
@@ -30,8 +30,11 @@
 package com.google.api.gax.grpc;
 
 import com.google.api.core.BetaApi;
+import com.google.api.core.InternalApi;
 import com.google.api.gax.retrying.RetrySettings;
 import com.google.api.gax.rpc.ApiCallContext;
+import com.google.api.gax.rpc.ApiExceptionFactory;
+import com.google.api.gax.rpc.EndpointContext;
 import com.google.api.gax.rpc.StatusCode;
 import com.google.api.gax.rpc.TransportChannel;
 import com.google.api.gax.rpc.internal.ApiCallContextOptions;
@@ -39,6 +42,7 @@ import com.google.api.gax.rpc.internal.Headers;
 import com.google.api.gax.tracing.ApiTracer;
 import com.google.api.gax.tracing.BaseApiTracer;
 import com.google.auth.Credentials;
+import com.google.auth.Retryable;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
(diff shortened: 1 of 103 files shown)


gax-java/gax/src/main/java/com/google/api/gax/rpc/ApiCallContext.java
```diff
@@ -266,4 +272,21 @@ public static <T> Key<T> create(String name) {
...
+   * invalid
+   *
+   * <p>For gRPC, if the client library is unable to retrieve the Universe Domain, the client
+   * library will attempt to retry with the defined retry bounds. If the retry bounds have been
+   * exceeded, the client library will return an Unavailable exception back.
+   *
+   * <p>For HttpJson, if the client library is unable to retrieve the Universe Domain, the client
+   * library will fail and return an exception.
+   */
+  void validateUniverseDomain();
```
lqiu96, line 291: Possible future enhancement: Both transports should have the same implementation for validating the universe domain. Migrate the transport specific validation code here and provide a default implementation in ApiCallContext (transport-agnostic).

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -95,6 +97,56 @@ public static Builder newBuilder() {
...
+    MtlsProvider mtlsProvider = input.mtlsProvider();
+    if (mtlsEndpoint != null) {
+      builder.setMtlsProvider(mtlsProvider);
+    }
+    builder.setUsingGDCH(input.usingGDCH());
+    try {
+      return builder.build();
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
```
lqiu96, line 147: Throw a RuntimeException so the checked IOException does not get bubbled up to calling methods.

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/HttpJsonDirectServerStreamingCallableTest.java
```diff
@@ -200,9 +206,9 @@ public void testServerStreamingStart() throws InterruptedException {
 
     Truth.assertThat(moneyObserver.controller).isNotNull();
     // wait for the task to complete, otherwise it may interfere with other tests, since they share
-    // the same MockService and unfinished request in this test may start reading messages
-    // designated for other tests.
-    Truth.assertThat(latch.await(60, TimeUnit.SECONDS)).isTrue();
+    // the same MockService and unfinished request in this tes may start readind messages designated
```
burkedavison, line 209:
```suggestion
    // the same MockService and unfinished request in this test may start reading messages designated
```

gax-java/gax/src/main/java/com/google/api/gax/rpc/ApiCallContext.java
```diff
@@ -266,4 +272,21 @@ public static <T> Key<T> create(String name) {
...
+   *
+   * <p>For both transports, a PermissionDenied exception will be thrown if the universe domain is
+   * invalid
+   *
+   * <p>For gRPC, if the client library is unable to retrieve the Universe Domain, the client
+   * library will attempt to retry with the defined retry bounds. If the retry bounds have been
+   * exceeded, the client library will return an Unavailable exception back.
+   *
+   * <p>For HttpJson, if the client library is unable to retrieve the Universe Domain, the client
+   * library will fail and return an exception.
```
burkedavison, line 289: These descriptions are good at defining the behavior, but please use the `@throws` Javadoc tag to define the methods' exception behavior, being explicit about the type of exception being thrown in each case.

Also, please be precise with language w.r.t. `throws an exception` vs `returns an exception`

lqiu96: Thanks, will update!

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -95,6 +111,56 @@ public static Builder newBuilder() {
...
+      // GDC-H has no universe domain, always return true.
+      return true;
+    } else if (credentials == null) {
+      // If passed with NoCredentialsProvider, match with GDU
+      return resolvedUniverseDomain().equals(Credentials.GOOGLE_DEFAULT_UNIVERSE);
+    }
+    return resolvedUniverseDomain().equals(credentials.getUniverseDomain());
+  }
+
+  public EndpointContext merge(EndpointContext input) {
```
blakeli0, line 126: I'm not sure we need a default `EndpointContext` or this `merge` method. The main reason we need one for `ApiCallContext` is that, comparing to the default `ApiCallContext` created from `ClientContext`, there could be different value for some settings(e.g. RetrySettings) for different RPCs. But I don't think we would have different    `EndpointContext` for different RPCs.

gax-java/gax/src/main/java/com/google/api/gax/rpc/ApiCallContext.java
blakeli0: Agree with the enhancement. I guess the reason we don't do it now is we don't have the getters for `credentials` in `ApiCallContext`?

lqiu96: The reason it's not done now is because gRPC and HttpJson have minor behavior differences regarding how to retry the Auth Exception (marking as Unavailable and retrying vs throwing an exception). In a future enhancement, gRPC and HttpJson should have the same behavior w.r.t. dealing with access token + universe domain failures.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
```diff
@@ -530,10 +605,42 @@ public <T> T getOption(Key<T> key) {
...
+            GrpcStatusCode.of(Status.Code.PERMISSION_DENIED),
+            false);
+      }
+    } catch (IOException e) {
+      // Check if it is an Auth Exception
+      if (e instanceof Retryable) {
+        Retryable retryable = (Retryable) e;
+        // Keep the behavior the same as gRPC-Java. Mark as Auth Exceptions as Unavailable
+        throw ApiExceptionFactory.createException(
+            new Throwable("Unable to retrieve the Universe Domain from the Credentials."),
```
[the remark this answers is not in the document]
lqiu96: I think it's possible to pass the IOException/ Retryable back, but may have some additional side effects. I believe IOException is a checked exception, which would require the retry framework to catch this specific exception. The retry framework also needs add an additional if/ else block to check for Retryables and checking for that also needs to be transport specific (as grpc and httpjson have slightly different behaviors for now).

The ApiException is a RuntimeException which is later caught in the retry framework.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
```diff
@@ -530,10 +605,42 @@ public <T> T getOption(Key<T> key) {
...
+            GrpcStatusCode.of(Status.Code.UNAVAILABLE),
+            retryable.isRetryable());
+      }
+      // Return the exception back to the user
+      throw new RuntimeException(e);
```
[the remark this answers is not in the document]
lqiu96: Makes sense about logging it. I probably need to change the method signature of `validateUniverseDomain()` as it currently prevents the call from executing by throwing an exception.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
```diff
@@ -530,10 +605,42 @@ public <T> T getOption(Key<T> key) {
     return options.getOption(key);
   }
 
+  /** {@inheritDoc} */
+  @Override
+  public void validateUniverseDomain() {
+    EndpointContext endpointContext = getEndpointContext();
+    Credentials credentials = getCredentials();
+    try {
+      if (!endpointContext.hasValidUniverseDomain(credentials)) {
```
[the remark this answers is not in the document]
lqiu96: I split the logic into two methods since I had a few issues
1. gRPC and HttpJson have slightly different behaviors with handling Auth Exception retries. Putting them in the EndpointContext would have them share the same behavior. 
2. ApiExceptionFactory requires a StatusCode which is in each transport-specific module (https://github.com/googleapis/sdk-platform-java/blob/ed16261ac8c90fe588026e30d62d948e7ca1af13/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcStatusCode.java#L40). It's inaccessible from the transport-neutral module.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
```diff
@@ -577,4 +687,18 @@ Metadata getMetadata() {
...
+  public EndpointContext getEndpointContext() {
+    return endpointContext;
+  }
+
+  // gRPC-Java uses CallCredentials (not Google Auth Library's Credentials). gRPC-Java doesn't
+  // expose a way to access the original Credentials from CallCredentials. For gRPC, the client
+  // library stores both the CallCredentials and Credentials object. This is so the client library
+  // can access any additional fields from the Credentials' class (i.e. Universe Domain).
+  @InternalApi
+  Credentials getCredentials() {
```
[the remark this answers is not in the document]
lqiu96: If validation can be done in endpointcontext, then this isn't needed. I had a few issue with doing that in the comment above.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
```diff
@@ -530,10 +605,42 @@ public <T> T getOption(Key<T> key) {
...
+    Credentials credentials = getCredentials();
+    try {
+      if (!endpointContext.hasValidUniverseDomain(credentials)) {
+        throw ApiExceptionFactory.createException(
+            new Throwable(
+                String.format(
+                    EndpointContext.INVALID_UNIVERSE_DOMAIN_ERROR_TEMPLATE,
+                    endpointContext.resolvedUniverseDomain(),
+                    credentials.getUniverseDomain())),
+            GrpcStatusCode.of(Status.Code.PERMISSION_DENIED),
```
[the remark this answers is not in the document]
lqiu96: I'll follow up with this. I think this might be a Java specific issue.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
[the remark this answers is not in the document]
blakeli0: I meant pass `retryable` instead of `new Throwable()` to the ApiExceptionFactory, which should still fit into the retry framework.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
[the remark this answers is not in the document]
blakeli0: 1. I think the different behavior is in the catch block, where we want to rethrow a retryable exception, but in the try block, it should be the same?
2. Yeah it's unfortunate that `StatusCode` is an interface, we can get around it by creating an inner class that overrides `getCode()` and returns `PERMISSION_DENIED`.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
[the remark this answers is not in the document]
lqiu96: I see. We have guidance regarding adding the error message so I wanted to include that text.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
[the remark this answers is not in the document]
lqiu96: > I think the different behavior is in the catch block, where we want to rethrow a retryable exception, but in the try block, it should be the same?

I might be misunderstanding what you mean. We check the retryable exception to see if it should be retried. The Retryable is bubbled up from the `endpointContext.hasValidUniverseDomain(...)` call in the try block. What do you mean by they should be the same?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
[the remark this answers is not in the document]
blakeli0: > The Retryable only contains a boolean field so I'm not sure it adds too much value.

The value to create an exception with the cause is that we don't lose the stacktrace from the original cause.
I think we can create an ApiException with both `cause` and `message` by using [this factory method](https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/gax/src/main/java/com/google/api/gax/rpc/ApiExceptionFactory.java#L44).

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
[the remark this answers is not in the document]
blakeli0: I see, if stopping the call from executing is required, then re-throwing makes sense. We can re-throw it as `UNAVAILABLE` as well, the only difference is that `retryable` would always be set to false if the cause is not a `Retryable` exception.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
```diff
@@ -530,10 +612,50 @@ public <T> T getOption(Key<T> key) {
...
+    try {
+      endpointContext.validateUniverseDomain(getCredentials(), UNAUTHENTICATED_STATUS_CODE);
+    } catch (IOException e) {
+      // Check if it is an Auth Exception (All instances of IOException from endpointContext's
+      // `validateUniverseDomain()` call should be an Auth Exception).
+      if (e instanceof Retryable) {
+        Retryable retryable = (Retryable) e;
+        // Keep the behavior the same as gRPC-Java. Mark as Auth Exceptions as Unavailable
+        throw ApiExceptionFactory.createException(
+            "Unable to retrieve the Universe Domain from the Credentials. Retrying the attempt.",
```
blakeli0, line 640: I don't think we want to mention `Retrying the attempt` as `retryable.isRetryable()` is not guaranteed to be true, and `UNAVAILABLE` may not be configured to be retryable by customers. I think this exception and the one below can share the same message, the only difference is the value of `retryable`.

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/HttpJsonClientCallsTest.java
```diff
@@ -0,0 +1,145 @@
+/*
+ * Copyright 2018 Google LLC
```
blakeli0, line 2:
```suggestion
 * Copyright 2023 Google LLC
```

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -42,6 +42,10 @@
 @InternalApi
 @AutoValue
 public abstract class EndpointContext {
+  public static final String INVALID_UNIVERSE_DOMAIN_ERROR_TEMPLATE =
```
blakeli0, line 45: Can these two templates be private or package private?

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -83,7 +87,7 @@ public abstract class EndpointContext {
 
   public abstract boolean usingGDCH();
 
-  abstract String resolvedUniverseDomain();
+  public abstract String resolvedUniverseDomain();
```
blakeli0, line 90: I think this can stay as package private?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
lqiu96: Makes sense. I used the same message for both and updated the message to be:
`"Unable to retrieve the Universe Domain from the Credentials."`

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
lqiu96: I made `INVALID_UNIVERSE_DOMAIN_ERROR_TEMPLATE` private, but `UNABLE_TO_RETRIEVE_CREDENTIALS_ERROR_MESSAGE` needs to be public as it's accessed from the gRPC and HttpJson sub-modules.

gax-java/gax-grpc/src/test/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProviderTest.java
```diff
@@ -284,14 +284,6 @@ public void testDirectPathXdsDisableByDefault() throws IOException {
     assertThat(provider.isDirectPathXdsEnabled()).isFalse();
   }
 
-  @Test
```
blakeli0, line 287: This is a new test added 4 days. Are we suppose to remove it?

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonCallContext.java
```diff
@@ -201,6 +213,10 @@ public HttpJsonCallContext merge(ApiCallContext inputCallContext) {
       newRetryableCodes = this.retryableCodes;
     }
 
+    // The EndpointContext is not updated as there should be no reason for a user
```
blakeli0, line 216: This a great explanation of `EndpointContext` itself, can you please move it to `EndpointContext` class?

blakeli0 (approved):

gax-java/gax-grpc/src/test/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProviderTest.java
lqiu96: Not sure why this was removed (possibly some mistake when merging). Adding it back.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonCallContext.java
lqiu96: Sure, will do.

PR #2331, opened 2024-01-02
feat: Add Universe Domain to ClientSettings

lqiu96:


gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientSettings.java
```diff
@@ -241,6 +246,12 @@ public B setClock(ApiClock clock) {
       return self();
     }
 
+    /** Sets the Universe Domain to configure the resolved endpoint */
+    public B setUniverseDomain(String universeDomain) {
```
blakeli0, line 250: This is probably where showcase tests can help us find issues earlier. Can you please add a few showcase tests to verify that universe domain set through `ClientSettings` and `StubSettings` can be used for resolving endpoints?

gax-java/gax/src/main/java/com/google/api/gax/rpc/StubSettings.java
```diff
@@ -544,6 +544,11 @@ public ApiClock getClock() {
       return clock;
     }
 
+    /** Gets the custom Universe Domain that was previously set on this Builder */
+    public String getUniverseDomain() {
```
blakeli0, line 548: This is for best practices but not actually used?

lqiu96: This was to keep consistency with the other getters in the Builders. I don't think there is much use for this getter inside the Builder, but I didn't want to be inconstant with the fields for public facing classes (in case there is a user that has a use case for getting the values in the Builder).

I was on the fence about adding it in. If you think we shouldn't add it, I'm ok to remove it as well.

lqiu96: Removed this.

gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientSettings.java
lqiu96: We've decided to park this. I'll create an issue to investigate aligning the stubsettings and clientsettings.

blakeli0 (approved):

Issue #2327, opened 2024-01-02
Verify update of cloudbuild-*.yaml files in next release (after 2.31.0)


mpeddada1:
The cloudbuild yaml files are set up to be updated by release-please using a combination of inline annotations ([example](https://github.com/googleapis/sdk-platform-java/blob/f499ced28a562cbb3ea49f14a4fa16eb6a8173cc/.cloudbuild/cloudbuild.yaml#L17)) and the extra-files settings in https://github.com/googleapis/sdk-platform-java/pull/2326.

Similar to the observations in https://github.com/googleapis/java-shared-config/issues/730, it is likely that these versions won't get automatically updated in the currently open [release PR](https://github.com/googleapis/sdk-platform-java/pull/2290) since the changes happened after the PR was opened. 

In the next SNAPSHOT and non SNAPSHOT release PR, we need to make sure that these versions are correctly updated

mpeddada1:
Closing issue. The `cloudbuild-*.yaml`  files have been getting updated (https://github.com/googleapis/sdk-platform-java/pull/2417/).

PR #2334, opened 2024-01-03
chore: one time manual docker images update for shared-deps 3.21.0 release

mpeddada1:
Preparing docker images for #2290. Similar to the observations in https://github.com/googleapis/java-shared-config/issues/730, release-please doesn't update these files despite the following configs being applied:
-  Inline annotation with `x-version-update:<artifact-name>:released` 
- Addition of `.cloudbuild/` yamls files to `extraFiles` parameter in `release-please.yaml` ([line](https://github.com/googleapis/sdk-platform-java/blob/99e51953f319c97b225a0209b093367fa440f9d3/.github/release-please.yml#L6))

These files were updated after the the release PR was opened. In the case of java-shared-config, the update to these extra files didn't happen until a new release PR was opened. Created a follow-up task to verify the automation in the next release #2327.

### DIFF
--- a/.cloudbuild/cloudbuild.yaml
+++ b/.cloudbuild/cloudbuild.yaml
@@ -14,7 +14,7 @@
 
 timeout: 7200s # 2 hours
 substitutions:
-  _SHARED_DEPENDENCIES_VERSION: '3.20.0' # {x-version-update:google-cloud-shared-dependencies:released}
+  _SHARED_DEPENDENCIES_VERSION: '3.21.0' # {x-version-update:google-cloud-shared-dependencies:released}
   _JAVA_SHARED_CONFIG_VERSION: '1.7.1'
 steps:
   # GraalVM A build
--- a/.cloudbuild/cloudbuild.yaml
+++ b/.cloudbuild/cloudbuild.yaml
@@ -14,7 +14,7 @@
 
 timeout: 7200s # 2 hours
 substitutions:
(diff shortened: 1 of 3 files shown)


.cloudbuild/cloudbuild.yaml
```diff
@@ -14,7 +14,7 @@
 
 timeout: 7200s # 2 hours
 substitutions:
-  _SHARED_DEPENDENCIES_VERSION: '3.20.0' # {x-version-update:google-cloud-shared-dependencies:released}
+  _SHARED_DEPENDENCIES_VERSION: '3.21.0' # {x-version-update:google-cloud-shared-dependencies:released}
```
burkedavison, line 17: Could you remind me why this one is `released` but the above two are `current`?

mpeddada1: The `cloudbuild.yaml` file is used for the Cloud Build trigger that actually publishes the docker images with every release whereas the `cloudbuild-test-*.yaml` files are mainly for verifying the docker images locally (therefore, they're also updated to SNAPSHOT). Since publishing the docker images for SNAPSHOT versions would be redundant, we only update `cloudbuild` when a release takes place.

mpeddada1: @burkedavison As discussed offline, modified the use of `released` to `current`. This will mitigate the risk of a manual trigger of the cloud build job accidentally overriding a published image.

burkedavison (approved):

PR #2337, opened 2024-01-04
test: Add a sample generate_config yaml file.

blakeli0:
Add a sample generation_config.yaml file for google-cloud-java. 

This is intended to be a sample file to make sure we agree on the high level design of the config file. This does not intend to be a exhaustive list of parameters at this moment, we will refine the parameters along the way.


library_generation/test/resources/google-cloud-java_config.yaml
```diff
@@ -0,0 +1,59 @@
...
+    #Optional.
+    rest_documentation:
+    #Optional.
+    rpc_documentation:
+    #Required.
+    services:
+        #Required. relative path to google3/third_party/googleapis/stable
+      - proto_path: google/cloud/asset/v1
+        #Required. Remove this parameter once https://github.com/googleapis/sdk-platform-java/issues/2050 is fixed.
+        proto_only: false
```
JoeWang1127, line 44: `proto_only` is parsed from BUILD.bazel, do you want to add this parameter here?

blakeli0: We don't need it if we can parse it, I thought we have to specify it. How do we parse it?

JoeWang1127: The logic is if we can find `java_gapic_library` target in BUILD, then set the value to `true`; otherwise set to `false`.

library_generation/test/resources/google-cloud-java_config.yaml
```diff
@@ -0,0 +1,61 @@
...
+#Required.
+python-version: 3.11.2
+#Optional. The root folder name of generated client libraries. If empty, modules will be created under current folder, useful for single module
+destination-path: google-cloud-java
+#Optional. We will create a new versions.txt file if it is not specified
+versions_path:
+#Required.
+apis:
+    #Required. The folder name of the module. This is the destination-name in new-client.py.
+  - name: java-asset
```
diegomarquezp, line 18: How are we planning to deal with owlbot files `owlbot.py` and .`OwlBot.yaml`? Are we just going to assume they are in `google-cloud-java/${name}`?

library_generation/test/resources/google-cloud-java_config.yaml
blakeli0: Sounds god. I'll remove it.

library_generation/test/resources/google-cloud-java_config.yaml
blakeli0: I think you can assume they exist for each api when the script make it to post-processing. If it does not exist, new-client.py or generate_library.sh should create it. cc: @JoeWang1127

library_generation/test/resources/google-cloud-java_config.yaml
```diff
@@ -0,0 +1,61 @@
...
+    #Optional.
+    product_documentation:
+    #Optional.
+    client_documentation:
+    #Optional.
+    rest_documentation:
+    #Optional.
+    rpc_documentation:
+    #Required.
+    services:
```
JoeWang1127, line 42: What happens if there's a new version defined in googleapis?
Do we want to dynamically get the versioned path or add a service in this file?

blakeli0: A new service will be added. Are new versions of APIs automatically updated now? I guess yes with proper owlbot.yaml config? if so, we may want to automatically update it with a separate process, or replace owlbot.yaml's functionality with this file. That's probably why @suztomo mentioned this idea in the meeting.

JoeWang1127: > Are new versions of APIs automatically updated now?

Yes, OwlBot cli copies all versions of the services using `.OwlBot.yaml`.

I'm thinking the wrapper file (the one reading the config file and generate libraries) can find versioned directory within `proto_path`.

`proto_path` in new-client.py is not a versioned directory, i.e., `google/cloud/alloydb`, while `proto_path` in generate_library.sh is a versioned directory, i.e., `google/cloud/alloydb/v1`.

blakeli0: I think it's better to explicit list them, and/or support a wildcard configuration like `google/cloud/asset/**`. Because some product like Bigtable needs to include `google/bigtable/admin/v2` and `google/bigtable/v2`, if we only supply `google/bigtable`, we will have to find all the services recursively and smartly, there might be some folders are excluded as well.

library_generation/test/resources/google-cloud-java_config.yaml
```diff
@@ -0,0 +1,52 @@
...
+#Required.
+python-version: 3.11.2
+#Optional. The root folder name of generated client libraries. If empty, modules will be created under current folder, useful for single module
+destination-path: google-cloud-java
+#Optional. This is a relative path to destination-path above. We will create a new versions.txt file if it is not specified
+versions_path: versions.txt
+#Required.
+products:
+    #Required. Can be used for populating the folder name java-{api_shortName}. This is also the destination-name in new-client.py.
+  - api_shortname: asset
```
JoeWang1127, line 18: Could you write an example of java-bigtable, which combines two clients: bigtable and bigtable-admin?

library_generation/test/resources/google-cloud-java_config.yaml
JoeWang1127: I think the wildcard/regex is fine. 
For example: `google/bigtable/v[1-9].*` and `google/bigtable/admin/v[1-9].*`

library_generation/test/resources/google-cloud-java_config.yaml
```diff
@@ -0,0 +1,52 @@
+#Required.
+gapic_generator_version: 2.30.0
```
JoeWang1127, line 2: What about protobuf and grpc versions?

library_generation/test/resources/google-cloud-java_config.yaml
```diff
@@ -0,0 +1,52 @@
...
+#Required.
+googleapis-commitish: 4512234113a18c1fda1fb0d0ceac8f4b4efe9801
+#Required.
+owlbot-cli-image: sha256:623647ee79ac605858d09e60c1382a716c125fb776f69301b72de1cd35d49409
+#Required.
+synthtool-commitish: 59fe44fde9866a26e7ee4e4450fd79f67f8cf599
+#Required.
+python-version: 3.11.2
+#Optional. The root folder name of generated client libraries. If empty, modules will be created under current folder, useful for single module
+destination-path: google-cloud-java
```
suztomo, line 12: Does this need to match repository name?

library_generation/test/resources/google-cloud-java_config.yaml
```diff
@@ -0,0 +1,52 @@
...
+#Optional. This is a relative path to destination-path above. We will create a new versions.txt file if it is not specified
+versions_path: versions.txt
```
suztomo, line 14: Is there any case this value is not versions.txt?

library_generation/test/resources/google-cloud-java_config.yaml
suztomo, line 1: Is there any mapping that defines that google-cloud-java_config.yaml is for https://github.com/googleapis/google-cloud-java ?

library_generation/test/resources/google-cloud-java_config.yaml
blakeli0: Good point, I think we can assume the location of versions.txt is always under the root folder of the repo? cc: @JoeWang1127 @diegomarquezp

diegomarquezp: Yes, it's always in the root folder.

library_generation/test/resources/google-cloud-java_config.yaml
blakeli0: Technically no, this is the local folder we are going to put generated files in, which could be different from the repo name. Theoretically and if possible, this config file and the whole hermetic build scripts should not care about anything regarding Github, there should be an automation script that calls hermetic build script, compare the result with remote branch, and create PRs if needed.

library_generation/test/resources/google-cloud-java_config.yaml
suztomo, line 1: Is there any mapping that defines what branches of the repositories would receive generated code?

blakeli0: Each branch would have its own config file, as long as the automation points to that branch, we should generate a new set of files accordingly.

library_generation/test/resources/google-cloud-java_config.yaml
```diff
@@ -0,0 +1,52 @@
...
+#Required.
+products:
```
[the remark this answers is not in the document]
blakeli0: I think that's acceptable for now, but we should remove it and have the whole repo generated with a single version of generator all the time once we switch to the new process.

library_generation/test/resources/google-cloud-java_config.yaml
blakeli0: No, as I mentioned in another [comment](https://github.com/googleapis/sdk-platform-java/pull/2337#discussion_r1449350919) as well, I don't think the config should have any knowledge about Github setup. Let me know if you feel differently.

JoeWang1127 (approved):
Let's merge the config file and refine it in later PRs as it's difficult to set an appropriate format at this moment.

library_generation/test/resources/google-cloud-java_config.yaml
blakeli0: Added.

library_generation/test/resources/google-cloud-java_config.yaml
blakeli0: Removed.

library_generation/test/resources/google-cloud-java_generation_config.yaml
```diff
@@ -0,0 +1,52 @@
...
+#Required.
+googleapis-commitish: 4512234113a18c1fda1fb0d0ceac8f4b4efe9801
+#Required.
+owlbot-cli-image: sha256:623647ee79ac605858d09e60c1382a716c125fb776f69301b72de1cd35d49409
+#Required.
+synthtool-commitish: 59fe44fde9866a26e7ee4e4450fd79f67f8cf599
+#Required.
+python-version: 3.11.2
+#Optional. The root folder name of generated client libraries. If empty, modules will be created under current folder, useful for single module
+destination-path: google-cloud-java
```
suztomo, line 16: > useful for single module

By reading the comment, I'm not sure when I want to set a value for `destination-path`. Is google-cloud-java repo or java-bigtable repo a single module?

library_generation/test/resources/google-cloud-java_generation_config.yaml
suztomo, line 1: > No, as I mentioned in another https://github.com/googleapis/sdk-platform-java/pull/2337#discussion_r1449350919 as well, I don't think the config should have any knowledge about Github setup. Let me know if you feel differently.

So in reality the real file is not going to be in sdk-platform-java, but the destination branch of the target repository. Can you write down the idea somwhere as source code comment?

library_generation/test/resources/google-cloud-java_generation_config.yaml
blakeli0: We can have detailed instruction later. I mentioned single module because if we set `destination-path` to `java-bigtable`, the code will be generated into `java-bigtable/java-bigtable`, which is not ideal.

library_generation/test/resources/google-cloud-java_generation_config.yaml
blakeli0: I think that is already clear in [Mike's high level design](https://docs.google.com/document/d/1HIKD848vzsNDnHZ6hPGoofmgYJRVjm4VJBeXpcJYKuk/edit?resourcekey=0-SD7fvILbkKj8myttYl73SQ&tab=t.0#heading=h.k3vfmlfnb0bc).

blakeli0:
@suztomo @JoeWang1127 @diegomarquezp, I'm merging this PR now. Please add additional comments regarding the parameters in [internal docs](https://docs.google.com/document/d/1LcUUJuH0xmTT40AcSSLPUWpUdcWEvkXEdnCla5wMTm0/edit?tab=t.0).

@diegomarquezp Please use this as a template and add more configs for libraries/GAPICs to the file, so that it can replace `proto_path_list.txt` in integration tests. Also, I'm expecting a separate config file for each handwritten library, looking at the content of `proto_path_list.txt`, we will need one for bigtable/logging/pubsub/storage. Supporting all handwritten libraries can be a separate PR, testing `java-bigtable` in this PR should be good enough.

library_generation/test/resources/google-cloud-java_generation_config.yaml
suztomo: Now they are connected. Thanks. I was confused by this file being in sdk-platform-java.

PR #2340, opened 2024-01-04
chore: Use Java Auth Library's GDU constant

lqiu96:


lqiu96:
Depends on: https://github.com/googleapis/sdk-platform-java/pull/2341 to bump in the `dependencies.properties`

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -200,7 +200,7 @@ private String determineEndpoint() throws IOException {
 
       // Check if mTLS is configured with non-GDU
       if (endpoint.equals(mtlsEndpoint())
-          && !resolvedUniverseDomain().equals(GOOGLE_DEFAULT_UNIVERSE)) {
+          && !resolvedUniverseDomain().equals(Credentials.GOOGLE_DEFAULT_UNIVERSE)) {
```
blakeli0, line 203: I might missed this in earlier PRs, since we only set resolvedUniverseDomain later, `resolvedUniverseDomain()` is always null here?

blakeli0: Never mind, `resolvedUniverseDomain` was already set in previous steps. It's still good to move `Credentials.GOOGLE_DEFAULT_UNIVERSE` to the left of the equals method though.

blakeli0 (approved):

PR #2347, opened 2024-01-05
feat: Mark `getDefaultEndpoint()` with @ObsoleteApi

lqiu96:
Client libraries no longer always use the default endpoint. The resolved endpoint will be determined by the EndpointContext and used by the client.

### DIFF
--- a/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceStubSettingsClassComposer.java
+++ b/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceStubSettingsClassComposer.java
@@ -18,6 +18,7 @@ import com.google.api.MonitoredResourceDescriptor;
 import com.google.api.core.ApiFunction;
 import com.google.api.core.ApiFuture;
 import com.google.api.core.BetaApi;
+import com.google.api.core.ObsoleteApi;
 import com.google.api.gax.batching.BatchingSettings;
 import com.google.api.gax.batching.FlowControlSettings;
 import com.google.api.gax.batching.FlowController.LimitExceededBehavior;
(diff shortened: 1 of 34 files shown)


gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceStubSettingsClassComposer.java
```diff
@@ -1188,6 +1189,13 @@ private List<MethodDefinition> createDefaultHelperAndGetterMethods(
             .setReturnType(returnType)
             .setName("getDefaultEndpoint")
             .setReturnExpr(ValueExpr.withValue(StringObjectValue.withValue(service.defaultHost())))
+            .setAnnotations(
+                ImmutableList.of(
+                    AnnotationNode.builder()
+                        .setType(FIXED_TYPESTORE.get(ObsoleteApi.class.getSimpleName()))
+                        .setDescription(
+                            "See https://github.com/googleapis/sdk-platform-java/issues/2346 for more information")
```
blakeli0, line 1197: Can we provide a brief information in the description first? So that customers don't have to click through the link when they hover over on the method? Something like "There is no concept of a default endpoint anymore with the introduction of universeDomain, please use ... instead". 

In addition, maybe we can add the annotation only until the alternative `getResolvedEndpoint()` is introduced later?

lqiu96: Sure, I'll can add a description and hold off on adding the annotation until a future date.

> until the alternative getResolvedEndpoint() is introduced

I don't think `getResolvedEndpoint()` is going to be a *100% true* alternative to `getDefaultEndpoint()`. Based on the previous configurations (prior to EndpointContext), the default endpoint may not be the resolved endpoint used by the client. We're no longer supporting a concept of a default endpoint so there really should not be an alternative to `getDefaultEndpoint()`.

lqiu96:
Removing the `getDefaultEndpoint()` from being set in this PR: https://github.com/googleapis/sdk-platform-java/pull/2348

mpeddada1 (approved):
LGTM but will let @blakeli0 give the final approval.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceStubSettingsClassComposer.java
blakeli0: I'm not a fan of having github links in comments, I think they should be used only if we have no other way. They could become invalid for a lot of reasons, repo access changes, Github URL format change etc., or maybe the source of truth is moved elsewhere. Can we add more info in the annotation description itself?

blakeli0 (approved):

PR #2348, opened 2024-01-05
feat: Remove setting the default endpoint in StubSettings

lqiu96:
### What does this do
1. This change removes the concept of a having a default endpoint. For GAPICs, the EndpointContext will resolve the endpoint. If there are no user configurations, the endpoint will resolve to `{serviceName}.googleapis.com`.
2. StubSettings' `getEndpoint()` will return the user set endpoint. If the user does not set an endpoint, the method will return the endpoint set by the service. 

### Showcase
Showcase Clients do not have a valid serviceName and cannot construct an endpoint via the EndpointContext without an explicit endpoint input. To run these locally, we must configure the endpoint explicitly to point to `localhost:7469` for gRPC and `http://localhost:7469` for httpjson.

### Google Cloud Libraries
For Google Cloud client libraries, the service name is stored in the EndpointContext and parsed from the proto or service yaml file.

BEGIN_COMMIT_OVERRIDE
feat: Do not set the default endpoint in StubSettings
feat: StubSettings' `getEndpoint()` will return the service's pre-configured endpoint if there are no user configurations
END_COMMIT_OVERRIDE

### DIFF
--- a/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/comment/SettingsCommentComposer.java
+++ b/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/comment/SettingsCommentComposer.java
@@ -52,6 +52,9 @@ public class SettingsCommentComposer {
   private static final String CLASS_HEADER_DEFAULTS_RETRIES_DESCRIPTION =
       "Retries are configured for idempotent methods but not for non-idempotent methods.";
 
+  public static final CommentStatement GET_ENDPOINT_COMMENT =
+      toSimpleComment(
+          "Returns the endpoint set by the user or the the service's default endpoint.");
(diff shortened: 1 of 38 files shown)


blakeli0:
Removing setting the default endpoint maybe breaking for customers, especially for some handwritten libraries like [Bigtable](https://github.com/search?q=repo%3Agoogleapis%2Fjava-bigtable%20getEndpoint&type=code), can you please verify? If yes, I don't think we want to change it until `getResolvedEndpoint()` is introduced publicly. Then we can probably move `EndpointContext` to `StubSettings` and do something like `setEndpoint(endpointContext.getResolvedEndpoint())`.

blakeli0 (approved):
LGTM. Before merging, please update the PR title and description to reflect the new changes, and change the type of the PR from `chore` to `feat`.

PR #2351, opened 2024-01-09
fix: adjust release-please configs for cloudbuild yaml updates

mpeddada1:
Following the `Update arbitrary yaml file` strategy is resulting in release-please making unexpected changes to the cloudbuild yaml files. See #2350.  

Modifying release-please configs to follow `Update arbitrary file` strategy combined with inline annotations as described in https://github.com/googleapis/release-please/blob/main/docs/customizing.md#updating-arbitrary-files: 
Screenshot 2024-01-09 at 1 19 08 PM

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -5,24 +5,7 @@
   "group-pull-request-title-pattern": "chore(${branch}): release ${version}",
   "packages": {
     ".": {
-      "extra-files": [
-        "WORKSPACE",
-        {
-          "type": "yaml",
-          "path": ".cloudbuild/cloudbuild.yaml",
-          "jsonpath": "$.substitutions._SHARED_DEPENDENCIES_VERSION"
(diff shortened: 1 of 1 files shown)


burkedavison:
Should there be associated `x-release-please-version` comments added to the cloudbuild files, as mentioned in the screenshot for this PR?

suztomo (approved):

mpeddada1:
> Should there be associated `x-release-please-version` comments added to the cloudbuild files, as mentioned in the screenshot for this PR?

Hmm yeah was thinking about this too. We have the `x-version-update` annotation which is slightly different from the ones listed in the readme but is the convention in most our repos including [java-shared-config](https://github.com/googleapis/java-shared-config) (which uses it to update the cloudbuild yaml files) so was curious to try this out as an initial attempt.

PR #2360, opened 2024-01-12
fix: Cancel the Timeout Task for HttpJson

lqiu96:
Fixes: https://github.com/googleapis/google-cloud-java/issues/10220

Currently, the executorService will wait for any previously submitted task to finish execution before being able to terminate. This PR attempts to fix the issue by cancelling the outstanding timeout task. If a response has been received prior to the timeout, the timeout task will be cancelled and the client should be able to terminate immediately afterwards.

This fixes an issue for Clients that have RPCs with a large timeout value (i.e. 10 min). The client would wait 10 minutes before being able to terminate completely.

## Local Testing
Running this with SNAPSHOT Gax version (2.41.1-SNAPSHOT) and latest Compute from Libraries-Bom v26.31.0
Logs:
```
}

	at com.google.api.client.http.HttpResponseException$Builder.build(HttpResponseException.java:293)
	at com.google.api.client.http.HttpRequest.execute(HttpRequest.java:1118)
	at com.google.api.gax.httpjson.HttpRequestRunnable.run(HttpRequestRunnable.java:115)
	... 6 more

Process finished with exit code 1
```

Running this with latest Gax version (fix not included) and latest Compute from Libraries-Bom v26.31.0:
```
  }
}

	at com.google.api.client.http.HttpResponseException$Builder.build(HttpResponseException.java:293)
	at com.google.api.client.http.HttpRequest.execute(HttpRequest.java:1118)
	at com.google.api.gax.httpjson.HttpRequestRunnable.run(HttpRequestRunnable.java:115)
	... 6 more
```

Missing the termination and exit code. It shows up after 10 minutes.


suztomo:
Would you try to express the problem https://github.com/googleapis/google-cloud-java/issues/10220 in a test? If it's difficult to write the test, would you describe the challenge in the pull request description?

lqiu96:
> Would you try to express the problem [googleapis/google-cloud-java#10220](https://github.com/googleapis/google-cloud-java/issues/10220) in a test? The test should fail before your fix.
> 
> If it's difficult to write the test, would you describe the challenge in the pull request description?

Will do, thanks! I think there should be a way to test this behavior.

lqiu96:
Added `ITClientShutdown` showcase tests to test that the client is terminating properly. I expect that the showcase tests will fail if it's using the HttpJsonClientCallsImpl in the main branch.

Showcase (Java 11): https://github.com/googleapis/sdk-platform-java/actions/runs/7560014731/job/20585113208?pr=2360
Showcase (Java 17): https://github.com/googleapis/sdk-platform-java/actions/runs/7560014731/job/20585114645?pr=2360

lqiu96:
Showcase (Java 11): https://github.com/googleapis/sdk-platform-java/actions/runs/7560111797/job/20585419373?pr=2360
Showcase (Java 17): https://github.com/googleapis/sdk-platform-java/actions/runs/7560111797/job/20585419851?pr=2360

I expect both these tests to fail and that `testHttpJson_rpcInvokedWithLargeTimeout_closeClientOnceResponseReceived` is the only test case that fails.

lqiu96:
Took a look at grpc-java's implementation and they keep track of the ScheduledFuture as well.

Interestingly, grpc-java does not interrupt if the timeout future if it is already running: https://github.com/grpc/grpc-java/blob/7f4c16e068eeb6c7049b7ab2f0420a72d2cc1fb2/core/src/main/java/io/grpc/internal/ClientCallImpl.java#L383

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITClientShutdown.java
```diff
@@ -0,0 +1,188 @@
...
+  public void testHttpJson_closeClient() throws Exception {
+    EchoClient httpjsonClient = TestClientInitializer.createHttpJsonEchoClient();
+    httpjsonClient.close();
+    // 500ms buffer time to properly terminate the client
+    httpjsonClient.awaitTermination(500, TimeUnit.MILLISECONDS);
+    Truth.assertThat(httpjsonClient.isShutdown()).isTrue();
+    Truth.assertThat(httpjsonClient.isTerminated()).isTrue();
+  }
+
+  // Test to ensure hte client can close + terminate after a quick RPC invocation
```
JoeWang1127, line 55: Is `hte` a typo?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITClientShutdown.java
```diff
@@ -0,0 +1,188 @@
...
+    grpcClient.echo(EchoRequest.newBuilder().setContent("Test").build());
+
+    grpcClient.close();
+    // 1s buffer time to properly terminate the client after RPC is invoked
+    grpcClient.awaitTermination(1, TimeUnit.SECONDS);
+    Truth.assertThat(grpcClient.isShutdown()).isTrue();
+    Truth.assertThat(grpcClient.isTerminated()).isTrue();
+  }
+
+  // Test to ensure hte client can close + terminate after a quick RPC invocation
```
JoeWang1127, line 69: Is `hte` a typo?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITClientShutdown.java
```diff
@@ -0,0 +1,188 @@
...
+    // the termination time should only take about 2s (time to receive a response) + time
+    // to close the client. Check that this takes less than 5s (2s request time + 3s
+    // buffer time).
+    long terminationTime = end - start;
+    Truth.assertThat(terminationTime).isLessThan(5000L);
+  }
+
+  // Loop until the client has terminated successfully. For tests that use this,
+  // try to ensure there is a timeout associated, otherwise this may run forever.
+  // Future enhancement: Use awaitility instead of busy waiting
```
JoeWang1127, line 181: Could you explain what presents us from this enhancement right now?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITClientShutdown.java
```diff
@@ -0,0 +1,188 @@
...
+    httpjsonClient.awaitTermination(1, TimeUnit.SECONDS);
+    Truth.assertThat(httpjsonClient.isShutdown()).isTrue();
+    Truth.assertThat(httpjsonClient.isTerminated()).isTrue();
+  }
+
+  // This test is to ensure that the client is able to close + terminate any resources
+  // once a response has been received. Set a max test duration of 15s to ensure that
+  // the test does not continue on forever.
+  @Test(timeout = 15000L)
+  public void testGrpc_rpcInvokedWithLargeTimeout_closeClientOnceResponseReceived()
```
JoeWang1127, line 87: I think this test and the following one, `testHttpJson_rpcInvokedWithLargeTimeout_closeClientOnceResponseReceived`, are similar test with different transport. Could you combine them in a parameterized test?

Also, what test will be used to test `grpc+rest`? Sorry I'm new to this area.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITClientShutdown.java
lqiu96: Yep, updating.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITClientShutdown.java
lqiu96: Yep, updating.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITClientShutdown.java
lqiu96: There is currently nothing actively blocking adding the Awaitility library. I think there are quite a few tests inside showcase and gax that could benefit from including it. I didn't include it in this PR b/c I thought it would make this scope of this PR much bigger.

I can create an issue to introduce Awaitility and using it can be done separately.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITClientShutdown.java
lqiu96: The tests are the same except for the transport. For the showcase tests, we've intentionally did not use parameterized tests for readability. IIRC, the reason being that we would be able to quickly read the test name and figure out what what is being tested. [ITIAM](https://github.com/googleapis/sdk-platform-java/blob/main/showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITIam.java) has the duplicated tests with the other difference being the transport. 

> Also, what test will be used to test grpc+rest? Sorry I'm new to this area.

I believe you're talking about the bazel rule input, right? `grpc+rest` tells the generator that generate grpc && httpjson stubs. You can't create a client with both transports at the same time (it's either gRPC or HttpJson).

JoeWang1127: > For the showcase tests, we've intentionally did not use parameterized tests for readability.

Maybe duplicate test cases are fine in this PR since refactoring tests is not the main purpose of this PR.

However, I suspect the gain of (possible?) readability outrun the downside of code duplication. Imagine you need to modify *all* test to accommodate a code change. 

> I believe you're talking about the bazel rule input, right? grpc+rest tells the generator that generate grpc && httpjson stubs. You can't create a client with both transports at the same time (it's either gRPC or HttpJson).

Got it, makes sense.

JoeWang1127 (dismissed):

JoeWang1127:
Sorry I missed one thing. According to [development guide](https://github.com/googleapis/sdk-platform-java/blob/main/DEVELOPMENT.md?plain=1#L52), changes to `gax` *must* include an unit test.

Could you add one or specify the challenge?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITClientShutdown.java
lqiu96: > However, I suspect the gain of (possible?) readability outrun the downside of code duplication. Imagine you need to modify all tests to accommodate a code change.

I understand the concern of having to make the same code change multiple times and ensuring that the changes all work. It was something brought up a while back when the showcase tests were first implemented and the preference was for readability and to not use parameterized tests.

lqiu96:
> Sorry I missed one thing. According to [development guide](https://github.com/googleapis/sdk-platform-java/blob/main/DEVELOPMENT.md?plain=1#L52), changes to `gax` _must_ include an unit test.
> 
> Could you add one or specify the challenge?

The challenge with unit tests for `HttpJsonClientCallsImpl` is that there are a ton of moving parts inside this class. Multiple executors, transports, and some call options that all need to be mocked and tuned to be work for the tests. I think changes inside this class would be better suited for an IT/ E2E using showcase to confirm that behavior is correct.

JoeWang1127:
> The challenge with unit tests for HttpJsonClientCallsImpl is that there are a ton of moving parts inside this class. Multiple executors, transports, and some call options that all need to be mocked and tuned to be work for the tests. I don't think it's impossible, but it seems very challenging and I don't think would provide coverage that Showcase ITs can't do. I think changes inside this class would be better suited for an IT/ E2E using showcase to confirm that behavior is correct (i.e. close() terminates the outstanding tasks which impacts client termination, deliver() is able to parse message and return results).

To me, the lack of unit test seems reasonable. Could you add it to the PR description?

JoeWang1127 (approved):
Explained why we don't need a unit test for the change. LGTM.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITClientShutdown.java
JoeWang1127: Do you think it makes sense to document the rational in https://github.com/googleapis/sdk-platform-java/blob/main/DEVELOPMENT.md#testing-guide in another PR (or create an issue)?

lqiu96: Personally, I don't have a strong opinion regarding this since I see the pros/cons of both. I am currently leaning to just keep as-is and follow the precedence set in existing tests. If future tests in files are better suited one way, then we can make a case for it then.

lqiu96:
> > The challenge with unit tests for HttpJsonClientCallsImpl is that there are a ton of moving parts inside this class. Multiple executors, transports, and some call options that all need to be mocked and tuned to be work for the tests. I don't think it's impossible, but it seems very challenging and I don't think would provide coverage that Showcase ITs can't do. I think changes inside this class would be better suited for an IT/ E2E using showcase to confirm that behavior is correct (i.e. close() terminates the outstanding tasks which impacts client termination, deliver() is able to parse message and return results).
> 
> To me, the lack of unit test seems reasonable. Could you add it to the PR description?

I'm looking through the code in grpc-java with ClientCallImpl and I think they are able to add some valid unit test inside. I'm going to try and add some similar cases

suztomo:
It seems that you were able to reproduce the problem in https://github.com/googleapis/google-cloud-java/issues/10220. In the pull request description, would you add observation where you confirmed the problem of java-compute is fixed?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITClientShutdown.java
```diff
@@ -0,0 +1,188 @@
...
+
+public class ITClientShutdown {
+
+  // Test to ensure the client can close + terminate properly
+  @Test
+  public void testGrpc_closeClient() throws Exception {
+    EchoClient grpcClient = TestClientInitializer.createGrpcEchoClient();
+    grpcClient.close();
+    // 500ms buffer time to properly terminate the client
+    grpcClient.awaitTermination(500, TimeUnit.MILLISECONDS);
```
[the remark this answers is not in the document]
suztomo: A good way to confirm flakiness is to run this test while running the google-cloud-java's "mvn clean install -T 4C" build in background.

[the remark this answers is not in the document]
burkedavison: > -T 4C

I like.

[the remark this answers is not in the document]
lqiu96: Ran this with 500 invocations locally while building google-cloud-java in the background using the command above. It did not fail the test. I will increase the timeout to try and prevent any future flakes.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonClientCallImpl.java
```diff
@@ -121,6 +122,12 @@ final class HttpJsonClientCallImpl<RequestT, ResponseT>
   @GuardedBy("lock")
   private volatile boolean closed;
 
+  // Store the timeout future created by the deadline schedule executor. The future
+  // can be cancelled if a response (either an error or valid payload) has been
+  // received before the timeout.
```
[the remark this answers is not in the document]
lqiu96: Will add, thanks.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITClientShutdown.java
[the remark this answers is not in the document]
suztomo: Thank you.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITClientShutdown.java
lqiu96: I don't think policy is set in stone for the showcase tests and I can create an issue to re-visit changing the tests to incorporate parameterized tests. It's been a while since the initial showcase tests were written and I think (?) there was a discussion a while back regarding using parameterized tests or duplicating tests for the transports.

suztomo (approved):
LGTM

burkedavison (approved):

PR #2361, opened 2024-01-12
fix: format method types and table in Client Overview

alicejli:

### DIFF
--- a/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/comment/ServiceClientCommentComposer.java
+++ b/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/comment/ServiceClientCommentComposer.java
@@ -31,6 +31,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
 public class ServiceClientCommentComposer {
   // Tokens.
@@ -231,15 +232,17 @@ public class ServiceClientCommentComposer {
     StringBuilder tableBuilder = new StringBuilder();
     tableBuilder
         .append("<table>\n")
+        .append("   <caption>Methods</caption>\n")
         .append("   <tr>\n")
         .append("     <th>Method</th>\n")
         .append("     <th>Description</th>\n")
-        .append("     <th>Method Variants</th>\n");
+        .append("     <th>Method Variants</th>\n")
+        .append("   </tr>\n");
     for (MethodAndVariants method : methodAndVariantsList) {
       tableBuilder
           .append("   <tr>\n")
           .append("     <td>")
-          .append(method.method)
+          .append(CommentFormatter.formatAsJavaDocComment(method.method, null))
           .append("</td>\n")
           .append("     <td>")
           .append(CommentFormatter.formatAsJavaDocComment(method.description, null))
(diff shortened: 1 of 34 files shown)


gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/comment/ServiceClientCommentComposer.java
```diff
@@ -231,15 +232,17 @@ private static String createTableOfMethods(List<MethodAndVariants> methodAndVari
...
         .append("     <th>Description</th>\n")
-        .append("     <th>Method Variants</th>\n");
+        .append("     <th>Method Variants</th>\n")
+        .append("   </tr>\n");
     for (MethodAndVariants method : methodAndVariantsList) {
       tableBuilder
           .append("   <tr>\n")
           .append("     <td>")
-          .append(method.method)
+          .append(CommentFormatter.formatAsJavaDocComment(method.method, null))
```
blakeli0, line 245: Let's see if we can refactor `CommentFormatter` so that it can be used for the whole table, otherwise this issue may happen again in the future. That being said, if it turns out to be more complicated that it supposed to, we can merge this for now and create another issue to track the refactoring effort.

suztomo (commented):
About the prevention of future problems, would you implement a test to verify the generated javadoc are good with regard to the Maven command in Java 17?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/comment/ServiceClientCommentComposer.java
alicejli: I investigated this a bit and I think it would be a little tricky to refactor `CommentFormatter` to be able to be applied to a whole table as I think it would need to have changes (or additions) to `JavadocComment` to only escape certain characters if they're not part of an html tag. 

I could rewrite the table in general to use `JavadocComment` instead of `StringBuilder` so at least it would be explicitly clear when we are *not* applying standard formatting (e.g. escaping HTML characters). Let me know if you think that's worth doing or if I should create an issue for the larger refactoring effort.

alicejli:
> About the prevention of future problems, would you implement a test to verify the generated javadoc are good with regard to the Maven command in Java 17?

Are you thinking something like the downstream tests we configured in java-shared-config: https://github.com/googleapis/java-shared-config/pull/662/files and applied to sdk-platform-java?

suztomo:
I didn't think of it. But it solves my ask.

alicejli:
> I didn't think of the downstream testing specifically. Yes, a check similar to the downstream check will address my ask. There are various ways to solve that. I'm asking "Next time when somebody breaks the javadoc generation logic, please block the pull request."

Cool. I created an issue https://github.com/googleapis/sdk-platform-java/issues/2380 to track that work as separate from this PR.

blakeli0 (approved):

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/comment/ServiceClientCommentComposer.java
alicejli: Tracking issue for refactor: https://github.com/googleapis/sdk-platform-java/issues/2385

PR #2371, opened 2024-01-17
feat: add a RetryInfoAlgorithm which uses server side retry delay

mutianf:


gax-java/gax/src/main/java/com/google/api/gax/rpc/ApiException.java
```diff
@@ -60,6 +60,18 @@ public ApiException(
     this.errorDetails = errorDetails;
   }
 
+  public ApiException(
```
blakeli0, line 63: Is this used?

mutianf: Yeah, it'll be used in bigtable client. Do you want me to move it to a separate pr?

gax-java/gax/src/main/java/com/google/api/gax/rpc/Callables.java
```diff
@@ -69,12 +79,21 @@ public static <RequestT, ResponseT> UnaryCallable<RequestT, ResponseT> retrying(
...
             new ExponentialRetryAlgorithm(settings.getRetrySettings(), clientContext.getClock()));
-    ScheduledRetryingExecutor<ResponseT> retryingExecutor =
+
+    return retrying(innerCallable, retryAlgorithm, clientContext);
+  }
+
+  public static <RequestT, ResponseT> ServerStreamingCallable<RequestT, ResponseT> retrying(
+      ServerStreamingCallable<RequestT, ResponseT> innerCallable,
+      ServerStreamingCallSettings<RequestT, ResponseT> callSettings,
+      StreamingRetryAlgorithm<Void> retryAlgorithm,
```
blakeli0, line 91: I know `StreamingRetryAlgorithm<Void>` always have a `Void` type before, do you know if it's by design? Asking because I think this method is only being called directly from Bigtable, it would be better to change it to `ResponseT`, but it may have other implications.

gax-java/gax/src/main/java/com/google/api/gax/rpc/ApiException.java
blakeli0: Can you point me to how Bigtable would use it?

mutianf: Actually let me revert this and put in a separate PR.

gax-java/gax/src/main/java/com/google/api/gax/rpc/Callables.java
mutianf: I think it's because RetryingServerStreamingCallable uses a ScheduledRetryingExecutor<Void>. It doesn't expect a ApiFuture<responseT> in the executor.createFuture() like the unary calls.

gax-java/gax/src/main/java/com/google/api/gax/rpc/Callables.java
```diff
@@ -51,6 +51,16 @@ public class Callables {
...
 
+  public static <RequestT, ResponseT> UnaryCallable<RequestT, ResponseT> retrying(
+      UnaryCallable<RequestT, ResponseT> innerCallable,
+      RetryAlgorithm<ResponseT> retryAlgorithm,
+      ClientContext clientContext) {
+    ScheduledRetryingExecutor<ResponseT> retryingExecutor =
+        new ScheduledRetryingExecutor<>(retryAlgorithm, clientContext.getExecutor());
+    return new RetryingCallable<>(
+        clientContext.getDefaultCallContext(), innerCallable, retryingExecutor);
+  }
```
lqiu96, line 62: Do you have an example of how BigTable is planning on using this? IIUC, most of the methods here are used by the TransportCallableFactory to build the callables and I believe the CallableFactory is invoked from the TransportServiceStub (which is generated).

mutianf: I staged the example here: https://github.com/googleapis/java-bigtable/pull/2055/files#diff-66bb41fee1ea55e7ab391ed8827b46e57035b81982c587251d577028866fade0, in the EnhancedBigtableStub class.

blakeli0: Thanks @mutianf, for tracking purposes, do you mind creating an issue and link the example above to the issue?

blakeli0: > Is this change b/c BigTable wants to set custom retry values that are different from what's in the grpc service config file and you don't want customers to set their own?

I'll give it a shot but Mattie can probably provide more info. In short, yes they want to set retry values from the server responses dynamically instead of a fixed value from service configs. See [internal docs](https://docs.google.com/document/d/1wo0KaD5FMrknOoyjSxPD_NGQ2NNJMb0WPVO9SzhwYCU/edit?tab=t.0#heading=h.x9snb54sjlu9) for details.

mutianf: > Is this change b/c BigTable wants to set custom retry values that are different from what's in the grpc service config file and you don't want customers to set their own?

We want to use a different RetryAlgorithm to look at the response RetryInfo: https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/gaxx/retrying/RetryInfoRetryAlgorithm.java. There's another internal doc that's focused on this use case: https://docs.google.com/document/d/1Kim-80vUSDgYxDVrFtkl8NkV-SsW8KAor_Ca1rpvbyo/edit?resourcekey=0-oV2CmU7_sZEMpPMFF9ZSOA&tab=t.0#heading=h.j0ev05qqzdu4.


@blakeli0  sure I'll create an issue.

lqiu96: Thanks Mattie! I'll read through the doc for more information regarding this. I have a better idea of the use case and just need to fully confirm my understanding.

PR #2373, opened 2024-01-17
feat: Otel Implementation 

ddixit14:


gax-java/gax/src/test/java/com/google/api/gax/tracing/MetricsTracerTest.java
```diff
@@ -0,0 +1,243 @@
...
+    private MetricsTracer metricsTracer;
+    @Mock
+    private MetricsRecorder metricsRecorder;
+
+    @Before
+    public void setUp() {
+        metricsTracer = new MetricsTracer(SpanName.of("FakeClient", "FakeMethod"), metricsRecorder);
+    }
+
+    // Ask Blake - how can we verify record{attempt/operation}Latency methods? we would need a timer for that?
```
ddixit14, line 94: @blakeli0 how can we verify record{}latency functions ? we will install a timer in this test class and somehow match the time taken ?

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracer.java
```diff
@@ -0,0 +1,164 @@
...
+    // imo, ideally, I think this should put the status_attribute in the map, and attempt_count
+    // and attempt_latency should be calculated.
+
+    attributes.put(STATUS_ATTRIBUTE, extractStatus(error));
+    metricsRecorder.recordAttemptLatency(attemptTimer.elapsed(TimeUnit.MILLISECONDS), attributes);
+    metricsRecorder.recordAttemptCount(1, attributes);
+  }
+
+  @Override
+  public void attemptPermanentFailure(Throwable error) {
```
blakeli0, line 137: Do we know if `operationFailed` would be called after `attemptPermanentFailure` or `attemptFailedRetriesExhausted`? 
Similarly, we need to make sure `operationCancelled` is called after `attemptCancelled`. If not, we need to record operation latency and operation count in these methods.

ddixit14: Confirmed operationCancelled is called after attemptCancelled. Now testing the last two scenarios.

PR #2377, opened 2024-01-17
chore: add custom artifacts

JoeWang1127:
In this PR:
- Add regex to exclude customized artifacts in firestore and datastore.

Context: two artifacts (`com.google.cloud:proto-google-cloud-firestore-bundle-v1` and `com.google.cloud.datastore:datastore-v1-proto-client`) are found by unmanaged dependency check in firestore ([log](https://github.com/googleapis/java-firestore/actions/runs/7559858977/job/20584621623?pr=1532#step:5:1559)) and datastore ([log](https://github.com/googleapis/java-datastore/actions/runs/7560229868/job/20585789472?pr=1290#step:5:1563)), respectively. Extend the regex in the check to exclude these artifacts.


suztomo (approved):
You can combine these pom.xml files into one and name "google-internal-artifact-test-case-pom.xml". Next time we find a case, for testing, we just add few lines into the file.

PR #2403, opened 2024-01-19
feat: Introduce interfaces for metrics instrumentation

blakeli0:
This PR introduces new public interfaces and refactors existing interfaces for metrics instrumentation. All the interfaces are framework-agnostic. Subsequent PRs should implement interfaces with an observability framework. For example, introduce an `OpenTelemetryMetricsRecorder` that implements `MetricsRecorder`. All new methods are created with Javadocs, they are expected to be updated in the implementation PR if needed.

See go/java-gapic-otel-metrics-design for details.


gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
```diff
@@ -77,7 +77,9 @@ public final class GrpcCallContext implements ApiCallContext {
   private static final GrpcStatusCode UNAUTHENTICATED_STATUS_CODE =
       GrpcStatusCode.of(Status.Code.UNAUTHENTICATED);
 
-  static final CallOptions.Key<ApiTracer> TRACER_KEY = CallOptions.Key.create("gax.tracer");
+  // This field is made public for handwritten libraries to easily access the tracer from
+  // CallOptions
+  public static final CallOptions.Key<ApiTracer> TRACER_KEY = CallOptions.Key.create("gax.tracer");
```
lqiu96, line 82: qq, having a `TRACER_KEY` in the CallOptions this a gRPC specific thing? Would it make sense to also have this for HttpJson and if so, would it make sense to make this as transport neutral as possible?

gax-java/gax/src/main/java/com/google/api/gax/tracing/BaseApiTracer.java
```diff
@@ -33,7 +33,10 @@
 import org.threeten.bp.Duration;
 
 /**
- * A base implementation of {@link ApiTracer} that does nothing.
+ * A base implementation of {@link ApiTracer} that does nothing. With the deprecation of Java 7
+ * support, all the methods in {@link ApiTracer} are now made default, we no longer need a base
+ * class that does nothing. This class should be removed once all the references to it are removed
+ * in Google Cloud Client Libraries.
```
lqiu96, line 39: Perhaps we can create a ticket for this so we don't forget about it. @ddixit14 

If I'm understanding this correctly, once stage 1 is complete, we can remove this class.

gax-java/gax/src/main/java/com/google/api/gax/tracing/MethodName.java
```diff
@@ -0,0 +1,65 @@
...
+/** A value class to represent the name of the RPC method in an {@link ApiTracer}. */
+@BetaApi
+@InternalApi
+@AutoValue
+public abstract class MethodName {
+  /**
+   * Creates a new instance of the RPC method name.
+   *
+   * @param serviceName The name of the service. In general this will be GAPIC generated service
+   *     name {@link StubSettings#getServiceName()}. However, in some cases, when the GAPIC
```
lqiu96, line 46: @blakeli0 I am now having some second thoughts on the `serviceName()` getter's name. I think there might be some slight discrepancy between the intended result and the function name.

`getServiceName()` returns the name that was configured in the default_host. For example, asset's will return cloudasset: https://github.com/googleapis/google-cloud-java/blob/9785f8cfd43db0a8968f086a1461242da9d16cb5/java-asset/google-cloud-asset/src/main/java/com/google/cloud/asset/v1/stub/AssetServiceStubSettings.java#L866-L868

even though the intended result for this probably should be `asset`. 

Not something blocking this PR, but just wondering your thoughts on this.

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsRecorder.java
```diff
@@ -0,0 +1,56 @@
...
+
+import com.google.api.core.BetaApi;
+import com.google.api.core.InternalApi;
+import java.util.Map;
+
+/**
+ * Provides an interface for metrics recording. The implementer is expected to use an observability
+ * framework, e.g. OpenTelemetry
+ */
+@BetaApi
```
lqiu96, line 41: Do we have a timeline of when to remove this Beta annotation? Same for the other annotations above.

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsRecorder.java
```diff
@@ -0,0 +1,56 @@
...
+@InternalApi
+public interface MetricsRecorder {
+
+  /** TODO: Add Javadoc */
```
lqiu96, line 45: Would it make sense to add some default Javadocs now (I think it would apply to all implementations)? Or is this meant for us to fill out later?

gax-java/gax/src/main/java/com/google/api/gax/tracing/MethodName.java
ddixit14: @lqiu96 why should the intended name be `asset` ?  
@blakeli0 Is there any way we could test this? that for a particular service, this is the intented serviceName ?

gax-java/gax/src/main/java/com/google/api/gax/tracing/BaseApiTracer.java
ddixit14: sure, I'll create a ticket for this.

gax-java/gax/src/main/java/com/google/api/gax/tracing/MethodName.java
lqiu96: > @lqiu96 why should the intended name be `asset` ? @blakeli0 Is there any way we could test this? that for a particular service, this is the intented serviceName ?

I don't want to say what the intended name *should* be. I think it might be a bit weird for customers since we wouldn't have consistent service names.

java-asset's serviceName is cloudasset (link above)
java-biglake's serviceName is biglake (https://github.com/googleapis/google-cloud-java/blob/458516b462e0a4494f32815fe9d0e6f0b30353f1/java-biglake/google-cloud-biglake/src/main/java/com/google/cloud/bigquery/biglake/v1/stub/MetastoreServiceStubSettings.java#L409-L411)

I think most services would have the service name of java-{serviceName} except for asset and possibly a few others.

It might not be a big concern after all, but something just doesn't sit right with me knowing that serviceName is pulled from a URI (default_host value) and not the package name or some config value (name_pretty or api name).

lqiu96: FYI, your link doesn't work for me. Seems like there is no defined `serviceName` and we've somehow settled on the same definition 😆 

Not blocking this PR, but it would be great if there was an official definition for serviceName. Perhaps a question for the core team.

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsRecorder.java
blakeli0: I'm expecting them to be removed when we GA Otel Metrics, maybe later this year. The deliverables in Q1 is only for Spanner hence I consider them as a private preview feature.

gax-java/gax/src/main/java/com/google/api/gax/tracing/MethodName.java
blakeli0: @lqiu96 I think the term `serviceName` you introduced still makes sense, because we call the yaml and json config for each versioned protos [service yaml](https://github.com/googleapis/googleapis/blob/master/google/cloud/asset/v1/cloudasset_v1.yaml) and [service config](https://github.com/googleapis/googleapis/blob/master/google/cloud/asset/v1/cloudasset_grpc_service_config.json) respectively. we may want to add more docs to explain that it is used for constructing service endpoints.  
@ddixit14 I would call `java-{serviceName}` a library name or repo name, as it could includes multiple versioned services or admin services.

gax-java/gax/src/main/java/com/google/api/gax/tracing/BaseApiTracer.java
blakeli0: I'm planning to deprecate `BaseApiTracer` in a follow up PR, it is an `InternalApi` and only used by Bigtable outside  of our team. So we can could deprecate it anytime want, and remove it after Bigtable removes the usage of it.

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsRecorder.java
blakeli0: Added

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
blakeli0: Yes, it makes sense. Currently it is only used by [GrpcChannelUUIDInterceptor](https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcChannelUUIDInterceptor.java#L47), if we have a need to use it in [HttpJsonHeaderInterceptor](https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonHeaderInterceptor.java#L52), then we can add it to HttpJsonCallOptions. We can refactor it when we have the need in the future.

ddixit14 (approved):

gax-java/gax/src/main/java/com/google/api/gax/tracing/MethodName.java
lqiu96: Ok, to wrap up the conversation above: I'll add some additional docs clarifying the intended use case (constructing the endpoint) and how the this value is constructed (from the config files).

It is fine that serviceName in the library (pulled from `getServiceName()`) doesn't 100% match `java-{serviceName}`.

lqiu96 (approved):

gax-java/gax/src/main/java/com/google/api/gax/tracing/MethodName.java
blakeli0: > It is fine that serviceName in the library (pulled from `getServiceName()`) doesn't 100% match `java-{serviceName}`.

Correct. The library/repo name may not always be in the format of `java-{serviceName}` and I would not refer `java-{serviceName}` as the format for library/repo name.

PR #2404, opened 2024-01-20
chore: renovate bot config validation

ddixit14:


ddixit14:
Commit (https://github.com/googleapis/sdk-platform-java/pull/2404/commits/948d3934fbd7249fd122795fd9b56a596e08efd0) resulted in check error, as expected. Hence confirmed it is workin on this repository.
image

ddixit14:
Confirming behaviour for one of downstream repos:

check [run](https://github.com/googleapis/sdk-platform-java/actions/runs/7597983367/job/20693461395?pr=2404) on java-logging-logback outputs [this](https://paste.googleplex.com/6226434417229824) config, which matches with the [actual config](https://github.com/googleapis/java-logging-logback/blob/main/renovate.json). So it should be working for others as well (hopefully). If needed, I can confirm the behaviour in the failure case as well (modify google-cloud-java renovate incorrectly, run the check again, verify, revert the changes).

.github/workflows/ci.yaml
```diff
@@ -321,3 +321,20 @@ jobs:
       uses: ./java-shared-dependencies/unmanaged-dependency-check
       with:
         bom-path: gapic-generator-java-bom/pom.xml
+
+  renovate_config_check:
```
suztomo, line 325: Can you make this as independent workflow?

Then I think of 2 options:

- workflow dispatch to periodically validate the config files in ~13 repositories. It's installed in sdk-platform-java.
- workflow for pull requests that touches renovate.json. It's installed in every ~13 repositories via synthtool template.

What is the trade off?

.kokoro/presubmit/common.sh
```diff
@@ -87,3 +87,9 @@ function find_last_release_version {
   parts=($(echo "$primary_artifact" | tr ":" "\n"))
   echo "${parts[1]}"
 }
+
+function validate_renovate_config {
```
suztomo, line 91: I see the command is in the workflow YAML file. This seems unnecessary.

.github/workflows/ci.yaml
ddixit14: synthool is better, periodically validating makes no sense if we can run it upon every touch to renovate.json
Raised https://github.com/googleapis/synthtool/pull/1920

ddixit14: Trade off: 
1) Using manual workflow dispatch for periodic validation is not a good idea because 
   (a) Manual invocation required everytime 
   (b) If there is an error in the config, we would get to know about it only when we run this check manually.
2) Using synthtool, this check will be installed as an independent workflow in each repository, running on every commit to config file, hence catching errors as and when they are about to be committed. This way we can catch the error faster and better. 

Hence going with synthool.

Issue #2411, opened 2024-01-23
Cloud Bigtable directpath integration test failing 


sushanb:
Cloud Bigtable directpath integration test has been failing due to a null pointer exception in this code

https://github.com/googleapis/sdk-platform-java/blob/acf52c77add061966823644b908e1d9f27082328/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L333

PR #2412, opened 2024-01-23
fix: Verify Universe Domain's DirectPath Compatibility after Endpoint Resolution

lqiu96:
Fixes: https://github.com/googleapis/sdk-platform-java/issues/2411

### DIFF
--- a/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
+++ b/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
@@ -333,12 +333,7 @@ public final class InstantiatingGrpcChannelProvider implements TransportChannelP
   // Universe Domain configuration is currently only supported in the GDU
   @VisibleForTesting
   boolean canUseDirectPathWithUniverseDomain() {
-    if (endpoint.contains(Credentials.GOOGLE_DEFAULT_UNIVERSE)) {
-      return true;
-    }
-    // This is only logged if DirectPath is enabled
-    LOG.log(Level.WARNING, "DirectPath will only work in the the googleapis.com Universe Domain");
-    return false;
+    return endpoint.contains(Credentials.GOOGLE_DEFAULT_UNIVERSE);
   }
 
   @VisibleForTesting
--- a/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
+++ b/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
@@ -295,11 +295,6 @@ public final class InstantiatingGrpcChannelProvider implements TransportChannelP
               Level.WARNING,
               "DirectPath is misconfigured. DirectPath is only available in a GCE environment.");
         }
-        // Case 4: Universe Domain is non-GDU
-        if (!canUseDirectPathWithUniverseDomain()) {
-          LOG.log(
-              Level.WARNING, "DirectPath will only work in the the googleapis.com Universe Domain");
-        }
       }
     }
   }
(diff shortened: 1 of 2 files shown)


lqiu96:
Looking into possible showcase tests to verify client initialization with certain settings.

lqiu96:
> Looking into possible showcase tests to verify client initialization with certain settings.

I think the configurations with this would revolve around setting flags in `InstantiatingGrpcChannelProvider` which would be covered in `InstantiatingGrpcChannelProviderTest`. I think dedicated e2e tests would be better suited in the future.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -334,8 +335,16 @@ static boolean isOnComputeEngine() {
...
-  private boolean canUseDirectPathWithUniverseDomain() {
-    return endpoint.contains("googleapis.com");
+  // Universe Domain configuration is currently only supported in the GDU
+  @VisibleForTesting
+  boolean canUseDirectPathWithUniverseDomain() {
+    // ClientContext will set an endpoint if it isn't provided to the TransportChannelProvider
+    // If the endpoint is null, then the endpoint hasn't been fully resolved yet (client not
+    // initialized). Before a call goes through with DirectPath, this validation must succeed.
+    if (endpoint == null) {
+      return true;
```
blakeli0, line 345: This would fix the issue but this is a little misleading IMO. If endpoint is null and this method returns true, it does not mean we `canUseDirectPathWithUniverseDomain`, but rather we can not validate it.  
Can we not log direct path misconfigure on client initialization? And only adding `endpoint.contains(Credentials.GOOGLE_DEFAULT_UNIVERSE)` in [channel creation](https://github.com/googleapis/sdk-platform-java/blob/371a69adec72845b7b9f12fd055f8964352d82f5/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L384) where endpoint is guaranteed not null? This should be good enough to meet the [requirement](https://docs.google.com/document/d/1MonFBfu_gQzWiY0tcr_gCFKLse8O5t_cRPGfRhZgXWo/edit?tab=t.0#heading=h.ckpknxgackem) for TPC.

lqiu96: > This would fix the issue but this is a little misleading IMO. If endpoint is null and this method returns true, it does not mean we canUseDirectPathWithUniverseDomain, but rather we can not validate it.
Can we not log direct path misconfigure on client initialization?

Makes sense. I think I can just remove the call from `logDirectPathMisconfig()` instead of trying to force it in there. 

> And only adding endpoint.contains(Credentials.GOOGLE_DEFAULT_UNIVERSE) in channel creation where endpoint is guaranteed not null? This should be good enough to meet the requirement for TPC.

sg, will update.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -334,8 +330,15 @@ static boolean isOnComputeEngine() {
...
-  private boolean canUseDirectPathWithUniverseDomain() {
-    return endpoint.contains("googleapis.com");
+  // Universe Domain configuration is currently only supported in the GDU
+  @VisibleForTesting
+  boolean canUseDirectPathWithUniverseDomain() {
+    if (endpoint.contains(Credentials.GOOGLE_DEFAULT_UNIVERSE)) {
+      return true;
+    }
+    // This is only logged if DirectPath is enabled
+    LOG.log(Level.WARNING, "DirectPath will only work in the the googleapis.com Universe Domain");
```
lqiu96, line 340: Kept this log statement as it only triggers if these three conditions above are true: https://github.com/googleapis/sdk-platform-java/blob/909bdf997dcaa23ec91da93121a1ed4a5bae9949/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L372-L374

Might be beneficial since the user is has all the other DirectPath configurations correct except for Universe Domain.

blakeli0: It could be beneficial, but this check is called on every channel resizing, which by default is every minute, hence could cause spam logs to customers. That's why the `logDirectPathMisconfigure` was moved to the constructor initially, so I think we can skip this logging for now.

blakeli0 (approved):

PR #2421, opened 2024-01-25
feat: MetricsTracer implementation

ddixit14:
This PR adds framework-neutral `MetricsTracer` to Gax. This class adds logics during the start and end of an RPC call to calculate the latencies and count the number of RPC calls. Metric recording responsibilities are delegated to the `MetricsRecorder`, providing flexibility for diverse observability frameworks.
 Unit tests are added for every implemented method, including tests for `MetricsTracerFactory` (class responsible for creating one MetricsTacer per operation).


gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracer.java
```diff
@@ -42,7 +51,111 @@
...
+  public void operationFailed(Throwable error) {
+    attributes.put(STATUS_ATTRIBUTE, extractStatus(error));
+    metricsRecorder.recordOperationLatency(
+        operationTimer.elapsed(TimeUnit.MILLISECONDS), attributes);
+    metricsRecorder.recordOperationCount(1, attributes);
+  }
+
+  @Override
+  public void attemptStarted(int attemptNumber) {
+    // no-op
```
blakeli0, line 95: How is this `attemptStarted(int attemptNumber)` used? If we don't need it, then we can remove this.

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracer.java
```diff
@@ -42,7 +51,111 @@
...
+  @Override
+  public void attemptCancelled() {
+
+    attributes.put(STATUS_ATTRIBUTE, StatusCode.Code.CANCELLED.toString());
+    metricsRecorder.recordAttemptLatency(attemptTimer.elapsed(TimeUnit.MILLISECONDS), attributes);
+    metricsRecorder.recordAttemptCount(1, attributes);
+  }
+
+  @Override
+  public void attemptFailed(Throwable error, Duration delay) {
```
blakeli0, line 120: What is this argument `Duration delay`?

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracer.java
ddixit14: This is getting called in [CheckingAttemptCallable.java](https://github.com/googleapis/sdk-platform-java/blob/30a366710e6b1afe26df1d1d255fee4abd341575/gax-java/gax/src/main/java/com/google/api/gax/rpc/CheckingAttemptCallable.java#L81). I agree, we do not need it. Removed.

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracer.java
ddixit14: [This](https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/gax/src/main/java/com/google/api/gax/tracing/ApiTracer.java#L115) is the amount of time to wait before the next attempt will start.

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracer.java
ddixit14: I think we will need this, this might be the issue for my test with multiple attempts failing.

gax-java/gax/src/test/java/com/google/api/gax/tracing/MetricsTracerTest.java
```diff
@@ -0,0 +1,247 @@
...
+  @Test
+  public void testErrorConversion() {
+    for (Code code : Code.values()) {
+      ApiException error = new ApiException("fake error", null, new FakeStatusCode(code), false);
+      String errorCode = metricsTracer.extractStatus(error);
+      assertThat(errorCode).isEqualTo(code.toString());
+    }
+  }
+
+  // this test needs detailed attention
```
ddixit14, line 212: @blakeli0 @lqiu96 Can you guys confirm if this a valid scenario ?

gax-java/gax/src/test/java/com/google/api/gax/tracing/MetricsTracerTest.java
```diff
@@ -0,0 +1,247 @@
...
+    verify(metricsRecorder).recordAttemptCount(count, attributes);
+    verify(metricsRecorder).recordOperationCount(count, attributes);
+    verify(metricsRecorder).recordAttemptLatency(anyDouble(), eq(attributes));
+    verify(metricsRecorder).recordOperationLatency(anyDouble(), eq(attributes));
+
+    verifyNoMoreInteractions(metricsRecorder);
+  }
+
+  @Test
+  public void testErrorConversion() {
```
blakeli0, line 204: We need more test cases for `extractStatus` method.

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracer.java
```diff
@@ -42,12 +51,123 @@
...
+  @Override
+  public void attemptPermanentFailure(Throwable error) {
+
+    attributes.put(STATUS_ATTRIBUTE, extractStatus(error));
+    metricsRecorder.recordAttemptLatency(attemptTimer.elapsed(TimeUnit.MILLISECONDS), attributes);
+    metricsRecorder.recordAttemptCount(1, attributes);
+  }
+
+  /** {@inheritDoc} */
+  @InternalApi("Visible for testing")
```
blakeli0, line 148: There is a `@VisibleForTesting` annotation.

gax-java/gax/src/test/java/com/google/api/gax/tracing/MetricsTracerTest.java
blakeli0, line 1: I see a few anti-patterns in this test file. 
1. We should focus on testing one single method in unit tests. For example, not testing `attemptSucceeded` and `operationSucceeded` in the same test case.
2. Test names can be more meaningful. e.g. `testSuccessExample` does not tell us what is the method or scenario it is testing, see [unit test naming](go/unit-testing-practices#naming) for details.

gax-java/gax/src/test/java/com/google/api/gax/tracing/MetricsTracerTest.java
blakeli0: I'm not sure we need this scenario. Can you try to re-organize this test file as I suggested in other comments first?

gax-java/gax/src/test/java/com/google/api/gax/tracing/MetricsTracerTest.java
ddixit14: Understood. I've added a separate test for each method now with a more meaningful test name.

gax-java/gax/src/test/java/com/google/api/gax/tracing/MetricsTracerTest.java
ddixit14: Added more cases.

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracer.java
ddixit14: Thank you, corrected.

gax-java/gax/src/test/java/com/google/api/gax/tracing/MetricsTracerTest.java
ddixit14: I've removed this test as it is testing multiple methods in the same test.

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracer.java
```diff
@@ -42,12 +52,123 @@
 @InternalApi
 public class MetricsTracer implements ApiTracer {
 
-  public MetricsTracer(MethodName methodName, MetricsRecorder metricsRecorder) {}
+  public static final String STATUS_ATTRIBUTE = "status";
```
lqiu96, line 55: nit: Can this be private?

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracer.java
```diff
@@ -42,12 +52,123 @@
...
+  /** {@inheritDoc} */
+  @Override
+  public void attemptPermanentFailure(Throwable error) {
+
+    attributes.put(STATUS_ATTRIBUTE, extractStatus(error));
+    metricsRecorder.recordAttemptLatency(attemptTimer.elapsed(TimeUnit.MILLISECONDS), attributes);
+    metricsRecorder.recordAttemptCount(1, attributes);
+  }
+
+  /** {@inheritDoc} */
```
lqiu96, line 148: What docs is this inheriting? I might be missing something, but don't think this Javadocs is valid.

Could this be expanded to explain the logic below (why we're capturing certain things like OK, CANCELLED, etc.)

gax-java/gax/src/test/java/com/google/api/gax/tracing/MetricsTracerFactoryTest.java
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  @Test
+  public void testNewTracer_notNull() {
+    // Call the newTracer method
+    ApiTracer apiTracer = metricsTracerFactory.newTracer(parent, spanName, OperationType.Unary);
+
+    // Assert that the apiTracer created has expected type and not null
+    assertTrue(apiTracer instanceof MetricsTracer);
+    assertNotNull(apiTracer);
```
lqiu96, line 68: nit: I believe inside Gax we use Google's [Truth](https://truth.dev/) fluent asserting library. 

Can this be changed to:
```
Truth.assertThat(apiTracer).isInstanceOf(MetricsTracer.class)
... etc
```

gax-java/gax/src/test/java/com/google/api/gax/tracing/MetricsTracerTest.java
```diff
@@ -0,0 +1,271 @@
...
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+import org.mockito.quality.Strictness;
+import org.threeten.bp.Duration;
+
+@RunWith(JUnit4.class)
+public class MetricsTracerTest {
+  @Rule
+  public final MockitoRule mockitoRule = MockitoJUnit.rule().strictness(Strictness.STRICT_STUBS);
```
lqiu96, line 63: I'm not too familiar with this Rule. Could you explain (+ document) why this is needed in this test?

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracer.java
```diff
@@ -42,12 +52,123 @@
...
+    }
+
+    return statusString;
+  }
 
   /**
    * Add attributes that will be attached to all metrics. This is expected to be called by
    * handwritten client teams to add additional attributes that are not supposed be collected by
    * Gax.
    */
```
lqiu96, line 170: Would you be able to add a bit more about when this should be used by handwritten libraries? I'm assuming that it most likely should be done prior to any attempt/ operation starting? Is there any impact if this is called during an operation or between attempts?

lqiu96:
Thanks @ddixit14, I think it's looking good. I have a general question about the recording the attempts.

Is there a reason why `attemptStarted()` doesn't record/ increment the number of attempts (i.e. runs `metricsRecorder.recordAttemptCount(1, attributes);`). I would assume that starting the attempt would record an attempt count.

I'm guessing there is a reason and I'm not too familiar with why, but it seems like that might be a better spot than having to do that in `attemptSucceeded` and `attemptCancelled` and `attemptFailed`, etc.

ddixit14:
> Thanks @ddixit14, I think it's looking good. I have a general question about the recording the attempts.
> 
> Is there a reason why `attemptStarted()` doesn't record/ increment the number of attempts (i.e. runs `metricsRecorder.recordAttemptCount(1, attributes);`). I would assume that starting the attempt would record an attempt count.
> 
> I'm guessing there is a reason and I'm not too familiar with why, but it seems like that might be a better spot than having to do that in `attemptSucceeded` and `attemptCancelled` and `attemptFailed`, etc.

According to me, if we record the attempt count when the attempt started, we cannot deduce its attributes (did it succeed/fail/got cancelled?). Attributes should be recorded only when we know the final status of the attempt. If we do it in attemptStarted(), there would be no "status" attributes to record. Does it make sense?

gax-java/gax/src/test/java/com/google/api/gax/tracing/MetricsTracerTest.java
ddixit14: This is just a stricter way of testing for early detection of unused stubs and argument mismatches. It has been used in tracer tests, so I thought it would be good to have for robust testing. I will add in some documentation for it.

gax-java/gax/src/test/java/com/google/api/gax/tracing/MetricsTracerTest.java
```diff
@@ -0,0 +1,271 @@
...
+  }
+
+  @Test
+  public void testAddAttributes_recordsAttributes()
+      throws NoSuchFieldException, IllegalAccessException {
+
+    metricsTracer.addAttributes("FakeTableId", "12345");
+
+    // Use reflection to access the private field
+    Field attributesMap = MetricsTracer.class.getDeclaredField("attributes");
```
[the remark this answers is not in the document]
ddixit14: That's potentially another way to test this. I can give it a try. What do you think @blakeli0 ? Which way would be better ? Is there a generalized way we test private fields in gax @lqiu96 ?

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracer.java
ddixit14: This is inheriting docs from https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/gax/src/main/java/com/google/api/gax/tracing/ApiTracer.java . Sure, added more detailed docs.

gax-java/gax/src/test/java/com/google/api/gax/tracing/MetricsTracerTest.java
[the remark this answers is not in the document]
blakeli0: +1 to what Lawrence said, using reflection is error prone and not recommended. I would expose a package private getter for it if we want to unit test it.

gax-java/gax/src/test/java/com/google/api/gax/tracing/MetricsTracerFactoryTest.java
ddixit14: Done, thank you.

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracer.java
ddixit14: That's a great question, which I don't have an answer for currently. But yes, as far as I can imagine, this will be add the attribute for every metric calculated during an instance of the metrics tracer.

lqiu96: Er, I meant that I see it's inheriting from ApiTracer, but I don't see an `extractStatus()` method in ApiTracer. I don't think there are any javadocs to inherit since there is no `super.extractStatus()`. I think you will have to write your own javadocs here.

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracer.java
lqiu96: Er, I meant that I see it's inheriting from ApiTracer, but I don't see an `extractStatus()` method in ApiTracer. I don't think there are any javadocs to inherit since there is no `super.extractStatus()`. I think you will have to write your own javadocs here.

ddixit14: yes yes, you're correct, I've added all the necessary docs now.

gax-java/gax/src/test/java/com/google/api/gax/tracing/MetricsTracerTest.java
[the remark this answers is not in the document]
ddixit14: Added a package-private method and used it in the test. (removed reflection code).

lqiu96:
> > Thanks @ddixit14, I think it's looking good. I have a general question about the recording the attempts.
> > Is there a reason why `attemptStarted()` doesn't record/ increment the number of attempts (i.e. runs `metricsRecorder.recordAttemptCount(1, attributes);`). I would assume that starting the attempt would record an attempt count.
> > I'm guessing there is a reason and I'm not too familiar with why, but it seems like that might be a better spot than having to do that in `attemptSucceeded` and `attemptCancelled` and `attemptFailed`, etc.
> 
> According to me, if we record the attempt count when the attempt started, we cannot deduce its attributes (did it succeed/fail/got cancelled?). Attributes should be recorded only when we know the final status of the attempt. If we do it in attemptStarted(), there would be no "status" attributes to record. Does it make sense?

I see. I think given the current interface of MetricsRecorder this makes sense. I wonder if MetricsRecorder should be enhanced with some methods like:
1. recordOperationResult
2. recordAttemptResult

I don't have a list of pro/cons with this, but it seems cleaner (?) or at least makes sense in my head. I think that recording the attempt should occur when the attempt started and not be dependent on getting a result. Given that we need to record the status, maybe that can be done separately from recording the attempt count?

This probably doesn't need to be fixed now, but just throwing some thoughts out.
CC: @blakeli0

ddixit14:
> > > Thanks @ddixit14, I think it's looking good. I have a general question about the recording the attempts.
> > > Is there a reason why `attemptStarted()` doesn't record/ increment the number of attempts (i.e. runs `metricsRecorder.recordAttemptCount(1, attributes);`). I would assume that starting the attempt would record an attempt count.
> > > I'm guessing there is a reason and I'm not too familiar with why, but it seems like that might be a better spot than having to do that in `attemptSucceeded` and `attemptCancelled` and `attemptFailed`, etc.
> > 
> > 
> > According to me, if we record the attempt count when the attempt started, we cannot deduce its attributes (did it succeed/fail/got cancelled?). Attributes should be recorded only when we know the final status of the attempt. If we do it in attemptStarted(), there would be no "status" attributes to record. Does it make sense?
> 
> I see. I think given the current interface of MetricsRecorder this makes sense. I wonder if MetricsRecorder should be enhanced with some methods like:
> 
> 1. recordOperationResult
> 2. recordAttemptResult
> 
> I don't have a list of pro/cons with this, but it seems cleaner (?) or at least makes sense in my head. I think that recording the attempt should occur when the attempt started and not be dependent on getting a result. Given that we need to record the status, maybe that can be done separately from recording the attempt count?
> 
> This probably doesn't need to be fixed now (or fixed at all), but just throwing some thoughts out. CC: @blakeli0

That's exactly what came to my mind when I thought about multiple attempts for a successful operation, should we keep a track of attempt_count somehow, with a counter maybe? But then I realized that we are calculating the metrics for each attempt, so keeping a track of number of attempts is irrelevant for us. For every attempt that resulted in something, we just increment the count by 1 with the corresponding attributes.

gax-java/gax/src/test/java/com/google/api/gax/tracing/MetricsTracerTest.java
```diff
@@ -0,0 +1,261 @@
+/*
+ * Copyright 2019 Google LLC
```
blakeli0, line 2:
```suggestion
 * Copyright 2024 Google LLC
```

gax-java/gax/src/test/java/com/google/api/gax/tracing/MetricsTracerTest.java
```diff
@@ -0,0 +1,261 @@
...
+
+  @Test
+  public void testAddAttributes_recordsAttributes() {
+
+    metricsTracer.addAttributes("FakeTableId", "12345");
+    Truth.assertThat(metricsTracer.getAttributes().get("FakeTableId").equals("12345"));
+  }
+
+  @Test
+  public void testExtractStatus_errorConversion() {
```
blakeli0, line 238: It would be better to test each scenario in its own test case, instead of testing all scenarios in a for loop.

ddixit14: I'm using a for loop for all cases of ApiExceptions, and then created two new tests for "OK" and "UNKNOWN". Is this okay?

gax-java/gax/src/test/java/com/google/api/gax/tracing/MetricsTracerTest.java
```diff
@@ -0,0 +1,263 @@
...
+
+    metricsTracer.addAttributes("FakeTableId", "12345");
+    Truth.assertThat(metricsTracer.getAttributes().get("FakeTableId").equals("12345"));
+  }
+
+  @Test
+  public void testExtractStatus_errorConversion_apiExceptions() {
+
+    // test all cases of ApiExceptions
+    for (Code code : Code.values()) {
```
blakeli0, line 240: We don't have to test all codes, using one code is good enough. Because we don't differentiate the statuses in this method.

blakeli0:
We don't have to mention OpenTelemetry in the title, as this implementation is framework neutral. Please add more details to the description as well.

gax-java/gax/src/test/java/com/google/api/gax/tracing/MetricsTracerTest.java
ddixit14: understood, changed to just one code.

ddixit14:
> We don't have to mention OpenTelemetry in the title, as this implementation is framework neutral. Please add more details to the description as well.

Understood, changed title and added a detailed description.

blakeli0 (approved):

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracer.java
```diff
@@ -42,12 +52,173 @@
 @InternalApi
 public class MetricsTracer implements ApiTracer {
 
-  public MetricsTracer(MethodName methodName, MetricsRecorder metricsRecorder) {}
+  private static final String STATUS_ATTRIBUTE = "status";
+
+  private Stopwatch attemptTimer;
+
+  private final Stopwatch operationTimer = Stopwatch.createStarted();
```
lqiu96, line 59: question: Is this initialized here because there is no `operationStarted`? Do you know if there would be any delay between when this started and when the RPC is kicked off (I'm assuming that would be considered operation start)?

Or is it roughly the same because MetricsTracer is created when the RPC is kicked off?

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracer.java
```diff
@@ -42,12 +52,173 @@
...
+
+    attributes.put(STATUS_ATTRIBUTE, StatusCode.Code.CANCELLED.toString());
+    metricsRecorder.recordAttemptLatency(attemptTimer.elapsed(TimeUnit.MILLISECONDS), attributes);
+    metricsRecorder.recordAttemptCount(1, attributes);
+  }
+
+  /**
+   * Adds an annotation that the attempt failed, but another attempt will be made after the delay.
+   *
+   * @param error the transient error that caused the attempt to fail.
```
lqiu96, line 150: nit: If this is called for any non-200 status code, then some errors may not be transient. Something like UNAUTHENTICATED or PERMISSION_DENIED wouldn't be transient. Probably just remove the transient keyword.

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracer.java
lqiu96: Ok, this doesn't need to figure out now or in this PR. I was just curious since it seems like whenever it is called, it would impact every following call.

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracer.java
```diff
@@ -42,12 +52,173 @@
...
    * Add attributes that will be attached to all metrics. This is expected to be called by
    * handwritten client teams to add additional attributes that are not supposed be collected by
    * Gax.
    */
-  public void addAttributes(String key, String value) {};
+  public void addAttributes(String key, String value) {
+    attributes.put(key, value);
+  };
+
+  /** This method is only used for testing */
```
lqiu96, line 220: Probably can just use the `@VisibleForTesting` annotation here.

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracer.java
ddixit14: yes, the latter statement is true.

gax-java/gax/src/test/java/com/google/api/gax/tracing/MetricsTracerFactoryTest.java
```diff
@@ -0,0 +1,81 @@
...
+    // Call the newTracer method
+    ApiTracer apiTracer = metricsTracerFactory.newTracer(parent, spanName, OperationType.Unary);
+
+    // Assert that the apiTracer created has expected type and not null
+    Truth.assertThat(apiTracer).isInstanceOf(MetricsTracer.class);
+    Truth.assertThat(apiTracer).isNotNull();
+
+    MetricsTracer metricsTracer = (MetricsTracer) apiTracer;
+    Truth.assertThat(
+        metricsTracer.getAttributes().get("method_name").equals("testService.testMethod"));
```
lqiu96, line 79: I think it should be 
`Truth.assertThat(metricsTracer.getAttributes().get("method_name")).isEqualTo("testService.testMethod");`?

lqiu96 (approved):
LGTM. I left a few nits if you could take a quick look at.

ddixit14:
> LGTM. I left a few nits if you could take a quick look at.

All changed and pushed. thank you.

PR #2429, opened 2024-01-26
fix: Endpoint resolution uses user set endpoint from ClientSettings

lqiu96:
Add new Internal method `getUserSetEndpoint()` that returns the user set endpoint from ClientSettings or null otherwise.

This is used as part of endpoint resolution to determine the proper endpoint.


blakeli0 (approved):
LGTM. Once we have getters for `resolvedEndpoint`, we should add more unit test cases.

PR #2430, opened 2024-01-29
fix: Move direct path misconfiguration log to before creating the first channel

### LINKED ISSUE #2423: DirectPath Misconfiguration emitted for incorrect Credentials but DirectPath still works
#### Environment details
1. GCE
1. artifact version(s): google-cloud-storage:2.32.1, gax 2.41.0

#### Issue

Run the following code in GCE with DirectPath enabled causes the following warning to be logged.

```
InstantiatingGrpcChannelProvider logDirectPathMisconfig
WARNING: DirectPath is misconfigured. Please make sure the credential is an instance of com.google.auth.oauth2.ComputeEngineCredentials .
```

#### Code example

```java
RetrySettings retrySettings = StorageOptions.getDefaultRetrySettings().toBuilder().build();
StorageOptions retryStorageOptions =
    StorageOptions.grpc().setRetrySettings(retrySettings).setAttemptDirectPath(true).build();
Storage storageClient = retryStorageOptions.getService();
```

blakeli0:
Fixes #2423 
The root cause of the issue is that `logDirectPathMisconfig()` is called in the builder of `InstantiatingGrpcChannelProvider`, which could be called multiple times before it is fully instantiated. We should only call `logDirectPathMisconfig()` right before `createChannel()` which creates the first channel. 

We can not move it to before `createSingleChannel()` because it is used for resizing channel regularly after a client is initialized, and we only want to log direct path misconfiguration once.

### DIFF
--- a/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
+++ b/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
@@ -272,8 +272,10 @@ public final class InstantiatingGrpcChannelProvider implements TransportChannelP
     return false;
   }
 
-  //This method should be called once per client initialization, hence can not be called in the builder
-  // or createSingleChannel, only in getTransportChannel which creates the first channel for a client.
+  // This method should be called once per client initialization, hence can not be called in the
+  // builder
+  // or createSingleChannel, only in getTransportChannel which creates the first channel for a
+  // client.
   private void logDirectPathMisconfig() {
     if (isDirectPathXdsEnabled()) {
       // Case 1: does not enable DirectPath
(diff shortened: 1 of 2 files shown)


lqiu96:
Thoughts on adding test which creates a InstantiatingGrpcChannelProvider, changes a property in the builder with it still misconfigured, and recreates it? Essentially runs `InstantiatingGrpcChannelProvider.{newBuilder|toBuilder()}.build()` twice. The FakeLogHandler should only show the misconfiguration log once since the logging should occur prior to channel creation and on `.build()`.

Might be beneficial so we don't accidentally refactor `logDirectPathMisconfig()` into the channel creation logic in the future.

lqiu96 (approved):

JesseLovelace (approved):

blakeli0:
> Thoughts on adding test which creates a InstantiatingGrpcChannelProvider, changes a property in the builder with it still misconfigured, and recreates it? Essentially runs `InstantiatingGrpcChannelProvider.{newBuilder|toBuilder()}.build()` twice. The FakeLogHandler should only show the misconfiguration log once since the logging should occur prior to channel creation and on `.build()`.

Thanks! Added a test case that calling builder should not log anything.

PR #2431, opened 2024-01-30
feat: add `generate_repo.py`

JoeWang1127:
In this PR:
- Add `generate_repo.py` to generate libraries defined in a configuration file.
  - Regenerate `.repo-metadata.json`
  - Generate `.OwlBot.yaml` and `owlbot.py` for new clients
  - Invoke `generate_composed_library.py` to generate multiple versions of a GAPIC library.
  - Apply repo-level post processing using python
    - Regenerate root `pom.xml`
    - Regenerate `gapic-libraries-bom/pom.xml`
- Format python script using `black`.
- Add unit tests for utility functions.
- Add an integration test using python, replacing shell integration tests.
- Add a python lint check in CI.

This is [step 2](https://docs.google.com/document/d/1JiCcG3X7lnxaJErKe0ES_JkyU7ECb40nf2Xez3gWvuo/edit?pli=1&tab=t.0#bookmark=id.h3n0hcp3ch2m) of milestone 2 of hermetic build project.


library_generation/model/library_config.py
```diff
@@ -0,0 +1,61 @@
...
+        rpc_documentation: Optional[str] = None,
+        cloud_api: Optional[bool] = True,
+        requires_billing: Optional[bool] = True,
+    ):
+        self.api_shortname = api_shortname
+        self.api_description = api_description
+        self.name_pretty = name_pretty
+        self.product_documentation = product_documentation
+        self.gapic_configs = gapic_configs
+        self.library_type = library_type if library_type else "GAPIC_AUTO"
```
JoeWang1127, line 50: Use `str` for `library_type` and `release_level` because yaml doesn't parse the version to Enum. Will refactor later.

.github/workflows/verify_library_generation.yaml
```diff
@@ -45,35 +43,13 @@ jobs:
...
         popd
-
-    - name: install utils (macos)
-      if: matrix.os == 'macos-12'
-      shell: bash
-      run: |
-        brew update --preinstall
-        # we need the `realpath` command to be available
-        brew install coreutils
-    - name: install docker (ubuntu)
```
JoeWang1127, line 56: We don't need to install docker because it's included in [ubuntu-22.04](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#tools)

library_generation/generate_repo.py
```diff
@@ -0,0 +1,111 @@
...
+    required=False,
+    default=".",
+    type=str,
+    help="""
+    If specified, the generated files will be sent to this location.
+    If not specified, the repository will be generated to the current working
+    directory.
+    """,
+)
+def generate(
```
JoeWang1127, line 60: Create this method to separate cli and logic.

library_generation/model/bom_config.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def __init__(
+        self, group_id: str, artifact_id: str, version: str, version_annotation
+    ):
+        self.group_id = group_id
+        self.artifact_id = artifact_id
+        self.version = version
+        self.version_annotation = version_annotation
+
+    def __lt__(self, another):
```
JoeWang1127, line 30: add `__lt__()` and `__eq__()` methods used in sorting BOM artifacts, which is useful in file comparison.

library_generation/test/integration_tests.py
```diff
@@ -0,0 +1,148 @@
...
+                )
+                compare_result = dircmp(
+                    f"{golden_dir}/{library_name}",
+                    f"{repo_dest}/{library_name}",
+                    ignore=[".repo-metadata.json"],
+                )
+                # compare source code
+                self.assertEqual([], compare_result.left_only)
+                self.assertEqual([], compare_result.right_only)
+                # self.assertEqual([], compare_result.diff_files)
```
JoeWang1127, line 78: The transport section in README.md is not rendered correctly because the value in template is grpc, http and both, rather than grpc, rest and grpc+rest.

PR to fix this issue: https://github.com/googleapis/synthtool/pull/1925

library_generation/generate_repo.py
```diff
@@ -0,0 +1,111 @@
...
+
+    for library_path, library in repo_config.libraries.items():
+        print(f"generating library {library.api_shortname}")
+
+        generate_composed_library(
+            config=config,
+            library_path=library_path,
+            library=library,
+            output_folder=repo_config.output_folder,
+            versions_file=repo_config.versions_file,
```
diegomarquezp, line 102: From design, we probably want to have a check for the existence of such file. AFAIU all repos have this file so it's a good heuristic to ensure that repository_path is valid

JoeWang1127: The versions.txt is checked in `prepare_repo` method in utilities.py.

diegomarquezp: Ah, I see. Thanks

library_generation/utilities.py
```diff
@@ -1,125 +1,503 @@
...
+    with open(versions_file, "r") as f:
+        for line in f.readlines():
+            if artifact_id in line:
+                return line.split(":")[index].strip()
+
+
+def main(argv: Sequence[str]) -> None:
+    if len(argv) < 1:
+        raise ValueError(
+            "Usage: python generate_composed_library_args.py function_name arg1...argN"
```
diegomarquezp, line 489: I think it should be `python utilities.py`
However this `main` function was made so shell scripts can call them. Since we are moving to ITs written in python, it may make sense to remove them

.github/workflows/verify_library_generation.yaml
diegomarquezp, line 1: What tool did you format the python files with? It'd be nice if that tool had a linter mode that fails on non-compliant files so we can add it as part of our linter step

JoeWang1127: > What tool did you format the python files with?

I use `black` to format python code and it also has a validate function.

>  It'd be nice if that tool had a linter mode that fails on non-compliant files so we can add it as part of our linter step

Sure, I'll add a lint step in CI.

This tool is included when I install all requirements.

library_generation/generate_composed_library.py
```diff
@@ -12,130 +27,148 @@
...
+
 def generate_composed_library(
     config: GenerationConfig,
+    library_path: str,
     library: LibraryConfig,
-    repository_path: str,
-    enable_postprocessing: bool = True,
+    output_folder: str,
+    versions_file: str,
+    language: str = "java",
```
JoeWang1127, line 49: Maybe remove this parameter.

library_generation/utilities.py
```diff
@@ -1,125 +1,503 @@
...
+    if library.api_reference:
+        repo_metadata["api_reference"] = library.api_reference
+    if library.issue_tracker:
+        repo_metadata["issue_tracker"] = library.issue_tracker
+    if library.rest_documentation:
+        repo_metadata["rest_documentation"] = library.rest_documentation
+    if library.rpc_documentation:
+        repo_metadata["rpc_documentation"] = library.rpc_documentation
+
+    # generate .repo-meta.json
```
JoeWang1127, line 396: Add more comments here.

library_generation/test/resources/integration/java-bigtable/generation_config.yaml
diegomarquezp, line 1: As discussed, we can either enable the IT for this split repo or leave it as a follow up, whichever is most convenient

library_generation/generate_composed_library.py
JoeWang1127: Removed

library_generation/utilities.py
JoeWang1127: Added.

library_generation/utilities.py
JoeWang1127: Removed.

.github/workflows/verify_library_generation.yaml
JoeWang1127: I added a python lint job in CI.

library_generation/test/resources/integration/java-bigtable/generation_config.yaml
JoeWang1127: I think I'll remove the split repo integration test for now, since we mainly focus on generating google-cloud-java.

diegomarquezp: I think it should be fine with leaving the yaml file, although not touched by our scripts

suztomo (approved):
It would be nice if I can see how the real main entry point (library_generation/generate_repo.py?) is invoked in verify_library_generation.yaml. No need to replace the Python-based integraiton tests.

library_generation/generate_repo.py
diegomarquezp, line 1: Maybe we can leave a follow up after this PR to update the README to reflect that the main entrypoint is this one and not the now-lower-level generate_library.sh

JoeWang1127: Sure, I'll follow up with this.

library_generation/test/resources/integration/java-bigtable/generation_config.yaml
JoeWang1127: Restored the config file for java-bigtable.

PR #2433, opened 2024-01-31
feat: OpentelemetryMetricsRecorder Implementation

ddixit14:


gax-java/gax/src/main/java/com/google/api/gax/tracing/OpentelemetryMetricsRecorder.java
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  public void recordOperationCount(long count, Map<String, String> attributes) {
+    operationCountRecorder.add(count, toOtelAttributes(attributes));
+  }
+
+  @VisibleForTesting
+  Attributes toOtelAttributes(Map<String, String> attributes) {
+
+    if (attributes == null) {
```
blakeli0, line 103: We could initialize the map to a new map in the constructor.

gax-java/gax/src/main/java/com/google/api/gax/tracing/OpentelemetryMetricsRecorder.java
```diff
@@ -0,0 +1,111 @@
...
+
+  private LongCounter attemptCountRecorder;
+
+  private Attributes attributes;
+
+  public OpentelemetryMetricsRecorder(Meter meter) {
+    this.meter = meter;
+    this.attemptLatencyRecorder =
+        meter
+            .histogramBuilder("attempt_latency")
```
blakeli0, line 59: The metric names should include service name, like `spanner/attempt_latency`, we should be able to get it from `StubSettings`

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracer.java
```diff
@@ -64,6 +65,7 @@ public class MetricsTracer implements ApiTracer {
 
   public MetricsTracer(MethodName methodName, MetricsRecorder metricsRecorder) {
     this.attributes.put("method_name", methodName.toString());
+    this.attributes.put("language", LANGUAGE);
```
lqiu96, line 68: nit: Could you document above in Javadocs (maybe in the class javadoc) that `method_name` and `language` will be autopopulated and that the value `Java` for language is by default.

I think it might be interesting if someone has like a kotlin app and uses our library with otel and sees that the language attribute is Java. Granted this is Java SDK and they can override this value themselves if they want.

gax-java/gax/src/main/java/com/google/api/gax/tracing/OpentelemetryMetricsRecorder.java
blakeli0: Never mind, I confused it with the field.

ddixit14: @blakeli0 yes, I was about to ask you this since I was getting confused as well..thank you!!

gax-java/gax/src/main/java/com/google/api/gax/tracing/OpentelemetryMetricsRecorder.java
```diff
@@ -0,0 +1,111 @@
...
+
+  private DoubleHistogram attemptLatencyRecorder;
+
+  private DoubleHistogram operationLatencyRecorder;
+
+  private LongCounter operationCountRecorder;
+
+  private LongCounter attemptCountRecorder;
+
+  private Attributes attributes;
```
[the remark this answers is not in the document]
ddixit14: no, will remove it. thank you!!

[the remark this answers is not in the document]
blakeli0: The original intention is that we could use it for some common client-level attributes, e.g. `isDirectPathEnabled`, but I see that it is not used at this moment. It could still be used for `language` attributes, so we don't have to add it for every operation in `MetricTracer`.

gax-java/pom.xml
```diff
@@ -158,6 +158,13 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
+      <dependency>
+        <groupId>io.opentelemetry</groupId>
+        <artifactId>opentelemetry-bom</artifactId>
+        <version>1.27.0</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
```
lqiu96, line 167: I'm not 100% sure about this. Would declaring the bom be better suited for shared-deps' third party module instead of gax? CC: @suztomo

suztomo: This should go to the shared dependencies BOM.

suztomo: @JoeWang1127 Shouldn't  the check "ci / unmanaged_dependency_check (pull_request) " fail here? I see it passed.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
```diff
@@ -0,0 +1,554 @@
...
+    // Finally, create the Tracer Factory
+    return new MetricsTracerFactory(otelMetricsRecorder);
+  }
+
+  private void setupOtelCollector(String scriptPath, String directoryPath, String configPath)
+      throws Exception {
+    Process process =
+        Runtime.getRuntime().exec(scriptPath + " " + directoryPath + " " + configPath);
+    process.waitFor();
+  }
```
lqiu96, line 553: I'm not sure that we should have the ITs create a new process or start the OTEL collector. I think  it would be better if this is run separately (i.e. manually for local testing).

The same way we do with the gapic-showcase server itself. We start up a local instance ourselves via (`gapic-showcase run`).

For the GH actions, I think you can create a step that run the startup scripts.

ddixit14: We need a new otelCollector for every test, that's why.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
```diff
@@ -0,0 +1,554 @@
...
+
+  @Test
+  public void testHttpJson_OperationSucceded_recordsMetrics() throws Exception {
+
+    // initialize the otel-collector
+    setupOtelCollector(
+        "../scripts/start_otelcol.sh",
+        "../directory1",
```
lqiu96, line 63: qq, why do these all need to be a in different directory?

`directory1` vs `directory2`?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
```diff
@@ -0,0 +1,554 @@
...
+
+    EchoRequest requestWithNoError =
+        EchoRequest.newBuilder()
+            .setError(Status.newBuilder().setCode(Code.OK.ordinal()).build())
+            .build();
+
+    client.echoCallable().futureCall(requestWithNoError).isDone();
+
+    // wait for the metrics to get uploaded
+    Thread.sleep(5000);
```
lqiu96, line 88: I'm a bit worried about flakes for this. I'm not sure if there is different way we can check for the status. Can you check if there a suggest solution to check for metrics? If not, perhaps we can busy-wait for like 5-10 seconds?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
ddixit14: I tried collecting them in one directory, but that overwrites the metrics of one test into another file. I can make this cleaner once the logic is approved.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
```diff
@@ -0,0 +1,554 @@
...
+        EchoSettings.newHttpJsonBuilder()
+            .setCredentialsProvider(NoCredentialsProvider.create())
+            .setTracerFactory(createOpenTelemetryTracerFactory("4317"))
+            .setTransportChannelProvider(
+                EchoSettings.defaultHttpJsonTransportProviderBuilder()
+                    .setHttpTransport(
+                        new NetHttpTransport.Builder().doNotValidateCertificate().build())
+                    .setEndpoint("http://localhost:7469")
+                    .build())
+            .build();
```
lqiu96, line 76: These configurations are look to be the same (barring transport and port). Can you move it to the TestInitializer class?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
```diff
@@ -0,0 +1,554 @@
...
+
+    EchoClient client = EchoClient.create(httpJsonEchoSettings);
+
+    EchoRequest requestWithNoError =
+        EchoRequest.newBuilder()
+            .setError(Status.newBuilder().setCode(Code.OK.ordinal()).build())
+            .build();
+
+    client.echoCallable().futureCall(requestWithNoError).isDone();
```
lqiu96, line 85: I don't think we need to check for `isDone()`. I think we can just shoo the request via like `client.echo(requestWithNoError)` since it's a sync call and it'll block for the response.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
```diff
@@ -0,0 +1,554 @@
...
+                    .setChannelConfigurator(ManagedChannelBuilder::usePlaintext)
+                    .build())
+            .setEndpoint("localhost:7469")
+            .build();
+
+    EchoClient client = EchoClient.create(grpcEchoSettings);
+
+    EchoRequest requestWithNoError =
+        EchoRequest.newBuilder()
+            .setError(Status.newBuilder().setCode(Code.OK.ordinal()).build())
```
lqiu96, line 124: I think you can do `setContent(...)` for a successful response for the echo RPC.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
```diff
@@ -0,0 +1,554 @@
...
+    client.echoCallable().futureCall(requestWithNoError).isDone();
+
+    // wait for the metrics to get uploaded
+    Thread.sleep(5000);
+
+    String filePath = "../directory1/testHttpJson_OperationSucceeded_metrics.txt";
+    // 1. Check if the log file exists
+    File file = new File(filePath);
+    Truth.assertThat(file.exists()).isTrue();
+    Truth.assertThat(file.length() > 0).isTrue();
```
lqiu96, line 94: I think it would be better if we can validate the contents inside the file (i.e. contains operations successful metrics/ data) rather than just verifying that it could write something to the file.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
```diff
@@ -0,0 +1,554 @@
...
+            .build();
+    EchoClient client = EchoClient.create(httpJsonEchoSettings);
+
+    EchoRequest requestWithNoError =
+        EchoRequest.newBuilder()
+            .setError(Status.newBuilder().setCode(Code.OK.ordinal()).build())
+            .build();
+
+    // explicitly cancel the request
+    client.echoCallable().futureCall(requestWithNoError).cancel(true);
```
lqiu96, line 170: There is probably a very small likelihood of this every happening, but I wonder if cancel is slow and that the Echo RPC returns a value back (since it is fast).

Perhaps for the cancel operation, we can use block (and block for like 1-2 seconds) before cancelling.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
```diff
@@ -0,0 +1,554 @@
...
+
+  @Test
+  public void testGrpc_attemptFailedRetriesExhausted_recordsMetrics() throws Exception {
+
+    setupOtelCollector(
+        "../scripts/start_otelcol.sh",
+        "../directory7",
+        "../opentelemetry-helper/configs/testGrpc_attemptFailedRetriesExhausted.yaml");
+
+    RetrySettings retrySettings = RetrySettings.newBuilder().setMaxAttempts(2).build();
```
lqiu96, line 317: qq, some tests use 2, 3, or 5 max attempts? Is there a reason for the different numbers?

ddixit14: No reason for these different numbers. The count appears in the metrics collected, so I will use these numbers in the verification.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
ddixit14: Yes, this is what I'm doing currently.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
```diff
@@ -0,0 +1,554 @@
...
+                        new NetHttpTransport.Builder().doNotValidateCertificate().build())
+                    .setEndpoint("http://localhost:7469")
+                    .build())
+            .build();
+    EchoClient client = EchoClient.create(httpJsonEchoSettings);
+
+    EchoRequest requestWithError =
+        EchoRequest.newBuilder()
+            .setError(Status.newBuilder().setCode(Code.PERMISSION_DENIED.ordinal()).build())
+            .build();
```
lqiu96, line 251: qq, also with this. Some tests are using PERMISSION_DENIED, AUTHENTICATED, and INVALID_ARGUMENT? Is there a reason for the different status code?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
```diff
@@ -0,0 +1,554 @@
...
+  private static ApiTracerFactory createOpenTelemetryTracerFactory(String port) {
+    // OTLP Metric Exporter setup
+    String endpoint = "http://localhost:" + port;
+    OtlpGrpcMetricExporter metricExporter =
+        OtlpGrpcMetricExporter.builder().setEndpoint(endpoint).build();
+
+    // Periodic Metric Reader configuration
+    PeriodicMetricReader metricReader =
+        PeriodicMetricReader.builder(metricExporter)
+            .setInterval(java.time.Duration.ofSeconds(3))
```
lqiu96, line 523: qq, Is there a reason specifically for 3s as the interval? Perhaps that's what seems like a good default value or something else?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
ddixit14: There is no specific reason. I will use these status codes to verify the metrics collected, so wanted to use as many different codes as I can.

showcase/scripts/cleanup_otelcol.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+# Directories to delete
+for ((i = 1; i <= 10; i++)); do
+    directory="../directory$i"
```
lqiu96, line 5: Given that we are probably adding in more metrics in the future (which also adds in more test cases to showcase), I don't think we should be hard coding the number of directories. I think in a comment above, it would be better if this could all be done in a single directory with different files, but there may be an issue with doing that (I'm not entirely sure, so let me know otherwise).

If directories are needed and the other option isn't possible, I think it would be better if the directories could be named (rather than having something like 1,2,3).

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
ddixit14: No specific reason, yes, that is like a good default value I could think of, given every test waits for 5 seconds for the metrics to get exported.

showcase/scripts/cleanup_otelcol.sh
ddixit14: Yes, working on making single directory possible.

showcase/opentelemetry-helper/configs/testGrpc_OperationCancelled.yaml
```diff
@@ -0,0 +1,17 @@
...
+  file:
+    path: "testGrpc_OperationCancelled_metrics.txt"
+
+service:
+  extensions: []
+  pipelines:
+    metrics:
+      receivers: [otlp]
+      processors: []
+      exporters: [file]
```
lqiu96, line 17: These yaml files look to be all the same (barring the endpoint which looks to be incremented from ~43xx) and the file path. 

Can we consolidate this somehow? Perhaps we have one template file and the required params are spliced into the file for each exporter?

I'm don't think that having one file for each metric (i.e. attempt, operation, retry) and duplicating it for each transport is maintainable/ scalable.

showcase/opentelemetry-helper/configs/testHttpJson_OperationCancelled.yaml
```diff
@@ -0,0 +1,17 @@
+receivers:
+  otlp:
+    protocols:
+      grpc:
+        endpoint: "localhost:4319"
```
lqiu96, line 5: Is this supposed to be grpc?

showcase/pom.xml
```diff
@@ -63,6 +74,61 @@
...
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-exporter-prometheus</artifactId>
+      <version>1.27.0-alpha</version>
+    </dependency>
+    <!--Used by Spring actuator-->
+    <dependency>
+      <groupId>io.micrometer</groupId>
+      <artifactId>micrometer-registry-prometheus</artifactId>
+      <version>1.9.7</version>
+    </dependency>
```
lqiu96, line 88: I might have missed this above, but where are we using this inside showcase?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
ddixit14: Done.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
ddixit14: Done.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
ddixit14: Now using only one directory.

showcase/scripts/cleanup_otelcol.sh
ddixit14: Done.

showcase/opentelemetry-helper/configs/testGrpc_OperationCancelled.yaml
ddixit14: I tried, but even if we have a template file, we need some sort of scripting to create a new file using it, supplying the endpoint and file path and use it as the config for the collector. That would require invoking another script for every test. Still thinking of a better solution..

showcase/pom.xml
ddixit14: not using it, forgot to remove. now removed. thank you.

showcase/opentelemetry-helper/configs/testHttpJson_OperationCancelled.yaml
ddixit14: yes.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
ddixit14: Added logic for verifying the attributes. The logic is generic, and we can verify the existence of any string (key-value pair) in the metrics file generated.

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracer.java
ddixit14: Done. added.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
ddixit14: This is done, moved the configurations to TestClientInitializer. Clients with retry settings are still getting initialized within the test though.

showcase/opentelemetry-helper/configs/testGrpc_OperationCancelled.yaml
ddixit14: We have one file for every test, not every metric. And for each transport, we have a similar test.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
ddixit14: The condition for busy-wait would be file existence + file not empty, and then I will verify the contents? I am not sure if this will work 100%, since it might return true as soon as some metrics are written to the file ?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
lqiu96: Having to create multiple otelCollectors is ok imo. My point above is that I don't think we should have a Java method that essentially executes a shell script in the test. I think it would be better to move all that otel initialization logic down to the ./startup_otel script and have that script create the necessary (10+ or however many) otel collectors

 > We need a new otelCollector for every test, that's why.

What about a single otelCollector for all the tests and the metrics in each test case have a unique identifier? Is something like that possible so that we won't need to create a new otel collector for every new test added?

showcase/opentelemetry-helper/configs/testGrpc_OperationCancelled.yaml
ddixit14: This is fixed. We do not have these files now. We're generating a new config file for each test, and destroying it during cleanup.

lqiu96: IMO, I think having multiple scripts, if needed, to create the startup logic for otel is fine (as long as we just need to call one entry point `./startup_otel.sh`). I just don't think we should have to copy and paste a new YAML file every time we want to add a new test case for otel showcase.

If we do need a new otel collector for every test, I think it would be better to have like a mapping somewhere and have the startup script read through the mapping and create the otel collectors from there. Not ideal given that we have to update both the Java unit tests and create a new entry in the script. That's why I have a bit of preference for a single otel collector if possible. If there are limitations with the single otel collector, then my preference would be not have a ton of YAML files.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
```diff
@@ -0,0 +1,517 @@
...
+    // wait for the metrics to get uploaded
+    Thread.sleep(5000);
+
+    String filePath = "../test_data/testGrpc_OperationSucceeded_metrics.txt";
+    String attribute1 =
+        "\"attributes\":[{\"key\":\"language\",\"value\":{\"stringValue\":\"Java\"}},{\"key\":\"method_name\",\"value\":{\"stringValue\":\"Echo.Echo\"}},{\"key\":\"status\",\"value\":{\"stringValue\":\"OK\"}}]";
+
+    String[] params = {filePath, attribute1};
+    int result = verify_metrics(params);
+    Truth.assertThat(result).isEqualTo(0);
```
lqiu96, line 119: I'm not sure that we need a separate shell script to do the verification. This looks JSON to me and I think you can just read the file and parse it with a JSON library (Gson for example). Validation can be used with Json.

Was there an issue you were facing regarding using Gson?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
```diff
@@ -0,0 +1,517 @@
...
+    setupOtelCollector("../test_data/testHttpJson_OperationCancelled.yaml");
+
+    EchoClient client =
+        TestClientInitializer.createHttpJsonEchoClientOpentelemetry(
+            createOpenTelemetryTracerFactory("4319"));
+
+    EchoRequest requestWithNoError =
+        EchoRequest.newBuilder()
+            .setError(Status.newBuilder().setCode(Code.OK.ordinal()).build())
+            .build();
```
lqiu96, line 142: FYI about EchoRequest, I think it would be better to use `setContent(...)` for a successful request/ response rather than using `setError(Code.OK)`. Same with the other occurances

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
```diff
@@ -0,0 +1,517 @@
...
+    EchoClient client =
+        TestClientInitializer.createHttpJsonEchoClientOpentelemetry(
+            createOpenTelemetryTracerFactory("4321"));
+
+    EchoRequest requestWithError =
+        EchoRequest.newBuilder()
+            .setError(Status.newBuilder().setCode(Code.UNKNOWN.ordinal()).build())
+            .build();
+
+    client.echoCallable().futureCall(requestWithError).isDone();
```
lqiu96, line 218: Also FYI about the echo RPCs. IIRC, they're sync calls (RPC that doesn't have the `async` suffix are sync) and you won't need to wait for `is.Done()`.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
lqiu96: Ok, I think that's fine. Can you document that in the comments or create a static constant for the 3s value?

showcase/scripts/start_otelcol.sh
```diff
@@ -0,0 +1,24 @@
...
+
+## in future iterations/improvement, make this version dynamic
+curl --proto '=https' --tlsv1.2 -fOL https://github.com/open-telemetry/opentelemetry-collector-releases/releases/download/v0.93.0/otelcol_0.93.0_linux_amd64.tar.gz
+tar -xvf otelcol_0.93.0_linux_amd64.tar.gz
+
+killall otelcol
+
+## Start OpenTelemetry Collector with the updated config file
+echo "Starting OpenTelemetry Collector with the updated config file: $2"
+nohup ./otelcol --config "$2" > /dev/null 2>&1 &
```
lqiu96, line 22: One more thing about the start up + clean up scripts. They probably need to be added to the showcase github actions ci yaml files. I think you'll need to create a separate step to call them.

showcase/scripts/start_otelcol.sh
```diff
@@ -0,0 +1,24 @@
...
+
+# Create the install directory if it doesn't exist
+mkdir -p "$install_dir"
+
+# Change directory to the install directory
+cd "$install_dir"
+
+## in future iterations/improvement, make this version dynamic
+curl --proto '=https' --tlsv1.2 -fOL https://github.com/open-telemetry/opentelemetry-collector-releases/releases/download/v0.93.0/otelcol_0.93.0_linux_amd64.tar.gz
+tar -xvf otelcol_0.93.0_linux_amd64.tar.gz
```
lqiu96, line 16: @ddixit14 I've figured out the issue when pulling this down locally. I'm running on my M1 Mac with ARM and this is pulling down linux_amd64.

What worked for me was to pull down `otelcol_0.94.0_darwin_arm64.tar.gz` and run that. Something to consider for the different machines that we're running on. Maybe we can use a docker image for otel collector locally

PR #2438, opened 2024-02-01
chore: change header year

JoeWang1127:
In this PR:
- Change copyright year to 2024.

### DIFF
--- a/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/comment/CommentComposer.java
+++ b/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/comment/CommentComposer.java
@@ -23,7 +23,7 @@ import java.util.List;
 
 public class CommentComposer {
   private static final String APACHE_LICENSE_STRING =
-      "Copyright 2023 Google LLC\n\n"
+      "Copyright 2024 Google LLC\n\n"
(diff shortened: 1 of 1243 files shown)


gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/comment/CommentComposer.java
```diff
@@ -23,7 +23,7 @@
 
 public class CommentComposer {
   private static final String APACHE_LICENSE_STRING =
-      "Copyright 2023 Google LLC\n\n"
+      "Copyright 2024 Google LLC\n\n"
```
lqiu96, line 26: Do you have any concerns if we compute the year? I'm not sure if there are any downsides with it.

Something like `"Copyright " + Calendar.getYear() + "..."` or something?

lqiu96 (approved):
Not sure if my comment is feasible. LGTM otherwise.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/comment/CommentComposer.java
JoeWang1127: Good idea, implemented that.

PR #2439, opened 2024-02-05
chore: remove unmanaged-dependencies-check-latest tag

JoeWang1127:
In this PR:
- Do not create `unmanaged-dependencies-check-latest` tag in a release. The tag is not deleted in remote branch so downstream repositories can still use it.

We decided to use shared dependencies tag (for example, `google-cloud-shared-dependencies/v1.2.3`) in unmanaged dependencies, so this tag is no longer needed.


suztomo:
As exchanged messages, Joe will confirm the tag "google-cloud-shared-dependencies/v1.2.3" will work with RenovateBot before deleting unmanaged-dependencies-check-latest tag.

PR #2440, opened 2024-02-05
feat: Client/StubSettings' getEndpoint() returns the resolved endpoint

lqiu96:
## Changes
- Client|StubSettings `getEndpoint()` will now return the fully resolved endpoint
- Client|StubSettings `getUniverseDomain()` will not return the fully resolved universe domain
- Remove the generated `getEndpoint()` method in the {Client}StubSettings class. The call to `getEndpoint()` will now always hit the parent StubSettings and return the fully resolved endpoint.


gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
```diff
@@ -152,7 +152,16 @@ private GrpcCallContext(
...
+    // Attempt to create an empty, non-functioning EndpointContext by default (if set to null).
+    // This will be configured and updated by the client with user configurations. Attempts to
+    // use the default endpointContext (not updated with configurations) will result in incorrect
+    // data back.
+    try {
+      this.endpointContext =
+          endpointContext == null ? EndpointContext.newBuilder().build() : endpointContext;
+    } catch (IOException ex) {
+      throw new RuntimeException(ex);
+    }
```
lqiu96, line 164: This was added because Java-Logging creates the ClientContext and creates a default CallContext inside the client: https://github.com/googleapis/java-logging/blob/447c82cafb77379a2dffee7155c58a098fa58e83/google-cloud-logging/src/main/java/com/google/cloud/logging/spi/v2/GrpcLoggingRpc.java#L122

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonCallContext.java
```diff
@@ -133,7 +133,16 @@ private HttpJsonCallContext(
...
+    // Attempt to create an empty, non-functioning EndpointContext by default (if set to null).
+    // This will be configured and updated by the client with user configurations. Attempts to
+    // use the default endpointContext (not updated with configurations) will result in incorrect
+    // data back.
+    try {
+      this.endpointContext =
+          endpointContext == null ? EndpointContext.newBuilder().build() : endpointContext;
+    } catch (IOException ex) {
+      throw new RuntimeException(ex);
+    }
```
lqiu96, line 145: Added for the same reason as listed in GrpcCallContext

gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
```diff
@@ -161,15 +161,16 @@ public static ClientContext create(StubSettings settings) throws IOException {
...
+    // The serviceName is overwritten in the child subclasses (GAPIC libraries will override the
+    // default StubSettings' serviceName). Client can determine if the GDC-H is being used via the
+    // Credentials. The Credentials object is resolved here and must be passed to the
+    // EndpointContext.
     EndpointContext endpointContext =
-        EndpointContext.newBuilder()
+        settings
+            .getEndpointContext()
+            .toBuilder()
             .setServiceName(settings.getServiceName())
```
lqiu96, line 173: The alternative to this is to create a new private static method inside the {Client}StubSettings class and set it there.  Currently, we can't set it inside with the non-static `getServiceName()` method as initializing the defaults is inside a static method: https://github.com/googleapis/sdk-platform-java/blob/c1cd0d1deb74d7aa81f1ad7f6400d193f1464fac/showcase/gapic-showcase/src/main/java/com/google/showcase/v1beta1/stub/EchoStubSettings.java#L700-L710

gax-java/gax/src/main/java/com/google/api/gax/rpc/StubSettings.java
```diff
@@ -450,8 +468,21 @@ public B setClock(ApiClock clock) {
       return self();
     }
 
+    /**
+     * This is intended to be used by the generated Client StubSettings' classes. Explicitly setting
+     * this value may result in unintended client behavior.
+     */
+    @InternalApi
+    public B setServiceName(@Nonnull String serviceName) {
```
blakeli0, line 476: Why do we have to expose a setter for it? Actually I don't remember why we need a field for it in the first place. Can we continue to call `getServiceName()` whenever we need it? Which should be populated in the generated libraries.

gax-java/gax/src/main/java/com/google/api/gax/rpc/StubSettings.java
```diff
@@ -338,6 +352,8 @@ protected Builder(ClientContext clientContext) {
         this.quotaProjectId = getQuotaProjectIdFromClientContext(clientContext);
         this.gdchApiAudience = clientContext.getGdchApiAudience();
         this.universeDomain = clientContext.getUniverseDomain();
+        this.endpointContextBuilder =
```
blakeli0, line 355: I know this is inside `StubSettings.Builder`, but instead of passing around the builder, we should pass the `EndpointContext` and call `toBuilder()` every time we need to set something, so that we won't have a incomplete `EndpointContext` passing around which could easily cause bugs later.
This means we would need to refactor other places, like setting the whole EndpointContext when we build `StubSettings`, instead of setting each field of it in setters.

gax-java/gax/src/main/java/com/google/api/gax/rpc/ApiCallContext.java
```diff
@@ -66,6 +66,9 @@ public interface ApiCallContext extends RetryingContext {
   /** Returns a new ApiCallContext with the given Endpoint Context. */
   ApiCallContext withEndpointContext(EndpointContext endpointContext);
 
+  /** Returns the EndpointContext stored in the ApiCallContext */
```
blakeli0, line 69: Why do we need to expose `EndpointContext` here?

gax-java/gax/src/main/java/com/google/api/gax/rpc/StubSettings.java
lqiu96: ServiceName is used to construct the resolved endpoint via the template: `https://{serviceName}.{universeDomain}:443`. For GAPICs (specifically non-showcase clients), we are able to generate the serviceName and override the `getServiceName()`. But for showcase clients, they do not have a serviceName and the value of `gteServiceName()` is `""`. 

This setter is for our showcase tests and (I think) a couple unit tests. I'll take another look to see if I can write a wrapper so that we don't need to expose this setter, but my first attempt wasn't able to find a way.

gax-java/gax/src/main/java/com/google/api/gax/rpc/StubSettings.java
lqiu96: I'm passing around an incomplete Builder because `EndpointContext.build()` throws an IOException. Passing the builder avoids needing to catch an IOException inside each impacted setter (endpoint, mtls endpoint, etc).

gax-java/gax/src/main/java/com/google/api/gax/rpc/ApiCallContext.java
lqiu96: I think there is a use-case where an user passes in an already created ClientContext: https://github.com/googleapis/sdk-platform-java/blob/c1cd0d1deb74d7aa81f1ad7f6400d193f1464fac/gax-java/gax/src/main/java/com/google/api/gax/rpc/StubSettings.java#L317-L341

I simply use the EndpointContext that was created in ClientContext because there is unfortunately a single param that is resolved inside the ClientContext: the `usingGDCH` param.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
lqiu96: >  if a handwritten library creates GrpcCallContext directly, then any gRPC calls would fail with null pointer exception

It would if we don't create an empty EndpointContext. It'll NPE when it does something like `ClientContext.getDefaultCallContext().getEndpointContext()...`

gax-java/gax/src/main/java/com/google/api/gax/rpc/StubSettings.java
blakeli0: Exposing a public setter in `StubSettings` could be harmful and confusing to customers, and could result in unintended behavior as you mentioned in the comment, even if we mark it as `@InternalApi`. If it is just for showcase, I would rather have some hacks in the generator rather than Gax. e.g. We can have logics saying that `If package name is "com.google.showcase.v1beta1", then we generate getServiceName with "showcase"`.

gax-java/gax/src/main/java/com/google/api/gax/rpc/ApiCallContext.java
lqiu96: This has been removed.

gax-java/gax/src/main/java/com/google/api/gax/rpc/StubSettings.java
lqiu96: Create the EndpointContext in the constructor.

gax-java/gax/src/main/java/com/google/api/gax/rpc/StubSettings.java
lqiu96: Created an ExtendedStubSettings class that returns the serviceName.

gax-java/gax/src/main/java/com/google/api/gax/rpc/StubSettings.java
```diff
@@ -99,17 +96,35 @@ protected StubSettings(Builder builder) {
...
+  /**
+   * Attempt to build the EndpointContext from the Builder based on all the user configurations
+   * passed in.
+   *
+   * @throws RuntimeException if there is an issue building the EndpointContext
+   */
+  private EndpointContext buildEndpointContext(Builder builder) {
+    try {
+      return EndpointContext.newBuilder()
+          .setServiceName(getServiceName())
```
lqiu96, line 118: Pass the serviceName to the EndpointContext. For GAPICs, the child StubSettings should have a valid serviceName.

gax-java/gax/src/main/java/com/google/api/gax/rpc/StubSettings.java
```diff
@@ -71,17 +71,14 @@ public abstract class StubSettings<SettingsT extends StubSettings<SettingsT>> {
   private final HeaderProvider internalHeaderProvider;
   private final TransportChannelProvider transportChannelProvider;
   private final ApiClock clock;
-  private final String serviceName;
-  private final String endpoint;
-  private final String mtlsEndpoint;
```
lqiu96, line 76: These are stored in the EndpointContext

gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
```diff
@@ -113,6 +109,9 @@ public abstract class ClientContext {
   @Nullable
   public abstract String getQuotaProjectId();
 
+  /** Package-Private as this is to be shared to StubSettings */
+  abstract EndpointContext getEndpointContext();
```
lqiu96, line 113: Pass the EndpointContext to the StubSettings. This is to set allow the StubSettings to be re-created with the original settings since `clientContext.getEndpoint()` returns the resolved endpoint and not the original config.

StubSettings will re-created the EndpointContext in the constructor and should be re-created with the settings users passed in.

lqiu96: Alternative: We pass the EndpointContext to ApiCallContext. We can expose a new public getter to the ApiCallContext (i.e. `getEndpointContext()`), but that would add a public method.

gax-java/gax/src/main/java/com/google/api/gax/rpc/StubSettings.java
```diff
@@ -142,34 +157,36 @@ public final ApiClock getClock() {
...
 
-  // Intended for Internal Use and Overriden by generated ServiceStubSettings classes.
-  // Meant to be shared between StubSettings and ClientContext.
+  /**
+   * Marked with Internal Api and meant to overriden by the generated subclasses. This getter is
+   * used to set the serviceName to the EndpointContext. The value in generated StubSettings
+   * subclasses comes from the proto files.
+   *
+   * <p>This should be effectively treated as an abstract method.
+   */
```
blakeli0, line 166: We might considered it before, if it is meant to be overridden by sub classes, can we make it `protected`?

gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
blakeli0: > Pass the created EndpointContext to the StubSettings. This is to set allow the StubSettings to be re-created with the original users settings

Really like it! No more public methods!

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITEndpointContext.java
```diff
@@ -50,17 +130,72 @@ public void endpointResolution_default() throws InterruptedException, IOExceptio
...
+              .setCredentialsProvider(NoCredentialsProvider.create())
+              .setUniverseDomain(customUniverseDomain)
+              .build();
+      echoClient = EchoClient.create(echoSettings);
+      // If user configured the universe domain, the endpoint is constructed from it
+      Truth.assertThat(echoClient.getSettings().getEndpoint()).isEqualTo("test.random.com:443");
+      Truth.assertThat(echoClient.getSettings().getUniverseDomain())
+          .isEqualTo(customUniverseDomain);
+    } finally {
+      if (echoClient != null) {
```
blakeli0, line 171: Extract `echoClient` to a field and put these logic in the clean up method?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITEndpointContext.java
```diff
@@ -50,17 +130,72 @@ public void endpointResolution_default() throws InterruptedException, IOExceptio
...
+        echoClient.close();
+        echoClient.awaitTermination(
+            TestClientInitializer.AWAIT_TERMINATION_SECONDS, TimeUnit.SECONDS);
+      }
+    }
+  }
+
+  @Test
+  public void endpointResolution_userSetUniverseDomain() throws IOException, InterruptedException {
```
blakeli0, line 156:
```suggestion
  public void endpointResolution_userSetUniverseDomainAndNoUserSetEndpoint() throws IOException, InterruptedException {
```

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITEndpointContext.java
```diff
@@ -50,17 +130,72 @@ public void endpointResolution_default() throws InterruptedException, IOExceptio
...
   @Test
-  public void endpointResolution_userConfiguration() throws InterruptedException, IOException {
+  public void endpointResolution_userSetEndpoint() throws InterruptedException, IOException {
     String customEndpoint = "test.com:123";
     EchoClient echoClient = null;
     try {
       EchoSettings echoSettings =
-          EchoSettings.newBuilder()
+          ExtendedEchoSettings.newBuilder()
+              .setCredentialsProvider(NoCredentialsProvider.create())
```
blakeli0, line 139: I'm wondering if it's possible to override `getUniverseDomain()` in `NoCredentials`, so that we can test the universe domain validation as well.

blakeli0 (commented):
Looking really good! The implementation is pretty much good to go, most of my comments are for tests only.

gax-java/gax/src/main/java/com/google/api/gax/rpc/StubSettings.java
lqiu96: Ah yep, will update!

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITEndpointContext.java
lqiu96: Ah good idea. Will do!

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITEndpointContext.java
lqiu96: I ended up creating two new wrappers: `UniverseDomainCredentials` which wraps Credentials and `UniverseDomainCredentialsProvider` which wraps CredentialsProvider.

Showcase tests can set the universe domain to the Credentials so we can test the validation logic by checking that the universe domain matches.

blakeli0 (approved):

PR #2444, opened 2024-02-07
test: fix flaky HttpJsonDirectServerStreamingCallableTest.testOnResponseError

meltsufin:
* increase timeout from 2s to 30s to reduce the chance of DEADLINE_EXCEEDED before the NOT_FOUND message is received
* keep separate low timeout for the testDeadlineExceededServerStreaming test
* remove mockService reset to speedup tests, as it's no longer necessary

Fixes: #1842.


suztomo (approved):
Checked the test with a heavy load and it passed.

PR #2353, opened 2024-01-10
feat: autopopulate fields in the request

alicejli:
Follow up to: https://github.com/googleapis/sdk-platform-java/pull/2312

This adds autopopulation of fields within the Request message for protos that are configured. This currently only supports UUID4.

PR #2446, opened 2024-02-07
chore: refactor golden tests for autopopulation feature

alicejli:
As a clean-up follow up to https://github.com/googleapis/sdk-platform-java/pull/2353, this PR refactors the unit golden tests for the autopopulation feature to a separate, new proto called `auto_populate_field_testing.proto`.


gapic-generator-java/src/test/proto/auto_populate_field_testing.proto
```diff
@@ -0,0 +1,144 @@
...
+    option (google.api.http) = {
+      post: "/v1beta1/AutoPopulateFieldTesting:collect"
+      body: "*"
+    };
+  }
+
+  // This bidirectional streaming method, upon receiving a request on the stream, will pass the same
+  // content back on the stream. This method should not have any autopopulated fields.
+  rpc AutoPopulateFieldTestingChat(stream AutoPopulateFieldTestingEchoRequest) returns (stream AutoPopulateFieldTestingEchoResponse);
+}
```
lqiu96, line 68: These don't have autopopulated fields b/c they're streaming RPCs, right?

lqiu96:
Can you also check if there should be goldens for ServiceStubClassComposerTest in grpcrest?

lqiu96:
Besides the two questions above, LGTM.

gapic-generator-java/src/test/proto/auto_populate_field_testing.proto
alicejli: Yup, that's correct.

alicejli:
> Can you also check if there should be goldens for ServiceStubClassComposerTest in grpcrest?

I've added in the goldens for them. Thanks!

gapic-generator-java/src/test/proto/auto_populate_field_testing.proto
lqiu96: Could you add a quick blurb that this is b/c it is a streaming RPC? I saw that it has `AutoPopulateFieldTestingEchoRequest` and forgot about streaming.

lqiu96:
> > Can you also check if there should be goldens for ServiceStubClassComposerTest in grpcrest?
> 
> I've added in the goldens for them. Thanks!

I took a look at the grpc+rest composers and we don't think that the auto populate PR impacts those composers. We probably don't need goldens there, so you can remove those golden cases.

lqiu96 (approved):
LGTM, if you can make the two quick changes. Thanks!

PR #2451, opened 2024-02-08
chore: update dev guide to get version of gapic-showcase to install.

zhumin8:
This addition removes the trailing ANSI escape sequence (`\x1b[0m`) on the returned version.
This regex `\x1b\[[0-9;]*m` matches ANSI escape sequences.

### DIFF
--- a/showcase/README.md
+++ b/showcase/README.md
@@ -18,7 +18,7 @@ update to a compatible client version in `./WORKSPACE`.
 ```shell
 # Install the showcase server version defined in gapic-showcase/pom.xml
 cd showcase
-go install github.com/googleapis/gapic-showcase/cmd/gapic-showcase@v"$(cd gapic-showcase;mvn help:evaluate -Dexpression=gapic-showcase.version -q -DforceStdout)"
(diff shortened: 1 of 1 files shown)


showcase/README.md
```diff
@@ -18,7 +18,7 @@ update to a compatible client version in `./WORKSPACE`.
 ```shell
 # Install the showcase server version defined in gapic-showcase/pom.xml
 cd showcase
-go install github.com/googleapis/gapic-showcase/cmd/gapic-showcase@v"$(cd gapic-showcase;mvn help:evaluate -Dexpression=gapic-showcase.version -q -DforceStdout)"
+go install github.com/googleapis/gapic-showcase/cmd/gapic-showcase@v"$(cd gapic-showcase;mvn help:evaluate -Dexpression=gapic-showcase.version -q -DforceStdout |sed 's/\x1b\[[0-9;]*m//g')"
```
lqiu96, line 21: Thanks, I just ran command his and ran into the same error. Could you just a comment about the sed command at the end (removing ANSI escape codes)?

Also, I wonder if there is a better way to get the showcase version...

lqiu96 (approved):

PR #2453, opened 2024-02-08
chore: use template excludes in generation config

JoeWang1127:
In this PR:
- Modify `generate_prerequisite_files` to render `owlbot.py` using template excludes from `GenerationConfig`, rather than hard code the excludes.
- Add unit tests to verify `from_yaml` method.


library_generation/generate_composed_library.py
```diff
@@ -54,8 +54,8 @@ def generate_composed_library(
     :param library_path: the path to which the generated file goes
     :param library: a LibraryConfig object contained inside config, passed here
     for convenience and to prevent all libraries to be processed
-    :param output_folder:
-    :param versions_file:
+    :param output_folder: the folder to where tools go
```
diegomarquezp, line 57: Later, we should consider renaming this folder because it is not where we create the resulting library/repo (which is the actual output of the script (?))

library_generation/test/unit_tests.py
```diff
@@ -104,6 +105,97 @@ def test_eprint_valid_input_succeeds(self):
         # print() appends a `\n` each time it's called
         self.assertEqual(test_input + "\n", result)
 
+    # parameterized tests need to run from the class, see
+    # https://github.com/wolever/parameterized/issues/37 for more info.
```
diegomarquezp, line 109:
```suggestion
    # https://github.com/wolever/parameterized/issues/37 for more info.
    # This test confirms that a ValueError with an error message about a missing key (specified in the first parameter of each `parameterized` tuple) when parsing a test configuartion yaml (second parameter) will be raised.
```

diegomarquezp (approved):
LGTM, thank you

library_generation/generate_composed_library.py
JoeWang1127: I agree, the name is a bit of confusing.

Do you have a better name in mind?

diegomarquezp: Although anything created by the script is output, in a functional perspective the output we care about is the one we have a final use for. Just clarifying this rationale because "anything is output" can be valid too (and this is more of a suggestion in the end).

library_generation/test/unit_tests.py
JoeWang1127: done

library_generation/generate_composed_library.py
JoeWang1127: Thanks. I'll keep it as-is for now (I need sometime to think of a better name). 

We can come up with a new name later.

diegomarquezp: > I agree, the name is a bit of confusing.
> 
> Do you have a better name in mind?

I find it slightly tough to answer. What about `workspace`?

JoeWang1127: Yeah, this sounds good.

PR #2455, opened 2024-02-09
fix: Apiary Host returns user set host if set

lqiu96:
Bug: https://github.com/googleapis/java-bigquery/issues/3125 (Removing `Fixes:` as I don't want to close the ticket until BigQuery is able to pull in a new version of shared-deps).

Following guidance in doc for Apiary (ping me internally for link).

If the user configures the host to be the `DEFAULT_HOST` (a non-valid endpoint for any service using java-core), then it should construct a valid service endpoint back using the universe domain.


java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
```diff
@@ -862,6 +862,11 @@ public String getResolvedHost(String serviceName) {
    */
   @InternalApi
   public String getResolvedApiaryHost(String serviceName) {
+    // The host value set to DEFAULT_HOST if the user didn't configure a host. If the
+    // user set a host the library uses that value, otherwise, construct the host for the user.
```
suztomo, line 866: These should be in Javadoc.

suztomo: And the javadoc shouldn't start with "Temporarily used for ...". It starts with "Returns ..."

suztomo (approved):

java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java
lqiu96: Added.

Issue #2050, opened 2023-09-28
gapic-generator-java exits with `IllegalStateException` when there's no service in protos


JoeWang1127:
The generator should return normally with no-op when there's no service in protos.

For example: when generating from `google/cloud/alloydb/connectors/v1alpha`, the generator exited with the following error:
```
Exception in thread "main" java.lang.IllegalStateException: No services found to generate
        at com.google.common.base.Preconditions.checkState(Preconditions.java:512)
        at com.google.api.generator.gapic.protoparser.Parser.parse(Parser.java:172)
        at com.google.api.generator.gapic.Generator.generateGapic(Generator.java:30)
        at com.google.api.generator.Main.main(Main.java:28)
--java_gapic_out: protoc-gen-java_gapic: Plugin failed with status code 1.
```

suztomo:
I would implement the script in a way that it would not call gapic-generator-java if there's no need to call it.

JoeWang1127:
> I would implement the script in a way that it would not call gapic-generator-java if there's no need to call it.

I'm currently working on this feature. 

However, `generate_library.sh` can't decide whether to invoke generator or not. This information is stored in `BUILD.bazel`: if there's no `java_gapic_library` target, then `generate_library.sh` should not call the generator.

The grpc java plugin is implemented in a way that if there's no service to generate, it will exit normally without generating any code. I think the generator (which is also a protoc plugin) should do the same.

blakeli0:
We discussed this in the project meeting and I agree with Joe. In addition, we have other use cases that can benefit from this change. For example, as @mpeddada1 mentioned in https://github.com/googleapis/sdk-platform-java/issues/2048, we may need the generator to generate reflect-config.json for proto-only modules.

blakeli0:
This is working as intended, changing the type from `bug` to `feature request`.

diegomarquezp:
cc: @lqiu96 @JoeWang1127 @alicejli 
We will address this issue by strictly causing a no-op (no output is generated) whenever no services are generated. We will leave the noop + reflect-config.json as a follow up to be addressed separately (https://github.com/googleapis/sdk-platform-java/issues/2048)

Some considerations:
 - Update documentation to explain that the generator will not fail but simply no-op
 - Add a log statement (stderr) indicating that no services will be generated

PR #2460, opened 2024-02-12
feat: gapic-generator-java to perform a no-op when no services are detected

diegomarquezp:
Fixes https://github.com/googleapis/sdk-platform-java/issues/2050

Adds behavior to gracefully perform a NOOP if no services are contained in the generation request.

## Confimation in hermetic build scripts

From `generate_library.sh` against `google/cloud/alloydb/connectors/v1`
```
+ /usr/local/google/home/diegomarquezp/Desktop/sdk2/sdk3/sdk-platform-java/library_generation/output/protobuf-25.2/bin/protoc --experimental_allow_proto3_optional --plugin=protoc-gen-java_gapic=/usr/local/google/home/diegomarquezp/.pyenv/versions/3.11.0/lib/python3.11/site-packages/library_generation/gapic-generator-java-wrapper --java_gapic_out=metadata:/usr/local/google/home/diegomarquezp/Desktop/sdk2/sdk3/sdk-platform-java/library_generation/output/temp_preprocessed/java_gapic_srcjar_raw.srcjar.zip --java_gapic_opt=transport=grpc,rest-numeric-enums,grpc-service-config=,gapic-config=,api-service-config=google/cloud/alloydb/connectors/v1/connectors_v1.yaml google/cloud/alloydb/connectors/v1/resources.proto google/cloud/common_resources.proto
Apr 05, 2024 9:33:22 PM com.google.api.generator.gapic.protoparser.Parser parse
WARNING: No services found to generate. This will produce an empty zip file
Apr 05, 2024 9:33:22 PM com.google.api.generator.gapic.composer.ClientLibraryPackageInfoComposer generatePackageInfo
WARNING: Generating empty package info since no services were found
+ did_generate_gapic=true
+ zipinfo -t /usr/local/google/home/diegomarquezp/Desktop/sdk2/sdk3/sdk-platform-java/library_generation/output/temp_preprocessed/java_gapic_srcjar_raw.srcjar.zip
Empty zipfile. 
+ did_generate_gapic=false
+ [[ false == \t\r\u\e ]]

```

I made some changes to library_generation but I moved them to a follow up PR (https://github.com/googleapis/sdk-platform-java/pull/2599) so the checks can pass here.


blakeli0:
Throwing an exception and catch it just to perform no-op and/or logging is not a good practice. It would be better if we can do the same thing gracefully. For example, can we remove this [check](https://github.com/googleapis/sdk-platform-java/blob/main/gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java#L178), log the scenario, and return an empty `GapicContext`?

diegomarquezp:
From

```
+ /home/runner/work/sdk-platform-java/sdk-platform-java/output/protobuf-25.2/bin/protoc --experimental_allow_proto3_optional --plugin=protoc-gen-java_gapic=/home/runner/work/sdk-platform-java/sdk-platform-java/library_generation/gapic-generator-java-wrapper --java_gapic_out=metadata:/home/runner/work/sdk-platform-java/sdk-platform-java/output/temp_preprocessed/java_gapic_srcjar_raw.srcjar.zip --java_gapic_opt=transport=grpc,rest-numeric-enums,grpc-service-config=,gapic-config=,api-service-config=google/cloud/alloydb/connectors/v1/connectors_v1.yaml google/cloud/alloydb/connectors/v1/resources.proto google/cloud/common_resources.proto
Error: Exception in thread "main" java.lang.IllegalStateException: No services found to generate
	at com.google.common.base.Preconditions.checkState(Preconditions.java:512)
	at com.google.api.generator.gapic.protoparser.Parser.parse(Parser.java:178)
	at com.google.api.generator.gapic.Generator.generateGapic(Generator.java:30)
	at com.google.api.generator.Main.main(Main.java:28)
```

I confirmed they pass locally. Not sure how to prove it in a PR that updates both the generator and the hermetic build scripts

gapic-generator-java/src/test/java/com/google/api/generator/gapic/protowriter/WriterTest.java
```diff
@@ -86,4 +99,69 @@ public void reflectConfig_isWritten() throws IOException {
...
     }
   }
+
+  @Test
+  public void writePackageInfo_emptyPackageInfo_writesEmptyString() throws IOException {
+    String result = Writer.writePackageInfo(GapicPackageInfo.empty(), visitor, jarOutputStream);
+    assertThat(result).isEmpty();
+    jarOutputStream.finish();
+    jarOutputStream.flush();
+    jarOutputStream.close();
```
JoeWang1127, line 109: Could you extract these three lines to a clean up method?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/model/GapicContext.java
```diff
@@ -84,11 +87,27 @@ public static Builder builder() {
     return new AutoValue_GapicContext.Builder()
         .setMixinServices(Collections.emptyList())
         .setGapicMetadataEnabled(false)
-        .setRestNumericEnumsEnabled(false);
+        .setRestNumericEnumsEnabled(false)
+        .setIsEmpty(false);
+  }
+
+  public static GapicContext empty() {
```
JoeWang1127, line 94: Similar to the above comment, maybe a static variable is better?

gapic-generator-java/src/main/java/com/google/api/generator/engine/ast/PackageInfoDefinition.java
```diff
@@ -24,6 +24,8 @@
 public abstract class PackageInfoDefinition implements AstNode {
   public abstract String pakkage();
 
+  public abstract boolean isEmpty();
+
```
lqiu96, line 28: I wonder if it's possible/ makes sense, but rather than having to set a new field to represent the state (empty or not), I wonder if we can just expose a method that checks the state.

i.e. something like
```
public boolean shouldGenerateFile() {
    return !fileHeader().isEmpty() || !headerCommentStatements.isEmpty() || !annotations().isEmpty();
}
```

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/Composer.java
```diff
@@ -239,7 +239,11 @@ protected static List<GapicClass> addApacheLicense(List<GapicClass> gapicClassLi
         .collect(Collectors.toList());
   }
 
-  private static GapicPackageInfo addApacheLicense(GapicPackageInfo gapicPackageInfo) {
+  @VisibleForTesting
+  protected static GapicPackageInfo addApacheLicense(GapicPackageInfo gapicPackageInfo) {
```
lqiu96, line 243: Can this be package-private scope?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/model/GapicContext.java
```diff
@@ -59,6 +60,8 @@ public GapicMetadata gapicMetadata() {
   @Nullable
   public abstract com.google.api.Service serviceYamlProto();
 
+  public abstract boolean isEmpty();
```
lqiu96, line 63: Similar to above in PackageInfoDefinition. Would it be possible to expose a method like

```
public boolean containsServices() {
    return !services().isEmpty();
}
```

instead of adding a new field to GapicContext? If not, I think this is fine.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -175,7 +180,10 @@ public static GapicContext parse(CodeGeneratorRequest request) {
             mixinServices,
             transport);
 
-    Preconditions.checkState(!services.isEmpty(), "No services found to generate");
+    if (services.isEmpty()) {
+      LOGGER.warning("No services found to generate. This will produce an empty generated srcjar");
+      return GapicContext.empty();
+    }
```
lqiu96, line 186: qq, if the proto files passed to the generator only have messages (no services), will the srcjar still empty?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -1099,7 +1107,8 @@ private static Map<String, FileDescriptor> getFilesToGenerate(CodeGeneratorReque
     return fileDescriptors;
   }
 
-  private static String parseServiceJavaPackage(CodeGeneratorRequest request) {
+  @VisibleForTesting
+  protected static String parseServiceJavaPackage(CodeGeneratorRequest request) {
```
lqiu96, line 1111: nit: Can this be package-private?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -1132,13 +1141,15 @@ private static String parseServiceJavaPackage(CodeGeneratorRequest request) {
...
-        !Strings.isNullOrEmpty(finalJavaPackage), "No service Java package found");
+    String finalJavaPackage = "";
+    Optional<Entry<String, Integer>> finalPackageEntry =
+        processedJavaPackageCount.entrySet().stream().max(Map.Entry.comparingByValue());
+    if (finalPackageEntry.isPresent()) {
+      finalJavaPackage = finalPackageEntry.get().getKey();
+    }
+    if (Strings.isNullOrEmpty(finalJavaPackage)) {
+      LOGGER.warning("No service Java package found");
+    }
```
lqiu96, line 1152: qq (for my curiosity), what is this block doing?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protowriter/Writer.java
```diff
@@ -167,8 +181,12 @@ private static void writeSamples(
     }
   }
 
-  private static String writePackageInfo(
+  @VisibleForTesting
+  static String writePackageInfo(
       GapicPackageInfo gapicPackageInfo, JavaWriterVisitor codeWriter, JarOutputStream jos) {
+    if (gapicPackageInfo.isEmpty()) {
+      return "";
+    }
```
lqiu96, line 189: nit: Can this check be done in the calling method? That way this doesn't need to return `""`? It just won't call `writePackageInfo` is the packageInfo object is empty.

lqiu96:
Thanks @diegomarquezp! I think this looks good. To refresh my memory, this is to deal with the scenario where we try to generate files for messages in java-common-protos (the proto files have no services), right?

meltsufin:
@diegomarquezp I'm a little late to this, but is there a related issue I can review that this is trying to solve?

diegomarquezp:
> @diegomarquezp I'm a little late to this, but is there a related issue I can review that this is trying to solve?

@meltsufin https://github.com/googleapis/sdk-platform-java/issues/2050

I'm adding it to the pr description

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
diegomarquezp: Yes. In theory you could pass a list of message protos, but if there are no services then it will noop

JoeWang1127:
What happens if you try to unzip `temp-codegen.srcjar`? 
Does it fail or produce an empty folder (I prefer this one because we can add logic to remove empty folders using `remove_empty_files` in utilities.sh, similarly to grpc plugin)?
```
+ zipinfo -t /usr/local/google/home/diegomarquezp/Desktop/sdk2/sdk-platform-java/library_generation/output/temp_preprocessed/temp-codegen.srcjar
Empty zipfile. 
+ did_generate_gapic=false
```

gapic-generator-java/src/main/java/com/google/api/generator/Main.java
```diff
@@ -21,6 +21,7 @@
 import java.io.IOException;
 
 public class Main {
+
```
blakeli0, line 24: Can we revert this empty line? Same thing for the empty line in `Generator` below.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
blakeli0: Do we have to add this block if we can reach `services.isEmpty()` [later](https://github.com/googleapis/sdk-platform-java/pull/2460/files#diff-648bf4fdb398cabe3694d2e383a4b273f690c2279af3247ede8c4d0e584a624cR183)? If we do, can we simply change the check to a WARNING log?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protowriter/Writer.java
```diff
@@ -69,16 +68,14 @@ public static CodeGeneratorResponse write(
...
 
-    writeMetadataFile(context, writePackageInfo(gapicPackageInfo, codeWriter, jos), jos);
-    writeReflectConfigFile(gapicPackageInfo.packageInfo().pakkage(), reflectConfigInfo, jos);
-
-    try {
-      jos.finish();
-      jos.flush();
-    } catch (IOException e) {
-      throw new GapicWriterException(e.getMessage(), e);
+    if (!gapicPackageInfo.isEmpty()) {
```
blakeli0, line 71: Can we use `context.isEmpty()` instead of `gapicPackageInfo.isEmpty()`? So that maybe we don't have add `isEmpty()` to `GapicPackageInfo`?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protowriter/Writer.java
```diff
@@ -88,6 +85,23 @@ public static CodeGeneratorResponse write(
     return response.build();
   }
 
+  public static CodeGeneratorResponse write(
```
blakeli0, line 88: Where is this method called?

diegomarquezp: This is the production method used by `Generator`:

https://github.com/googleapis/sdk-platform-java/blob/56775a61a054fb67a15023481cd3d0a0cf2b0496/gapic-generator-java/src/main/java/com/google/api/generator/gapic/Generator.java#L35

The one I created is for testing purposes, where we can provide our own output stream.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
diegomarquezp: @lqiu96 I think it gets the java package that is not a mixin and was found to be referenced the most times in the file descriptors.

diegomarquezp: @blakeli0 I realized that WARNING is redundant, but we still need this logic because we need the `ResourceNames` to be ready before parsing (and counting) the services.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -175,7 +180,10 @@ public static GapicContext parse(CodeGeneratorRequest request) {
             mixinServices,
             transport);
 
-    Preconditions.checkState(!services.isEmpty(), "No services found to generate");
+    if (services.isEmpty()) {
+      LOGGER.warning("No services found to generate. This will produce an empty generated srcjar");
```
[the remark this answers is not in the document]
diegomarquezp: I think the srcjar creation is part of `protoc`. I tried the following main function:

```java
  public static void main(String[] args) throws IOException {
    ExtensionRegistry registry = ExtensionRegistry.newInstance();
    ProtoRegistry.registerAllExtensions(registry);
    CodeGeneratorRequest request = CodeGeneratorRequest.parseFrom(System.in, registry);
    CodeGeneratorResponse response = Generator.generateGapic(request);
    if (response.getFileCount() > 0) {
      // we only write to stdout when we have files to generate. This is in order
      // to avoid creating an empty srcjar
      response.writeTo(System.out);
    }
  }
```

This still produces an empty srcjar. I'm looking into where do we parse `java_gapic_out`

gapic-generator-java/src/main/java/com/google/api/generator/engine/ast/PackageInfoDefinition.java
```diff
@@ -39,15 +41,22 @@ public void accept(AstNodeVisitor visitor) {
 
   public static Builder builder() {
     return new AutoValue_PackageInfoDefinition.Builder()
+        .setIsEmpty(false)
         .setFileHeader(Collections.emptyList())
         .setHeaderCommentStatements(Collections.emptyList())
         .setAnnotations(Collections.emptyList());
   }
 
+  public static PackageInfoDefinition empty() {
```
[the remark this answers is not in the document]
diegomarquezp: Done

gapic-generator-java/src/test/java/com/google/api/generator/gapic/protowriter/WriterTest.java
diegomarquezp: Done

gapic-generator-java/src/main/java/com/google/api/generator/engine/ast/PackageInfoDefinition.java
diegomarquezp: Done, I modified other `isEmpty()` methods as well

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/Composer.java
diegomarquezp: Done

gapic-generator-java/src/main/java/com/google/api/generator/gapic/model/GapicContext.java
diegomarquezp: Done

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
diegomarquezp: Done

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protowriter/Writer.java
diegomarquezp: Done

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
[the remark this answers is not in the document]
diegomarquezp: Correction - this is done in `Writer.write`. I'll adjust the logic.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
diegomarquezp: I modified this to not generate a srcjar at all if no services are found

diegomarquezp:
> What happens if you try to unzip `temp-codegen.srcjar`? Does it fail or produce an empty folder (I prefer this one because we can add logic to remove empty folders using `remove_empty_files` in utilities.sh, similarly to grpc plugin)?
> 
> ```
> + zipinfo -t /usr/local/google/home/diegomarquezp/Desktop/sdk2/sdk-platform-java/library_generation/output/temp_preprocessed/temp-codegen.srcjar
> Empty zipfile. 
> + did_generate_gapic=false
> ```

We will now NOT generate a srcjar if no services are found. This means that the wrapper zip `java_gapic_srcjar_raw.srcjar.zip` will be empty. If we try to `unzip` it, it will fail (and the script will exit due to `set -e`), so we still need that `zipinfo` check. So to answer the question, it would not generate an empty folder, it would just fail.

gapic-generator-java/src/main/java/com/google/api/generator/engine/ast/PackageInfoDefinition.java
```diff
@@ -22,8 +22,14 @@
 
 @AutoValue
 public abstract class PackageInfoDefinition implements AstNode {
+  public static final PackageInfoDefinition EMPTY = builder().setPakkage("").build();
+
   public abstract String pakkage();
 
+  public boolean shouldGenerateFile() {
+    return pakkage().isEmpty();
+  }
```
lqiu96, line 31: Should this be `!pakkage().isEmpty()`? If so, this only won't generate only pakkage is set to `""`? 

Do we need to also check fileHeaders, headerCommentStatements, and annotations?

gapic-generator-java/src/main/java/com/google/api/generator/engine/writer/JavaWriterVisitor.java
```diff
@@ -1034,6 +1034,9 @@ public void visit(ClassDefinition classDefinition) {
 
   @Override
   public void visit(PackageInfoDefinition packageInfoDefinition) {
+    if (packageInfoDefinition.shouldGenerateFile()) {
+      return;
+    }
```
lqiu96, line 1039: Same thing: `!packageInfoDefinition.shouldGenerateFile()`? If we shouldn't generate, then we exit quickly?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/ClientLibraryPackageInfoComposer.java
```diff
@@ -44,7 +48,10 @@ public class ClientLibraryPackageInfoComposer {
   private static final String SERVICE_DESCRIPTION_HEADER_PATTERN = "Service Description: %s";
 
   public static GapicPackageInfo generatePackageInfo(GapicContext context) {
-    Preconditions.checkState(!context.services().isEmpty(), "No services found to generate");
+    if (context.services().isEmpty()) {
```
lqiu96, line 51: I think GapicContext has `containsServices()`. Probably can be `!context.containsServices()`?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/model/GapicPackageInfo.java
```diff
@@ -20,8 +20,16 @@
...
 public abstract class GapicPackageInfo {
+
+  public static final GapicPackageInfo EMPTY =
+      builder().setPackageInfo(PackageInfoDefinition.EMPTY).build();
+
   public abstract PackageInfoDefinition packageInfo();
 
+  public boolean isEmpty() {
+    return packageInfo().shouldGenerateFile();
+  }
```
lqiu96, line 31: nit: I see from the comment above that this is `A thin wrapper around PackageInfoDefinition`. If it makes sense, perhaps we can name it the same?

GapicPackageInfo's shouldGenerateFile() which calls PackageInfoDefinition's shouldGenerateFile()

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/ComposerTest.java
```diff
@@ -154,6 +168,22 @@ public void composeSamples_parseProtoPackage() {
     }
   }
 
+  @Test
+  public void testEmptyGapicContext_succeeds() {
```
JoeWang1127, line 172: From what I read from the test, you want to verify no exception is thrown in `Composer.composeServiceClasses(GapicContext.EMPTY)`

How about write tests verifying the result, rather assert the exception is null?

gapic-generator-java/src/test/java/com/google/api/generator/gapic/protowriter/WriterTest.java
```diff
@@ -86,4 +101,56 @@ public void reflectConfig_isWritten() throws IOException {
...
+            Collections.emptyList(),
+            "temp-codegen.srcjar",
+            jarOutputStream,
+            output);
+    assertTrue(output.size() == 0);
+    closeJarOutputStream();
+  }
+
+  @Test
+  public void productionWrite_emptyGapicContext_succeeds() throws IOException {
```
JoeWang1127, line 138: Same as comment on `testEmptyGapicContext_succeeds`.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protowriter/Writer.java
blakeli0: I see, thanks.

blakeli0:
> > What happens if you try to unzip `temp-codegen.srcjar`? Does it fail or produce an empty folder (I prefer this one because we can add logic to remove empty folders using `remove_empty_files` in utilities.sh, similarly to grpc plugin)?
> > ```
> > + zipinfo -t /usr/local/google/home/diegomarquezp/Desktop/sdk2/sdk-platform-java/library_generation/output/temp_preprocessed/temp-codegen.srcjar
> > Empty zipfile. 
> > + did_generate_gapic=false
> > ```
> 
> We will now NOT generate a srcjar if no services are found. This means that the wrapper zip `java_gapic_srcjar_raw.srcjar.zip` will be empty. If we try to `unzip` it, it will fail (and the script will exit due to `set -e`), so we still need that `zipinfo` check. So to answer the question, it would not generate an empty folder, it would just fail.

What creates the `java_gapic_srcjar_raw.srcjar.zip`? The generator or the hermetic build scripts? Can we not generate anything at all, which would be a true no-op? 
In general, I think the changes in `GapicContext` and `Parser` makes sense, but the we may not need changes in `Composer` and `Writer`. For example, can we return null `CodeGeneratorResponse` and skip [writing to disk](https://github.com/googleapis/sdk-platform-java/blob/main/gapic-generator-java/src/main/java/com/google/api/generator/Main.java#L29) all together?

gapic-generator-java/src/main/java/com/google/api/generator/engine/ast/PackageInfoDefinition.java
diegomarquezp: That's right. However we decided to have this PR being less invasive regarding package info, so we removed this change.

gapic-generator-java/src/main/java/com/google/api/generator/engine/writer/JavaWriterVisitor.java
diegomarquezp: The purpose of this change was to avoid producing a srcjar because this function will return a non-empty string even if we have no services. I now changed this to simply no-op as `Writer` will exit early and not create a file when no services are found.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/ClientLibraryPackageInfoComposer.java
diegomarquezp: Good catch, I used the `containsServices` instead.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/model/GapicPackageInfo.java
diegomarquezp: As mentioned above, changes to package info are being dropped in favor of a no-op at the `Writer` level

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/ComposerTest.java
diegomarquezp: Thanks, this simplifies the test. Done.

gapic-generator-java/src/test/java/com/google/api/generator/gapic/protowriter/WriterTest.java
diegomarquezp: Done

diegomarquezp:
> > > What happens if you try to unzip `temp-codegen.srcjar`? Does it fail or produce an empty folder (I prefer this one because we can add logic to remove empty folders using `remove_empty_files` in utilities.sh, similarly to grpc plugin)?
> > > ```
> > > + zipinfo -t /usr/local/google/home/diegomarquezp/Desktop/sdk2/sdk-platform-java/library_generation/output/temp_preprocessed/temp-codegen.srcjar
> > > Empty zipfile. 
> > > + did_generate_gapic=false
> > > ```
> > 
> > 
> > We will now NOT generate a srcjar if no services are found. This means that the wrapper zip `java_gapic_srcjar_raw.srcjar.zip` will be empty. If we try to `unzip` it, it will fail (and the script will exit due to `set -e`), so we still need that `zipinfo` check. So to answer the question, it would not generate an empty folder, it would just fail.
> 
> What creates the `java_gapic_srcjar_raw.srcjar.zip`? The generator or the hermetic build scripts? Can we not generate anything at all, which would be a true no-op? In general, I think the changes in `GapicContext` and `Parser` makes sense, but the we may not need changes in `Composer` and `Writer`. For example, can we return null `CodeGeneratorResponse` and skip [writing to disk](https://github.com/googleapis/sdk-platform-java/blob/main/gapic-generator-java/src/main/java/com/google/api/generator/Main.java#L29) all together?

That sounds good. I modified `Writer` to return `null` if the context is empty, rendering the changes in package info unnecessary.

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/ComposerTest.java
```diff
@@ -53,8 +56,13 @@ class ComposerTest {
           .build();
   private List<Sample> ListofSamples = Arrays.asList(new Sample[] {sample});
 
+  @Before
```
JoeWang1127, line 59: Could you change this tag to `@BeforeEach`?

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/ComposerTest.java
```diff
@@ -53,8 +56,13 @@ class ComposerTest {
...
+  public void initialSanityCheck() {
+    assertTrue(context.containsServices());
+  }
+
   @Test
-  void gapicClass_addApacheLicense() {
+  public void gapicClass_addApacheLicense_validInput_succeeds() {
```
JoeWang1127, line 65: Could you remove the `public` in the signature?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protowriter/Writer.java
```diff
@@ -69,15 +69,13 @@ public static CodeGeneratorResponse write(
       writeSamples(gapicClazz, getSamplePackage(gapicClazz), classPath, jos);
     }
 
+    // package info may come as null if we have no services, but we will exit
+    // this function early if so.
```
lqiu96, line 73: qq, is this also still true? I don't see `writePackageInfo()` exiting early.

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/ComposerTest.java
```diff
@@ -154,6 +162,16 @@ void composeSamples_parseProtoPackage() {
     }
   }
 
+  @Test
+  public void testEmptyGapicContext_doesNotThrow() {
+    Composer.composeServiceClasses(GapicContext.EMPTY);
```
lqiu96, line 167: nit: Can this be something like assert empty list or assert null instead?

gapic-generator-java/src/test/java/com/google/api/generator/gapic/protoparser/ParserTest.java
```diff
@@ -609,6 +611,17 @@ void parseNestedProtoTypeName() {
             "google.ads.googleads.v3.resources.MutateJob.MutateJobMetadata"));
   }
 
+  @Test
+  public void testParse_noServices_returnsEmptyGapicContext() {
+    GapicContext result = Parser.parse(CodeGeneratorRequest.newBuilder().build());
+    assertFalse(result.containsServices());
```
lqiu96, line 617: nit: to match the test method, can this be assert that the result == GapicContext.Empty?

lqiu96 (approved):
LGTM, added a few nits regarding the comments and tests

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/ComposerTest.java
diegomarquezp: Done

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/ComposerTest.java
diegomarquezp: Done

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -175,7 +179,10 @@ public static GapicContext parse(CodeGeneratorRequest request) {
             mixinServices,
             transport);
 
-    Preconditions.checkState(!services.isEmpty(), "No services found to generate");
+    if (services.isEmpty()) {
+      LOGGER.warning("No services found to generate. This will produce an empty zip file");
```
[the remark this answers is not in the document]
diegomarquezp: It's not correct anymore. I corrected the log entry. Thanks for the catch.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protowriter/Writer.java
diegomarquezp: Not true anymore either. I removed the comment.

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/ComposerTest.java
diegomarquezp: Done

gapic-generator-java/src/test/java/com/google/api/generator/gapic/protoparser/ParserTest.java
diegomarquezp: Done

blakeli0 (approved):

PR #2468, opened 2024-02-13
fix: Fix watchdog to start with WAITING state

### LINKED ISSUE #2498: Watchdog with automatic inbound flow control should start with WAITING state
There are 2 states in watchdog, IDLE and WAITING. IDLE indicates the stream is waiting for caller to take some action, and WAITING indicates the stream is waiting for server to return something. We update the state whenever there's some activity from the server or caller, and record timestamps when the state changes. When a stream is inactive for too long, watchdog will cancel the stream and throw an exception.

In the current implementation when we create a `WatchdogStream`, we always set the stream state to "IDLE". The "IDLE" state is flipped to "WAITING" state when caller calls `request()`. However, this implementation is assuming that automatic inbound flow control is disabled. In the case where automatic inbound flow control is enabled, caller won't call `request()` for the next response and in this case starting the `WatchdogStream` in the IDLE state is incorrect.

mutianf:
Watchdog should start with WAITING state, and only switch to `idle` if auto flow control was disabled. Before the fix, when auto flow control was enabled, we wait for server to return a response without calling `onRequest()` and watchdog would report the timeout exception because of idle timeout, which is incorrect and causes confusion.

Fixes #2498 ☕️


gax-java/gax/src/main/java/com/google/api/gax/rpc/Watchdog.java
```diff
@@ -220,6 +220,16 @@ public void onStartImpl(StreamController controller) {
             public void disableAutoInboundFlowControl() {
               Preconditions.checkState(
                   !hasStarted, "Can't disable automatic flow control after the stream has started");
+
+              synchronized (lock) {
```
igorbernstein2, line 224: I dont think you need to lock here. While we are in !hasStarted mode, the only relevant thread is the caller's

gax-java/gax/src/main/java/com/google/api/gax/rpc/Watchdog.java
```diff
@@ -220,6 +220,16 @@ public void onStartImpl(StreamController controller) {
...
+                // WatchdogStream is created with waiting state. If auto flow control is disabled,
+                // set the state to be IDLE, which will be updated to WAITING when onRequest() is
+                // called.
+                if (state == State.WAITING) {
+                  state = State.IDLE;
+                }
```
igorbernstein2, line 230: I dont think you need the conditional either. The state transition for a response observer is strictly: constructed -> start -> everything else
So you should just blindly overwrite it here

igorbernstein2 (changes requested):

gax-java/gax/src/main/java/com/google/api/gax/rpc/Watchdog.java
```diff
@@ -194,7 +194,7 @@ class WatchdogStream<ResponseT> extends StateCheckingResponseObserver<ResponseT>
     private volatile StreamController innerController;
 
     @GuardedBy("lock")
-    private State state = State.IDLE;
```
igorbernstein2, line 197: Please add comment that by default that a stream has automatic inbound flow control enabled. So the stream will not wait for the caller to request() a message, so thats why this defaults to WAITING

gax-java/gax/src/main/java/com/google/api/gax/rpc/Watchdog.java
```diff
@@ -220,6 +220,12 @@ public void onStartImpl(StreamController controller) {
             public void disableAutoInboundFlowControl() {
               Preconditions.checkState(
                   !hasStarted, "Can't disable automatic flow control after the stream has started");
+
+              // WatchdogStream is created with waiting state. If auto flow control is disabled,
+              // set the state to be IDLE, which will be updated to WAITING when onRequest() is
+              // called.
```
igorbernstein2, line 226: I think you want to move the first part of this info to the initialization of the variable and keep the second part here:
"When a caller disables automatic flow control, they are responsible for requesting the first message. So the initial state for the watchdog is idle.

igorbernstein2 (approved):
LGTM after the doc comments are resolved

gax-java/gax/src/test/java/com/google/api/gax/rpc/WatchdogTest.java
```diff
@@ -154,6 +154,31 @@ public void testTimedOutBeforeStart() throws InterruptedException {
...
+    call1.getController().getObserver().onError(new CancellationException("cancelled"));
+
+    Throwable actualError = null;
+    try {
+      downstreamObserver.done.get();
+    } catch (ExecutionException e) {
+      actualError = e.getCause();
+    }
+    assertThat(actualError).isInstanceOf(WatchdogTimeoutException.class);
+    assertThat(actualError.getMessage()).contains("waiting for next response");
```
lqiu96, line 179: qq, why is this a WatchdogTimeoutException and not the CancellationException above?

lqiu96:
Thanks @mutianf, I think the changes make sense and lgtm. I'm not the most familiar with the Watchdog service so I just have a few questions about the test and the watchdog flow in general. Would you also be able to create an issue in this repo, add some details, and link it to this PR? Thanks!

gax-java/gax/src/test/java/com/google/api/gax/rpc/WatchdogTest.java
```diff
@@ -154,6 +154,31 @@ public void testTimedOutBeforeStart() throws InterruptedException {
...
+    AutoFlowControlObserver<String> downstreamObserver = new AutoFlowControlObserver<>();
+
+    autoFlowControlCallable.call("request", watchdog.watch(downstreamObserver, waitTime, idleTime));
+    MockServerStreamingCall<String, String> call1 = autoFlowControlCallable.popLastCall();
+
+    clock.incrementNanoTime(idleTime.toNanos() + 1);
+    watchdog.run();
+    assertThat(downstreamObserver.done.isDone()).isFalse();
+    assertThat(call1.getController().isCancelled()).isTrue();
+    call1.getController().getObserver().onError(new CancellationException("cancelled"));
```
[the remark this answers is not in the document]
mutianf: Yes, it's cancelled by watchdog on line 349 in cancelIfStale() method

gax-java/gax/src/test/java/com/google/api/gax/rpc/WatchdogTest.java
mutianf: Calling onError on the observer will call onErrorImpl in the WatchdogStream (line 299). And if the stream is already cancelled by watchdog, the error on this watchdog stream will be the WatchdogException (line 348).

mutianf:
Thanks @lqiu96, I created https://github.com/googleapis/sdk-platform-java/issues/2498 and responded to your questions, PTAL!

gax-java/gax/src/test/java/com/google/api/gax/rpc/WatchdogTest.java
lqiu96: > the error on this watchdog stream will be the WatchdogException

Gothca. I took a look at the lines again and just have a clarifying quesiton. The WatchDogException you linked above is from the `cancelIfStale()` call, right? If the stream was already cancelled from watchdog, why is `cancelIfStale()` still run? I think it should have been removed from the openstreams from the `onErrorImpl` call right?

mutianf: watchdog is a runnable that calls cancelIfStale (line 131). Which in this test is called on line 167 `watchdog.run()`

lqiu96: This was resolved in an offline discussion with Mattie. 

The follow question was how the CancellationException became a WatchdogTimeoutException. It's from this line: https://github.com/googleapis/sdk-platform-java/blob/8dc7c52e63f23e9be6dde82aa54f9f077a80b2f8/gax-java/gax/src/main/java/com/google/api/gax/rpc/Watchdog.java#L300-L303

lqiu96 (approved):

PR #2485, opened 2024-02-14
feat: Universe Domain Environment Variable Support

lqiu96:
Support for reading the `GOOGLE_CLOUD_UNIVERSE_DOMAIN` env var.

Since we are still on JUnit 4 and there is no dedicated support for testing Env Vars, we created a new step in the CI to test this Env Var test:
image


lqiu96:
On a second look, I think this might cause issues when running this locally. This test will fail if the local environment does not have the `GOOGLE_CLOUD_UNIVERSE_DOMAIN` env var set. I think I'll change this to a maven profile that needs to be activated (and is disabled by default).

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -283,6 +285,11 @@ String mtlsEndpointResolver(
     }
 
     public EndpointContext build() throws IOException {
+      // If the universe domain wasn't configured explicitly in the settings, check the
```
blakeli0, line 288: Can we move the logic to `determineUniverseDomain()`?

gax-java/gax/src/test/java/com/google/api/gax/rpc/EndpointContextTest.java
```diff
@@ -340,6 +340,21 @@ public void endpointContextBuild_gdchFlow_noUniverseDomain_customEndpoint() thro
         .isEqualTo(Credentials.GOOGLE_DEFAULT_UNIVERSE);
   }
 
+  // This Universe Domain should match the `GOOGLE_CLOUD_UNIVERSE_DOMAIN` Env Var
+  // For this test running locally or in CI, check that the Env Var is set properly.
+  // This test should only run when the maven profile `EnvVarTest` is enabled.
+  @Test
+  public void endpointContextBuild_universeDomainEnvVarSet() throws IOException {
```
blakeli0, line 347: Can you please add one more test case that if universe domain is set on both ClientSettings and Env Var, then ClientSettings should take precedence.

lqiu96: Yep, added a test case that will be run with `-PenvVarTest`

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
lqiu96: Sure, moved it to `determineUniverseDomain()`

PR #2500, opened 2024-02-23
feat: Introduce OpenTelemetry Metrics Recording

lqiu96:
Builds off of https://github.com/googleapis/sdk-platform-java/pull/2433, based on the design in go/java-gapic-otel-metrics-design.

Discovered two issues via showcase tests:
1. https://github.com/googleapis/sdk-platform-java/issues/2502
2. https://github.com/googleapis/sdk-platform-java/issues/2503

These issues are not blocking for this PR.


gax-java/gax/BUILD.bazel
```diff
@@ -18,6 +18,8 @@ _COMPILE_DEPS = [
     "@com_google_code_findbugs_jsr305//jar",
     "@com_google_errorprone_error_prone_annotations//jar",
     "@com_google_guava_guava//jar",
+    "@io_opentelemetry_opentelemetry_api//jar",
```
blakeli0, line 21: Is anything going to break if we don't add the dependencies to this file?

lqiu96: I added these deps to the BUILD.bazel file since it's was failing test the golden ITs (the `bazelisk test //...` command. I'm not sure how the golden ITs are linked to gax's BUILD.bzel file + dependencies.properties, but added it to resolve the issue (at least temporarily).

gax-java/gax/src/test/java/com/google/api/gax/tracing/OpentelemetryMetricsRecorderTest.java
```diff
@@ -0,0 +1,271 @@
...
+
+    // Meter Creation
+    Meter meter =
+        openTelemetry
+            .meterBuilder("OtelMetricsRecorderTest")
+            .setInstrumentationVersion(GaxProperties.getGaxVersion())
+            .build();
+
+    otelMetricsRecorder = new OpentelemetryMetricsRecorder(meter);
+  }
```
lqiu96, line 88: On the fence about using an InMemory(Reader|Exporter) here. Might not be needed since I believe this should be covered in showcase tests, but provides the extra benefit of testing each individual metric in isolation (multiple invocations + each invocation records attributes)..

gax-java/gax/src/main/java/com/google/api/gax/tracing/OpentelemetryMetricsRecorder.java
```diff
@@ -0,0 +1,145 @@
...
+   *
+   * <p>The operation count should always be 1 and this should be invoked once.
+   *
+   * @param count The number of operations made
+   * @param attributes Map of the attributes to store
+   */
+  @Override
+  public void recordOperationCount(long count, Map<String, String> attributes) {
+    operationCountRecorder.add(count, toOtelAttributes(attributes));
+  }
```
lqiu96, line 136: Both Operation record calls should only occur once. Do we want to validate this/ throw an exception if recordOperation(...) is invoked more than once?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
```diff
@@ -0,0 +1,490 @@
...
+    httpClient.awaitTermination(TestClientInitializer.AWAIT_TERMINATION_SECONDS, TimeUnit.SECONDS);
+  }
+
+  private void verifyMetricData(
+      List<MetricData> metricDataList, int operationCount, int attemptCount) {
+    for (MetricData metricData : metricDataList) {
+      Data<?> data = metricData.getData();
+      List<PointData> points = new ArrayList<>(data.getPoints());
+      Truth.assertThat(points.size()).isEqualTo(1);
+      PointData pointData = points.get(0);
```
lqiu96, line 143: When testing this, all the metrics have a list of PointData entries and the size of the list is one. My assumption with this is that normally the OtelSdk is created with a PeriodicMetricReader as opposed to the InMemoryMetricsReader used in this test. The PeriodicMetricReader would be reading the data as set intervals and each interval would be a new PointData object (i.e. If the interval is 10s and data is read over 60s, there would be a list of 6 PointData entries).

I think the `collectAllMetrics()` aggregates the PointData values together for each metrics and this is why there is only one entry in the list. I couldn't find anything in the Otel docs to confirm this though.

gax-java/gax/src/main/java/com/google/api/gax/tracing/OpentelemetryMetricsRecorder.java
```diff
@@ -0,0 +1,153 @@
...
+   * <ul>
+   *   <li>Attempt Latency: Histogram
+   *   <li>Operation Latency: Histogram
+   *   <li>Attempt Count: Counter
+   *   <li>Operation Count: Counter
+   * </ul>
+   *
+   * @param meter Otel Meter used to create the instruments of measurements
+   */
+  public OpentelemetryMetricsRecorder(Meter meter) {
```
blakeli0, line 69: This should take an instance of `OpenTelemetry`, and we create a `Meter` in the constructor.

gax-java/gax/src/main/java/com/google/api/gax/tracing/OpentelemetryMetricsRecorder.java
```diff
@@ -0,0 +1,153 @@
...
+    this.attemptLatencyRecorder =
+        meter
+            .histogramBuilder("attempt_latency")
```
blakeli0, line 72: The metric name should be of format `{service_name}/{metric_name}`, can we take `service_name` as a constructor parameter and build the full metric name with it?

gax-java/gax/src/main/java/com/google/api/gax/tracing/OpentelemetryMetricsRecorder.java
blakeli0: I think we should rely on the existing framework. If there is a existing bug, it should be the responsibility of the recorder to validate it as well, we should probably validate it in the `MetricsTracer`.

gax-java/pom.xml
```diff
@@ -158,6 +158,13 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
+      <dependency>
+        <groupId>io.opentelemetry</groupId>
+        <artifactId>opentelemetry-bom</artifactId>
+        <version>1.34.1</version>
```
blakeli0, line 164: I think we already have it in the [shared-dependencies](https://github.com/googleapis/sdk-platform-java/blob/1463d645edf96087335618542f7bbca722a97b71/java-shared-dependencies/third-party-dependencies/pom.xml#L41), can we try to remove it from there?

showcase/pom.xml
```diff
@@ -38,6 +38,13 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
+      <dependency>
+        <groupId>io.opentelemetry</groupId>
+        <artifactId>opentelemetry-bom</artifactId>
+        <version>1.34.1</version>
```
blakeli0, line 44: I don't think we need it since it's already in Gax.

gax-java/gax/src/main/java/com/google/api/gax/tracing/OpentelemetryMetricsRecorder.java
lqiu96: Makes sense. Updating.

gax-java/gax/src/main/java/com/google/api/gax/tracing/OpentelemetryMetricsRecorder.java
lqiu96: IIUC, the user is going to be creating the the MetricsRecorder. Should they be expected to supply the serviceName in themselves?

gax-java/pom.xml
lqiu96: Ah yeah, let me try and remove.

gax-java/gax/src/main/java/com/google/api/gax/tracing/OpentelemetryMetricsRecorder.java
blakeli0: Customers wouldn't do it by themselves. For Spanner, it is going to pass it in java-spanner. For pure GAPICs, I'm expecting that we are going to have a "enableOtelMetrics" in ClientSetting, and we will create an `OpentelemetryMetricsRecorder` in Gax for them during client initialization if it is enabled.

gax-java/pom.xml
lqiu96: It seems like only two jars were added in shared-deps: https://github.com/googleapis/sdk-platform-java/blob/1463d645edf96087335618542f7bbca722a97b71/java-shared-dependencies/third-party-dependencies/pom.xml#L185-L194

CC: @JoeWang1127 IIRC you mentioned this to me before. What was the reason for only these and not the BOM?

lqiu96: Hmm, also. It seems the the pom for Gax flows like Gax -> Gax-Parent -> Generator-Pom-Parent -> Shared-Configs. I might be missing where it's pulling in shared-deps (or does it?)

JoeWang1127: > What was the reason for only these and not the BOM?

I thought only these two artifacts are needed.

> It seems the the pom for Gax flows like Gax -> Gax-Parent -> Generator-Pom-Parent -> Shared-Configs

I think this is right, everything you add to the gax, it's eventually be part of shared-dependencies.

gax-java/gax/src/main/java/com/google/api/gax/tracing/OpentelemetryMetricsRecorder.java
lqiu96: I see, makes sense. I'll a second param to the ctor.

gax-java/pom.xml
lqiu96: > I thought only these two artifacts are needed.

Gotcha. I'm looking to add `opentelemetry-sdk-testing` with test scope for now and I think in the future we'll need to add `opentelemetry-sdk`. Would you have an objections if update shared-deps to pull in the BOM?

> I think this is right, everything you add to the gax, it's eventually be part of shared-dependencies.

If I'm understanding this correctly, this means that we'll need to add it to both gax and shared-deps, right? They should have the same version as renovate-bot should update them together?

showcase/pom.xml
lqiu96: I think I can remove this if shared-deps brings in otel-bom: https://github.com/googleapis/sdk-platform-java/pull/2500#discussion_r1513065599

gax-java/pom.xml
blakeli0: We should not need `opentelemetry-sdk`, the only thing we need should be `opentelemetry-api`. Is there a code reference to anything other than `opentelemetry-api`?

lqiu96: Discussed offline. We don't need to create the OpenTelemetry client so no need to have the `opentelemetry-sdk` package.

gax-java/gax/src/test/java/com/google/api/gax/tracing/OpentelemetryMetricsRecorderTest.java
lqiu96: Revert to not use InMemory for this unit test.

gax-java/gax/BUILD.bazel
lqiu96: Just keep the opentelemetry-api

showcase/pom.xml
lqiu96: Shared-Deps will have Otel Bom from Gax which is pulled in. No need to define it here.

gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientSettings.java
```diff
@@ -120,6 +122,16 @@ public final String getGdchApiAudience() {
...
 
+  /**
+   * Gets the configured {@link ApiTracerFactory} that will be used to generate traces for
+   * operations.
+   */
+  @BetaApi("The surface for tracing is not stable yet and may change in the future.")
+  @Nonnull
+  public ApiTracerFactory getTracerFactory() {
+    return stubSettings.getTracerFactory();
+  }
```
lqiu96, line 133: Remove this.

gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientSettings.java
```diff
@@ -284,6 +296,16 @@ public B setGdchApiAudience(@Nullable String gdchApiAudience) {
...
 
+    /**
+     * Sets the ApiTracerFactory for the client instance. To enable default metrics, users need to
+     * create an instance of metricsRecorder and pass it to the metricsTracerFactory, and set it
+     * here.
+     */
+    public B setTracerFactory(@Nullable ApiTracerFactory tracerFactory) {
+      stubSettings.setTracerFactory(tracerFactory);
+      return self();
+    }
```
lqiu96, line 307: Remove this.

gax-java/dependencies.properties
```diff
@@ -39,6 +39,7 @@ maven.com_google_api_grpc_proto_google_common_protos=com.google.api.grpc:proto-g
 maven.com_google_api_grpc_grpc_google_common_protos=com.google.api.grpc:grpc-google-common-protos:2.33.0
 maven.com_google_auth_google_auth_library_oauth2_http=com.google.auth:google-auth-library-oauth2-http:1.23.0
 maven.com_google_auth_google_auth_library_credentials=com.google.auth:google-auth-library-credentials:1.23.0
+maven.io_opentelemetry_opentelemetry_api=io.opentelemetry:opentelemetry-api:1.34.1
```
blakeli0, line 42:
```suggestion
maven.io_opentelemetry_opentelemetry_api=io.opentelemetry:opentelemetry-api:1.35.0
```

gax-java/gax/src/main/java/com/google/api/gax/tracing/OpenTelemetryMetricsRecorder.java
```diff
@@ -0,0 +1,161 @@
...
+   *   <li>Operation Count: Counter
+   * </ul>
+   *
+   * @param openTelemetry OpenTelemetry instance
+   * @param serviceName Service Name
+   */
+  public OpenTelemetryMetricsRecorder(OpenTelemetry openTelemetry, String serviceName) {
+    Meter meter =
+        openTelemetry
+            .meterBuilder("Gax-OtelMetrics")
```
blakeli0, line 75: Let's use `gax-java`. Based on the [Javadoc](https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/metrics/MeterProvider.java#L21-L30), the value here should be the library name.

lqiu96: Sure, updated.

gax-java/gax/src/main/java/com/google/api/gax/tracing/OpenTelemetryMetricsRecorder.java
```diff
@@ -0,0 +1,161 @@
...
+/**
+ * OpenTelemetry implementation of recording metrics. This implementation collections the
+ * measurements related to the lifecyle of an RPC.
+ *
+ * <p>For the Otel implementation, an attempt is a single RPC invocation and an operation is the
+ * collection of all the attempts made before a response is returned (either as a success or an
+ * error). A single call (i.e. `EchoClient.echo()`) should have an operation_count of 1 and may have
+ * an attempt_count of 1+ (depending on the retry configurations).
+ */
+public class OpenTelemetryMetricsRecorder implements MetricsRecorder {
```
blakeli0, line 53:
```suggestion
@BetaApi
@InternalApi
public class OpenTelemetryMetricsRecorder implements MetricsRecorder {
```

blakeli0 (approved):

gax-java/gax/src/main/java/com/google/api/gax/tracing/OpenTelemetryMetricsRecorder.java
```diff
@@ -0,0 +1,163 @@
...
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package com.google.api.gax.tracing;
+
+import com.google.api.gax.core.GaxProperties;
```
blakeli0, line 33:
```suggestion
import com.google.api.core.BetaApi;
import com.google.api.core.InternalApi;
import com.google.api.gax.core.GaxProperties;
```

PR #2509, opened 2024-02-27
feat: improve batching summary errors

igorbernstein2:
To avoid data loss, the Batching api throws an exception when the batcher is closed when at least 1 entry failed. To help debugging, the BatchingException tries to be helpful by giving some details about the errors. Since the Batcher lifetime can extend indefinitely, the Batcher implementation tries to keep a bound on the amount of resources it uses to track the errors. Previously it would only track exception types and counts. The idea being that if the customer has the need for fine grained details, the customer can retrieve the details from the ApiFuture that was returned when an entry was added. However this hasn't panned out and creates confusion.

This PR stores a sample of the error messages. The sample is by default capped to 50 entry and 50 rpc error messages. This can be adjusted by setting system properties


gax-java/gax/src/main/java/com/google/api/gax/batching/BatcherStats.java
```diff
@@ -52,6 +54,13 @@ class BatcherStats {
   private final Map<Class, Integer> entryExceptionCounts = new HashMap<>();
   private final Map<Code, Integer> entryStatusCounts = new HashMap<>();
 
+  private final EvictingQueue<String> sampleOfRpcErrors =
+      EvictingQueue.create(
+          Integer.getInteger("com.google.api.gax.batching.errors.max-samples", 50));
```
blakeli0, line 59: Can I get more info about this system property `com.google.api.gax.batching.errors.max-samples`? Like where is it documented?

Separately, can we extract `Integer.getInteger("com.google.api.gax.batching.errors.max-samples", 50)` to a common field?

igorbernstein2: Added some doc and extracted the constant

blakeli0 (approved):

PR #2539, opened 2024-03-06
chore: add `jq` to hermetic build image definition

diegomarquezp:
From https://github.com/diegomarquezp/google-cloud-java/actions/runs/8178920720/job/22363860250

```
+ /usr/local/lib/python3.11/site-packages/library_generation/owlbot/bin/write_clirr_ignore.sh /usr/local/lib/python3.11/site-packages/library_generation
/workspace/google-cloud-java/java-datastore/proto-google-cloud-datastore-v1 /workspace/google-cloud-java/java-datastore
/workspace/google-cloud-java/java-datastore/proto-google-cloud-datastore-v1/src/main/java /workspace/google-cloud-java/java-datastore/proto-google-cloud-datastore-v1 /workspace/google-cloud-java/java-datastore
/usr/local/lib/python3.11/site-packages/library_generation/owlbot/bin/write_clirr_ignore.sh: line 29: jq: command not found
Library postprocessing failed
Error: Process completed with exit code 1.
```

This PR adds jq to the docker image

## Cause
`write_clirr_ignore` only acts on the repo whenever there is no clirr_ignored_differences found. This occurs when generating a new library from scratch, and we've only testing against existing libraries (i.e. we didn't reach this call to `jq` in any of our previous testing).

https://github.com/googleapis/sdk-platform-java/blob/92e91bc0d91a2619a87abf9b072229a3ab3f4d91/library_generation/owlbot/bin/write_clirr_ignore.sh#L24-L29

### DIFF
--- a/.cloudbuild/library_generation/library_generation.Dockerfile
+++ b/.cloudbuild/library_generation/library_generation.Dockerfile
@@ -17,7 +17,7 @@ FROM gcr.io/cloud-devrel-public-resources/python
 
 # install tools
 RUN apt-get update && apt-get install -y \
-	unzip openjdk-17-jdk rsync maven \
+	unzip openjdk-17-jdk rsync maven jq \
 	&& apt-get clean
 
 COPY library_generation /src


.cloudbuild/library_generation/library_generation.Dockerfile
```diff
@@ -17,7 +17,7 @@ FROM gcr.io/cloud-devrel-public-resources/python
 
 # install tools
 RUN apt-get update && apt-get install -y \
-	unzip openjdk-17-jdk rsync maven \
+	unzip openjdk-17-jdk rsync maven jq \
```
blakeli0, line 20: What's the use case for it? Do we need to process Json in command line? In general, we should try to limit the dependencies in our docker images, because every dependency would bring more potential security concerns. e.g. I see that the [latest JQ release notes](https://github.com/jqlang/jq/releases/tag/jq-1.7.1) contains two CVE fixes.

blakeli0: I see, I thought this is something we introduced, let's keep it as it is then.

blakeli0 (approved):

PR #2531, opened 2024-03-01
feat: get PR description from googleapis commits

JoeWang1127:
In this PR:
- Add `generate_pr_description.py` to get commit messages from googleapis repository.
  - Parse configuration to get latest googleapis commit (C) and proto_path (P).
  - Combine commit message from C to a baseline commit (B), exclusively. Only commits that change files in P are considered.
- Add unit tests for utility functions.
- Add integration tests for `generate_pr_description.py`.

This is [step 5](https://docs.google.com/document/d/1JiCcG3X7lnxaJErKe0ES_JkyU7ECb40nf2Xez3gWvuo/edit?pli=1&tab=t.0#bookmark=id.g8qkyq11ygpx) in milestone 2 of hermetic build project.

PR #2548, opened 2024-03-08
chore: write description to a file

JoeWang1127:
In this PR:
- Write PR description into a file which lives the same directory with the configuration.

Context:
After #2531, we can generate pull request description from generation config and a baseline commit.

However, the return value (string) can not be retrieved when running in docker environment. Therefore, write the description to a file. Put multi-line description to a file also helps editing the pr body ([link](https://github.com/cli/cli/issues/595#issuecomment-608372734)).

### DIFF
--- a/library_generation/test/integration_tests.py
+++ b/library_generation/test/integration_tests.py
@@ -203,7 +203,7 @@ class IntegrationTest(unittest.TestCase):
             if sub_dir.is_file():
                 continue
             repo = sub_dir.name
-            if repo == "golden":
+            if repo == "golden" or repo == "java-bigtable":
                 continue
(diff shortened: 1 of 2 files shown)


library_generation/test/integration_tests.py
```diff
@@ -203,7 +203,7 @@ def __get_config_files(cls, path: str) -> List[tuple[str, str]]:
             if sub_dir.is_file():
                 continue
             repo = sub_dir.name
-            if repo == "golden":
+            if repo == "golden" or repo == "java-bigtable":
```
JoeWang1127, line 206: Disable integration test in java-bigtable ([issue](https://github.com/googleapis/sdk-platform-java/issues/2549))

PR #2554, opened 2024-03-11
chore: remove `google/shopping/type/types.proto` when generating shopping APIs

JoeWang1127:
After #2553 is merged, we do not need to add `google/shopping/type/types.proto` in `generate_library.sh` as a special case.


diegomarquezp:
The `owl-bot-copy` [output](https://paste.googleplex.com/6219027103350784) does not show `rm` statements for some reason

blakeli0 (approved):

PR #2555, opened 2024-03-12
feat: skip build only commit

JoeWang1127:
In this PR:
- When generate pull request description, do not include a commit that only changes `BUILD.bazel`.

### DIFF
--- a/library_generation/generate_pr_description.py
+++ b/library_generation/generate_pr_description.py
@@ -148,8 +148,8 @@ def __get_commit_messages(
 def __filter_qualified_commit(paths: Dict[str, str], commit: Commit) -> (Commit, str):
     """
     Returns a tuple of a commit and libray_name.
-    A qualified commit means at least one file changes in that commit is
-    within the versioned proto_path in paths.
+    A qualified commit means at least one file, excluding BUILD.bazel, changes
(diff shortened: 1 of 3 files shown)


diegomarquezp (approved):

library_generation/generate_pr_description.py
```diff
@@ -158,7 +158,7 @@ def __filter_qualified_commit(paths: Dict[str, str], commit: Commit) -> (Commit,
     """
     for file in commit.stats.files.keys():
         versioned_proto_path = find_versioned_proto_path(file)
-        if versioned_proto_path in paths:
+        if versioned_proto_path in paths and (not file.endswith("BUILD.bazel")):
```
blakeli0, line 161: Do we have a unit test for this case? Or it is tested indirectly in the golden test?

JoeWang1127: It's tested in the golden test.

I changed the commit in the test configuration file and `chore(ruby)` didn't make it in the golden file.

blakeli0: OK, it would be great if we can add unit tests for this class in the future so that don't have to update the goldens for un-related changes, but it's fine for now.

blakeli0 (approved):

PR #2557, opened 2024-03-13
chore: refactor utility functions

JoeWang1127:
In this PR:
- Code Refactor 
  - Refactor pom generation functions to a separate file.
  - Refactor monorepo post processing functions to a separate file.
  - Refactor unit tests to separate files.
- Expose CLI tool used in google-cloud-java
  - Generate root pom.xml
  - Generate gapic-libraries-bom/pom.xml

To use these tools in google-cloud-java (through docker container):
```
python library_generation/cli/generate_monorepo_root_pom.py generate --repository-path=.

python library_generation/cli/generate_monorepo_gapic_bom.py generate --repository-path=. --versions-file=./versions.txt
```


suztomo (approved):

library_generation/utils/pom_generator.py
```diff
@@ -0,0 +1,157 @@
...
+                version_annotation = artifact_id[:index]
+                bom_configs.append(
+                    BomConfig(
+                        group_id=group_id,
+                        artifact_id=artifact_id,
+                        version=version,
+                        version_annotation=version_annotation,
+                    )
+                )
+    # handle edge case: java-grafeas
```
blakeli0, line 113: I know this was done in previous PRs, what is special about these three modules?

library_generation/test/unit_tests.py
blakeli0, line 1: Can this file be renamed to utilities_tests.py? Or it still contains more than just unit tests for utilities.py?

library_generation/utils/pom_generator.py
JoeWang1127: These three modules are not generated any more.

However, these modules are still part of the pom, so I need to add them using different approach since they don't have similar structure as other modules.

library_generation/test/unit_tests.py
JoeWang1127: Done.

library_generation/utils/pom_generator.py
blakeli0: I see, do we have to do the same for every handwritten libraries in `google-cloud-java`?

blakeli0 (approved):

library_generation/utils/pom_generator.py
JoeWang1127: No. I don't think this step is needed in handwritten libraries.

We need to refactor to only applicable in monorepo.

PR #2559, opened 2024-03-13
fix(deps): add detector-resource-support dependencies

mutianf:
b/329410901


java-shared-dependencies/third-party-dependencies/pom.xml
```diff
@@ -184,13 +185,15 @@
       </dependency>
       <dependency>
         <groupId>io.opentelemetry</groupId>
-        <artifactId>opentelemetry-api</artifactId>
+        <artifactId>opentelemetry-bom</artifactId>
```
blakeli0, line 188: Hey @mutianf, we plan to merge our own Otel [PR](https://github.com/googleapis/sdk-platform-java/pull/2500/files) as well in this cycle, and it includes the bom, do you think it would cover your use case?

mutianf: Perfect! Yes! I do also need the other `detector-resources-support` dependency though. Do you think you can also add that in your PR?

blakeli0: Since we don't use `detector-resources-support`, I think it's better to leave it in this PR. I'll let you know once we merged our Otel PR.

blakeli0: @mutianf The PR is merged now, do you mind modifying this PR accordingly?

mutianf: Updated! However I noticed that the open telemetry version in shared-dependencies pom is already 1.36.0 but in gax it was still 1.35.0. I think we should update the version in gax :)

blakeli0: The OpenTelemetry version in [Gax](https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/pom.xml#L164) should come from the same [opentelemetry.version](https://github.com/googleapis/sdk-platform-java/blob/main/gapic-generator-java-pom-parent/pom.xml#L35) property, where did you find that it is still 1.35.0? In addition, I think we might be able to remove `opentelemetry-bom` section from this pom because it is already declared in Gax?

blakeli0 (approved):

PR #2587, opened 2024-03-16
feat: add generation config comparator

JoeWang1127:
In this PR:
- Add a generation config comparator to compare the change type and which libraries are affected by the change.
- Refactor `LibraryConfig`.
- Add unit tests (no integration tests because the comparator has not been used).

The comparator makes assumptions that the following library level parameters will not change:
- googleapis commit (no use case)

This is the first step to [improve](https://docs.google.com/document/d/1JiCcG3X7lnxaJErKe0ES_JkyU7ECb40nf2Xez3gWvuo/edit?tab=t.g3vua2kd06gx#heading=h.pygigzqg78jp) performance of hermetic code generation.


diegomarquezp (approved):
LGTM, thanks for this PR. Just a couple of minor comments.

library_generation/utils/generation_config_comparator.py
```diff
@@ -0,0 +1,320 @@
...
+    list is empty.
+
+    :param baseline_config: the baseline GenerationConfig object
+    :param latest_config: the latest GenerationConfig object
+    :return: a mapping from ChangeType to a list of library_name of affected
+    libraries.
+    """
+    diff = {}
+    if baseline_config.googleapis_commitish != latest_config.googleapis_commitish:
+        diff[ChangeType.GOOGLEAPIS_COMMIT] = []
```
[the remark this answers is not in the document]
JoeWang1127: Thanks for the suggestion.

library_generation/utils/generation_config_comparator.py
```diff
@@ -0,0 +1,320 @@
...
+    # 3rd round comparison.
+    __compare_changed_libraries(
+        diff=diff,
+        baseline_libraries=baseline_libraries,
+        latest_libraries=latest_libraries,
+        changed_libraries=changed_libraries,
+    )
+
+
+def __convert(libraries: List[LibraryConfig]) -> Dict[str, HashLibrary]:
```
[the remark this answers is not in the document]
JoeWang1127: Done. Thanks for the suggestion.

library_generation/utils/generation_config_comparator.py
```diff
@@ -0,0 +1,324 @@
...
+
+from library_generation.model.gapic_config import GapicConfig
+from library_generation.model.generation_config import GenerationConfig
+from library_generation.model.library_config import LibraryConfig
+from library_generation.utilities import get_library_name
+
+ALL_LIBRARIES = []
+
+
+class ChangeType(Enum):
```
blakeli0, line 26: Why do we need so many different types? Do you plan to have different behaviors(e.g. commit messages) per type?
I think we could have a few types, but having so many seems overkill and hard to maintain. e.g. All repo level configs might be grouped in to one type other than `GOOGLEAPIS_COMMIT` change. Same thing for all library level configs other than `LIBRARIES_ADDITION` and `LIBRARIES_REMOVAL`.

library_generation/utils/generation_config_comparator.py
```diff
@@ -0,0 +1,324 @@
...
+    CODEOWNER_TEAM = 17
+    EXCLUDED_DEPENDENCIES = 18
+    EXCLUDED_POMS = 19
+    CLIENT_DOCS = 20
+    ISSUE_TRACKER = 21
+    REST_DOCS = 22
+    RPC_DOCS = 23
+    REQUIRES_BILLING = 24
+    EXTRA_VERSIONED_MODULES = 25
+    VERSION_ADDITION = 26
```
blakeli0, line 52: To keep the terms consistent, this can be `GAPIC_ADDITION` and `GAPIC_REMOVAL`

library_generation/utils/generation_config_comparator.py
```diff
@@ -0,0 +1,324 @@
...
+    """
+    Compare each library with the same library_name to find what parameters are
+    changed.
+
+    Some parameters should not change:
+
+    - api_shortname
+    - library_name
+    - distribution_name
+    - group_id
```
blakeli0, line 187: Technically anything can be changed, even including `library_name`. If `library_name` changes, we can treat it as a `removal of existing library + new client library`. And I have seen  service teams try to change `group_id` in the past. Is there challenges that prevent us from doing so? Or it's just we think they should not be changed logically?

library_generation/utils/generation_config_comparator.py
JoeWang1127: > Do you plan to have different behaviors(e.g. commit messages) per type?

That's the reason for these different types of change.

> I think we could have a few types, but having so many seems overkill and hard to maintain. e.g. All repo level configs might be grouped in to one type other than GOOGLEAPIS_COMMIT change.

If we just have one type for repo level parameter change (except googleapis commit), we can only generate something like `repo level configuration change`, but not `generator version update` or `protobuf version update` in release note.

library_generation/utils/generation_config_comparator.py
JoeWang1127: Done.

library_generation/utils/generation_config_comparator.py
JoeWang1127: I need to have a key to determine whether a library is changed or removed/added and I choose the key to be `library_name`.

The logic is generic enough such that if `baseline_config` has a library, `a` (library_name), and `latest_config` has a library, `b` (library_name), then the comparison will be `a` is removed from `baseline_config` and `b` is added to `lastest_config`.

blakeli0: I agree that the existing logic should be able to handle this. That's why I don't think we want to mention `library_name` can not be changed. For other things like `api_shortname`, I don't think we want to mention them either, because our scripts should just generate accordingly, is that right?

library_generation/utils/generation_config_comparator.py
blakeli0: We can design the type generically. e.g. The type is `REPO_LEVEL_CHANGE`, the enum type can take two parameters `config_name` and `config_version`. For all changes in this type, we will regenerate the whole repo and the commit message will of the same format, like `Update {config_name} to {config_version}`.

library_generation/utils/generation_config_comparator.py
JoeWang1127: IIUC, if the `library_name` or `api_shorname` is changed, we should treat this case as `LIBRARY_REMOVAL` and `LIBRARY_ADDITION`, right?

> And I have seen service teams try to change group_id in the past. Is there challenges that prevent us from doing so? Or it's just we think they should not be changed logically?

I don't know we have these cases. I think we can handle this case as distribution name change or group id change.

library_generation/utils/generation_config_comparator.py
JoeWang1127: For `config_name`, we still need an enum or do you think string type is suffice?

blakeli0: String should be good enough. We should be able to at least reduce the Enum types, even if the lines of code are not reduced, it would be much easier to understand the code now, as other developers may think there are 20+ ways of different commit messages, compare to maybe 5. And the enum types will grow at a much faster pace if we add one enum type for every new parameter.

JoeWang1127: SG, I'll update the enum types.

library_generation/test/utils/generation_config_comparator_unit_tests.py
```diff
@@ -0,0 +1,477 @@
...
+        self.latest_config.libraries = []
+        result = compare_config(
+            baseline_config=self.baseline_config,
+            latest_config=self.latest_config,
+        )
+        self.assertTrue(len(result[ChangeType.LIBRARIES_REMOVAL]) == 1)
+        config_change = result[ChangeType.LIBRARIES_REMOVAL][0]
+        self.assertEqual("existing_library", config_change.library_name)
+
+    def test_compare_config_api_shortname_update(self):
```
JoeWang1127, line 197: Add this test case to verify the behavior of `api_shortname` change.

library_generation/test/utils/generation_config_comparator_unit_tests.py
```diff
@@ -0,0 +1,477 @@
...
+            latest_config=self.latest_config,
+        )
+        self.assertTrue(len(result[ChangeType.LIBRARIES_REMOVAL]) == 1)
+        config_change = result[ChangeType.LIBRARIES_REMOVAL][0]
+        self.assertEqual("existing_library", config_change.library_name)
+        self.assertTrue(len(result[ChangeType.LIBRARIES_ADDITION]) == 1)
+        config_change = result[ChangeType.LIBRARIES_ADDITION][0]
+        self.assertEqual("new_api_shortname", config_change.library_name)
+
+    def test_compare_config_library_name_update(self):
```
JoeWang1127, line 210: Add this test case to verify the behavior of `library_name` change.

library_generation/utils/generation_config_comparator.py
JoeWang1127: Done, added unit tests to verify these parameters.

library_generation/utils/generation_config_comparator.py
JoeWang1127: Done, reduce change types.

library_generation/utils/generation_config_comparator.py
```diff
@@ -0,0 +1,379 @@
...
+from library_generation.model.generation_config import GenerationConfig
+from library_generation.model.library_config import LibraryConfig
+from library_generation.utilities import get_library_name
+
+
+class ChangeType(Enum):
+    GOOGLEAPIS_COMMIT = 1
+    REPO_LEVEL_CHANGE = 2
+    LIBRARIES_ADDITION = 3
+    LIBRARIES_REMOVAL = 4
```
blakeli0, line 29: Is the scripts going to handle `LIBRARIES_REMOVAL` correctly? IIRC, the scripts is going to skip a library if it is not listed in the `generation_config.yaml`.

library_generation/utils/generation_config_comparator.py
```diff
@@ -0,0 +1,379 @@
...
+
+    :param baseline_config: the baseline GenerationConfig object
+    :param latest_config: the latest GenerationConfig object
+    :return: a mapping from ConfigChange to a list of ConfigChange objects.
+    """
+    diff = defaultdict(list[ConfigChange])
+    if baseline_config.googleapis_commitish != latest_config.googleapis_commitish:
+        diff[ChangeType.GOOGLEAPIS_COMMIT] = []
+    if baseline_config.gapic_generator_version != latest_config.gapic_generator_version:
+        config_change = ConfigChange(
```
blakeli0, line 69: Can we iterate over all fields in `GenerationConfig`? Same thing for the `LibraryConfig` below. It could be a separate PR.

library_generation/test/utils/generation_config_comparator_unit_tests.py
blakeli0: Why do we treat `api_shortname` change a library removal? If `library_name` stays the same and `api_shortname` changes, it should be a library level change?

library_generation/utils/generation_config_comparator.py
JoeWang1127: This PR only creates change types of two configurations, it doesn't specify how to handle each type.

For library removal, I expect `generate_pr_body.py` will create an release note entry like `feat/chore: xxx is removed`. However, the actual code in google-cloud-java needs to be manually removed.

library_generation/utils/generation_config_comparator.py
JoeWang1127: > Can we iterate over all fields in GenerationConfig? Same thing for the LibraryConfig below.

Does the goal is to reduce the amount of code?

library_generation/test/utils/generation_config_comparator_unit_tests.py
JoeWang1127: > If library_name stays the same and api_shortname changes, it should be a library level change?

In this test case, the `library_name` is changed because `library_name` is None and falls back to `api_shortname`.

Do you think we should add a test case to verify what happened if `api_shorname` is changed but `library_name` is set and remains the same? I expect this change will be a library level change.

library_generation/utils/generation_config_comparator.py
blakeli0: Yes, the logic is duplicated for every field.

library_generation/utils/generation_config_comparator.py
blakeli0: Would it be better if we ignore this case and add the `feat/chore: xxx is removed` in the corresponding PR? Or we can think about how to remove libraries automatically, but it would probably be a larger effort.

library_generation/utils/generation_config_comparator.py
JoeWang1127: To iterate field of a class, I think we need a enum type or a list of strings to specify the field?

library_generation/test/utils/generation_config_comparator_unit_tests.py
blakeli0: > Do you think we should add a test case to verify what happened if api_shorname is changed but library_name is set and remains the same? I expect this change will be a library level change.

Yes, please. Please rename this test case as well, as it currently only indicates api_shortname update. In addition, this makes me thinking that we should probably put the logic that figures out the `library_name` into GenerationConfig itself, because this logic is already somewhere in the scripts and we don't have to duplicate it again here.

library_generation/utils/generation_config_comparator.py
JoeWang1127: Personally, I think it's better to output this change type here because this is the responsible of the comparator (to produce the change type, but not how to handle each type).

This type can serve as a trigger for automatic removal in further improvement and we don't need to refactor the comparator for this case. WDYT?

blakeli0: Sounds reasonable to me. But what is going to be the title for the PR that removes a library? Unless you want to make it a `chore` which would be a little confusing, it would still be duplicated in the release-note. I feel ignoring this case for now is better than documenting this behavior and tell release managers to not use `feat` in the removal PR. We don't have to remove the code, they can be commented out until we figure this out.

library_generation/utils/generation_config_comparator.py
blakeli0: Does something like [this](https://stackoverflow.com/questions/11637293/iterate-over-object-attributes-in-python) help?

library_generation/test/utils/generation_config_comparator_unit_tests.py
blakeli0: > Is it better to put it into `LibraryConfig`?

Yes, I meant `LibraryConfig` not `GenerationConfig`.

> Do you have a use case of changing api_shortname and does this case actual change it or just attempt to change it? This parameter is also in .OwlBot.yaml and I'm not sure what will happen if we only change this in configuration but not in .OwlBot.yaml.

That's a good point, if we don't feel comfortable our scripts can handle `api_shortname` change, we should throw exceptions when we detect such changes.

library_generation/utils/generation_config_comparator.py
JoeWang1127: SG. I'll do some refactor then.

library_generation/utils/generation_config_comparator.py
JoeWang1127: I don't know python can do this, thanks for the help. I'll do some refactor then.

library_generation/utils/generation_config_comparator.py
JoeWang1127: I commented out `LIBRARIES_REMOVAL` and `GAPIC_REMOVAL`.

library_generation/test/utils/generation_config_comparator_unit_tests.py
JoeWang1127: I've moved `get_library_name` to `LibraryConfig`.

Added a test case to verify the script should raise `ValueError` when `api_shorname` is changed but `library_name` remains the same.

library_generation/test/utils/generation_config_comparator_unit_tests.py
```diff
@@ -0,0 +1,459 @@
...
+        self.latest_config.libraries[0].api_shortname = "new_api_shortname"
+        result = compare_config(
+            baseline_config=self.baseline_config,
+            latest_config=self.latest_config,
+        )
+        self.assertTrue(len(result[ChangeType.LIBRARIES_ADDITION]) == 1)
+        config_change = result[ChangeType.LIBRARIES_ADDITION][0]
+        self.assertEqual("new_api_shortname", config_change.library_name)
+
+    def test_compare_config_api_shortname_update_with_library_name_raise_error(self):
```
JoeWang1127, line 197: Add this test case to verify the script should raise `ValueError` when `api_shorname` is changed but `library_name` remains the same.

library_generation/utils/generation_config_comparator.py
blakeli0: I thought `GAPIC_REMOVAL` can be handled automatically?

JoeWang1127: @diegomarquezp after #2572 is merged, can we handle GAPIC removal automatically, i.e., deep-remove regex should work at library level?

diegomarquezp: Deep remove does not remove the entire library but the src folders inside that gapic, grpc and proto "sub libraries". It's meant to replace old code in the sublibraries. It wouldn't automatically delete any sublibrary if it stops being listed in the config yaml.

JoeWang1127: @blakeli0 does this answer your question?

blakeli0 (approved):

PR #2592, opened 2024-03-19
feat: Validate the Universe Domain inside Java-Core

lqiu96:
Validate the Universe Domain prior to the request being initialized and executed. The request will throw an `IllegalStateException` if the validation fails (configured Universe Domain does not match the Credentials' Universe Domain).


java-core/google-cloud-core-http/src/main/java/com/google/cloud/http/HttpTransportOptions.java
```diff
@@ -42,6 +42,7 @@
 public class HttpTransportOptions implements TransportOptions {
 
   private static final long serialVersionUID = 7890117765045419810L;
+  private static final String GOOGLE_CLOUD_UNIVERSE_DOMAIN = "GOOGLE_CLOUD_UNIVERSE_DOMAIN";
```
blakeli0, line 45: Can we expose the same [String](https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java#L50) as public in Gax?

java-core/google-cloud-core-http/src/main/java/com/google/cloud/http/HttpTransportOptions.java
```diff
@@ -153,11 +154,46 @@ public HttpRequestInitializer getHttpRequestInitializer(
         serviceOptions.getMergedHeaderProvider(internalHeaderProvider);
 
     return new HttpRequestInitializer() {
+
+      /**
+       * Helper method to resolve the Universe Domain. First checks the user configuration from
+       * ServiceOptions, then the Environment Variable. If both haven't been set, resolve the value
+       * to be the Google Default Universe.
+       */
+      private String determineUniverseDomain() {
```
blakeli0, line 163: This is probably a larger scope, have we considered using `EndpointContext` in java-core? The logic here and below are pretty much duplicated from `EndpointContext`, if we can wrap all the info to `EndpointContext`, we can reuse  `determineUniverseDomain()` from it. Or even better, we can just call `validateUniverseDomain()` it directly if possible.

lqiu96: Ah, that's a good point. I had not considered using EndpointContext in java-core before, mostly because EndpointContext was written with GAPICs in mind and some of EndpointContext concepts (i.e. parsing serviceName) were written with that in mind.

I took a look and I think there are a few open questions that I'll need to figure out from a first glance:
1. Conversion between host -> endpoint
2. Handwritten team probably need to manually set the serviceName somewhere
3. I'm not sure where the mTLS configs are set (I don't see them inside SerivceOptions)
4. Do Handwritten Libraries know how to handle ApiExceptions?
5. Somewhere in Java-Core needs to check for GDC-H

I think these are possible to do in Java-Core so that we can use EndpointContext, but I'll need to do more investigating.

java-core/google-cloud-core-http/src/main/java/com/google/cloud/http/HttpTransportOptions.java
lqiu96: Ah, yep. I think so. I'll try it.

java-core/google-cloud-core-http/src/main/java/com/google/cloud/http/HttpTransportOptions.java
blakeli0: Thanks for the quick response! 

- 1 and 2 are valid concerns. 
- For 3 and 5, I don't think the libraries use java-core care about mTLS and GDC-H, so we can leave them empty. 
- For 4, I don't think we have to handle ApiExceptions, we are throwing `IllegalStateException` in this PR, we will be throwing `UnauthenticatedException` if we follow the same way we did in [HttpJsonCallContext](https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonCallContext.java#L432-L433). Actually it might be better to change the exception type to `UnauthenticatedException` even if we don't use `EndpointContext`.

lqiu96: Makes sense. If mTLS and GDC-H aren't a concern for now then I think that should be easier. I'll need to figure out how to handle 1 & 2 (and potentially 4).

I don't think we need to handle ApiException, but I don't know how to create the ApiException's retryable value. IIRC for GAPICs, we are checking the retryable status codes and setting it accordingly, but this probably is different from how handwritten libraries are doing it. I'll probably need to check for the retry codes they have configured somewhere (to keep the behavior the same) and this may differ between each library. It's possible that it's done inside ServiceOptions or something shared from java-core, but I'm not entirely sure.

I created a FR ticket to track this: https://github.com/googleapis/sdk-platform-java/issues/2631

Would you be fine with keeping this duplicate logic in Java-Core for now? I think this will require some more investigation.

I'll update the exception to return `UnauthenticatedException`.

blakeli0: > Would you be fine with keeping this duplicate logic in Java-Core for now?

Yes, this sounds good to me.

java-core/google-cloud-core-http/src/main/java/com/google/cloud/http/HttpTransportOptions.java
```diff
@@ -153,11 +154,45 @@ public HttpRequestInitializer getHttpRequestInitializer(
...
+
+        // delegate is always HttpCredentialsAdapter or null (NoCredentials)
+        if (delegate != null) {
+          HttpCredentialsAdapter httpCredentialsAdapter = (HttpCredentialsAdapter) delegate;
+          credentialsUniverseDomain = httpCredentialsAdapter.getCredentials().getUniverseDomain();
+        }
+
+        // Validate the universe domain before initializing the request
+        if (!configuredUniverseDomain.equals(credentialsUniverseDomain)) {
+          throw new IllegalStateException(
```
blakeli0, line 185: Can we change the exception type to `UnauthenticatedException` to match the behavior in Gax?

java-core/google-cloud-core-http/src/test/java/com/google/cloud/http/HttpTransportOptionsTest.java
```diff
@@ -42,6 +77,144 @@ public class HttpTransportOptionsTest {
...
+  // Credentials' getUniverseDomain() returns `random.com`
+  private Credentials customCredentials;
+  private HttpRequest defaultHttpRequest;
+
+  /**
+   * The following interfaces and classes are from ServiceOptionsTest. Copied over here as
+   * ServiceOptions resides inside google-cloud-core test folder and is not accessible from
+   * google-cloud-core-http.
+   */
+  interface TestService extends Service<TestServiceOptions> {}
```
blakeli0, line 95: nit: It might be subjective, usually I prefer to put these private test interfaces/classes to the bottom of the file so it's easier to focus on the actual test cases.

PR #2598, opened 2024-03-21
feat: generate selected libraries

JoeWang1127:
In this PR:
- Generate a selected list of libraries based on `library_name`

This PR is a preparation for improve nightly generation.

Note that we need to change the [new library generation workflow](https://github.com/googleapis/google-cloud-java/blob/main/.github/workflows/generate_new_client_hermetic_build.yaml) in google-cloud-java to accommodate this change because `library_generation/generate_repo.py` has replaced `--target-library-api-shortname` with `--target-library-names`.


library_generation/cli/generate_repo.py
blakeli0, line 1: So this is supposed to be the endtrypoint in the future? If it is, can we rename this file to avoid confusion? Otherwise I think keeping the cli together with the logic is fine.

diegomarquezp (approved):
LGTM

library_generation/generate_repo.py
```diff
@@ -118,5 +64,21 @@ def generate_from_yaml(
...
+    config: GenerationConfig, target_library_names: list[str] = None
+) -> list[LibraryConfig]:
+    """
+    Returns LibraryConfig objects whose library_name is in target_library_names.
+
+    :param config: a GenerationConfig object.
+    :param target_library_names: library_name of target libraries.
+    :return:
+    """
+    if target_library_names is None:
```
blakeli0, line 77: What if the list is empty? Does it mean we generate nothing or generate everything?

JoeWang1127: > What if the list is empty

We generate nothing.

The rational is if we pass a list, then the libraries will be selected from the list. Therefore, an empty list means no generation.

If we don't pass a list, the default value is None and we generate everything.

library_generation/cli/generate_repo.py
JoeWang1127: > So this is supposed to be the entrypoint in the future?

Yes.

We can change the name, then before combining `generate_pr_description.py`, we will have some like this in google-cloud-java
```
entrypoint.py ... // generate libraries
generate_pr_description.py ... // generate PR description
```

Do you think this is fine?

blakeli0: Yep, I think that's better than having two `generate_repo.py`. In general, if you have a design in your mind but will not implement it immediately, it's better to define the interfaces first so that others are easy to follow.

library_generation/generate_repo.py
blakeli0: SG. Can you mention the behavior in the comment above? I wonder how it is going to be implemented in `generate_repo.py` later, maybe just no-op if the list is empty?

library_generation/cli/generate_repo.py
JoeWang1127: OK, I think this (having two `generate_repo.py` is confusing) makes sense.

I decided to remove `cli/generate_repo.py` and keep the cli part in `generate_repo.py` (as before).

I'll introduce `entrypoint.py` in follow PR.

library_generation/generate_repo.py
JoeWang1127: Adde comments.

blakeli0 (approved):

PR #2599, opened 2024-03-22
chore: gapic-generator-java noop follow up on hermetic build scripts

diegomarquezp:
This is a follow up from https://github.com/googleapis/sdk-platform-java/pull/2460

Since GGJ will NOOP when no services are found in the passed `proto_path`s, we remove the `proto_only` flag entirely so we can rely on this new behavior instead.

### DIFF
--- a/library_generation/test/resources/integration/google-cloud-java/generation_config.yaml
+++ b/library_generation/test/resources/integration/google-cloud-java/generation_config.yaml
@@ -37,6 +37,7 @@ libraries:
     - proto_path: google/cloud/alloydb/v1beta
 
   - api_shortname: alloydb
+    # this is a proto_only library with which the gapic-generator should no-op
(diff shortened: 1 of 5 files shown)


library_generation/test/resources/integration/google-cloud-java/generation_config.yaml
```diff
@@ -55,4 +56,4 @@ libraries:
     api_description: Provides insights about your customers and their Assured Workloads based on your Sovereign Controls by Partners offering.
     GAPICs:
     - proto_path: google/cloud/cloudcontrolspartner/v1
-    - proto_path: google/cloud/cloudcontrolspartner/v1beta
\ No newline at end of file
+    - proto_path: google/cloud/cloudcontrolspartner/v1beta
```
JoeWang1127, line 59: any benefits of adding this line?

PR #2604, opened 2024-03-26
feat: add config change

JoeWang1127:
In this PR:
- Add `config_change.py` to get:
  - Changed libraries, which will be passed to `generate_repo.py` to generate libraries selectedly.
  - Qualified commits, which will be passed to `generate_pr_description.py` to generate PR description.
- Refactor part of utility methods to `proto_path_utils.py`.
- Refactor `generation_config_comparator.py` to move data class to `config_change.py`.
- Refactor `utilities.py` and `utilities.sh` to `utils/utilities.py` and `utils/utilities.sh`.
- Add unit tests.

Follow up of #2587

For the goal of series of PRs, please refer to [improvement proposal](https://docs.google.com/document/d/1JiCcG3X7lnxaJErKe0ES_JkyU7ECb40nf2Xez3gWvuo/edit?tab=t.g3vua2kd06gx#bookmark=id.72s3ukwwzevo).


library_generation/model/config_change.py
```diff
@@ -0,0 +1,155 @@
...
+        latest_config: GenerationConfig,
+    ):
+        self.change_to_libraries = change_to_libraries
+        self.baseline_config = baseline_config
+        self.latest_config = latest_config
+
+    def get_changed_libraries(self) -> Optional[list[str]]:
+        """
+        Returns library name of changed libraries.
+        None if there is a repository level change.
```
diegomarquezp, line 76: `None` has a deep meaning related to all libraries being processed. I think it would help to explain what it means. For example "None represents that all libraries are affected", or return a constant `ALL_LIBRARIES_CHANGED=None` to make it self-explanatory.

library_generation/model/config_change.py
```diff
@@ -0,0 +1,155 @@
...
+    ) -> list[QualifiedCommit]:
+        """
+        Returns qualified commits from configuration change.
+
+        A qualified commit is a commit that changes at least one file (excluding
+        BUILD.bazel) within a versioned proto path in the given proto_paths.
+        :param repo_url: the repository contains the commit history.
+        :return: QualifiedCommit objects.
+        """
+        tmp_dir = "/tmp/repo"
```
diegomarquezp, line 103: Is it possible to download googleapis in the `output` folder? This is just because we will convert our scripts to only use `output` and not `/tmp` in favor of running docker containers in parallel.

library_generation/model/config_change.py
```diff
@@ -0,0 +1,155 @@
...
+                break
+            commit = commit_parents[0]
+        shutil.rmtree(tmp_dir, ignore_errors=True)
+        return qualified_commits
+
+    def __get_library_name_from_qualified_commits(self) -> list[str]:
+        qualified_commits = self.get_qualified_commits()
+        library_names = []
+        for qualified_commit in qualified_commits:
+            library_names.extend(qualified_commit.libraries)
```
diegomarquezp, line 128: If I understand correctly, `library_names` could end up with repeated library names. Is this handled in any way?
If not, we can remove duplicates with something like `return list(set(library_names))`

library_generation/model/config_change.py
```diff
@@ -0,0 +1,155 @@
...
+            if qualified_commit is not None:
+                qualified_commits.append(qualified_commit)
+            commit_parents = commit.parents
+            if len(commit_parents) == 0:
+                break
+            commit = commit_parents[0]
+        shutil.rmtree(tmp_dir, ignore_errors=True)
+        return qualified_commits
+
+    def __get_library_name_from_qualified_commits(self) -> list[str]:
```
diegomarquezp, line 124:
```suggestion
    def __get_library_names_from_qualified_commits(self) -> list[str]:
```

(This may imply more changes than just changing this line)

library_generation/model/config_change.py
```diff
@@ -0,0 +1,155 @@
...
+        return library_names
+
+    @staticmethod
+    def __create_qualified_commit(
+        proto_paths: dict[str, str], commit: Commit
+    ) -> Optional[QualifiedCommit]:
+        """
+        Returns a qualified commit from the given Commit object; otherwise None.
+
+        :param proto_paths:
```
diegomarquezp, line 138: Would you mind expanding this comment to tell what are the expected contents of each key and value in `proto_paths`?

library_generation/test/model/config_change_unit_tests.py
```diff
@@ -0,0 +1,224 @@
...
+                        changed_param="test",
+                        latest_value="test",
+                        library_name="test-library",
+                    )
+                ],
+            },
+            baseline_config=ConfigChangeTest.__get_a_gen_config(),
+            latest_config=ConfigChangeTest.__get_a_gen_config(),
+        )
+        self.assertIsNone(config_change.get_changed_libraries())
```
diegomarquezp, line 42: Similar to one of the comments above, the test would be a bit more readable if we added something like `self.assertEquals(ALL_LIBRARIES_AFFECTED, config_change.get_changed_libraries())`

library_generation/test/model/config_change_unit_tests.py
```diff
@@ -0,0 +1,224 @@
...
+                        ],
+                    ),
+                ],
+            ),
+        )
+        qualified_commits = config_change.get_qualified_commits()
+        self.assertEqual(3, len(qualified_commits))
+        self.assertEqual({"gke-backup"}, qualified_commits[0].libraries)
+        self.assertEqual(
+            "b8691edb3f1d3c1583aa9cd89240eb359eebe9c7",
```
diegomarquezp, line 159: nit: can we use variable names for these commit hashes? This is just in order to make the test more self-explanatory.

library_generation/utils/proto_path_utils.py
```diff
@@ -0,0 +1,61 @@
...
+    paths = {}
+    for library in config.libraries:
+        for gapic_config in library.gapic_configs:
+            paths[gapic_config.proto_path] = library.get_library_name()
+    return paths
+
+
+def find_versioned_proto_path(file_path: str) -> str:
+    """
+    Returns a versioned proto_path from a given file_path; or file_path itself
```
diegomarquezp, line 48: Can we clarify that `file_path` should be a proto file?

library_generation/model/config_change.py
JoeWang1127: Done.

library_generation/model/config_change.py
JoeWang1127: Done.

library_generation/test/model/config_change_unit_tests.py
JoeWang1127: Done.

library_generation/model/config_change.py
JoeWang1127: Good catch.

I updated the method to return a list of unique library names in ascending order.

library_generation/utils/proto_path_utils.py
JoeWang1127: Done.

library_generation/model/config_change.py
JoeWang1127: I changed to `sh_util("get_output_folder")`

library_generation/model/config_change.py
```diff
@@ -0,0 +1,159 @@
...
+            return ConfigChange.ALL_LIBRARIES_CHANGED
+        library_names = set()
+        for change_type, library_changes in self.change_to_libraries.items():
+            if change_type == ChangeType.GOOGLEAPIS_COMMIT:
+                library_names.update(self.__get_library_names_from_qualified_commits())
+            else:
+                library_names.update(
+                    [library_change.library_name for library_change in library_changes]
+                )
+        return sorted(list(library_names))
```
JoeWang1127, line 93: Return a list of unique, sorted library names

diegomarquezp (approved):

library_generation/utils/proto_path_utils.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+    version_pattern = "^v[1-9]"
+    index = proto_path.rfind("/")
+    version = proto_path[index + 1 :]
+    if re.match(version_pattern, version):
+        return proto_path[:index]
+    return proto_path
+
+
+def get_proto_paths(config: GenerationConfig) -> dict[str, str]:
```
blakeli0, line 33: I would move this to `GenerationConfig`class, and/or refactor it further to a function `get_library_name_from_proto_path(proto_path:str)`.

As a general best practice, we should try to reduce the number of utils file and move the logics to the relative model classes if possible

library_generation/model/config_change.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def get_qualified_commits(
+        self,
+        repo_url: str = "https://github.com/googleapis/googleapis.git",
+    ) -> list[QualifiedCommit]:
+        """
+        Returns qualified commits from configuration change.
+
+        A qualified commit is a commit that changes at least one file (excluding
+        BUILD.bazel) within a versioned proto path in the given proto_paths.
```
blakeli0, line 103: I know we made the decision to exclude BUILD.bazel previously, but we may need to revisit it for scenarios that the changes include transport/numeric_enums/include_samples. It means we may have to parse BUILD.bazel in this step, which would definitely make it more complicated. It was fine because we regenerate the whole repo anyway, but if we start excluding libraries from regeneration, then we need to be more careful.

library_generation/utilities.py
blakeli0, line 1: I would move this file to under `/utils` folder.

library_generation/model/config_change.py
```diff
@@ -0,0 +1,159 @@
...
+        :param repo_url: the repository contains the commit history.
+        :return: QualifiedCommit objects.
+        """
+        tmp_dir = sh_util("get_output_folder")
+        shutil.rmtree(tmp_dir, ignore_errors=True)
+        os.mkdir(tmp_dir)
+        # we only need commit history, thus shadow clone is enough.
+        repo = Repo.clone_from(url=repo_url, to_path=tmp_dir, filter=["blob:none"])
```
blakeli0, line 111: I know this was done in previous PRs. Based on our recent discussions, is it possible to get the commit history remotely? Without cloning the repo? Creating new temp folders could have other implications like preventing us from running images in parallel.

JoeWang1127: I did some digging but I can't find one without `git clone`.

I don't think we need to focus on parallel execution as it's not the main goal of this PR. Also, we can achieve parallelism if we can randomize the temp folder so that no thread will clone the repo into the same folder.

library_generation/model/config_change.py
JoeWang1127: I think we exclude `BUILD.bazel` because we don't want changes in other languages appear in commit messages, e.g., `[ruby] xxx`.

How about I change the regex in filtering commit message to exclude irrelevant message, rather than excluding from generation.

library_generation/utils/proto_path_utils.py
JoeWang1127: Done.

library_generation/utilities.py
JoeWang1127: Done.

library_generation/model/config_change.py
JoeWang1127: > I think we may have to parse the BUILD.bazel file to see if there is any changes to transport/numeric_enums/include_samples in Java targets. Of course this can be a separate a PR and slightly lower priority as they don't change often.

I agree. I'll add a task as the next step and keep this part as-is.

library_generation/model/config_change.py
blakeli0: I don't think it's a focus either but something we should keep in mind. Creating temp folders and writing files has other implications, e.g. it's usually slower than making a request, it complicates the scripts that we have to clean the temp folders and files up etc. 
That being said, it's OK if there is no easy way at this moment. We can refactor how we pull the proto definitions in general in the future. e.g. I see that we are currently [downloading proto definitions for every library](https://github.com/googleapis/sdk-platform-java/blob/main/library_generation/utilities.py#L161-L194), where we could download the whole googleapis repo as the first step of generate_repo.py. This actually brings another question that if googleapis_commitish is overridden on library level, our release notes might be generated incorrectly. 
In short, this PR is good to go as it is, but we need to put the above tasks into consideration and track them somewhere.

library_generation/utils/utilities.py
```diff
@@ -65,7 +65,7 @@ def sh_util(statement: str, **kwargs) -> str:
         kwargs["stderr"] = subprocess.PIPE
     output = ""
     with subprocess.Popen(
-        ["bash", "-exc", f"source {script_dir}/utilities.sh && {statement}"],
+        ["bash", "-exc", f"source {script_dir}/../utilities.sh && {statement}"],
```
blakeli0, line 68: How about move `utilities.sh` as well?

blakeli0 (approved):
LGTM. Please track the refactoring and potential issues discussed in this PR in our project plan.

library_generation/model/config_change.py
JoeWang1127: > I see that we are currently [downloading proto definitions for every library](https://github.com/googleapis/sdk-platform-java/blob/main/library_generation/utilities.py#L161-L194),

The googleapis repo is downloaded only once unless there are googleapis commit in library level.

Since this use case is rarely used, how about removing it?

library_generation/utils/utilities.py
JoeWang1127: Done

library_generation/model/config_change.py
JoeWang1127: Based on discussion, we keep this piece of code as-is.

PR #2609, opened 2024-03-27
chore: use subprocess in integration tests

JoeWang1127:
In this PR:
- Use subprocess to run python command in integration tests.

The CLI is being used in downstream libraries, so it's better to test CLI directly.

### DIFF
--- a/library_generation/test/integration_tests.py
+++ b/library_generation/test/integration_tests.py
@@ -15,6 +15,7 @@ import json
 import os
 import shutil
 import unittest
+from click.testing import CliRunner
 from distutils.dir_util import copy_tree
 from distutils.file_util import copy_file
 from filecmp import cmp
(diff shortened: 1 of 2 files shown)


library_generation/test/integration_tests.py
```diff
@@ -103,8 +109,14 @@ def test_generate_repo(self):
                     copy_file(f"{repo_dest}/pom.xml", golden_dir)
                 else:
                     copy_tree(f"{repo_dest}", f"{golden_dir}/{library_name}")
-            generate_from_yaml(
-                generation_config_yaml=config_file, repository_path=repo_dest
+            subprocess.run(
```
blakeli0, line 112: I understand that we have to change the way we call `generate_repo.py`, but why the [diff check](https://github.com/googleapis/sdk-platform-java/blob/main/library_generation/test/integration_tests.py#L124-L135) below didn't fail?

JoeWang1127: There are two methods in `generate_repo.py`: `generate_from_yaml`, which is the implementation and `generate`, which is a wrapper of `generate_from_yaml` and decorated with `click`.

The IT is calling `generate_from_yaml` directly so there's no failure since this method is not changed.

When running in docker container in google-cloud-java, it calls `generate`. Even though `generate` didn't change, the deletion of `main` in the file causing the call failed without warning (or didn't call `generate` at all!).

The reason for setting two methods is that if `generate_from_yaml` decorated with `click`, it can't be called in python code.

blakeli0: I thought the problem was that the generation process was not triggered, otherwise how could it finish in 3 minutes?

JoeWang1127: The deletion of `main` in `generate_repo.py` only affected the CLI (`generate` in this case).

The method is called, just without any real effect.

JoeWang1127:
We will run integration tests using `docker run`.

PR #2613, opened 2024-03-29
chore: add integration test running in docker container

JoeWang1127:
In this PR:
- In the integration test, run `generate_repo.py` and `generate_pr_description.py` in docker container.
- Remove test image in cloud build.

Context: 
In this [log](https://github.com/googleapis/google-cloud-java/actions/runs/8445620169/job/23133097820), the nightly generation job only took 3 mins. After investigation, we found out that removing `main` method in `generate_repo.py` (#2598) caused the CLI is not running despite that the job returns successfully.

This non-explicit failure demonstrated that there's difference between calling CLI (`generate` method in `generate_repo.py`) and its implementation (`generate_from_yaml` method in `generate_repo.py`).

We decided to change the integration test to run the CLI in docker container, rather than testing `generate_from_yaml` because it is the containerized CLI will be used in downstream libraries.


JoeWang1127:
I'll remove [this](https://pantheon.corp.google.com/cloud-build/triggers;region=global/edit/5783744f-0820-419f-bc5e-abbbece4be0b?e=13803378&mods=monitoring_api_prod&project=cloud-devrel-kokoro-resources) cloud build trigger once the PR is approved.

library_generation/test/integration_tests.py
```diff
@@ -181,44 +162,137 @@ def test_generate_repo(self):
...
+                "type=none",
+                "--opt",
+                f"device={device_dir}",
+                "--opt",
+                "o=bind",
+            ]
+        )
+
+    @classmethod
+    def __run_entry_point_in_docker_container(
```
JoeWang1127, line 235: After creating `entry_point.py`, there will be only once `docker run`.

diegomarquezp (approved):
LGTM

library_generation/test/integration_tests.py
```diff
@@ -12,103 +12,84 @@
...
-# within config_dir named {repo}/generation_config.yaml, where repo is
-# the name of the repository the target libraries live.
-config_dir = f"{script_dir}/resources/integration"
-golden_dir = f"{config_dir}/golden"
+golden_dir = os.path.join(script_dir, "resources", "integration", "golden")
+repo_root_dir = os.path.join(script_dir, "..", "..")
+build_file = os.path.join(
+    repo_root_dir, ".cloudbuild", "library_generation", "library_generation.Dockerfile"
+)
+image_tag = f"test-image:latest"
```
blakeli0, line 36: We don't have to use f string as this is a hardcoded value.

library_generation/test/integration_tests.py
```diff
@@ -12,103 +12,84 @@
...
-            self.assertTrue(
-                cmp(
-                    f"{config_dir}/{repo}/pr-description-golden.txt",
-                    f"{description_file}",
-                ),
-                "The generated PR description does not match the expected golden file",
-            )
-            os.remove(f"{description_file}")
+    def test_entry_point_running_in_container(self):
+        self.__build_image(docker_file=build_file, tag=image_tag, cwd=repo_root_dir)
```
blakeli0, line 52: Unless we want to expose `tag` as a parameter of `IntegrationTest` in the future, I don't think we need to pass around `tag` in this script, we can always use `image_tag`.

library_generation/test/integration_tests.py
JoeWang1127: Done

library_generation/test/integration_tests.py
JoeWang1127: Done.

blakeli0 (approved):

PR #2614, opened 2024-03-29
feat: Add DirectPath Compatible Getter to gRPC channel provider

lqiu96:


gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -446,6 +443,24 @@ && canUseDirectPathWithUniverseDomain()) {
...
+   * settings and a few other configurations/settings must also be valid for the request to go
+   * through DirectPath.
+   *
+   * <p>Checks: 1. Credentials are compatible 2.Running on Compute Engine 3. Universe Domain is
+   * configured to for the Google Default Universe
+   *
+   * @return if DirectPath is enabled for the client AND if the configurations are valid
+   */
+  @InternalApi
+  public boolean canUseDirectPath() {
```
blakeli0, line 457: Is there any existing unit tests that test these logics indirectly? Theoretically we should create new unit tests that cover the combinations of these cases, but I know it may not be easy as some of them involves system properties.

blakeli0, line 457: How do we imagine Spanner use this getter? I think it might be better for us to create a way in `MetricTracer` to take common client level attributes, otherwise they have to pass this value around to each RPC calls.

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracerFactory.java
```diff
@@ -45,13 +47,31 @@
...
   public ApiTracer newTracer(ApiTracer parent, SpanName spanName, OperationType operationType) {
-    return new MetricsTracer(
-        MethodName.of(spanName.getClientName(), spanName.getMethodName()), metricsRecorder);
+    MetricsTracer metricsTracer =
+        new MetricsTracer(
+            MethodName.of(spanName.getClientName(), spanName.getMethodName()), metricsRecorder);
+    for (Map.Entry<String, String> attributeEntrySet : attributes.entrySet()) {
+      metricsTracer.addAttributes(attributeEntrySet.getKey(), attributeEntrySet.getValue());
+    }
+    return metricsTracer;
```
lqiu96, line 75: Possible options:
1. Use MetricTracer's `addAttributes()` to add *only* add client level attributes. The clientName and methodName attributes will continue to be passed via MethodName via the constructor.
2. Add MethodName attribute with `addAttributes()` and refactor `MetricsTracer` to only take in metricsRecorder

Probably going to keep it this way (option 1).

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracerFactory.java
```diff
@@ -45,13 +47,31 @@
...
+  }
+
+  /**
+   * Pass in a Map of client level attributes which will be added to every single MetricsTracer
+   * created from the ApiTracerFactory.
+   */
+  public MetricsTracerFactory(MetricsRecorder metricsRecorder, Map<String, String> attributes) {
     this.metricsRecorder = metricsRecorder;
+    this.attributes = attributes;
   }
```
lqiu96, line 65: Additional constructor to take in user configured in client level attributes. 

In the future, we can have Gax automatically collect attributes and this would need a constructor to take in a set of attribute flags. Class is marked with InternalApi and could be open to more refactoring if we want to pursue this.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
lqiu96: I'll look to see how feasible it is to add some new unit tests to cover this.

lqiu96: Discussed offline. Adding a new map to be able to pass client level attributes to the MetricsTracerFactory.

lqiu96: Tried creating unit tests for this and I'm currently unable to find a way to mock the `isOnComputeEngine()` functionality as that reads from a file. I'll take a second look at this in a bit.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -84,11 +84,20 @@
  */
 @InternalExtensionOnly
 public final class InstantiatingGrpcChannelProvider implements TransportChannelProvider {
```
lqiu96, line 86: This is a final class and I'm unable to subclass this in the unit tests. For now, I've decided to keep the final keyword and workaround this by including package-private setters for the EnvProvider and the SystemProductNameReader.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
```diff
@@ -778,4 +787,58 @@ public void testHttpJson_multipleFailedAttempts_successfulOperation() throws Exc
...
+    grpcClient.echo(echoRequest);
+
+    List<MetricData> metricDataList = getMetricDataList(inMemoryMetricReader);
+    Map<String, String> attributeMapping =
+        ImmutableMap.of(
+            MetricsTracer.METHOD_NAME_ATTRIBUTE,
+            "Echo.Echo",
+            MetricsTracer.LANGUAGE_ATTRIBUTE,
+            MetricsTracer.DEFAULT_LANGUAGE,
+            "directpath_enabled",
```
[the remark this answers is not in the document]
lqiu96: Will do

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
```diff
@@ -778,4 +787,58 @@ public void testHttpJson_multipleFailedAttempts_successfulOperation() throws Exc
...
+            "false",
+            "testing",
+            "showcase");
+    verifyDefaultMetricsAttributes(metricDataList, attributeMapping);
```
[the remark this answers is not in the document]
lqiu96: Sure! will do

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracerFactory.java
```diff
@@ -45,13 +47,32 @@
 public class MetricsTracerFactory implements ApiTracerFactory {
   protected MetricsRecorder metricsRecorder;
 
+  // These are client attributes and pertain every single trace
+  private final Map<String, Object> attributes;
```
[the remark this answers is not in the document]
lqiu96: I went back and forth between `Map<String, String>` and `Map<String, Object>`. Since `canUseDirectPath()` returns a boolean: https://github.com/googleapis/sdk-platform-java/blob/b1be846e2a830a19c9f9ad920336a63283b734cd/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L466-L471, I would need to case it back to a String before adding it to the map. I was thinking that we accept whatever and then cast it ourselves with `String.valueOf()`.

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracerFactory.java
```diff
@@ -45,13 +47,32 @@
...
+    this(metricsRecorder, ImmutableMap.of());
+  }
+
+  /**
+   * Pass in a Map of client level attributes which will be added to every single MetricsTracer
+   * created from the ApiTracerFactory.
+   */
+  public MetricsTracerFactory(MetricsRecorder metricsRecorder, Map<String, Object> attributes) {
     this.metricsRecorder = metricsRecorder;
+    this.attributes = attributes;
```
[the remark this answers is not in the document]
lqiu96: Sure. Will update.

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracerFactory.java
```diff
@@ -45,13 +47,32 @@
...
   }
 
   @Override
   public ApiTracer newTracer(ApiTracer parent, SpanName spanName, OperationType operationType) {
-    return new MetricsTracer(
-        MethodName.of(spanName.getClientName(), spanName.getMethodName()), metricsRecorder);
+    MetricsTracer metricsTracer =
+        new MetricsTracer(
+            MethodName.of(spanName.getClientName(), spanName.getMethodName()), metricsRecorder);
+    for (Map.Entry<String, Object> attributeEntrySet : attributes.entrySet()) {
```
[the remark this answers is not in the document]
lqiu96: Sure, will update.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -753,6 +780,18 @@ public Builder setAttemptDirectPathXds() {
       return this;
     }
 
+    @VisibleForTesting
+    Builder setEnvProvider(EnvironmentProvider envProvider) {
```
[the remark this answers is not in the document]
lqiu96: I am using it for these unit tests: https://github.com/googleapis/sdk-platform-java/blob/b1be846e2a830a19c9f9ad920336a63283b734cd/gax-java/gax-grpc/src/test/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProviderTest.java#L681-L734

To cover the system env variable tests

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -117,6 +126,7 @@ public final class InstantiatingGrpcChannelProvider implements TransportChannelP
   @Nullable private final Boolean allowNonDefaultServiceAccount;
   @VisibleForTesting final ImmutableMap<String, ?> directPathServiceConfig;
   @Nullable private final MtlsProvider mtlsProvider;
+  private final SystemProductNameReader systemProductNameReader;
```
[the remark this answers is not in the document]
lqiu96: There is a bit of change added just for testing. I think there is quite a bit of added benefit, namely that we get to test multiple functions and also to get to test how they interact together: https://github.com/googleapis/sdk-platform-java/blob/b1be846e2a830a19c9f9ad920336a63283b734cd/gax-java/gax-grpc/src/test/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProviderTest.java#L668-L809. I believe the behavior is kept the same and I am adding wrappers for the same functionality.

The added classes and methods are all package-private scope and should not impact the customers at all. And since they are package-private, we should be able to easily remove and refactor to use mockito v5 for this same purpose.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
lqiu96: Did you mean that I should remove the annotation? Or that I should keep the everything the same?

blakeli0: We should remove the annotation.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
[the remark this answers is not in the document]
blakeli0: The logic is the same but the behavior is slightly different. `isOnComputeEngine()` used to be a static method, now it is not because we have to initialize an instance of `SystemProductNameReader`. It would be great if we can keep `isOnComputeEngine()` static. Or even better, `isOnComputeEngine()` does not have to be a static method, it could be a static block so that it does not get executed every time we call it. The benefit is marginal though since `InstantiatingGrpcChannelProvider` should only be called during client initialization.

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracerFactory.java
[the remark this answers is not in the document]
blakeli0: I think it is better to keep it consistent with the what we currently support in [MetricsTracer](https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracer.java#L64). We could change the type in `MetricsTracer` to `Map<String, Object>` as well, but that mean more work in [OpenTelemetryMetricsRecorder](https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/gax/src/main/java/com/google/api/gax/tracing/OpenTelemetryMetricsRecorder.java#L159-L164) because OpenTelemetry does not support put an attribute with `Object` value.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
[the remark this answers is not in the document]
lqiu96: I see, I have overlooked the intention behind `isOnComputeEngine()` being static. Let me see if I can find a way around this (either with keeping it static or moving it to a static clock).

I would also prefer to not change the behavior if possible.

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracerFactory.java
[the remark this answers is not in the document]
lqiu96: Ok, I will revert back to Map<String, String>. Users will be expected to cast their value to a String themselves.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -446,6 +450,24 @@ && canUseDirectPathWithUniverseDomain()) {
...
+   * settings and a few other configurations/settings must also be valid for the request to go
+   * through DirectPath.
+   *
+   * <p>Checks: 1. Credentials are compatible 2.Running on Compute Engine 3. Universe Domain is
+   * configured to for the Google Default Universe
+   *
+   * @return if DirectPath is enabled for the client AND if the configurations are valid
+   */
+  @InternalApi
+  public boolean canUseDirectPath() {
```
surbhigarg92, line 464: We should also add a check for [isDirectPathXdsEnabled](https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L261)

lqiu96: @mohanli-ml I believe you had helped implement this logic before. We're trying to expose a getter for the conditions that would enable DirectPath for this gRPC channel. Should `isDirectPathXdsEnabled()` be added here?

I copied over the original configs set: https://github.com/googleapis/sdk-platform-java/blob/16aec03465be61e9732a2213d0df2fd2181c7149/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L373-L376

lqiu96:
> Error:
> 
> ```
> Error:  Failures: 
> Error:    OpencensusTracerFactoryTest.testImplicitParentSpan:102 
> Wanted but not invoked:
> internalTracer.spanBuilderWithExplicitParent(
>     <any string>,
>     same(Mock for Span, hashCode: 125992315)
> );
> -> at com.google.api.gax.tracing.OpencensusTracerFactoryTest.testImplicitParentSpan(OpencensusTracerFactoryTest.java:102)
> 
> However, there were exactly 2 interactions with this mock:
> internalTracer.getCurrentSpan();
> -> at com.google.api.gax.tracing.OpencensusTracerFactory.newTracer(OpencensusTracerFactory.java:101)
> 
> internalTracer.spanBuilderWithExplicitParent(
>     "FakeClient.FakeMethod",
>     null
> );
> -> at com.google.api.gax.tracing.OpencensusTracerFactory.newTracer(OpencensusTracerFactory.java:110)
> 
> 
> Error:    OpencensusTracerTest.testLongRunningExample:152 
> Argument(s) are different! Wanted:
> span.addAnnotation(
>     "Operation started",
>     {}
> );
> -> at com.google.api.gax.tracing.OpencensusTracerTest.testLongRunningExample(OpencensusTracerTest.java:152)
> Actual invocations have different arguments:
> span.addAnnotation(
>     "Scheduling next poll",
>     {"delay ms" = AttributeValueLong{longValue=5}, "attempt" = AttributeValueLong{longValue=0}, "status" = AttributeValueString{stringValue=OK}}
> );
> -> at com.google.api.gax.tracing.OpencensusTracer.attemptFailed(OpencensusTracer.java:353)
> span.addAnnotation(
>     "Operation started"
> );
> -> at com.google.api.gax.tracing.OpencensusTracer.lroStartSucceeded(OpencensusTracer.java:290)
> span.addAnnotation(
>     "Polling completed",
>     {"attempt" = AttributeValueLong{longValue=1}}
> );
> -> at com.google.api.gax.tracing.OpencensusTracer.attemptSucceeded(OpencensusTracer.java:323)
> span.putAttributes(
>     {"attempt count" = AttributeValueLong{longValue=2}}
> );
> -> at com.google.api.gax.tracing.OpencensusTracer.operationSucceeded(OpencensusTracer.java:254)
> span.end(
>     
> );
> -> at com.google.api.gax.tracing.OpencensusTracer.operationSucceeded(OpencensusTracer.java:255)
> 
> [INFO] 
> Error:  Tests run: 614, Failures: 2, Errors: 0, Skipped: 0
> ```
> 
> It seems I have somehow broke OpenCensus with this change.

This seems to due to the added `mockito-inline` dependency used to mock static methods. I'm not sure why it's causing issues in non-related unit tests, so I'm going to remove this and look for a different way to test.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -82,13 +81,25 @@
...
+  static {
+    try {
+      systemProductName =
+          Files.asCharSource(new File("/sys/class/dmi/id/product_name"), StandardCharsets.UTF_8)
+              .readFirstLine();
+    } catch (IOException e) {
+      // Keep existing behavior the same (null means it is not on compute engine)
+      systemProductName = null;
+    }
+  }
```
lqiu96, line 97: SystemProductName logic moved to a static block and initialized once. Stored in a variable so that it can be overriden.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -753,6 +777,16 @@ public Builder setAttemptDirectPathXds() {
...
 
+    /**
+     * Package-Private scope as it is used to test DirectPath functionality in tests. This overrides
+     * the computed systemProductName when the class is initialized.
+     */
+    @VisibleForTesting
+    Builder setSystemProductName(String systemProductName) {
+      this.systemProductName = systemProductName;
+      return this;
+    }
```
lqiu96, line 788: Package-Private setter to allow tests to override the computed SystemProductName

lqiu96:
Based on https://github.com/googleapis/sdk-platform-java/issues/2818, I'll need to remove junit-pioneer and add mocks for EnvProvider.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -147,6 +158,19 @@ private InstantiatingGrpcChannelProvider(Builder builder) {
...
 
+  /**
+   * Package-Private constructor that is only visible for testing DirectPath functionality inside
+   * tests. This overrides the computed systemProductName when the class is initialized to help
+   * configure the result of {@link #isOnComputeEngine()} check.
+   *
+   * <p>If productName is null, that represents the result of an IOException
+   */
+  @VisibleForTesting
+  InstantiatingGrpcChannelProvider(Builder builder, String productName) {
```
lqiu96, line 169: Adding the EnvProvider as another param would need to make the envProvider variable non-final. I kept the package-private setter for now until we can migrate to using junit-pioneer to set the env vars.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
blakeli0: I think there are conditions that we have `canUseDirectPath` is true but `isDirectPathXdsEnabled` is false based on the current logic, maybe we can expose `isDirectPathXdsEnabled` as a public method, and the Spanner team can set a client level attribute based on `canUseDirectPath() && isDirectPathXdsEnabled()`?

blakeli0 (approved):

PR #2616, opened 2024-04-02
feat: add entry point

JoeWang1127:
In this PR:
- Create `entry_point.py` to combine `generate_repo.py` and `generate_pr_description.py`.
- Change Integration test.

Follow up of #2604.

For the goal of series of PRs, please refer to [improvement proposal](https://docs.google.com/document/d/1JiCcG3X7lnxaJErKe0ES_JkyU7ECb40nf2Xez3gWvuo/edit?tab=t.g3vua2kd06gx#bookmark=id.72s3ukwwzevo).


library_generation/cli/entry_point.py
```diff
@@ -0,0 +1,94 @@
...
+    type=str,
+    help="""
+    Path to generation_config.yaml that contains the metadata about
+    library generation.
+    The googleapis commit in the configuration is the baseline commit,
+    exclusively, from which the commit message is considered.
+    """,
+)
+@click.option(
+    "--latest-generation-config",
```
blakeli0, line 43: Can we all it something more generic?Because theoretically we can compare any two configs, not always compare with the latest. 

In addition, what if the `latest-generation-config` has a googleapis commitish that is prior to it in `basedline-generation-config`? I don't think we have to support this case, but we should at least handle this case so our script does not blow up, maybe by adding validations that the googleapis commitish in one of them should be higher than the other.

library_generation/cli/entry_point.py
```diff
@@ -0,0 +1,94 @@
...
+@click.pass_context
+@click.version_option(message="%(version)s")
+def main(ctx):
+    pass
+
+
+@main.command()
+@click.option(
+    "--baseline-generation-config",
+    required=True,
```
blakeli0, line 33: Can we make the interface as simple as possible? For example, both of the configs can be optional. We can use the existence of baseline config to decide if we want to to generate PR descriptions, and assume the current config to be exist in the current folder?

library_generation/cli/entry_point.py
JoeWang1127: > Can we all it something more generic?Because theoretically we can compare any two configs, not always compare with the latest.

I think these two names are intuitive, do you have other candidates?

> what if the latest-generation-config has a googleapis commitish that is prior to it in basedline-generation-config?

The problem is we don't the relationship between two commitish without traversing from one to another. The worst scenario is the description has too many characters (due to lots of commits) and `gh` command returns error.

JoeWang1127: We decided to use `current_generation_config` and `baseline_generation_config` as parameter names.

The googleapis commitish in `current_generation_config` should be newer than `baseline_generation_config` and we'll verify this by commit time.

library_generation/cli/entry_point.py
JoeWang1127: If both are optional, then we have:
1. both are not specified, we assume there's a `generation_config.yaml` in the current dir and do not generate pr description.
2. one is specified, we use it to generate repo and do not generate pr description.
3. both are specified, we use `current-generation-config` to generate repo and generate pr description.

library_generation/cli/entry_point.py
```diff
@@ -0,0 +1,118 @@
...
+@main.command()
+@click.option(
+    "--baseline-generation-config",
+    required=False,
+    default=None,
+    type=str,
+    help="""
+    Path to generation_config.yaml that contains the metadata about
+    library generation.
+    The googleapis commit in the configuration is the oldest commit,
```
blakeli0, line 39: Can we include some high level description for `baseline-generation-config`? For example, something like "We compare baseline-generation-config with the current generation config, and generate commit messages based on the diff of the googleapis-commitish between the two generation configs." 
Same comment for `current-generation-config` below.

library_generation/cli/entry_point.py
```diff
@@ -0,0 +1,118 @@
...
+    default_generation_config = f"{os.getcwd()}/generation_config.yaml"
+    if baseline_generation_config is None and current_generation_config is None:
+        if not os.path.isfile(default_generation_config):
+            raise FileNotFoundError(
+                f"{default_generation_config} does not exist when "
+                f"both baseline_generation_config and current_generation_config"
+                f" are not specified."
+            )
+        current_generation_config = default_generation_config
+    elif current_generation_config is None:
```
blakeli0, line 81: If `current_generation_config` is none, and `baseline_generation_config` is provided, I think we should throw an error. `current_generation_config` should be the source of truth for library generation.

library_generation/cli/entry_point.py
```diff
@@ -0,0 +1,118 @@
...
+        # make sure current_generation_config is not None if only one config
+        # is specified.
+        current_generation_config = baseline_generation_config
+        baseline_generation_config = None
+    current_generation_config = os.path.abspath(current_generation_config)
+    repository_path = os.path.abspath(repository_path)
+    if not baseline_generation_config:
```
blakeli0, line 88: Can we add a test for this scenario? Either in unit test or integration test.

library_generation/cli/entry_point.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+@main.command()
+@click.option(
+    "--baseline-generation-config",
+    required=False,
+    default=None,
+    type=str,
+    help="""
+    Path to generation_config.yaml that contains the metadata about
```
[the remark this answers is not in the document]
JoeWang1127: Relative path is fine because all path will be converted to absolute path later.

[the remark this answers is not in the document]
blakeli0: So either relative path or absolute path should work? If that's the case, let's document it here as well.

library_generation/generate_pr_description.py
```diff
@@ -121,7 +131,15 @@ def __get_commit_messages(
     shutil.rmtree(tmp_dir, ignore_errors=True)
     os.mkdir(tmp_dir)
     repo = Repo.clone_from(repo_url, tmp_dir)
-    commit = repo.commit(latest_commit)
+    commit = repo.commit(current_commit)
+    current_commit_time = __get_commit_timestamp(commit)
+    baseline_commit_time = __get_commit_timestamp(repo.commit(baseline_commit))
+    if current_commit_time < baseline_commit_time:
```
blakeli0, line 137: Like this validation!  I think we should add `=` though.
```suggestion
    if current_commit_time <= baseline_commit_time:
```

library_generation/cli/entry_point.py
```diff
@@ -0,0 +1,118 @@
...
+    baseline_generation_config: str,
+    current_generation_config: str,
+    repository_path: str,
+):
+    default_generation_config = f"{os.getcwd()}/generation_config.yaml"
+    if baseline_generation_config is None and current_generation_config is None:
+        if not os.path.isfile(default_generation_config):
+            raise FileNotFoundError(
+                f"{default_generation_config} does not exist when "
+                f"both baseline_generation_config and current_generation_config"
```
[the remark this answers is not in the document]

library_generation/cli/entry_point.py
JoeWang1127: Done.

library_generation/cli/entry_point.py
JoeWang1127: Done.

library_generation/generate_pr_description.py
JoeWang1127: Done.

Added unit tests to verify the behavior.

library_generation/cli/entry_point.py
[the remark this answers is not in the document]

library_generation/cli/entry_point.py
JoeWang1127: I added high level description of `generate` method in doc string.
User can read the docs through
```
python library_generation/cli/entry_point.py generate --help
```

blakeli0: Thanks, looks good! I think we still need more info to distinguish between `baseline-generation-config` from `current-generation-config`. For example, both descriptions include ```Absolute or relative path to generation_config.yaml that contains the
    metadata about library generation.```, in fact, only the info in `current-generation-config` is used for generation, `baseline-generation-config` is only used for generating commit history. Also the additional info regarding inclusive/exclusive of googleapis commitish might be a little too detailed.

library_generation/cli/entry_point.py
```diff
@@ -0,0 +1,144 @@
...
+    config_change = compare_config(
+        baseline_config=from_yaml(baseline_generation_config),
+        current_config=from_yaml(current_generation_config),
+    )
+    generate_from_yaml(
+        config=config_change.current_config,
+        repository_path=repository_path,
+        target_library_names=config_change.get_changed_libraries(),
+    )
+    generate_pr_descriptions(
```
blakeli0, line 136: We probably want to rename `generate_pr_description` to something like `generate_commit_history`, because we just happen to use PR description to capture the commit history, and we could use other ways. It could be a separate PR.

library_generation/cli/entry_point.py
JoeWang1127: Changed the comment.

blakeli0 (approved):

library_generation/README.md
```diff
@@ -184,20 +188,23 @@ libraries:
...
 
 ```bash
 # install python module (allows the `library_generation` module to be imported from anywhere)
 python -m pip install -r library_generation/requirements.in
+# install library_generation module
+python -m pip install library_generation
 # generate the repository
-python -m library_generation/generate_repo.py generate \
---generation-config-yaml=/path/to/config-file \
+python -m library_generation/entry_point.py generate \
```
blakeli0, line 199: FYI, we have to remove `-m` here, this is what tripped me when I run locally, I updated it in my [PR](https://github.com/googleapis/sdk-platform-java/pull/2619)

PR #2619, opened 2024-04-03
refactor: Rename requirements.in to requirements.txt.

blakeli0:
Rename requirements.in to requirements.txt. It is a more standard way to manage Python dependencies, it can be recognized automatically by some IDEs.

Add environment setup section to README for running hermetic build scripts.


library_generation/README.md
```diff
@@ -10,6 +10,8 @@ configuration file.
 - Java runtime environment (8 or above)
 - Apache Maven (used in formatting source code)
 - Python (3.11.6 or above)
+- Docker 
+- Git
```
JoeWang1127, line 14: Why do we need `git`?

JoeWang1127 (approved):

library_generation/README.md
blakeli0: There are some [post processing scripts](https://github.com/googleapis/sdk-platform-java/blob/main/library_generation/postprocess_library.sh#L136-L143) perform git operations, I don't think we should  continue this way and should refactor it in the future, but currently we do.

PR #2630, opened 2024-04-10
feat: Parser to consume the api-versioning value from proto

zhumin8:
This is part one to resolve b/330980553, this pr allows read and parse api versioning from proto to service model.

Changes included in this pr:
- rename current apiVersion to packageVersion. This packageVersion is parsed from package name and exposed in #1075. It was intended to be used in Spring codegen, but not actually used. It is used in generating sample region tags (see details [9ec3531](https://github.com/googleapis/sdk-platform-java/pull/2630/commits/9ec3531ea267c641ed265d685d4d4fecbe2285f3)) 
- exposing apiVersion from Service model.
- Parser.java to parse apiVersion from `ClientProto.apiVersion` to service.


lqiu96:
> rename current apiVersion to apiMajorVersion. This apiMajorVersion is parsed from package name and exposed in https://github.com/googleapis/sdk-platform-java/pull/1075. It was intended to be used in Spring codegen, but not actually used. It is used in generating sample region tags (see details [9ec3531](https://github.com/googleapis/sdk-platform-java/pull/2630/commits/9ec3531ea267c641ed265d685d4d4fecbe2285f3))

I know you wrote that it's not actually used inside Spring codegen, but for sample region tags. Would this break any functionality for the generated samples in Spring? It sounds like it probably won't, but can double check this to confirm?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/model/Service.java
```diff
@@ -52,6 +55,10 @@ public boolean hasDescription() {
     return !Strings.isNullOrEmpty(description());
   }
 
+  public boolean hasApiVersion() {
+    return !Strings.isNullOrEmpty(apiVersion());
+  }
```
lqiu96, line 60: Do you know if it's possible where the apiVersion is set, but it's set to an empty string?

i.e. the Service proto has `google.api.api_version = ""`. It's not our responsibility to validate, but I'm wondering if it's possible

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -483,26 +483,33 @@ public static List<Service> parseService(
...
+                              serviceYamlProtoOpt,
+                              outputArgResourceNames,
+                              transport));
+
+              String apiVersion = null;
+              if (serviceOptions.hasExtension(ClientProto.apiVersion)) {
+                apiVersion = serviceOptions.getExtension(ClientProto.apiVersion);
+                service.setApiVersion(apiVersion);
+              }
+              return service.build();
```
lqiu96, line 512: nit: Since Service's apiVersion is nullable, can we move this extraction section to before the Builder?

```
String apiVersion = null;
if (serviceOptions.hasExtension(ClientProto.apiVersion)) {
  apiVersion = serviceOptions.getExtension(ClientProto.apiVersion);
}

return serviceBuilder
  ...
  .setApiVersion(apiVersion)
  .build()
```

gapic-generator-java/src/test/java/com/google/api/generator/gapic/model/ServiceTest.java
```diff
@@ -50,6 +46,19 @@ public class ServiceTest {
...
+    testServiceBuilder =
+        Service.builder()
+            .setName("Echo")
+            .setDefaultHost("localhost:7469")
+            .setOauthScopes(Arrays.asList("https://www.googleapis.com/auth/cloud-platform"))
+            .setPakkage(SHOWCASE_PACKAGE_NAME)
+            .setProtoPakkage(SHOWCASE_PACKAGE_NAME)
+            .setOriginalJavaPackage(SHOWCASE_PACKAGE_NAME)
+            .setOverriddenName("Echo");
+  }
```
lqiu96, line 60: Do we need to recreate the Service Builder before every test?

gapic-generator-java/src/test/proto/echo.proto
```diff
@@ -44,6 +44,7 @@ service Echo {
   // This service is meant to only run locally on the port 7469 (keypad digits
   // for "show").
   option (google.api.default_host) = "localhost:7469";
+  option (google.api.api_version) = "v1_20230601";
```
lqiu96, line 47: Depending on if `(google.api.api_version)=""` if allowed or not, can you add test case for that?

Ignore this comment if the above is not allowed.

lqiu96: Also, can you add a test case where there are multiple services being parsed and each service has a different api_version? It could also be where it doesn't necessarily follow `v{major version}_date` format. Maybe one could have a version like `(google.api.api_version)=5.3` or something

lqiu96:
Thanks @zhumin8! I think it look good! I left a few nits and a few open questions that popped in while I reviewing. I think we can just double check with the team to confirm.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/model/Service.java
parthea: Treating `api_version=""` the same as `api_version=null` sounds good to me.

zhumin8:
> > rename current apiVersion to apiMajorVersion. This apiMajorVersion is parsed from package name and exposed in #1075. It was intended to be used in Spring codegen, but not actually used. It is used in generating sample region tags (see details [9ec3531](https://github.com/googleapis/sdk-platform-java/pull/2630/commits/9ec3531ea267c641ed265d685d4d4fecbe2285f3))
> 
> I know you wrote that it's not actually used inside Spring codegen, but for sample region tags. Would this break any functionality for the generated samples in Spring (and/or any other generated samples)? It sounds like it probably won't, but can double check this to confirm?

For non-Spring samples, it's not breaking anything after the renaming. 
For Spring generator, [code search](https://github.com/search?q=repo%3AGoogleCloudPlatform%2Fspring-cloud-gcp+path%3A%2F%5Espring-cloud-generator%5C%2F%2F+.apiVersion%28%29&type=code) confirms `apiVersion()` is not used, only `apiShortName()` is used [here](https://github.com/search?q=repo%3AGoogleCloudPlatform%2Fspring-cloud-gcp+path%3A%2F%5Espring-cloud-generator%5C%2F%2F++apiShortName%28%29&type=code).

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -470,6 +470,11 @@ public static List<Service> parseService(
                 }
               }
 
+              if (serviceOptions.hasExtension(ClientProto.apiVersion)) {
+                String apiVersion = serviceOptions.getExtension(ClientProto.apiVersion);
+                serviceBuilder.setApiVersion(apiVersion);
+              }
```
lqiu96, line 476: nit:
```
              String apiVersion = null;
              if (serviceOptions.hasExtension(ClientProto.apiVersion)) {
                apiVersion = serviceOptions.getExtension(ClientProto.apiVersion);
              }
...
              return serviceBuilder
                       .setName(serviceName)
                       ....
                       .setApiVersion(apiVersion)
```

I believe the Service class can take a null or empty ApiVersion

gapic-generator-java/src/test/proto/echo.proto
```diff
@@ -142,6 +143,30 @@ service Echo {
...
+    };
+    option (google.api.method_signature) = "content";
+    option (google.api.method_signature) = "error";
+    option (google.api.method_signature) = "content,severity";
+    option (google.api.method_signature) = "name";
+    option (google.api.method_signature) = "parent";
+    option (google.api.method_signature) = "";
+  }
+}
+
```
lqiu96, line 169: Can you break this into a new proto file? Similar to https://github.com/googleapis/sdk-platform-java/blob/01bdd890bc71b89a52052336519206db1cd40f92/gapic-generator-java/src/test/proto/auto_populate_field_testing.proto

so we don't need to change the assertions for existing echo tests?

lqiu96 (approved):
LGTM. I added a few nits if you could resolve before merging.

lqiu96:
Can you also update the description: `rename current apiVersion to apiMajorVersion`. I believe it should be `rename current apiVersion to packageVersion`.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
zhumin8: I had similar concerns at first. But I believe both null and empty are covered. (I'll create additional test cases in ParserTest for it).
Note that If value is not set for serviceBuilder, it is null by default.

zhumin8:
@lqiu96 Moved testing to a dedicated new proto as you suggested. This`api_version_testing.proto` may contain some redundant definitions at this point. I plan to address either as part of the next pr, or in a followup pr.

lqiu96:
> @lqiu96 Moved testing to a dedicated new proto as you suggested. This`api_version_testing.proto` may contain some redundant definitions at this point. I plan to address either as part of the next pr, or in a followup pr.

Thanks, LGTM.

PR #2633, opened 2024-04-12
fix: exclude path_to_yaml when compares configs

JoeWang1127:
In this PR:
- Exclude `path_to_yaml` when compares generation configs because this parameter is not used in library generation.

### DIFF
--- a/library_generation/utils/generation_config_comparator.py
+++ b/library_generation/utils/generation_config_comparator.py
@@ -38,10 +38,12 @@ def compare_config(
     :return: a ConfigChange objects.
     """
     diff = defaultdict(list[LibraryChange])
+    # these params are not used in config comparison.
+    excluded_params = {"path_to_yaml"}
     baseline_params = __convert_params_to_sorted_list(baseline_config)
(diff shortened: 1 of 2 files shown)


library_generation/utils/generation_config_comparator.py
```diff
@@ -38,10 +38,12 @@ def compare_config(
     :return: a ConfigChange objects.
     """
     diff = defaultdict(list[LibraryChange])
+    # these params are not used in config comparison.
+    excluded_params = {"path_to_yaml"}
```
blakeli0, line 42: Why do we need this parameter in `GenerationConfig` class? How is it used? I thought `GenerationConfig` is a pure representation for the yaml, but looks like it includes more things.

JoeWang1127: This parameter is added in #2443.

Basically, this parameter is used to run template exclusion using our own implementation of synthtool.

JoeWang1127:
We'll try to remove this parameter out of `GenerationConfig`.

PR #2634, opened 2024-04-12
chore: remove `is_monorepo` and `path_to_yaml` from `GenerationConfig`

JoeWang1127:
In this PR:
- Ronvert `is_monorepo` to a method in `GenerationConfig`.
- Remove `path_to_yaml` from `GenerationConfig`.

These two parameters are not used for library generation. Therefore, they should not be a parameter in `GenerationConfig`.


library_generation/cli/entry_point.py
```diff
@@ -113,6 +113,7 @@ def generate(
         # baseline_generation_config is not specified.
         # Do not generate pull request description.
         generate_from_yaml(
+            config_path=current_generation_config,
```
blakeli0, line 116: Not part of this PR, `current_generation_config` and `baseline_generation_config` are actually paths not the content, so probably better to rename them as something like `current_generation_config_path`.

blakeli0 (approved):

diegomarquezp (approved):

library_generation/cli/entry_point.py
JoeWang1127: Raised #2638.

PR #2639, opened 2024-04-14
feat: add `libraries_bom_version` to generation configuration 

JoeWang1127:
In this PR:
- Add `libraries_bom_version` as a required parameter in generation config.

The integration test is verified using a feature branch of synthtool: https://github.com/googleapis/synthtool/pull/1956.

The synthtool commit in integration test is changed to `63cc541da2c45fcfca2136c43e638da1fbae174d` after the above feature branch is merged.


library_generation/README.md
```diff
@@ -97,6 +97,7 @@ They are shared by library level parameters.
 | protobuf_version        |    No    | inferred from the generator if not specified |
 | grpc_version            |    No    | inferred from the generator if not specified |
 | googleapis-commitish    |   Yes    |                                              |
+| libraries_bom_version   |   Yes    |                                              |
```
blakeli0, line 100: Can you please add it to our spec doc as well?

blakeli0 (approved):
Thanks, LGTM! The nightly generation is going to fail if we don't add it to the generation_config.yaml in `google-cloud-java` since this is a required parameter. I think we need to pin the docker image version in `google-cloud-java` soon, otherwise it will constantly be interrupted by changes like this and the OwlBot renaming PR.

library_generation/test/resources/integration/google-cloud-java/baseline_generation_config.yaml
```diff
@@ -1,8 +1,9 @@
-gapic_generator_version: 2.37.0
+gapic_generator_version: 2.38.1
 protobuf_version: 25.2
 googleapis_commitish: a17d4caf184b050d50cacf2b0d579ce72c31ce74
+libraries_bom_version: 26.37.0
 owlbot_cli_image: sha256:623647ee79ac605858d09e60c1382a716c125fb776f69301b72de1cd35d49409
-synthtool_commitish: 5e1fb2032fa44bc170677b38713023b4fec51a4e
+synthtool_commitish: 63cc541da2c45fcfca2136c43e638da1fbae174d
```
diegomarquezp, line 6: Just a side note that the old commitish `5e1fb2` belongs to a branch (not up to date with `main`) that had the java templates removed. Using a committish that contains java templates should not affect the generation because we have this logic in postprocessing:

https://github.com/googleapis/sdk-platform-java/blob/e542af8a8548fa336e2d3cd15111a0672358c84a/library_generation/owlbot/bin/entrypoint.sh#L51-L54

diegomarquezp (approved):
LGTM, just one note

PR #2642, opened 2024-04-15
chore: Change all the references of OwlBot.yaml to OwlBot-hermetic.yaml.

blakeli0:
This is follow up of https://github.com/googleapis/google-cloud-java/pull/10696. The hermetic build generation process would fail without this PR due to unable to find OwlBot.yaml files. 
This PR changes all the references of `OwlBot.yaml` to `OwlBot-hermetic.yaml`, removes `OwlBot.yaml` from a few places that don't need to reference it in the first places.


library_generation/postprocess_library.sh
```diff
@@ -44,7 +44,7 @@ for required_input in "${required_inputs[@]}"; do
   fi
 done
 
-for owlbot_file in ".repo-metadata.json" "owlbot.py" ".OwlBot.yaml"
+for owlbot_file in ".repo-metadata.json" "owlbot.py" ".OwlBot-hermetic.yaml"
```
JoeWang1127, line 47: Could you extract these values to variables?

blakeli0: For String literals, it is usually preferred to use it directly as it is more readable, especially if it is only used once. I do see `.OwlBot.yaml` being used multiple times, so I'll extract it and leave the other two as they are.

JoeWang1127 (approved):

diegomarquezp (approved):
LGTM, thank you

library_generation/postprocess_library.sh
```diff
@@ -74,12 +75,12 @@ if [[ "${is_monorepo}" == "true" ]]; then
   # - "/google-.*/src"
 
   library_name=$(basename "${postprocessing_target}")
-  cat "${postprocessing_target}/.OwlBot.yaml" \
+  cat "${postprocessing_target}/${owlbot_yaml_file_name}" \
     | sed "s/- \"\/${library_name}/ - \"/" \
     > "${postprocessing_target}/.OwlBot.hermetic.yaml"
```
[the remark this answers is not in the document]
blakeli0: Agreed!

PR #2615, opened 2024-04-02
chore: bake synthtool and owl-bot into library_generation docker image

diegomarquezp:
## Context
[Milestone 7 of Hermetic Build](https://docs.google.com/document/d/1v-sJBmdNEVBRryL8n90XK8OIiqVphaJemXVhANw14Jk/edit?pli=1&resourcekey=0-QGUb2do-JBlDWKvptWBp_g&tab=t.0#bookmark=id.bkbj04ib2d4n)

## In this PR
 - Bake synthtool into the docker image (we won't `git clone` it anymore)
 - Bake the owlbot CLI into the docker image (we won't use docker inside docker anymore)
 - Create a DEVELOPMENT.md file with instructions for local setup (example: install owlbot CLI locally)

## Follow ups
 - [ ] modify google-cloud-java's workflows to correct the volume mapping being used

PR #2644, opened 2024-04-16
chore: remove generated files in tests

JoeWang1127:
In this PR:
- Remove generated files after running tests.

Note that files generated by `docker run` have root permission, thus can't be removed by this PR. This issue will be resolved in #2615.


.gitignore
```diff
@@ -15,14 +15,12 @@ target/
...
-
 # Python
 **/__pycache__/
 .venv
 
 # library generation
-**/output/
+output/
 **/googleapis
 library_generation/test/resources/test_monorepo_postprocessing/
```
blakeli0, line 25: Do we know what's the effort to remove this line as well?

JoeWang1127: Removed.

library_generation/test/integration_tests.py
```diff
@@ -176,6 +187,11 @@ def __build_image(cls, docker_file: str, cwd: str):
             cwd=cwd,
         )
 
+    @classmethod
+    def __remove_generated_files(cls):
+        shutil.rmtree(f"{output_dir}", ignore_errors=True)
```
blakeli0, line 192: I tried this in the past and it did not work due to permissions. Can you remove `ignore_errors=True` and see if it works? Separately, why did we add `ignore_errors=True` in the first place? I think we should error out instead of ignore errors in integration tests.

JoeWang1127: Right now the permission issue is still there and will be resolved in #2615.

If the remove failed, the integration test will fail when running `git clone` in an non-empty directory.

blakeli0: So does it work or not at this moment? If we still have permission issues, I don't think we should add this line right now.

.gitignore
blakeli0: This line is still here, do you plan to remove it?

library_generation/test/integration_tests.py
JoeWang1127: I comment out this line.

.gitignore
JoeWang1127: Sorry I missed it, now it's removed.

diegomarquezp (approved):

library_generation/generate_repo.py
```diff
@@ -77,9 +77,11 @@ def generate(
         config_path=generation_config_yaml,
         config=config,
         repository_path=repository_path,
-        target_library_names=target_library_names.split(",")
-        if target_library_names is not None
-        else target_library_names,
+        target_library_names=(
+            target_library_names.split(",")
```
blakeli0, line 81: If I'm understanding correctly, this is splitting a String to a list by comma, and both the String field and the list field have the same name `target_library_names`. Can we reorganize the code and/or rename the fields to make it clear?

library_generation/test/utilities_unit_tests.py
```diff
@@ -373,6 +375,7 @@ def test_prepare_repo_split_repo_success(self):
         self.assertEqual("output", Path(repo_config.output_folder).name)
         library_path = sorted([Path(key).name for key in repo_config.libraries])
         self.assertEqual(["misc"], library_path)
+        shutil.rmtree(repo_config.output_folder, ignore_errors=True)
```
blakeli0, line 378: Why we can not use `__remove_generated_files` here?

library_generation/generate_repo.py
JoeWang1127: This reformat is done by `black`. It updates to a new version so this line needs to reformat to pass python linter check.

This piece of code will be removed once `google-cloud-java` uses `entry_point.py`.

Can we leave it as-is?

library_generation/test/utilities_unit_tests.py
JoeWang1127: `__remove_generated_files` is used to delete `.repo-metadata.json`, `.OwlBot-hermetic.yaml` and `owlbot.py` and I changed the method name to `__remove_prerequisite_files`.

library_generation/test/integration_tests.py
```diff
@@ -168,6 +175,7 @@ def test_entry_point_running_in_container(self):
                     "The generated PR description does not match the expected golden file",
                 )
                 print("  PR description comparison succeed.")
+        self.__remove_generated_files()
```
JoeWang1127, line 178: Remove the generated files after the integration test succeeds.

blakeli0 (approved):

PR #2354, opened 2024-01-10
deps: update gapic-showcase to v0.30.0

alicejli:
Steps: 

1) Updated the gapic-showcase version in `gapic-showcase/pom.xml` to 0.30.0
2) Ran `mvn compile -P update`

PR #2653, opened 2024-04-17
deps: update gapic-showcase to v0.33.0

zhumin8:
update gapic-showcase so that we can pickup changes in https://github.com/googleapis/gapic-showcase/pull/1484

Steps: 
1) Updated the gapic-showcase version in `gapic-showcase/pom.xml` to 0.33.0
2) Ran `mvn compile -P update`

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -106,17 +106,6 @@ http_archive(
     urls = ["https://github.com/googleapis/disco-to-proto3-converter/archive/%s.zip" % _disco_to_proto3_converter_commit],
 )
 
-# Showcase
-_showcase_version = "0.28.2"
-
-http_archive(
-    name = "com_google_gapic_showcase",
-    strip_prefix = "gapic-showcase-%s" % _showcase_version,
-    urls = [
-        "https://github.com/googleapis/gapic-showcase/archive/refs/tags/v%s.zip" % _showcase_version,
-    ],
-)
-
 http_archive(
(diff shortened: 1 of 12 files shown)


lqiu96:
@burkedavison Do you remember what this showcase version value in the WORKSPACE file is for? https://github.com/googleapis/sdk-platform-java/blob/main/WORKSPACE#L110

Do we need to keep this updated/ is this used at all?

zhumin8:
2 failed cis:
- showcase-clirr - okay to ignore according to https://github.com/googleapis/sdk-platform-java/pull/2354#issuecomment-1889954320
- [gapic-generator-java-root] sonarcloud. 
  - For coverage on new code, other than api-version, most changes comes from proto change related to request-id. cc. @alicejli for context. 
  - For duplication and reliability, consider these are generated code, Wonder what's our general approach here? cc. @lqiu96 @blakeli0 

image

lqiu96:
> For duplication and reliability, consider these are generated code, Wonder what's our general approach here?

I think it's fine for us to merge it in. We can't change the protoc generated code and the generated output from the gapic-generator probably needs a larger refactor to "fix".

zhumin8:
@blakeli0 @lqiu96 Could you take another look?
I'd like to merge this in if it looks good.

lqiu96 (approved):
LGTM, thanks!

PR #2671, opened 2024-04-19
feat: Update Gapic generator and Gax to emit api-versioning via header

zhumin8:
This is part 2, following changes in https://github.com/googleapis/sdk-platform-java/pull/2630

Changes in this pr:
- ApiClientHeaderProvider: add key and token setter for api version.
- AbstractServiceStubSettingsClassComposer: add logic to set this header when api-version is present (not null or empty)
- Unit and golden tests to reflect above changes. (This includeds a few new golden files created for tests based on the dedicated proto created in #2630)

This pr does not include gapic-showcased testing for now.
---
manual tested with [gapic-showcase](https://github.com/googleapis/gapic-showcase) v0.33.0

steps: 
- Use gapic-showcase v0.33.0 (which includes update in 1484)
- use `gapic-showcase run -v` which prints the request header
- From showcase folder, run `mvn verify -P enable-integration-tests -Dit.test=ITAutoPopulatedFields.java -pl=gapic-showcase`

Printed header in manual testing
Grpc
```
2024/04/16 10:26:41 Received Unary Request for Method: /google.showcase.v1beta1.Echo/Echo
2024/04/16 10:26:41     Request headers:
2024/04/16 10:26:41       grpc-accept-encoding: gzip
2024/04/16 10:26:41       content-type: application/grpc
2024/04/16 10:26:41       x-goog-api-version: v1_20240408
2024/04/16 10:26:41       :authority: localhost:7469
2024/04/16 10:26:41       user-agent: grpc-java-netty/1.62.2
2024/04/16 10:26:41       x-goog-api-client: gl-java/17.0.7__Eclipse-Adoptium__Temurin-17.0.7-7 gapic/0.0.1-SNAPSHOT gax/2.46.2-SNAPSHOT grpc/1.62.2
2024/04/16 10:26:41     Request:  error:{code:2}  request_id:"1e26ee16-ed07-4eeb-be0e-49b769678779"  other_request_id:"40883e5d-3b20-4add-854c-9808bd009260"
2024/04/16 10:26:41     Returning Error: rpc error: code = Unknown desc = 
```
Httpjson
```
2024/04/16 10:26:41 Received POST request matching '/v1beta1/echo:echo': "/v1beta1/echo:echo"                                                               
2024/04/16 10:26:41   urlPathParams (expect 0, have 0): map[]                 
2024/04/16 10:26:41   urlRequestHeaders:                                                                                                                    
    User-Agent: "Google-HTTP-Java-Client/1.44.1 (gzip)"                                                                                                     
    Content-Type: "application/json; charset=utf-8"                           
    Connection: "keep-alive"                                                                                                                                
    Accept-Encoding: "gzip"                                                   
    X-Goog-Api-Client: "gl-java/17.0.7__Eclipse-Adoptium__Temurin-17.0.7-7 gapic/0.0.1-SNAPSHOT gax/2.46.2-SNAPSHOT rest/"                                  
    X-Goog-Api-Version: "v1_20240408"                                                                                                                       
    Accept: "text/html, image/gif, image/jpeg, *; q=.2, */*; q=.2"            
    Content-Length: "129"                                                     
2024/04/16 10:26:41   request: {                                                                                                                            
  "error":  {                                                                                                                                               
    "code":  500,                                                                                                                                           
    "message":  "",                                                                                                                                         
    "details":  []                                                            
  },                                                                                                                                                        
  "severity":  "UNNECESSARY",                                                                                                                               
  "header":  "",                                                                                                                                            
  "otherHeader":  "",                                                         
  "requestId":  "e690a84b-176d-421e-9e5d-ba752f68fec8",                       
  "otherRequestId":  "47e0fbde-056a-43ae-bba6-8b6770d861f7"                   
}    
```

### DIFF
--- a/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceStubSettingsClassComposer.java
+++ b/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceStubSettingsClassComposer.java
@@ -370,6 +370,16 @@ public abstract class AbstractServiceStubSettingsClassComposer implements ClassC
             .setReturnType(returnType)
             .build();
 
+    if (!Strings.isNullOrEmpty(service.apiVersion())) {
+
+      returnExpr =
+          MethodInvocationExpr.builder()
+              .setExprReferenceExpr(returnExpr)
+              .setMethodName("setApiVersionToken")
+              .setArguments(ValueExpr.withValue(StringObjectValue.withValue(service.apiVersion())))
+              .setReturnType(returnType)
+              .build();
+    }
     return MethodDefinition.builder()
         .setScope(ScopeNode.PUBLIC)
         .setIsStatic(true)
--- a/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceStubSettingsClassComposer.java
+++ b/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceStubSettingsClassComposer.java
@@ -370,7 +370,7 @@ public abstract class AbstractServiceStubSettingsClassComposer implements ClassC
             .setReturnType(returnType)
             .build();
 
-    if (!Strings.isNullOrEmpty(service.apiVersion())) {
+    if (service.hasApiVersion()) {
 
       returnExpr =
           MethodInvocationExpr.builder()
--- a/gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpcrest/ServiceStubSettingsClassComposerTest.java
+++ b/gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpcrest/ServiceStubSettingsClassComposerTest.java
@@ -23,37 +23,51 @@ import com.google.api.generator.test.framework.GoldenFileWriter;
 import com.google.api.generator.test.protoloader.GrpcRestTestProtoLoader;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.Arrays;
+import java.util.Collection;
 import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
 
+@RunWith(Parameterized.class)
 public class ServiceStubSettingsClassComposerTest {
+  @Parameterized.Parameters
+  public static Collection<Object[]> data() {
+    return Arrays.asList(
+        new Object[][] {
+          {"EchoStubSettings.golden", GrpcRestTestProtoLoader.instance().parseShowcaseEcho()},
+          {"WickedStubSettings.golden", GrpcRestTestProtoLoader.instance().parseShowcaseWicked()}
+        });
+  }
+
+  @Parameterized.Parameter public String goldenFileName;
+
+  @Parameterized.Parameter(1)
+  public GapicContext context;
+
   @Test
   public void generateServiceClasses() {
-    GapicContext context = GrpcRestTestProtoLoader.instance().parseShowcaseEcho();
     Service echoProtoService = context.services().get(0);
     GapicClass clazz =
         ServiceStubSettingsClassComposer.instance().generate(context, echoProtoService);
 
     JavaWriterVisitor visitor = new JavaWriterVisitor();
     clazz.classDefinition().accept(visitor);
-    GoldenFileWriter.saveCodegenToFile(this.getClass(), "EchoStubSettings.golden", visitor.write());
-    Path goldenFilePath =
-        Paths.get(GoldenFileWriter.getGoldenDir(this.getClass()), "EchoStubSettings.golden");
+    GoldenFileWriter.saveCodegenToFile(this.getClass(), goldenFileName, visitor.write());
+    Path goldenFilePath = Paths.get(GoldenFileWriter.getGoldenDir(this.getClass()), goldenFileName);
     Assert.assertCodeEquals(goldenFilePath, visitor.write());
   }
 
   @Test
   public void generateServiceClassesWicked() {
-    GapicContext context = GrpcRestTestProtoLoader.instance().parseShowcaseWicked();
     Service wickedProtoService = context.services().get(0);
     GapicClass clazz =
         ServiceStubSettingsClassComposer.instance().generate(context, wickedProtoService);
 
     JavaWriterVisitor visitor = new JavaWriterVisitor();
     clazz.classDefinition().accept(visitor);
-    GoldenFileWriter.saveCodegenToFile(
-        this.getClass(), "WickedStubSettings.golden", visitor.write());
-    Path goldenFilePath =
-        Paths.get(GoldenFileWriter.getGoldenDir(this.getClass()), "WickedStubSettings.golden");
+    GoldenFileWriter.saveCodegenToFile(this.getClass(), goldenFileName, visitor.write());
+    Path goldenFilePath = Paths.get(GoldenFileWriter.getGoldenDir(this.getClass()), goldenFileName);
     Assert.assertCodeEquals(goldenFilePath, visitor.write());
   }
 }
--- a/gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpcrest/ServiceStubSettingsClassComposerTest.java
+++ b/gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpcrest/ServiceStubSettingsClassComposerTest.java
@@ -57,17 +57,4 @@ public class ServiceStubSettingsClassComposerTest {
     Path goldenFilePath = Paths.get(GoldenFileWriter.getGoldenDir(this.getClass()), goldenFileName);
     Assert.assertCodeEquals(goldenFilePath, visitor.write());
   }
-
-  @Test
-  public void generateServiceClassesWicked() {
-    Service wickedProtoService = context.services().get(0);
-    GapicClass clazz =
-        ServiceStubSettingsClassComposer.instance().generate(context, wickedProtoService);
-
-    JavaWriterVisitor visitor = new JavaWriterVisitor();
-    clazz.classDefinition().accept(visitor);
-    GoldenFileWriter.saveCodegenToFile(this.getClass(), goldenFileName, visitor.write());
-    Path goldenFilePath = Paths.get(GoldenFileWriter.getGoldenDir(this.getClass()), goldenFileName);
-    Assert.assertCodeEquals(goldenFilePath, visitor.write());
-  }
 }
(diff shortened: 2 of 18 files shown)


zhumin8:
@lqiu96 As we talked off-line, I am still working on some details, plus integration testing. But feel free to take a first look.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceStubSettingsClassComposer.java
```diff
@@ -370,6 +370,16 @@ protected MethodDefinition createApiClientHeaderProviderBuilderMethod(
             .setReturnType(returnType)
             .build();
 
+    if (!Strings.isNullOrEmpty(service.apiVersion())) {
```
lqiu96, line 373: I know part 1 is not merged yet. Adding a note here to remind us for `service.hasApiVersion()`

gax-java/gax/src/main/java/com/google/api/gax/rpc/ApiClientHeaderProvider.java
```diff
@@ -68,6 +70,9 @@ protected ApiClientHeaderProvider(Builder builder) {
       headersBuilder.put(QUOTA_PROJECT_ID_HEADER_KEY, builder.getQuotaProjectIdToken());
     }
 
+    if (builder.getApiVersionToken() != null) {
```
lqiu96, line 73: I think it should be `Token != null && token != ""`

gax-java/gax/src/test/java/com/google/api/gax/rpc/ApiClientHeaderProviderTest.java
```diff
@@ -40,6 +40,7 @@ public class ApiClientHeaderProviderTest {
 
   private static final String X_GOOG_API_CLIENT = "x-goog-api-client";
   private static final String CLOUD_RESOURCE_PREFIX = "google-cloud-resource-prefix";
+  private static final String API_VERSION_HEADER_KEY = "x-goog-api-version";
 
   @Test
   public void testServiceHeaderDefault() {
```
lqiu96, line 46: Can you add an assertion to the default test case to show that `provider.getHeaders()` does not have `API_VERSION_HEADER_KEY`

gax-java/gax/src/test/java/com/google/api/gax/rpc/ApiClientHeaderProviderTest.java
```diff
@@ -138,4 +139,11 @@ public void testQuotaProjectHeader() {
     assertThat(provider.getHeaders().get(ApiClientHeaderProvider.QUOTA_PROJECT_ID_HEADER_KEY))
         .matches(quotaProjectHeaderValue);
   }
+  @Test
+  public void testApiVersionHeader() {
```
lqiu96, line 143: Add a case if the user sets the ApiVersionToken to ""

gax-java/gax/src/test/java/com/google/api/gax/rpc/ApiClientHeaderProviderTest.java
```diff
@@ -40,6 +40,7 @@ public class ApiClientHeaderProviderTest {
 
   private static final String X_GOOG_API_CLIENT = "x-goog-api-client";
   private static final String CLOUD_RESOURCE_PREFIX = "google-cloud-resource-prefix";
+  private static final String API_VERSION_HEADER_KEY = "x-goog-api-version";
```
lqiu96, line 43: I think the token above is package-private and we *should* be able to access it there. Better to define it at one spot if possible.

gax-java/gax/src/main/java/com/google/api/gax/rpc/ApiClientHeaderProvider.java
zhumin8: I am not exactly sure I am following your thoughts here. 
If you are talking about the public setter in the Builder, then I think we are safe here. This constructor is protected and is meant to enforce construction through the builder.

or is this what you have in mind?
```
ApiClientHeaderProvider emptyProvider = ApiClientHeaderProvider.newBuilder().setApiVersionToken("")
        .build();
```
I don't think it's this HeaderProvider's responsibility to make any judgement on the header values. This logic belongs to `AbstractServiceStubSettingsClassComposer.java` via `service.hasApiVersion()`. I'll add in relevant tests to demonstrate.

zhumin8 (commented):
Working through testing changes after separated out api_version_testing.proto. Will update once ready.

gax-java/gax/src/main/java/com/google/api/gax/rpc/ApiClientHeaderProvider.java
zhumin8: I am not exactly sure I am following your thoughts here. 
If you are talking about the public setter in the Builder, then I think we are safe here. This constructor is protected and is meant to enforce construction through the builder.

or is this what you have in mind?
```
ApiClientHeaderProvider emptyProvider = ApiClientHeaderProvider.newBuilder().setApiVersionToken("")
        .build();
```
I don't think it's this HeaderProvider's responsibility to make any judgement on the header values. This logic belongs to `AbstractServiceStubSettingsClassComposer.java` via `service.hasApiVersion()`. I'll add in relevant tests to demonstrate.

lqiu96: > I don't think it's this HeaderProvider's responsibility to make any judgement on the header values. This logic belongs to AbstractServiceStubSettingsClassComposer.java via service.hasApiVersion(). I'll add in relevant tests to demonstrate.

The composer only checks the api versionvalue from the proto file. This case is for any runtime configurations that the user adds to the client library. We currently treat any service configuration where api_version of `""` as null and I am currently thinking that it would make sense for the same behavior here.

IIUC, ClientContext does make judgement on the header values and does some validation: https://github.com/googleapis/sdk-platform-java/blob/fe0481ec599f26f6cfe4df48b404da20ee447f57/gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java#L297-L308. I was thinking that it should be enforced, but that isn't required from this PR.

zhumin8:
Force pushed after rebase with main + mainly test updates.

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpcrest/ServiceStubSettingsClassComposerTest.java
```diff
@@ -23,37 +23,51 @@
...
     JavaWriterVisitor visitor = new JavaWriterVisitor();
     clazz.classDefinition().accept(visitor);
-    GoldenFileWriter.saveCodegenToFile(
-        this.getClass(), "WickedStubSettings.golden", visitor.write());
-    Path goldenFilePath =
-        Paths.get(GoldenFileWriter.getGoldenDir(this.getClass()), "WickedStubSettings.golden");
+    GoldenFileWriter.saveCodegenToFile(this.getClass(), goldenFileName, visitor.write());
+    Path goldenFilePath = Paths.get(GoldenFileWriter.getGoldenDir(this.getClass()), goldenFileName);
     Assert.assertCodeEquals(goldenFilePath, visitor.write());
   }
```
lqiu96, line 72: Now that this is included as a parameterized test. I don't think we need this method anymore.

gax-java/gax/src/test/java/com/google/api/gax/rpc/ApiClientHeaderProviderTest.java
```diff
@@ -138,4 +138,18 @@ public void testQuotaProjectHeader() {
...
+    ApiClientHeaderProvider provider =
+        ApiClientHeaderProvider.newBuilder().setApiVersionToken("fake-version-string").build();
+    assertThat(provider.getHeaders().size()).isEqualTo(2);
+    assertThat(provider.getHeaders().get(ApiClientHeaderProvider.API_VERSION_HEADER_KEY))
+        .matches("fake-version-string");
+
+    ApiClientHeaderProvider emptyProvider =
+        ApiClientHeaderProvider.newBuilder().setApiVersionToken("").build();
+    assertThat(
+        emptyProvider.getHeaders().get(ApiClientHeaderProvider.API_VERSION_HEADER_KEY).isEmpty());
```
lqiu96, line 153: My personal opinion is that I think this case should result in the ApiVersion header not being passed (similar to how it would behave if the service proto doesn't have this option). I know the client library can't reasonably do validation all values passed in to the ApiClientHeader, but empty string and null are cases which we know aren't valid ApiVersions.

The reason I have this case in mind is because it is something we check for inside the EndpointContext: https://github.com/googleapis/sdk-platform-java/blob/75352d197f1744435cca83ac788edea51de787f8/gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java#L220-L222

zhumin8: Thanks for the context. 

The way I see it currently is that, this logic of validating api version should not be repeated if possible. One benefit is that in case of future changes to it, say additional logic to treat a group of versions in a way (totally hypothetical here), change can be centralized. 
Currently, this apiVersion information flows from:
proto --> via Parser --> model (Service) --> via composer --> output code (StubSettings) --> creates a ApiClientHeaderProvider instance.
`ApiClientHeaderProvider.java` does not need to know the details about how to validate api versions because it's already validated upstream.

WDYT?

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpcrest/ServiceStubSettingsClassComposerTest.java
zhumin8: Ah. good catch.

gax-java/gax/src/test/java/com/google/api/gax/rpc/ApiClientHeaderProviderTest.java
lqiu96: I think we might be on two different pages here. I don't intend for ApiClientHeaderProvider to try and validate the value that is set by the service team inside the proto file. If that value is not set by the team or is set as empty, then I trust that the generator has the correct logic to determine when the `setApiVersionToken()` code should or should not be generated.

My concern is that both ApiClientHeaderProvider and the ApiVersionToken setter is public. Users can create an ApiClientHeaderProvider separately and then pass it into the client via StubSettings (https://github.com/googleapis/sdk-platform-java/blob/849bb503b9e09310e8225c2997a5ceeb866dab31/gax-java/gax/src/main/java/com/google/api/gax/rpc/StubSettings.java#L442-L449). I believe it's possible given that we previously have suggested to users to set their own header provider values for certain workaround use cases.

I agree that's it's a extremely rare possibility that users are both creating a custom HeaderProvider and intentionally setting the ApiVersionToken to `""`, but because it's possible, I don't want to rule that possibility out. IMO, we should keep consistent with the default behavior done automatically for GAPICs (i.e. if it is set to be null or empty string, then it's not set in the header).

Granted this is a small nit pick that I think is a better UX. If you feel strongly otherwise, I don't want to continue to block on this small part. I think the other bits LGTM and I will do another pass before approving. Thanks for all the changes :)

lqiu96 (approved):

gax-java/gax/src/test/java/com/google/api/gax/rpc/ApiClientHeaderProviderTest.java
zhumin8: Followup discussions directed off-line. 
Merging this pr for now. May raise followup based on discussion result.

PR #2688, opened 2024-04-24
fix: change folder prefix for adding headers

JoeWang1127:
In this PR:
- Change post processor to add header when the folder name starts with `proto-` or `grpc-`.


blakeli0 (approved):
LGTM. This is one more reason that we need to have golden integration tests in sdk-platform-java.

PR #2691, opened 2024-04-25
feat: add a CLI tool to validate generation configuration 

JoeWang1127:
In this PR:
- Add a CLI tool to validate generation config.

Downstream libraries, e.g., google-cloud-java, can write a workflow job like:
```
docker run ...
 python /src/cli/entry_point.py \
 validate-generation-config \
 --generation-config-path=path/to/generation_config.yaml
```

### DIFF
--- a/library_generation/model/generation_config.py
+++ b/library_generation/model/generation_config.py
@@ -91,12 +91,16 @@ def from_yaml(path_to_yaml: str) -> GenerationConfig:
         config = yaml.safe_load(file_stream)
 
     libraries = __required(config, "libraries", REPO_LEVEL_PARAMETER)
+    if not libraries:
+        raise ValueError(f"Library is None in {path_to_yaml}.")
 
     parsed_libraries = list()
     for library in libraries:
         gapics = __required(library, "GAPICs")
 
         parsed_gapics = list()
+        if not gapics:
+            raise ValueError(f"GAPICs is None in {library}.")
         for gapic in gapics:
             proto_path = __required(gapic, "proto_path", GAPIC_LEVEL_PARAMETER)
             new_gapic = GapicConfig(proto_path)
--- a/library_generation/model/generation_config.py
+++ b/library_generation/model/generation_config.py
@@ -46,6 +46,7 @@ class GenerationConfig:
         self.libraries = libraries
         self.grpc_version = grpc_version
         self.protoc_version = protoc_version
+        self.__validate()
 
     def get_proto_path_to_library_name(self) -> dict[str, str]:
         """
@@ -61,6 +62,17 @@ class GenerationConfig:
 
     def is_monorepo(self) -> bool:
         return len(self.libraries) > 1
+
+    def __validate(self) -> None:
+        seen_library_names = dict()
+        for library in self.libraries:
+            library_name = library.get_library_name()
+            if library_name in seen_library_names:
+                raise ValueError(
+                    f"{library.name_pretty} has the same library name with "
+                    f"{seen_library_names.get(library_name)}"
+                )
+            seen_library_names[library_name] = library.name_pretty
 
 
 def from_yaml(path_to_yaml: str) -> GenerationConfig:
@@ -130,7 +142,9 @@ def from_yaml(path_to_yaml: str) -> GenerationConfig:
 
 def __required(config: Dict, key: str):
     if key not in config:
-        raise ValueError(f"required key {key} not found in yaml")
+        raise ValueError(
+            f"required key {key} not found in {config} " f"when parsing yaml"
+        )
     return config[key]
 
 
--- a/library_generation/model/generation_config.py
+++ b/library_generation/model/generation_config.py
@@ -19,8 +19,9 @@ from typing import List, Optional, Dict
 from library_generation.model.library_config import LibraryConfig
 from library_generation.model.gapic_config import GapicConfig
 
-LIBRARY_LEVEL_PARAMETER = "Library-level parameter"
-REPO_LEVEL_PARAMETER = "Repo-level parameter"
+REPO_LEVEL_PARAMETER = "Repo level parameter"
+LIBRARY_LEVEL_PARAMETER = "Library level parameter"
+GAPIC_LEVEL_PARAMETER = "GAPIC level parameter"
 
 
 class GenerationConfig:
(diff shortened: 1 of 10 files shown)


library_generation/utils/utilities.py
```diff
@@ -130,10 +130,6 @@ def prepare_repo(
         # use absolute path because docker requires absolute path
         # in volume name.
         absolute_library_path = str(Path(library_path).resolve())
-        if absolute_library_path in libraries:
```
JoeWang1127, line 133: We don't need this check because the Generation object is validated when reaches this point.

library_generation/model/generation_config.py
```diff
@@ -62,6 +63,17 @@ def get_proto_path_to_library_name(self) -> dict[str, str]:
     def is_monorepo(self) -> bool:
         return len(self.libraries) > 1
 
+    def __validate(self) -> None:
+        seen_library_names = dict()
+        for library in self.libraries:
+            library_name = library.get_library_name()
+            if library_name in seen_library_names:
+                raise ValueError(
+                    f"{library.name_pretty} has the same library name with "
```
blakeli0, line 72: Can we make the error message more verbose and actionable? Something like
```
f"Both {library.name_pretty} and {seen_library_names.get(library_name)} have the same library_name: 
{library_name}, please update one of the library to have a different library_name."
```

blakeli0: Another edge case is that two libraries may have the same name_pretty, but I think the error message is already good enough, so we may not have to consider it.

library_generation/model/generation_config.py
```diff
@@ -130,7 +142,9 @@ def from_yaml(path_to_yaml: str) -> GenerationConfig:
 
 def __required(config: Dict, key: str):
     if key not in config:
-        raise ValueError(f"required key {key} not found in yaml")
+        raise ValueError(
+            f"required key {key} not found in {config} " f"when parsing yaml"
```
blakeli0, line 146: Is {config} going to print out the whole config? If yes, I don't think we want to do that. Something like 
```
f"Required repo level config {key} not found"
```
should be good enough.

blakeli0: Separately, there are a few library level required field, and I don't see any validation for them yet. They can be addressed in a separate PR.

blakeli0:
How do we plan to use it? I think we can add it to the CI as a required check for every PR, as this should be a very lightweight check.

library_generation/model/generation_config.py
JoeWang1127: Done.

library_generation/model/generation_config.py
JoeWang1127: I remove `config` in the err message.

> Separately, there are a few library level required field, and I don't see any validation for them yet. They can be addressed in a separate PR.

Library level required fields are tests in `utilities_unit_tests.py`. I moved it to `generation_config_unit_test.py`.

JoeWang1127:
> How do we plan to use it? I think we can add it to the CI as a required check for every PR, as this should be a very lightweight check.

Yes, the check should be finished quickly (~). 

I added how to use the check in the description.

library_generation/test/model/generation_config_unit_test.py
```diff
@@ -133,3 +134,82 @@ def test_is_monorepo_with_two_libraries_returns_true(self):
...
+
+    # parameterized tests need to run from the class, see
+    # https://github.com/wolever/parameterized/issues/37 for more info.
+    # This test confirms that a ValueError with an error message about a
+    # missing key (specified in the first parameter of each `parameterized`
+    # tuple) when parsing a test configuration yaml (second parameter) will
+    # be raised.
+    @parameterized.expand(
+        [
+            ("libraries", f"{test_config_dir}/config_without_libraries.yaml"),
```
blakeli0, line 176: I'm still not a fan of parameterized tests, but I know this test was pre-existing, and in this case, the name of test yamls are kind of self-explanatory, so it's fine.

library_generation/model/generation_config.py
blakeli0: I see that library level config is validated in the same way as repo level config now. However, it makes the error message not clear for library level config. e.g. If `api-shortname` is missing for a specific library, we don't know which library is missing `api-shortname` from `required key {key} not found`.

blakeli0 (commented):
Can the command just be 
```
docker run ...
 python /src/cli/entry_point.py validate-generation-config
```
 without specifying the yaml location? I think it should be fine since we have a default location?

JoeWang1127:
> I see that library level config is validated in the same way as repo level config now. However, it makes the error message not clear for library level config. e.g. If api-shortname is missing for a specific library, we don't know which library is missing api-shortname from required key {key} not found.

I added a parameter, `level`, so that library config is shown in the raised exception. Missing a repo level parameter will not print the whole config (existing behavior).

blakeli0:
> > I see that library level config is validated in the same way as repo level config now. However, it makes the error message not clear for library level config. e.g. If api-shortname is missing for a specific library, we don't know which library is missing api-shortname from required key {key} not found.
> 
> I added a parameter, `level`, so that library config is shown in the raised exception. Missing a repo level parameter will not print the whole config (existing behavior).

The main concern is not that we don't know if the config is at library or repo level, the concern is that we don't know **which** library is missing a config if it's a library level config.

JoeWang1127:
> The main concern is not that we don't know if the config is at library or repo level, the concern is that we don't know which library is missing a config if it's a library level config.

When parsing from yaml, we don't know which parameter maybe missing so I think the best we can do is printing the library dict and hopefully some of the values can let us find which library is missing parameter.

library_generation/model/generation_config.py
```diff
@@ -114,23 +132,36 @@ def from_yaml(path_to_yaml: str) -> GenerationConfig:
...
 
     return parsed_config
 
 
-def __required(config: Dict, key: str):
+def __required(config: Dict, key: str, level: str = LIBRARY_LEVEL_PARAMETER):
     if key not in config:
-        raise ValueError(f"required key {key} not found in yaml")
+        message = (
+            f"{level}, {key}, is not found in {config} in yaml."
```
blakeli0, line 160: Can we add a unit test that verifies config is printed out for library level?

blakeli0 (approved):
LGTM other than one minor issue for unit test. Separately I think we should error out on unknown parameters, which I don't think we are doing currently.

library_generation/model/generation_config.py
JoeWang1127: Done.

JoeWang1127:
> Separately I think we should error out on unknown parameters, which I don't think we are doing currently.

I'll address this issue in a follow-up PR.

library_generation/test/model/generation_config_unit_test.py
JoeWang1127: I replaced this parameterized test to several individual tests.

PR #2698, opened 2024-04-29
chore: run library generation jobs conditionally

JoeWang1127:
In this PR:
- Add a job to get changed directories in pull request.
- Run library generation tests only if the changed directories contains `library_generation`.

Verified in
- [Run tests](https://github.com/googleapis/sdk-platform-java/actions/runs/8887270660/job/24402267953)
- [Don't run tests](https://github.com/googleapis/sdk-platform-java/actions/runs/8887279933/job/24402294618)

The goal is to make library generation tests as a required check. Moreover, we don't want these tests run against every pull requests because only pull requests change library_generation directory should be tested.

Currently, the workflow can only run if the files changed are within `paths`. Therefore, we can't set this workflow as required because it will get stuck and prevent merging for pull requests that don't change files within paths.

Using this approach, this workflow runs on every pull request and only run actual tests when the condition is met (determined by the first job).

Official docs: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks

We can make the workflow required after merging this PR and the tests in the workflow will only run if the changed directories contains `library_generation`.

### DIFF
--- a/.github/workflows/verify_library_generation.yaml
+++ b/.github/workflows/verify_library_generation.yaml
@@ -31,7 +31,7 @@ jobs:
       env:
         base_ref: ${{ github.event.pull_request.base.ref }}
         head_ref: ${{ github.event.pull_request.head.ref }}
-  integration_tests:
+  library-generation-integration-tests:
     runs-on: ubuntu-22.04
     needs: should-run-library-generation-tests
     if: needs.should-run-library-generation-tests.outputs.should_run == 'true'
@@ -65,7 +65,7 @@ jobs:
       run: |
         set -x
         python -m unittest library_generation/test/integration_tests.py
-  unit_tests:
+  library-generation-unit-tests:
     runs-on: ${{ matrix.os }}
     needs: should-run-library-generation-tests
     if: needs.should-run-library-generation-tests.outputs.should_run == 'true'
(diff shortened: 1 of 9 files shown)


.github/workflows/verify_library_generation.yaml
```diff
@@ -3,14 +3,38 @@ on:
     branches:
     - main
   pull_request:
-    paths:
-    - library_generation/**
 
   workflow_dispatch:
 name: verify_library_generation
 jobs:
+  should-run-library-generation-tests:
```
blakeli0, line 10: What is the difference between this approach and adding `paths` to `pull_request` above?

JoeWang1127: The workflow can only run if the files changed are within `paths`. Therefore, you can't set this workflow as required because it will get stuck and prevent merging for pull requests that don't change files within `paths`.

diegomarquezp (approved):

.github/workflows/verify_library_generation.yaml
blakeli0: I see, thanks for the explanation, it's very clear to me now! Do you mind adding this info somewhere in the PR description?

blakeli0 (approved):
LGTM. Please update the PR description before merging.

PR #2720, opened 2024-05-02
feat: generate proto-only repository

JoeWang1127:
In this PR:
- Enable hermetic build script to generate `java-common-protos` and `java-iam` in this repo.
- Change `libraries_bom_version` to optional.
- Refactor unit tests.
- Remove `java-common-protos/codecov.yaml` and `java-iam/codecov.yaml` as they are no longer needed.


blakeli0:
> java-iam/README.md is excluded from templating, should we not generate it?

I don't think it makes sense for both java-iam and java-common-protos to have READMEs, I think we can remove them and exclude them from generation. WDYT @suztomo ?

suztomo:
README.md files help people (including us) to understand the modules, regardless of it's generated or manually maintained.

generation_config.yaml
```diff
@@ -0,0 +1,60 @@
+gapic_generator_version: 2.40.1-SNAPSHOT
```
blakeli0, line 1: How do we plan to update this? We probably want a different strategy, like updating it with release-please, also not triggering regeneration if `gapic_generator_version` is the only parameter updated.

generation_config.yaml
```diff
@@ -0,0 +1,60 @@
...
+protoc_version: '25.3'
+googleapis_commitish: 5a0ae652df1ef39b7849bf2bd6ba417376ce3204
+libraries_bom_version: 26.38.0
```
blakeli0, line 4: I don't think this is needed for this repo.

generation_config.yaml
```diff
@@ -0,0 +1,60 @@
...
+owlbot_cli_image: sha256:623647ee79ac605858d09e60c1382a716c125fb776f69301b72de1cd35d49409
```
blakeli0, line 5: I think we can remove these two parameters as they are part of the generation image now? @diegomarquezp

library_generation/generate_repo.py
```diff
@@ -39,6 +41,16 @@ def generate_from_yaml(
     If specified with an empty list, then no library will be generated.
     If not specified, all libraries in the configuration yaml will be generated.
     """
+    # Determine whether the generated repo is a proto-only repository.
+    # Using libraries in generation config without filtering out target library
+    # names because the remaining library may not qualify as proto-only
+    # library, e.g., iam.
+    proto_only_repo = False
```
blakeli0, line 48: Can we move this logic to `is_monorepo()`? Or we need it for something else other than the monorepo post-processing?

library_generation/owlbot/templates/java_library/README.md
```diff
@@ -272,7 +272,7 @@ Java is a registered trademark of Oracle and/or its affiliates.
 [developer-console]: https://console.developers.google.com/
 [create-project]: https://cloud.google.com/resource-manager/docs/creating-managing-projects
 [cloud-cli]: https://cloud.google.com/cli
-[troubleshooting]: https://github.com/googleapis/google-cloud-java/blob/main/TROUBLESHOOTING.md
```
blakeli0, line 275: I think this one is expected to be in google-cloud-java, as the `TROUBLESHOOTING.md` is shared by all libraries. See [java-bigtable](TROUBLESHOOTING.md)

generation_config.yaml
JoeWang1127: When I added this parameter, I didn't think it may not be needed in `README.md`.

I'll change it to optional.

library_generation/owlbot/templates/java_library/README.md
JoeWang1127: Sure, I'll restore the change.

library_generation/generate_repo.py
JoeWang1127: It is used to determine the monorepo name, google-cloud-java if `proto_only_repo` is false; sdk-platform-java otherwise. If we have a third monorepo, we need to revise this logic.

Also, do not populate `api_id` in `.repo-metadata.json` if `proto_only_repo` is `true`. From the `README.md` of common-protos and iam, they don't have an api_id.

generation_config.yaml
JoeWang1127: Done.

library_generation/owlbot/templates/java_library/README.md
JoeWang1127: Done.

generation_config.yaml
JoeWang1127: > We probably want a different strategy, like updating it with release-please

Yes, I think release-please should update this value.

> also not triggering regeneration if gapic_generator_version is the only parameter updated

We only have two libraries in this repo, is this an overkill?

JoeWang1127:
> README.md files help people (including us) to understand the modules, regardless of it's generated or manually maintained.

I'll exclude `README.md` from generation and keep them as-is.

generation_config.yaml
JoeWang1127: On a second thought, `libraries_bom_version` is used in google-cloud-java and handwritten libraries. Making it an optional parameter may cause the README.md rendered incorrectly.

Should we just use an empty string to indicate this value is actually not used in this repo?

JoeWang1127:
Owlbot postprocessor failed:
````
Status: Downloaded newer image for gcr.io/cloud-devrel-public-resources/owlbot-java@sha256:a57d2ea6d1a77aa96c17ad0850b779ec6295f88b6c1da3d214b2095d140a2066
gcr.io/cloud-devrel-public-resources/owlbot-java:latest@sha256:a57d2ea6d1a77aa96c17ad0850b779ec6295f88b6c1da3d214b2095d140a2066
ls: cannot access '*/.OwlBot.yaml': No such file or directory
Processing a split repo
Generating templates and retrieving files from owl-bot-staging directory...
...done
Generating missing pom.xml...
/owlbot
working directory: /workspace/sdk-platform-java
Traceback (most recent call last):
  File "/owlbot/src/fix-poms.py", line 504, in <module>
    main()
  File "/owlbot/src/fix-poms.py", line 279, in main
    with open(".repo-metadata.json", "r") as fp:
FileNotFoundError: [Errno 2] No such file or directory: '.repo-metadata.json'
```

blakeli0:
> README.md files help people (including us) to understand the modules, regardless of it's generated or manually maintained.

I don't disagree with the usefulness of READMEs, but if you look at the content of the [README](https://github.com/googleapis/sdk-platform-java/blob/main/java-common-protos/README.md) of java-common-protos, and [README](https://github.com/googleapis/sdk-platform-java/blob/main/java-iam/README.md) of java-iam, they are awfully wrong. For java-common-protos, I think it was a mistake to generate README in the first place as it is not a service. For java-iam, I can see some value if it is properly maintained. But currently it is not maintained and provide incorrect information like still support Java 7.

library_generation/generate_repo.py
```diff
@@ -39,6 +41,16 @@ def generate_from_yaml(
...
+    for library in config.libraries:
+        if library.get_library_name() in PROTO_ONLY_LIBRARIES:
```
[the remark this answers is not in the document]
diegomarquezp: In other words, what's the difference between checking for all libraries being `proto_only` vs having a single `common-protos` in the library list?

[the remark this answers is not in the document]
JoeWang1127: The `proto-only` libraries means they are dependencies of client libraries (which may or may not have a service defined in proto). Maybe `common_proto_libraries` makes more sense?

When you said proto-only libraries in google-cloud-java, do you mean `java-alloydb-connectors`? I don't think `java-alloydb-connectors` is a proto-only library using the above definition because it is not a dependency of other client libaries.

generation_config.yaml
JoeWang1127: Let's change it to optional.

java-common-protos/.OwlBot-hermetic.yaml
```diff
@@ -22,50 +22,50 @@ deep-preserve-regex:
 - "/java-common-protos/google-.*/src/test/java/com/google/cloud/.*/v.*/it/IT.*Test.java"
 
 deep-copy-regex:
-- source: "/google/api/google-api-java/proto-google-api-java/src"
+- source: "/google/api/.*-java/proto-google-.*/src"
```
JoeWang1127, line 25: @diegomarquezp could you explain why do we need to build owlbot cli source folder?

java-common-protos/grpc-google-common-protos/src/main/java/com/google/cloud/location/LocationsGrpc.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Google LLC
+ * Copyright 2024 Google LLC
```
JoeWang1127, line 2: @diegomarquezp do you know why this copyright is not updated by post processor?

library_generation/generate_repo.py
```diff
@@ -48,18 +60,19 @@ def generate_from_yaml(
...
             output_folder=repo_config.output_folder,
             versions_file=repo_config.versions_file,
+            proto_only_repo=proto_only_repo,
         )
 
-    # we skip monorepo_postprocessing if not in a monorepo
-    if not config.is_monorepo():
+    # we skip monorepo_postprocessing if it is not a monorepo
+    # or has proto-only libraries.
+    if not config.is_monorepo() or proto_only_repo:
```
JoeWang1127, line 75: this parameter should be `is_gapic_monorepo`.

JoeWang1127:
Split this PR to 2:
- script change in library generation
- generation config, copyright year and `.OwlBot.yaml`

library_generation/generate_repo.py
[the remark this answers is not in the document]
JoeWang1127: We may not need this parameter.

library_generation/generate_repo.py
JoeWang1127: Done.

generation_config.yaml
JoeWang1127: Done.

library_generation/generate_repo.py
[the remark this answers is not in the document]
JoeWang1127: I introduced `contains_common_protos` in generation config object.

library_generation/model/generation_config.py
```diff
@@ -61,6 +63,15 @@ def get_proto_path_to_library_name(self) -> dict[str, str]:
     def is_monorepo(self) -> bool:
         return len(self.libraries) > 1
 
+    def contains_common_protos(self) -> bool:
```
JoeWang1127, line 66: I added this method to determine whether there are common protos in the configuration.

We can't just have one `is_gapic_monorepo` method since we have three cases:
- sdk-platform-java, `contains_common_protos` returns `True`
- google-cloud-java, `contains_common_protos` returns `False` and `is_monorepo` returns `True`
- handwritten libraries, `contains_common_protos` returns `False` and `is_monorepo` returns `False`

library_generation/generate_repo.py
```diff
@@ -48,18 +50,17 @@ def generate_from_yaml(
...
             library_path=library_path,
             library=library,
             output_folder=repo_config.output_folder,
             versions_file=repo_config.versions_file,
+            gapic_repo=not config.contains_common_protos(),
         )
 
-    # we skip monorepo_postprocessing if not in a monorepo
-    if not config.is_monorepo():
+    if not config.is_monorepo() or config.contains_common_protos():
```
JoeWang1127, line 63: Only perform monorepo post processing if this is a gapic monorepo.

java-common-protos/.OwlBot-hermetic.yaml
JoeWang1127: The folder name is coming BUILD target and we need to parse BUILD.bazel to know the folder name.

java-common-protos/grpc-google-common-protos/src/main/java/com/google/cloud/location/LocationsGrpc.java
JoeWang1127: Maybe we need to update synthtool commit.

library_generation/generate_repo.py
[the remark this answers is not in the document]
diegomarquezp: Thank you. I think this change would make it necessary to change the PR description as well. Maybe to something that explains that this PR enables support for generating sdk-platform-java.

diegomarquezp (approved):
LGTM, just some minor comments.

library_generation/generate_repo.py
```diff
@@ -18,6 +18,8 @@
 from library_generation.model.library_config import LibraryConfig
 from library_generation.utils.monorepo_postprocessor import monorepo_postprocessing
 
+PROTO_ONLY_LIBRARIES = ["common-protos"]
```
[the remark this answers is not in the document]
JoeWang1127: This variable is actually not used, I removed it.

library_generation/generate_repo.py
```diff
@@ -48,18 +48,17 @@ def generate_from_yaml(
...
         print(f"generating library {library.get_library_name()}")
-
         generate_composed_library(
             config_path=config_path,
             config=config,
             library_path=library_path,
             library=library,
             output_folder=repo_config.output_folder,
             versions_file=repo_config.versions_file,
+            gapic_repo=not config.contains_common_protos(),
```
blakeli0, line 58: Why do we call this `gapic_repo`? I see that `config` is already being passed around, can we just use `config.contains_common_protos()`? I feel it is easier to understand.

library_generation/test/model/gapic_inputs_unit_tests.py
blakeli0, line 1: Thanks for backfilling this test!

library_generation/model/generation_config.py
```diff
@@ -20,6 +20,7 @@
 REPO_LEVEL_PARAMETER = "Repo level parameter"
 LIBRARY_LEVEL_PARAMETER = "Library level parameter"
 GAPIC_LEVEL_PARAMETER = "GAPIC level parameter"
+COMMON_PROTOS = "common-protos"
```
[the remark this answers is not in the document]
JoeWang1127: Done.

library_generation/generate_repo.py
JoeWang1127: I removed this parameter as we already pass the config.

library_generation/model/generation_config.py
JoeWang1127: I removed the setter method.

library_generation/utils/utilities.py
```diff
@@ -250,6 +252,8 @@ def generate_prerequisite_files(
         "library_type": library.library_type,
         "requires_billing": library.requires_billing,
     }
+    if config.contains_common_protos():
```
blakeli0, line 255: nit: Can we add a comment saying that this is for `java-common-protos` and `java-iam` modules in `sdk-platform-java`? And/or change the logic to `if repo == `"googleapis/sdk-platform-java"?

library_generation/utils/utilities.py
```diff
@@ -211,11 +212,12 @@ def generate_prerequisite_files(
         else f"{library.group_id}:google-{cloud_prefix}{library_name}"
     )
     distribution_name_short = re.split(r"[:/]", distribution_name)[-1]
-    repo = (
-        "googleapis/google-cloud-java"
-        if config.is_monorepo()
-        else f"googleapis/{language}-{library_name}"
-    )
+    if config.contains_common_protos():
```
blakeli0, line 215: Future consideration: _Maybe_ we want to introduce a repo level config called `repo` or `repo_name` to represent this.

blakeli0 (approved):
LGTM.

library_generation/utils/utilities.py
JoeWang1127: Done.

PR #2731, opened 2024-05-03
fix: only append `.api.grpc` suffix to group id if the artifact id starts with `proto-` or `grpc-`

JoeWang1127:
In this PR:
- Only append `.api.grpc` suffix to group id if the corresponding artifact id starts with `proto-` or `grpc-`.

We need to support `ad-manager`, which is the first artifact id doesn't start with `proto-`, `grpc-` and `google-`.

### DIFF
--- a/library_generation/test/owlbot/fix_poms_unit_tests.py
+++ b/library_generation/test/owlbot/fix_poms_unit_tests.py
@@ -1,0 +1,59 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import os
+import shutil
+import unittest
+from library_generation.owlbot.src.fix_poms import main
+from library_generation.test.compare_poms import compare_xml
+from library_generation.test.test_utils import FileComparator
+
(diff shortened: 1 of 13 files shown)


library_generation/test/owlbot/fix_poms_unit_tests.py
```diff
@@ -0,0 +1,59 @@
...
+            self.__remove_file_in_subdir(ad_manager_resource, sub_dir)
+
+    @classmethod
+    def __copy__golden(cls, base_dir: str, subdir: str):
+        golden = os.path.join(base_dir, subdir, "pom-golden.xml")
+        pom = os.path.join(base_dir, subdir, "pom.xml")
+        shutil.copyfile(golden, pom)
+
+    @classmethod
+    def __compare_pom_in_subdir(cls, base_dir: str, subdir: str):
```
blakeli0, line 47: Maybe this can be moved to `compare_poms.py`.

library_generation/test/owlbot/fix_poms_unit_tests.py
```diff
@@ -0,0 +1,59 @@
...
+from library_generation.test.compare_poms import compare_xml
+from library_generation.test.test_utils import FileComparator
+
+script_dir = os.path.dirname(os.path.realpath(__file__))
+resources_dir = os.path.join(script_dir, "..", "resources", "test-owlbot")
+file_comparator = FileComparator()
+
+
+class FixPomsTest(unittest.TestCase):
+    def test_fix_poms_update_poms_for_ad_manager_correctly(self):
```
blakeli0, line 27: The test case is not specific to `ad_manager`, more like `non-cloud`, or `group_id_does_not_start_with_google` to be more specific.

JoeWang1127: Done.

library_generation/test/owlbot/fix_poms_unit_tests.py
JoeWang1127: Done.

blakeli0 (approved):

PR #2737, opened 2024-05-06
test: add showcase test for api-version

zhumin8:
This pr updates gapic-showcase version to 0.35.0 and adds showcase tests to verify behavior of api version headers being emitted (changes in #2630 and #2671).

### DIFF
--- a/showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiVersionHeaders.java
+++ b/showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiVersionHeaders.java
@@ -30,6 +30,7 @@ import com.google.showcase.v1beta1.stub.EchoStubSettings;
 import io.grpc.*;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.concurrent.TimeUnit;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
@@ -50,6 +51,7 @@ public class ITApiVersionHeaders {
   private static final String EXPECTED_EXCEPTION_MESSAGE =
       "Header provider can't override the header: "
           + ApiClientHeaderProvider.API_VERSION_HEADER_KEY;
+  private static final int DEFAULT_AWAIT_TERMINATION_SEC = 10;
 
   // Implement a client interceptor to retrieve the trailing metadata from response.
   private static class GrpcCapturingClientInterceptor implements ClientInterceptor {
@@ -182,11 +184,16 @@ public class ITApiVersionHeaders {
   }
 
   @After
-  public void destroyClient() {
+  public void destroyClient() throws InterruptedException {
     grpcClient.close();
     httpJsonClient.close();
     grpcComplianceClient.close();
     httpJsonComplianceClient.close();
+
+    grpcClient.awaitTermination(DEFAULT_AWAIT_TERMINATION_SEC, TimeUnit.SECONDS);
+    httpJsonClient.awaitTermination(DEFAULT_AWAIT_TERMINATION_SEC, TimeUnit.SECONDS);
+    grpcComplianceClient.awaitTermination(DEFAULT_AWAIT_TERMINATION_SEC, TimeUnit.SECONDS);
+    httpJsonComplianceClient.awaitTermination(DEFAULT_AWAIT_TERMINATION_SEC, TimeUnit.SECONDS);
   }
 
   @Test
--- a/showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiVersionHeaders.java
+++ b/showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiVersionHeaders.java
@@ -27,13 +27,16 @@ import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 
+// TODO: add testing on error responses once feat is implemented in showcase.
+//  https://github.com/googleapis/gapic-showcase/pull/1456
 public class ITApiVersionHeaders {
-  private static final String SPLIT_TOKEN = "&";
   private static final String API_VERSION_HEADER_STRING = "x-goog-api-version";
   private static final String HTTP_RESPONSE_HEADER_STRING =
       "x-showcase-request-" + API_VERSION_HEADER_STRING;
   private static final Metadata.Key<String> API_VERSION_HEADER_KEY =
       Metadata.Key.of(API_VERSION_HEADER_STRING, Metadata.ASCII_STRING_MARSHALLER);
+
+  private static final String EXPECTED_ECHO_API_VERSION = "v1_20240408";
 
   // Implement a client interceptor to retrieve the trailing metadata from response.
   private static class GrpcCapturingClientInterceptor implements ClientInterceptor {
@@ -177,7 +180,7 @@ public class ITApiVersionHeaders {
   public void testGrpc_matchesApiVersion() {
     grpcClient.echo(EchoRequest.newBuilder().build());
     String headerValue = grpcInterceptor.metadata.get(API_VERSION_HEADER_KEY);
-    assertThat(headerValue).isEqualTo("v1_20240408");
+    assertThat(headerValue).isEqualTo(EXPECTED_ECHO_API_VERSION);
   }
 
   @Test
@@ -186,7 +189,7 @@ public class ITApiVersionHeaders {
     ArrayList headerValues =
         (ArrayList) httpJsonInterceptor.metadata.getHeaders().get(HTTP_RESPONSE_HEADER_STRING);
     String headerValue = (String) headerValues.get(0);
-    assertThat(headerValue).isEqualTo("v1_20240408");
+    assertThat(headerValue).isEqualTo(EXPECTED_ECHO_API_VERSION);
   }
 
   @Test
--- a/showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiVersionHeaders.java
+++ b/showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiVersionHeaders.java
@@ -18,6 +18,7 @@ package com.google.showcase.v1beta1.it;
 import static com.google.common.truth.Truth.assertThat;
 
 import com.google.api.gax.httpjson.*;
+import com.google.api.gax.rpc.ApiClientHeaderProvider;
 import com.google.common.collect.ImmutableList;
 import com.google.showcase.v1beta1.*;
 import com.google.showcase.v1beta1.it.util.TestClientInitializer;
@@ -30,11 +31,11 @@ import org.junit.Test;
 // TODO: add testing on error responses once feat is implemented in showcase.
 //  https://github.com/googleapis/gapic-showcase/pull/1456
 public class ITApiVersionHeaders {
-  private static final String API_VERSION_HEADER_STRING = "x-goog-api-version";
   private static final String HTTP_RESPONSE_HEADER_STRING =
-      "x-showcase-request-" + API_VERSION_HEADER_STRING;
+      "x-showcase-request-" + ApiClientHeaderProvider.API_VERSION_HEADER_KEY;
   private static final Metadata.Key<String> API_VERSION_HEADER_KEY =
-      Metadata.Key.of(API_VERSION_HEADER_STRING, Metadata.ASCII_STRING_MARSHALLER);
+      Metadata.Key.of(
+          ApiClientHeaderProvider.API_VERSION_HEADER_KEY, Metadata.ASCII_STRING_MARSHALLER);
 
   private static final String EXPECTED_ECHO_API_VERSION = "v1_20240408";
 
--- a/showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiVersionHeaders.java
+++ b/showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiVersionHeaders.java
@@ -16,6 +16,7 @@
 package com.google.showcase.v1beta1.it;
 
 import static com.google.common.truth.Truth.assertThat;
+import static org.junit.Assert.assertThrows;
 
 import com.google.api.gax.httpjson.*;
 import com.google.api.gax.rpc.ApiClientHeaderProvider;
@@ -46,6 +47,9 @@ public class ITApiVersionHeaders {
 
   private static final String EXPECTED_ECHO_API_VERSION = "v1_20240408";
   private static final String CUSTOM_API_VERSION = "user-supplied-version";
+  private static final String EXPECTED_EXCEPTION_MESSAGE =
+      "Header provider can't override the header: "
+          + ApiClientHeaderProvider.API_VERSION_HEADER_KEY;
 
   // Implement a client interceptor to retrieve the trailing metadata from response.
   private static class GrpcCapturingClientInterceptor implements ClientInterceptor {
@@ -218,7 +222,7 @@ public class ITApiVersionHeaders {
         .isNotIn(httpJsonComplianceInterceptor.metadata.getHeaders().keySet());
   }
 
-  @Test(expected = IllegalArgumentException.class)
+  @Test
   public void testGrpcEcho_userApiVersionThrowsException() throws IOException {
     StubSettings stubSettings =
         grpcClient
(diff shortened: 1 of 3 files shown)


showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiVersionHeaders.java
```diff
@@ -0,0 +1,208 @@
...
+                    metadata = trailers;
+                  }
+                  super.onClose(status, trailers);
+                }
+              };
+
+          super.start(wrappedListener, headers);
+        }
+      };
+    }
```
lqiu96, line 63: Not something that has to be done in this PR, but since using client interceptors is something that multiple tests have implemented, I think we can try and migrate this to TestClientInitializer if possible (I don't know if there are an limitations with that, but just something to consider).

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiVersionHeaders.java
```diff
@@ -0,0 +1,208 @@
...
+import com.google.common.collect.ImmutableList;
+import com.google.showcase.v1beta1.*;
+import com.google.showcase.v1beta1.it.util.TestClientInitializer;
+import io.grpc.*;
+import java.util.ArrayList;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+public class ITApiVersionHeaders {
```
lqiu96, line 30: Can you add a quick comment, like a TODO, above this that we can use echo to echo back the once this gapic-showcase PR makes it in? https://github.com/googleapis/gapic-showcase/pull/1456

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiVersionHeaders.java
```diff
@@ -0,0 +1,208 @@
...
+  private static final String SPLIT_TOKEN = "&";
+  private static final String API_VERSION_HEADER_STRING = "x-goog-api-version";
```
lqiu96, line 32: nit: Personal opinion is that we can consider making the constant public: https://github.com/googleapis/sdk-platform-java/blob/43e62b998958abd4ca96b83800380c9696ff693f/gax-java/gax/src/main/java/com/google/api/gax/rpc/ApiClientHeaderProvider.java#L45 so that we don't redefine it multiple spots. 

We tried package-private, but showcase needs it as well.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiVersionHeaders.java
```diff
@@ -0,0 +1,208 @@
...
+    httpJsonClient.close();
+    grpcComplianceClient.close();
+    httpJsonComplianceClient.close();
+  }
+
+  @Test
+  public void testGrpc_matchesApiVersion() {
+    grpcClient.echo(EchoRequest.newBuilder().build());
+    String headerValue = grpcInterceptor.metadata.get(API_VERSION_HEADER_KEY);
+    assertThat(headerValue).isEqualTo("v1_20240408");
```
lqiu96, line 180: nit: Perhaps a constant value for here and the one below? i.e. something like `DEFAULT_SHOWCASE_API_VERSION`?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiVersionHeaders.java
zhumin8: mmh, sounds a good idea. These interceptors are slightly different, but I guess we can create an interceptor that does it all.
To keep current pr from being cluttered, I would file a separate issue to look into and refactor these.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiVersionHeaders.java
zhumin8: Will do. I'll probably also file an issue to track, so we don't forget about it.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiVersionHeaders.java
```diff
@@ -0,0 +1,208 @@
...
+    grpcComplianceInterceptor = new GrpcCapturingClientInterceptor();
+    grpcComplianceClient =
+        TestClientInitializer.createGrpcComplianceClient(
+            ImmutableList.of(grpcComplianceInterceptor));
+
+    // Create HttpJson ComplianceClient and Interceptor
+    httpJsonComplianceInterceptor = new HttpJsonCapturingClientInterceptor();
+    httpJsonComplianceClient =
+        TestClientInitializer.createHttpJsonComplianceClient(
+            ImmutableList.of(httpJsonComplianceInterceptor));
```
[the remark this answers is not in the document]
zhumin8: In current setup, we _could_ override with `.setHeaderProvider()` when creating the test client. I am reluctant to do so though. 
- I feel like showcase integration testing should test expected behavior of un-modified service behaviors E2E.
- As we discussed in doc comment last time, I am inclined to not allow user override this "x-goog-api-version" header with client libraries. (Will raise a separate pr) In this case, "unsets the showcase apiVersion value" should not work.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiVersionHeaders.java
zhumin8: I tend to agree. But I was not 100% on if this a breaking change to gax. Even though we are exposing a new public field, this is a final field and no breaking to existing usage, so I suppose it's okay?

lqiu96: It was package-private so we can change it since no one else should have access to it.

It is now public field, but I think that's fine since we're doing it so that this constant isn't re-defined in multiple spots.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiVersionHeaders.java
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  @After
+  public void destroyClient() {
+    grpcClient.close();
+    httpJsonClient.close();
+    grpcComplianceClient.close();
+    httpJsonComplianceClient.close();
```
lqiu96, line 179: Can you also add `awaitTermination()` here after the `.close()` calls?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiVersionHeaders.java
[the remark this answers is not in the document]
lqiu96: > I feel like showcase integration testing should test expected behavior of un-modified service behaviors E2E.

IMO, it shouldn't just be limited to un-modified service behaviors. We should try to account for customizations where we can (i.e. If new features are added or showcase has new functionality). We're not in state where every modification/ feature is covered, but I think we should strive to get as close as possible. There are going to be certain spots we miss and new edge cases that are discovered, but we should try to backfill/ add them in as we discover them.

> I am inclined to not allow user override this "x-goog-api-version" header with client libraries.

Makes sense. Let's get some insight from the core team to see if this is prioritized (and if not, we can always prioritize this for Java clients ourselves). The public setter method is pubic as of now and removing will be challenging/ impossible. We can explore some options to see how we want to proceed.

Being that removing may be challenging/impossible, I think it may make sense to include both Echo and Compliance clients in then since we test for the behavior of add custom headers to both clients that that have an auto generated ApiVersion vs clients that don't have this generated.

i.e. I believe Echo with custom headers set should have a conflict error message and Compliance with custom headers should not have a conflict error message.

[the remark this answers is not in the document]
zhumin8: I agree with most of these, but wanted to clarify a few:
- This [ITApiVersionHeaders.java](https://github.com/googleapis/sdk-platform-java/pull/2737/files/219c8fdb150a33a9f2e44c442f4e91a586000ead#diff-0c74baec42410162926db3ec4be8e8f2c04fd9d6cf214ef5bbf407e7e350f0b5) tests for behavior when client has an auto generated ApiVersion (Echo) and gets correctly send via header on requests, and when client don't have this generated (Compliance) then no such header sent. User header are not covered.
- To test behavior of adding user header to both clients also makes sense, I will add based on current behavior. This may change in the future. In fact, I have a todo item in https://github.com/googleapis/sdk-platform-java/pull/2745 for it if we decide to go with this approach.
- "The public setter method" >> I assume you are referring to [this setter](https://github.com/googleapis/sdk-platform-java/blob/c9d8ded1a8af3d7cbb9a647cb802a71412c8d7d7/gax-java/gax/src/main/java/com/google/api/gax/rpc/ApiClientHeaderProvider.java#L218)? If so, this is only one HeaderProvider the client library code uses, users could use other HeaderProvider. But to guard users to not override  "x-goog-api-version" header with `setHeaderProvider()`, I am thinking adding to conflict resolution as in [pr2745](https://github.com/googleapis/sdk-platform-java/pull/2745) 
- "Let's get some insight from the core team if this is prioritized " >> Other than comment on proposal doc, I've started a thread on the aip [here](https://github.com/aip-dev/google.aip.dev/pull/1331/files#r1591764838) 

To summarize, I am keeping both clients in this test to cover scenario of service opted-in for api-version and not. Also adding tests for setting custom api version headers for these clients, these behavior may change via a separate pr.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiVersionHeaders.java
zhumin8: IIUC, `.close()` is same as `.shutdown()` ([code](https://github.com/googleapis/sdk-platform-java/blob/8381dc58c3bc0d5821c9c22bb7af49b3ad9e544e/gax-java/gax/src/main/java/com/google/api/gax/core/BackgroundResourceAggregation.java#L93)). We should `.shutdown()` a client or `.awaitTermination()`, not both, and the main difference is that `.awaitTermination()` has timeout ([comment here](https://github.com/googleapis/sdk-platform-java/blob/8381dc58c3bc0d5821c9c22bb7af49b3ad9e544e/gax-java/gax/src/main/java/com/google/api/gax/core/BackgroundResource.java#L70-L74)).
However, if we want to prevent a test hanging, this may not work, as these are called within `@After`  and I believe if a test hangs and does not finish, the `@After` method will not be executed. An alternative is to add `@Test(timeout = ..)` to tests. But I wonder if you have a specific concern about these tests? Or do you prefer to implement timeouts for all integration tests as precautious?

lqiu96: We typically use `close()` in conjunction with `awaitTermination()`, with `close()` being called first. You're right `close()` just calls `shutdown()` and all `shutdown()` does to prevent *new* tasks from being scheduled, but still allows currently in-flight tasks to continue. We simply use `awaitTermination()` to allow the current in-flight tasks to finish (it blocks until they are done or until the timeout has been reached).

We had previously seen some grpc-java warnings about managed channels still being allocated without trying to wait to the tasks to terminated. i.e something like:
```
io.grpc.internal.ManagedChannelImpl$ManagedChannelReference cleanQueue
SEVERE: *~*~*~ Channel io.grpc.internal.ManagedChannelImpl-2096 for target 127.0.0.1:8771 was not shutdown properly!!! ~*~*~*
    Make sure to call shutdown()/shutdownNow() and awaitTermination().
java.lang.RuntimeException: ManagedChannel allocation site
```

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiVersionHeaders.java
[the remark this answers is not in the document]
lqiu96: > This [ITApiVersionHeaders.java](https://github.com/googleapis/sdk-platform-java/pull/2737/files/219c8fdb150a33a9f2e44c442f4e91a586000ead#diff-0c74baec42410162926db3ec4be8e8f2c04fd9d6cf214ef5bbf407e7e350f0b5) tests for behavior when client has an auto generated ApiVersion (Echo) and gets correctly send via header on requests, and when client don't have this generated (Compliance) then no such header sent. User header are not covered.

Yep, I think that's good! The test cases do look good. I don't mean that we should try to cover modifying *every* user header, but rather the relevant one to the test (api-version header).

> In fact, I have a todo item in https://github.com/googleapis/sdk-platform-java/pull/2745

Let's have an internal discussion. I do prefer this behavior but I believe this does technically put us at a "behavior breaking change" which is something that we try not to do (if possible). I think this may be a case where we can argue for it given that it's a relatively new feature.

> Also adding tests for setting custom api version headers for these clients, these behavior may change via a separate pr.

I'll do another round of review on this PR. No worries about adding modifications to the header in this PR if you haven't already. I believe it's already a pretty niche case and it can be added and/or tested based on what happens to #2745

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiVersionHeaders.java
```diff
@@ -0,0 +1,311 @@
...
+  @Test
+  public void testHttpJson_noApiVersion() {
+    RepeatRequest request =
+        RepeatRequest.newBuilder().setInfo(ComplianceData.newBuilder().setFString("test")).build();
+    httpJsonComplianceClient.repeatDataSimplePath(request);
+    assertThat(API_VERSION_HEADER_KEY)
+        .isNotIn(httpJsonComplianceInterceptor.metadata.getHeaders().keySet());
+  }
+
+  @Test(expected = IllegalArgumentException.class)
```
lqiu96, line 221: Can you use `assertThrows()` instead of `@Test(expected = ...)`? Something like: https://github.com/googleapis/sdk-platform-java/blob/3611e8f8982070f6b1aa14b6272677608aa5f6fb/showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITUnaryCallable.java#L70-L71

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiVersionHeaders.java
```diff
@@ -0,0 +1,311 @@
...
+            .getStubSettings()
+            .toBuilder()
+            .setHeaderProvider(
+                FixedHeaderProvider.create(
+                    ApiClientHeaderProvider.API_VERSION_HEADER_KEY, CUSTOM_API_VERSION))
+            .build();
+
+    try (EchoClient echo =
+        EchoClient.create(EchoSettings.create((EchoStubSettings) stubSettings))) {
+      assertThat(true).isFalse();
```
lqiu96, line 235: What is this line doing?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiVersionHeaders.java
```diff
@@ -0,0 +1,311 @@
...
+    }
+  }
```
lqiu96, line 255: Same as the two comments above.

lqiu96:
LGTM. Just a few nits/ questions, but I'll approve once those are addressed.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiVersionHeaders.java
zhumin8: Oh, nvm. 
I was not aware the `assertThrows()` as you mentioned in above comment is available in Junit 4 too, and was trying to be more precise on where exception is thrown. Switching to `assertThrows()` approach.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiVersionHeaders.java
zhumin8: same above.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiVersionHeaders.java
lqiu96: Bumping this.

lqiu96 (approved):
LGTM. Small addition for `@After` if you could resolve

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiVersionHeaders.java
zhumin8: Thanks for explaining. Added.

PR #2754, opened 2024-05-08
chore: use Junit5 in java core

### LINKED ISSUE #2726: Migrate Junit 4 to Junit 5: java-core

JoeWang1127:
Fix #2726.

`BaseSerializationTest` will not migrate to Junit 5 because downstream libraries, e.g., java-logging, are extending this class and these libraries still use Junit 4. Migrating this class to Junit 5 will cause test failures in downstream libraries.


java-core/pom.xml
```diff
@@ -37,13 +37,14 @@
...
-        <artifactId>junit</artifactId>
-        <version>4.13.2</version>
-        <scope>test</scope>
+        <groupId>org.junit</groupId>
+        <artifactId>junit-bom</artifactId>
+        <version>5.10.2</version>
+        <type>pom</type>
+        <scope>import</scope>
       </dependency>
+      <!-- Test dependencies -->
```
blakeli0, line 47: Can we move this line before junit-bom?

blakeli0:
Can you please add more info regarding the skipped test classes? Maybe add comments to [BaseSerializationTest](https://github.com/googleapis/sdk-platform-java/blob/2072a8c899f147b55e5f05fa6ce4296575cbb472/java-core/google-cloud-core/src/test/java/com/google/cloud/BaseSerializationTest.java#L38) and PR description?

java-core/pom.xml
JoeWang1127: Done.

java-core/google-cloud-core/src/test/java/com/google/cloud/BaseSerializationTest.java
```diff
@@ -34,6 +34,10 @@
  * {@code restorableObjects()} method to return all restorable objects whose state must be tested
  * for proper serialization. Both methods can return {@code null} if no such object needs to be
  * tested.
+ *
+ * <p>This class has not migrated to Junit 5 because downstream libraries, e.g., java-logging, are
+ * extending this class and these libraries still use Junit 4. Migrating this class to Junit 5 will
+ * cause test failures in downstream libraries.
```
lqiu96, line 40: If a issue for this doesn't exist already, can you create an so we don't forget about this.

java-core/pom.xml
```diff
@@ -39,10 +39,11 @@
...
       <dependency>
-        <groupId>junit</groupId>
-        <artifactId>junit</artifactId>
-        <version>4.13.2</version>
-        <scope>test</scope>
+        <groupId>org.junit</groupId>
+        <artifactId>junit-bom</artifactId>
+        <version>5.10.2</version>
+        <type>pom</type>
+        <scope>import</scope>
```
lqiu96, line 46: Might be missing some context about dependency management with java-core and/or test jars, but I would have assumed this version would be been defined in pom-parent or shared-deps instead of having to declare it here.

lqiu96 (approved):
LGTM

java-core/google-cloud-core/src/test/java/com/google/cloud/BaseSerializationTest.java
JoeWang1127: Issue created: https://github.com/googleapis/sdk-platform-java/issues/2763

java-core/pom.xml
JoeWang1127: I think declare a test dependency in java-core is fine since it allows other modules to use different testing framework.

That being said, we can declare test dependencies in a centralized place, e.g., third-party dependency. However, this should be a separate task.

blakeli0: @suztomo Do you have any suggestions of what's the best way to manage junit 5 dependencies?

suztomo: Avoid declaring junit bom in the dependencyManagement section of pom.xml files that would go to the Libraries BOM, such as the pom-parent or gapic-generator-java-bom. That would control library users' junit usage.

Third-party-dependencies is good but java-core, I believe, does not import it.

How about declaring the JUnit version as a property, not dependencyManagement section, in the pom-parent?

JoeWang1127: @suztomo thanks for the suggestion.

java-core/pom.xml
```diff
@@ -39,10 +39,11 @@
...
       <dependency>
-        <groupId>junit</groupId>
-        <artifactId>junit</artifactId>
-        <version>4.13.2</version>
-        <scope>test</scope>
+        <groupId>org.junit</groupId>
+        <artifactId>junit-bom</artifactId>
+        <version>${junit.version}</version>
+        <type>pom</type>
+        <scope>import</scope>
```
blakeli0, line 46: Per @suztomo's suggestion, I don't think we want to import junit bom in java-core, as I think it would go into [shared-dependencies](https://github.com/googleapis/sdk-platform-java/blob/main/java-shared-dependencies/first-party-dependencies/pom.xml#L45-L51). So I think we should declare those three test dependencies individually and manage the version through the parent pom, is that correct @suztomo ?

suztomo: > I don't think we want to import junit bom in java-core, as I think it would go into [shared-dependencies](https://github.com/googleapis/sdk-platform-java/blob/main/java-shared-dependencies/first-party-dependencies/pom.xml#L45-L51)

java-core it not imported by the shared dependencies BOM.

Having JUnit in third-party-dependencies is a good place. Because third-party-dependencies BOM does not go to the Libraries BOM.

blakeli0: > java-core it not imported by the shared dependencies BOM.

I think it is included in the [first-party-dependencies](https://github.com/googleapis/sdk-platform-java/blob/main/java-shared-dependencies/first-party-dependencies/pom.xml#L45-L51), am I missing something?

JoeWang1127: java-core imports shared dependencies in [parant pom](https://github.com/googleapis/sdk-platform-java/blob/d3d6af29aaab425fe3e7e72ef6899eb3b24df599/java-core/pom.xml#L35), why this doesn't cause a circular dependency?

suztomo: > am I missing something?

Importing a BOM and declaring dependencies are different. For a BOM, only imported BOMs matters.

blakeli0: IIUC, first-party-dependencies imports google-cloud-core bom, which also imports junit bom, so junit bom would ends up in libraries bom?

JoeWang1127: > IIUC, first-party-dependencies imports google-cloud-core bom, which also imports junit bom, so junit bom would ends up in libraries bom?

Actually, junit bom will not end up in libraries bom since only dependencies declared in google-clour-core will end up in libraries bom. junit, however, is a transitive dependency.

JoeWang1127: > So I think we should declare those three test dependencies individually and manage the version through the parent pom

I think use junit-bom is better because `junit-platform-launcher` has a different version than `junit-jupiter-engine` and `junit-vintage-engine`.

java-core/google-cloud-core/src/test/java/com/google/cloud/ServiceOptionsTest.java
```diff
@@ -378,36 +377,38 @@ public void testBuilder_quotaProjectServiceOptionTakesPrecedence() {
...
   @Test
-  public void testBuilderNoCredentials() {
+  void testBuilderNoCredentials() {
     assertEquals(NoCredentials.getInstance(), OPTIONS_NO_CREDENTIALS.getCredentials());
-    assertTrue(NoCredentials.getInstance().equals(OPTIONS_NO_CREDENTIALS.getCredentials()));
-    assertFalse(NoCredentials.getInstance().equals(OPTIONS.getCredentials()));
-    assertFalse(NoCredentials.getInstance().equals(null));
+    assertEquals(NoCredentials.getInstance(), OPTIONS_NO_CREDENTIALS.getCredentials());
+    assertNotEquals(NoCredentials.getInstance(), OPTIONS.getCredentials());
+    assertNotEquals(null, NoCredentials.getInstance());
```
lqiu96, line 388: Perhaps assertNotNull?

java-core/google-cloud-core/src/test/java/com/google/cloud/TimestampTest.java
```diff
@@ -157,67 +154,67 @@ public void toFromSqlTimestamp() {
...
-      Assert.fail();
+      Assertions.fail();
     } catch (IllegalArgumentException ex) {
       assertNotNull(ex.getMessage());
     }
   }
 
   @Test
-  public void boundsNanosMax() {
+  void boundsNanosMax() {
```
lqiu96, line 187: Perhaps these can be converted to assertThrows()?

java-core/google-cloud-core/src/test/java/com/google/cloud/testing/VersionTest.java
```diff
@@ -39,57 +39,57 @@ public void testFromString() {
...
     } catch (IllegalArgumentException ex) {
       assertNotNull(ex.getMessage());
     }
   }
 
   @Test
-  public void testFromStringMissingMinor() {
+  void testFromStringMissingMinor() {
     try {
       Version.fromString("2016");
```
lqiu96, line 64: Same with these. Maybe we can use `assertThrows()`?

blakeli0 (approved):

java-core/google-cloud-core/src/test/java/com/google/cloud/testing/VersionTest.java
JoeWang1127: Done.

java-core/google-cloud-core/src/test/java/com/google/cloud/TimestampTest.java
JoeWang1127: Done

java-core/google-cloud-core/src/test/java/com/google/cloud/ServiceOptionsTest.java
JoeWang1127: Done.

lqiu96 (approved):

PR #2757, opened 2024-05-09
chore: Migrate Junit 4 to Junit 5 for showcase

### LINKED ISSUE #2728: Migrate Junit 4 to Junit 5: showcase

zhumin8:
fixes #2728, attempt to remove Junit 4 support after migration.
Other than POM dependency migrate, changes include:
- package name changes
- Junit 5 syntax upgrades, e.g. `@Before` --> `@BeforeEach`, Replace assertion methods 
- remove public modifier on tests and test classes.
- Refactor JUnit 4 TemporaryFolder `@Rule` in [ITGdch.java](https://github.com/googleapis/sdk-platform-java/pull/2757/files#diff-6ae7755a0b038e1a2febae2d27e36c762f6751b8c7db577421667069399884b4) to JUnit 5 `@TempDir`
- Replace `@Test(timeout = 15000L)` in [ITClientShutdown.java](https://github.com/googleapis/sdk-platform-java/pull/2757/files#diff-70d1df57471178a7a63302f82e4a4855ffbbd642ea67d92d501bd1f7008957ca) with `@Timeout(15)`
- Update `@RunWith(Parameterized.class)` test in [ITHttpAnnotation.java](https://github.com/googleapis/sdk-platform-java/pull/2757/files#diff-03d420650ecc9fe78ad4887761043c4fdceaa978f464ce30cfc4ed5f8be9b64d) to `@ParameterizedTest` with `@MethodSource("data")`

~~Note: #2737 creates a new test class with JUnit4 syntax. Depending on merging order, I will either update in this pr, or #2737.~~ Updated.

Due to truth library depending on junit 4 ([see issue](https://github.com/google/truth/issues/333)), junit 4 cannot be completely removed, or will encounter `java.lang.ClassNotFoundException: org.junit.runner.notification.RunListener` running tests with maven surefire. To keep things cleaner, excluding the implicitly junit brought in from truth and `junit-vintage-engine`. We could also do the reverse, and make a comment if that's prefered.


showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/ComplianceClientHttpJsonTest.java
```diff
@@ -50,20 +50,20 @@
...
 @Generated("by gapic-generator-java")
-public class ComplianceClientHttpJsonTest {
+class ComplianceClientHttpJsonTest {
   private static MockHttpService mockService;
   private static ComplianceClient client;
 
-  @BeforeClass
-  public static void startStaticServer() throws IOException {
+  @BeforeAll
+  static void startStaticServer() throws IOException {
```
lqiu96, line 66: IIRC, these unit test classes (i.e. ComplianceClientHttpJsonTest, ComplianceClientTest, etc..) are generated by the generator. I think these files will be overwritten on the next showcase regeneration so we'll probably need to coordinate a change in the generator + shared-deps/pom-parent for these.

blakeli0: Yes, let's only migrate the integration tests in `it` folder.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
```diff
@@ -76,27 +74,29 @@ public class ITGdch {
...
     }
   }
 
-  private void prepareCredentials() throws IOException {
+  private void prepareCredentials(Path tempDir) throws IOException {
     // Copy file so it can be referenced by Path even in native-image builds
-    File caCertFile = tempFolder.newFile(CA_CERT_FILENAME);
+    File caCertFile = new File(tempDir.toFile(), CA_CERT_FILENAME);
+    //    Path caCertPath = tempDir.resolve(CA_CERT_FILENAME);
+    //    File caCertFile = Files.createFile(caCertPath);
```
lqiu96, line 99: Were these lines left in for testing locally?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITAutoPopulatedFields.java
```diff
@@ -120,7 +120,7 @@ public void sendMessage(ReqT message) {
   private EchoClient httpJsonClient;
   private EchoClient httpJsonClientWithRetries;
 
-  @Before
+  @BeforeEach
   public void createClients() throws Exception {
```
lqiu96, line 124: Can these methods all be package-private scope?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/ComplianceClientHttpJsonTest.java
blakeli0: > If it doesn't exist already, can you create a ticket to update the generator to use JUnit5 when generating these unit tests?

I purposely excluded generated unit tests from the scope at this moment, because the change would apply to google-cloud-java and all handwritten libraries, we need to coordinate the change and update their poms accordingly. Also since the main benefit is to make unit testing in `sdk-platform-java` easier, I think it's better to exclude the effort at this moment.
On the other hand, creating a ticket to track the future effort is always a good idea.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITGdch.java
zhumin8: oops. Thanks for catching this.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/ComplianceClientHttpJsonTest.java
zhumin8: Good point. Did not realize these are generated tests. Will revert the changes. 

Created https://github.com/googleapis/sdk-platform-java/issues/2762 to track updating generator for unit tests gen.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiVersionHeaders.java
```diff
@@ -158,8 +158,8 @@ public void onClose(int statusCode, HttpJsonMetadata trailers) {
   private ComplianceClient grpcComplianceClient;
   private ComplianceClient httpJsonComplianceClient;
 
-  @Before
-  public void createClients() throws Exception {
+  @BeforeEach
```
lqiu96, line 161: Might have missed this before/ don't remember our conversation, but was there a reason this was is @BeforeEach/@AfterEach and not @BeforeAll/@AfterAll?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITAutoPopulatedFields.java
```diff
@@ -120,8 +120,8 @@ public void sendMessage(ReqT message) {
   private EchoClient httpJsonClient;
   private EchoClient httpJsonClientWithRetries;
 
-  @Before
-  public void createClients() throws Exception {
+  @BeforeEach
```
lqiu96, line 123: Same with this. I know it isn't a test you wrote, but could we try with @BeforeAll? If it doesn't work, let's keep it as is.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITHttpAnnotation.java
```diff
@@ -39,38 +39,33 @@
...
 
-  @Parameterized.Parameters(name = "Compliance Group Name: {0}")
-  public static String[] data() {
-    return new String[] {
-      "Fully working conversions, resources",
-      "Binding selection testing",
-      "Cases that apply to non-path requests",
-      "Fully working conversions, no resources"
-    };
+  public static Stream<Arguments> data() {
```
lqiu96, line 55: nit: Probably doesn't matter, but to keep it consistent perhaps this can be private?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
```diff
@@ -141,8 +141,8 @@ private OpenTelemetryMetricsRecorder createOtelMetricsRecorder(
     return new OpenTelemetryMetricsRecorder(openTelemetry, SERVICE_NAME);
   }
 
-  @Before
-  public void setup() throws Exception {
+  @BeforeEach
```
lqiu96, line 144: Ah this case I remember, it's because the InMemoryMetricRecorder needs to be cleared every run. It was a limitation when I was writing the test so it should be kept this way.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/update.sh
```diff
@@ -0,0 +1,54 @@
...
+    sed -i 's/import org.junit.AfterClass;/import org.junit.jupiter.api.AfterAll;/g' "$file"
+    sed -i 's/import org.junit.Ignore;/import org.junit.jupiter.api.Disabled;/g' "$file"
+    sed -i 's/import org.junit.Assert;/import org.junit.jupiter.api.Assertions;/g' "$file"
+    sed -i 's/import static org.junit.Assert.assertEquals;/import static org.junit.jupiter.api.Assertions.assertEquals;/g' "$file"
+    sed -i 's/import static org.junit.Assert.assertTrue;/import static org.junit.jupiter.api.Assertions.assertTrue;/g' "$file"
+    sed -i 's/import static org.junit.Assert.assertNotSame;/import static org.junit.jupiter.api.Assertions.assertNotSame;/g' "$file"
+    sed -i 's/import static org.junit.Assert.assertThrows;/import static org.junit.jupiter.api.Assertions.assertThrows;/g' "$file"
+
+    # Replace annotations (with potential whitespace variations and end-of-word check)
+    sed -i 's/@Test(timeout = 15000L)/@Test\n  @Timeout(15)/g' "$file"
```
lqiu96, line 21: Nice script! I think this file can be removed from this PR.

lqiu96 (approved):
LGTM. Added a few nits. If you could test them locally and see if they work.

lqiu96:
Since Min is OOO, I will take a look a final few things for this issue and merge if everything else looks fine.

showcase/gapic-showcase/pom.xml
```diff
@@ -132,14 +132,35 @@
 
     <!-- Test dependencies -->
     <dependency>
-      <groupId>junit</groupId>
-      <artifactId>junit</artifactId>
+      <groupId>org.junit.platform</groupId>
```
blakeli0, line 135: We probably don't need this `junit-platform-launcher`.

lqiu96:
```
[INFO] --- maven-failsafe-plugin:3.2.5:integration-test (default) @ gapic-showcase ---
[INFO] Using configured provider org.apache.maven.surefire.junitcore.JUnitCoreProvider
[INFO] 
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 0, Failures: 0, Errors: 0, Skipped: 0
```

Looks like the ITs aren't actually running.

showcase/gapic-showcase/pom.xml
lqiu96: Removed

lqiu96:
I see it is running the tests now:
```
[INFO] --- clirr-maven-plugin:2.8:check (default) @ gapic-showcase ---
[INFO] Skipping execution
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for GAPIC Showcase Client Core Parent 0.0.1-SNAPSHOT:
[INFO] 
[INFO] GAPIC Showcase Client Core Parent .................. SUCCESS [  2.109 s]
[INFO] proto-gapic-showcase-v1beta1 ....................... SUCCESS [  1.008 s]
[INFO] grpc-gapic-showcase-v1beta1 ........................ SUCCESS [  2.228 s]
[INFO] GAPIC Showcase Client .............................. SUCCESS [01:39 min]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  01:45 min
[INFO] Finished at: 2024-05-17T19:11:27Z
[INFO] ------------------------------------------------------------------------
```

showcase/pom.xml
```diff
@@ -54,10 +54,11 @@
...
       <dependency>
-        <groupId>junit</groupId>
-        <artifactId>junit</artifactId>
-        <version>4.13.2</version>
-        <scope>test</scope>
+        <groupId>org.junit</groupId>
+        <artifactId>junit-bom</artifactId>
+        <version>5.10.2</version>
+        <type>pom</type>
+        <scope>import</scope>
```
lqiu96, line 61: As this doesn't inherit from shared-deps or pom-parent, we declare the version here.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITServerSideStreaming.java
```diff
@@ -151,7 +151,7 @@ public void testGrpc_serverWaitTimeout_watchdogCancelsStream() throws Exception
       for (EchoResponse response : responseStream) {
         responses.add(response.getContent());
       }
-      Assert.fail("No exception was thrown");
+      Assertions.fail("No exception was thrown");
```
JoeWang1127, line 154: Can this be replaced by `assertThrows`?

JoeWang1127 (approved):

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITServerSideStreaming.java
lqiu96: Yep! Updating.

showcase/pom.xml
```diff
@@ -116,6 +117,15 @@
           <plugin>
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-failsafe-plugin</artifactId>
+            <!-- Shared-Configs defines JUnit Provider 4.7 as default: https://github.com/googleapis/java-shared-config/blob/465bb399aef9aa8383f11c23f10a97df49c1d057/java-shared-config/pom.xml#L86-L92 -->
+            <!-- Override it here for showcase only as other libraries may still be reliant on JUnit4 -->
+            <dependencies>
```
blakeli0, line 122: Do we know why the integration tests would not run without this change?

blakeli0 (approved):

showcase/pom.xml
lqiu96: I believe it's due the failsafe using junit-provider 4.7 (which works for junit 4.7+) as the selected provider instead of junit-platform: https://maven.apache.org/surefire/maven-failsafe-plugin/examples/junit-platform.html#provider-selection

PR #2761, opened 2024-05-09
fix: prepare to generate grafeas

JoeWang1127:
In this PR:
- Add suffix to sample source folder
- Add unit tests

Context:
- The package name of grafeas is `io.grafeas.v1` which is not starts with `com`. When copying samples to destination directory, we need to change the suffix of source directory.
- grafeas doesn't have a valid api_id (empty string). Before this change, the `api_id` will be populated as `{library.api_shortname}.googleapis.com`, which is not correct. `api_id` should not be populated to `.repo-metadata.json` if set to empty string.

With the library entry added in generation configuration:
```
- api_shortname: containeranalysis
  name_pretty: Grafeas
  product_documentation: https://grafeas.io
  api_description: n/a
  client_documentation: "https://cloud.google.com/java/docs/reference/grafeas/latest/overview"
  release_level: stable
  distribution_name: "io.grafeas:grafeas"
  api_id: ""
  codeowner_team: "@googleapis/aap-dpes"
  library_name: grafeas
  requires_billing: false
  GAPICs:
  - proto_path: grafeas/v1
 ```

The generated java-grafeas is in https://github.com/googleapis/google-cloud-java/pull/10820

### CODE REVIEW

File: library_generation/utils/utilities.py [hunk 1]
```diff
@@ -216,9 +217,12 @@ def generate_prerequisite_files(
         if config.is_monorepo()
         else f"googleapis/{language}-{library_name}"
     )
-    api_id = (
-        library.api_id if library.api_id else f"{library.api_shortname}.googleapis.com"
-    )
+    if library.api_id == EMPTY_API_ID:
```
blakeli0, line 220: Thanks for fixing this empty `api_id` issue! However, I think it was a mistake to have an empty `api_id` in repo-metadata.json, as it is specified in the [service yaml](https://github.com/googleapis/googleapis/blob/master/grafeas/v1/grafeas_v1.yaml#L3C7-L3C39). I created a [PR](https://github.com/googleapis/google-cloud-java/pull/10831) to fix it.

library_generation/generate_library.sh
```diff
@@ -224,7 +224,11 @@ if [[ "${proto_only}" == "false" ]]; then
   mv_src_files "gapic" "test" "${temp_destination_path}"
   if [ "${include_samples}" == "true" ]; then
     # move java_gapic_srcjar/samples/snippets to samples/snippets
-    mv_src_files "samples" "main" "${temp_destination_path}"
+    if [[ "${proto_path}" == "grafeas/v1" ]]; then
```
blakeli0, line 227: Shall we just skip generating samples for grafeas as they are currently not included in the repo?

library_generation/utils/utilities.py
JoeWang1127: Thanks, I didn't notice the api_id should not be empty.

I'll revert this change.

library_generation/generate_library.sh
JoeWang1127: If we don't want to add special case like this, we need to change `include_samples` in BUILD.

library_generation/utils/utilities.py
JoeWang1127: Done.

diegomarquezp (approved):

blakeli0 (approved):
LGTM. Please update description to reflect the latest change.

PR #2765, opened 2024-05-12
chore: centralize hermetic build workflow implementations

diegomarquezp:
This PR transfers the workflows and their implementations involved in hermetic library generation (i.e. `update_googleapis_committish.yaml` and `hermetic_library_generation.yaml`).

The purpose is to simplify the hermetic build setup requirements in HW libraries and google-cloud-java to simply declare two workflows that will delegate all the logic to the workflows and their corresponding scripts stored in sdk-platform-java. 

### Follow ups:
 - In google-cloud-java, remove `generation/update_googleapis_commit.sh` and `hermetic_library_generation.sh`, plus the workflows that use them.

### DIFF
--- a/.github/workflows/hermetic_library_generation.yaml
+++ b/.github/workflows/hermetic_library_generation.yaml
@@ -15,7 +15,7 @@
 # downstream client libraries before they are released.
 name: Hermetic library generation upon generation config change through pull requests
 on:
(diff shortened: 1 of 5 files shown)


diegomarquezp:
We will use synthtool templates for this

PR #2772, opened 2024-05-15
chore: [hermetic-build] simplify usage of synthtool 

diegomarquezp:
This PR eliminates the need for:
 * `library_generation/owlbot/src/apply_repo_templates.py`
 * Passing the config yaml path to the postprocessing scripts

 ### How is it achieved?
 We set the environment variable `SYNTHTOOL_TEMPLATES` to point to a
 temp dir where we copy the contents of `owlbot/templates` plus a
 modification of the template using `sed` (instead of
 `synthtool.common_templates`). In this way, we avoid the usage of
 `apply_repo_templates.py`.
 generation_config.template_excludes.

 #### Why don't we use common_templates to pass libraries_bom_version?
 Because `owlbot.py` is a legacy script that was written when
 `java.common_templates` didn't take `libraries_bom_version` as a
 parameter. If we don't like this approach of having two template
 replacement workflows we could

 1. Modify all owlbot.py files
 2. Pass libraries_bom_version as an environment variable (not bad)

 #### What about template_excludes in the config yaml?
 We still need it to generate new libraries. This is done by rendering a
 new `owlbot.py` via templates.


library_generation/owlbot/bin/entrypoint.sh
```diff
@@ -36,22 +39,19 @@ if [[ "${is_monorepo}" == "true" ]]; then
...
 echo "Retrieving files from owl-bot-staging directory..."
 if [ -f "owlbot.py" ]
 then
-  # we use an empty synthtool folder to prevent cached templates from being used
+  # we copy the templates to a temp folder because we need to do a special
+  # modification regarding libraries_bom_version that can't be handled by the
+  # synthtool library considering the way owlbot.py files are written
   export SYNTHTOOL_TEMPLATES=$(mktemp -d)
+  cp -r ${scripts_root}/owlbot/templates/* "${SYNTHTOOL_TEMPLATES}"
+  sed -i 's/\$\$__libraries_bom_version__\$\$/'"${libraries_bom_version}"'/g' "${SYNTHTOOL_TEMPLATES}/java_library/README.md"
```
JoeWang1127, line 51: What if we migrate java.py to sdk-platform-java? Do you think of another solution?

diegomarquezp: It would be ideal. A couple considerations:
 * `owlbot.py` imports `synthtool`, meaning that we would need to not only transfer `java.py` but other scripts such as `common.py` (or we could merge all the logic in a single script)
 * If we migrate synthtool/java.py, we would still need a way to pass `libraries_bom_version` to `owlbot.py`. We can either modify `owlbot.py` or we can rely on an environment variable consumed in java.py

diegomarquezp: Another more hacky approach would be to pass the version in repo-metadata

library_generation/owlbot/src/apply_repo_templates.py
blakeli0, line 1: What was this file used for? I think it was not only for `libraries_bom_version`? Or maybe it was a mistake that we rendered the templates twice?

library_generation/owlbot/bin/entrypoint.sh
blakeli0: The logic here seems not easy to maintain, as it is tightly coupled with a magic string `__libraries_bom_version__`. 
I'm leaning towards the environment variable approach you mentioned. If we go with that approach, I guess we should not need any changes in this file or the template, we just need to load the libraries bom version from environment variable and add it to the kwargs in [synthtool](https://github.com/googleapis/synthtool/blob/0b86c72fe652dd7e52ba05a63f61bc1399ad5d65/synthtool/languages/java.py#L505)?

diegomarquezp: I added https://github.com/googleapis/synthtool/pull/1967 to infer the bom version from the environment. Once that's merged, we can enable the env var approach.

library_generation/owlbot/src/apply_repo_templates.py
diegomarquezp: It allowed us to run `common_templates` with the list of `template_excludes` listed in the config yaml

JoeWang1127 (approved):

PR #2779, opened 2024-05-16
test: use Junit5 in api-common

### LINKED ISSUE #2727: Migrate Junit 4 to Junit 5: api-common-java

alicejli:
Fixes #2727  ☕️

### DIFF
--- a/api-common-java/src/test/java/com/google/api/core/ApiFuturesTest.java
+++ b/api-common-java/src/test/java/com/google/api/core/ApiFuturesTest.java
@@ -38,12 +38,12 @@ import java.util.concurrent.CancellationException;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.atomic.AtomicInteger;
-import org.junit.Test;
+import org.junit.jupiter.api.Test;
 
 public class ApiFuturesTest {
 
   @Test
-  public void testAddCallback() throws Exception {
+  public void testAddCallback() {
     final AtomicInteger flag = new AtomicInteger();
     SettableApiFuture<Integer> future = SettableApiFuture.<Integer>create();
     ApiFutures.addCallback(
--- a/api-common-java/src/test/java/com/google/api/core/SettableApiFutureTest.java
+++ b/api-common-java/src/test/java/com/google/api/core/SettableApiFutureTest.java
@@ -30,11 +30,11 @@
 package com.google.api.core;
 
 import com.google.common.truth.Truth;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
-import org.junit.Test;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
 
 public class SettableApiFutureTest {
   @Test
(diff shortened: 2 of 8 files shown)


api-common-java/src/test/java/com/google/api/core/ApiFuturesTest.java
```diff
@@ -38,12 +38,12 @@
...
 import java.util.concurrent.Executor;
 import java.util.concurrent.atomic.AtomicInteger;
-import org.junit.Test;
+import org.junit.jupiter.api.Test;
 
 public class ApiFuturesTest {
 
   @Test
-  public void testAddCallback() throws Exception {
+  public void testAddCallback() {
```
JoeWang1127, line 46: You can remove `public` since this is not required in Junit 5.

alicejli: Updated.

api-common-java/src/test/java/com/google/api/core/ApiFuturesTest.java
```diff
@@ -38,12 +38,12 @@
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.atomic.AtomicInteger;
-import org.junit.Test;
+import org.junit.jupiter.api.Test;
 
 public class ApiFuturesTest {
```
lqiu96, line 43: I believe the class can also be package-private: i.e. `class ApiFuturesTest`

alicejli: Ah okay - I'll update all of them to be package-private, assuming there's no reason for test classes to be public?

api-common-java/src/test/java/com/google/api/core/SettableApiFutureTest.java
```diff
@@ -57,15 +57,19 @@ public void testCancel() throws Exception {
...
 
-  @Test(expected = ExecutionException.class)
-  public void testException() throws Exception {
-    SettableApiFuture<Integer> future = SettableApiFuture.<Integer>create();
-    future.setException(new Exception());
-    future.get();
+  @Test
+  void testException() {
+    Assertions.assertThrows(
+        Exception.class,
```
lqiu96, line 63: Hmm, odd. The old test had `@Test(expected = ExecutionException.class)`. I wonder what changed since it looks like it's throwing Exception back now.

api-common-java/src/test/java/com/google/api/resourcenames/UntypedResourceNameTest.java
```diff
@@ -82,10 +79,10 @@ public void testInsertIntoFieldValuesMap() {
...
 
   @Test
-  public void testNullName() {
+  void testNullName() {
     assertFalse(UntypedResourceName.isParsableFrom(null));
     try {
-      UntypedResourceName fooName = UntypedResourceName.parse(null);
+      UntypedResourceName.parse(null);
     } catch (NullPointerException e) {
     }
```
lqiu96, line 87: Can this be converted to `assertThrows`?

api-common-java/src/test/java/com/google/api/core/ApiFuturesTest.java
lqiu96: I think they were required for Junit4, but Junit5 no longer requires them to be public (just not private).

api-common-java/src/test/java/com/google/api/core/SettableApiFutureTest.java
alicejli: I think since `ExecutionException` is a subtype of `Exception`, both are technically correct (I tested both and they both pass). But to your point of keeping it the same as the previous one, I've updated to have it `assertThrows(ExecutionException.class)`

lqiu96 (approved):
LGTM

PR #2784, opened 2024-05-17
chore: remove Owlbot yaml

JoeWang1127:


blakeli0:
I think we should only remove these two files after we introduce `generation_config.yaml` to `sdk-platform-java`.

JoeWang1127:
Merge it after we introducing library generation workflow.

blakeli0 (approved):

PR #2793, opened 2024-05-21
chore: remove resquests dependency from library_generation

JoeWang1127:


diegomarquezp (approved):
Thanks for the dep cleanup! Can you please update the title to something that explains that this is a dependency we are removing from the hermetic build scripts? For example:
 - chore: [hermetic-build] remove `requests` dependency
 - chore: remove `resquests` dependency from `library_generation`

blakeli0 (approved):

Issue #2821, opened 2024-05-24
HeaderProvider reference is lost during ServiceOptions.toBuilder


burkedavison:
`serviceOptions.toBuilder().build()` should not cause field values to be unset.

blakeli0:
Thanks Burke! How did you notice this issue?

burkedavison:
Working through specific requirements for https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/2906 by creating a test for it. I found I couldn't modify LoggingOptions via toBuilder without erasing its header provider. 
https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/d8d7168e57683adf76b1e56b1a3e7d87fb7d8bdf/spring-cloud-gcp-logging/src/main/java/com/google/cloud/spring/logging/LoggingAppender.java#L43

PR #2822, opened 2024-05-24
fix: retain all fields during ServiceOptions.tobuilder

burkedavison:
See #2821 

This PR originally created to ensure headerProvider is correctly copied, but expanded to ensure all fields are used in equals and hashCode comparisons for the associated copy tests.


burkedavison:
Downstream checks failing on serialization tests.

PR #2825, opened 2024-05-28
chore: add generation workflow

JoeWang1127:
In this PR:
- Add workflow and shell scripts used in library generation.


.github/workflows/hermetic_library_generation.yaml
```diff
@@ -0,0 +1,54 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+        token: ${{ secrets.CLOUD_JAVA_BOT_TOKEN }}
+    - name: Build image
+      shell: bash
+      run: |
+        docker build \
```
blakeli0, line 33: Compare to the workflow in google-cloud-java, is this the only main change?

library_generation/scripts/hermetic_library_generation.sh
blakeli0, line 1: I feel it's better to put these two shell scripts under `.github/` as they are only used by github workflows, WDYT? Same thing for the [scripts](https://github.com/googleapis/google-cloud-java/blob/main/generation/hermetic_library_generation.sh) in google-cloud-java.

.github/workflows/hermetic_library_generation.yaml
JoeWang1127: Changes compared to google-cloud-java
1. build local image
2. install gapic-generator-java-pom-parent

library_generation/scripts/hermetic_library_generation.sh
JoeWang1127: I think it's fine.

.github/workflows/hermetic_library_generation.yaml
diegomarquezp: An idea to get rid of the `.m2` volume mapping.
Our workflows build and publish a docker image every time we merge a PR in this repo. The tag is the commit SHA of the merged (squashed) commit. 

Would it be useful to run `mvn install -pl ggj-pom-parent` in the Dockerfile and label the image as `x.x.x-SNAPSHOT` as well (we can have two tags)?

diegomarquezp (approved):

library_generation/scripts/hermetic_library_generation.sh
JoeWang1127: Done.

.github/workflows/hermetic_library_generation.yaml
JoeWang1127: This step is specific to generate common protos. 

Putting `mvn install` in the Dockerfile will apply to all libraries, which is a redundant step.

blakeli0 (approved):

Issue #2832, opened 2024-05-30
Update method_name Otel Attribute to method


lqiu96:
The method_name default attribute for each trace should be changed to be method to be consistent with the mapping in Cloud Monitoring.

PR #2833, opened 2024-05-30
fix: Update default Otel Attribute from method_name to method

lqiu96:
Fixes https://github.com/googleapis/sdk-platform-java/issues/2832

### DIFF
--- a/gax-java/gax/clirr-ignored-differences.xml
+++ b/gax-java/gax/clirr-ignored-differences.xml
@@ -47,4 +47,10 @@
     <className>com/google/api/gax/rpc/StubSettings</className>
     <method>* getServiceName()</method>
   </difference>
+  <!-- Update default method_name Otel attribute to method -->
+  <difference>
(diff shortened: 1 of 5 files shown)


gax-java/gax/clirr-ignored-differences.xml
```diff
@@ -47,4 +47,10 @@
     <className>com/google/api/gax/rpc/StubSettings</className>
     <method>* getServiceName()</method>
   </difference>
+  <!-- Update default method_name Otel attribute to method -->
```
blakeli0, line 50: Since `MetricsTracer` is an internal API, can we make exclude the whole class?

lqiu96: I wasn't find a way to ignore all differences, so I kept it as `<differenceType>6011</differenceType>`. I just used a wildcard for `<field>`.

blakeli0 (approved):

PR #2840, opened 2024-05-31
fix: HttpJson tests generate request for Mixins with custom paths

lqiu96:
[WIP] Draft PR to explore the possible way to fix this issue.

Fixes: https://github.com/googleapis/sdk-platform-java/issues/1839


gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
```diff
@@ -442,6 +454,79 @@ public static Expr createSimpleMessageBuilderValue(
...
+    // logic will try to match a resource that has the pattern `service/*`.
+
+    // Resource reference (google.api.resource_reference) is a string type with a field name of
+    // `parent` or `name`. It won't have anything nested in the bindings (i.e. `{name.field=*}`).
+    // This check is needed as a message can contain fields with valid resource references that
+    // aren't used in the path. If the field isn't used in the path, simply return the matched
+    // resource.
+    boolean isFieldUsedInPath =
+        bindings != null && bindings.getPathParametersValuePatterns().containsKey(field.name());
+    if (isFieldUsedInPath) {
```
blakeli0, line 506: Consider refactor this to
```
if (! isFieldUsedInPath) {
  return matchedResources.get(0);
}
...
```
to reduce indentation.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
```diff
@@ -442,6 +454,79 @@ public static Expr createSimpleMessageBuilderValue(
...
+      for (ResourceName resourceName : matchedResources) {
+        // If the resource is matched as a wildcard `*`, the field's resource_reference is a
+        // wildcard. Attempt to match with a known resource.
+        if (resourceName.isOnlyWildcard()) {
```
blakeli0, line 510: What exactly does `isOnlyWildcard` mean?

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposerTest.java
blakeli0, line 1: Can we create a golden test with a test service yaml that mimics the exact scenario(overrides mixin path)?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
```diff
@@ -442,6 +454,79 @@ public static Expr createSimpleMessageBuilderValue(
...
+    }
+    List<ResourceName> matchedResources = ImmutableList.of(resourceNames.get(resourceTypeString));
+    // Use the parent resource if the field is a `child_type`. This is for certain RPCs like `List`
+    // and `Create` which have a resource reference to the parent collection.
+    if (field.resourceReference().isChildType()) {
+      // A resource may have multiple parents. Check the possibility for each parent resource
+      List<ResourceName> parentResources =
+          findParentResources(matchedResources.get(0), new ArrayList<>(resourceNames.values()));
+      if (!parentResources.isEmpty()) {
+        matchedResources = parentResources;
```
blakeli0, line 484: The logic here is OK, but I feel we may want to differentiate the concept of `matchedResources` and `derivedResources`/`flattenedResources`. The matched resource is actually always one, but we could derive more resources from it if it is of child type. Hence putting the matched resource to a list and reusing/reassigning the same list is a little confusing.

lqiu96: I see. Let me rename this from matchedResources into something less confusing since I do think I'm using the term `match` to mean multiple things.

For fields with direct resource_references, matched resources is always one (assuming that the resource exists and is known). However, it is possible that the pattern doesn't match to the bindings so it really isn't "matched" yet.

For child_type fields, we *must* try to use the parent resources. However, we still need to match the parent resources with the bindings, so they're technically not matched yet either.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
lqiu96: Sounds good. Will do.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
```diff
@@ -442,6 +454,79 @@ public static Expr createSimpleMessageBuilderValue(
...
+
+    // Resource reference (google.api.resource_reference) is a string type with a field name of
+    // `parent` or `name`. It won't have anything nested in the bindings (i.e. `{name.field=*}`).
+    // This check is needed as a message can contain fields with valid resource references that
+    // aren't used in the path. If the field isn't used in the path, simply return the matched
+    // resource.
+    boolean isFieldUsedInPath =
+        bindings != null && bindings.getPathParametersValuePatterns().containsKey(field.name());
+    if (isFieldUsedInPath) {
+      for (ResourceName resourceName : matchedResources) {
```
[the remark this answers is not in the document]
lqiu96: Sounds good. I'm also thinking about changing this method to return all matched resource names

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
lqiu96: I believe If the resource's pattern is `*`

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
```diff
@@ -442,6 +454,79 @@ public static Expr createSimpleMessageBuilderValue(
...
+              .orElse(null);
+        }
+
+        // If resource matches the HttpBindings, use the resource
+        if (resourceName.getMatchingPattern(bindings) != null) {
+          return resourceName;
+        }
+      }
+      // Unable to match the possible resources with the HttpBindings
+      return null;
```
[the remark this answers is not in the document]
lqiu96: It will create a default value with the createValue() method. The tests will pass since the default value will just be a value that matches the HttpBindings.

i.e. The tests will go from something like
```
setResource(ProjectName.of("[Project]"]))
```
to
```
setResource("projects/project-134")
```

where the resource value is just the string version of the resource.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
```diff
@@ -249,10 +251,19 @@ private static Optional<ResourceName> findParentResource(
...
       if (parentPattern.isPresent() && patternToResourceName.containsKey(parentPattern.get())) {
-        return Optional.of(patternToResourceName.get(parentPattern.get()));
+        parentResources.add(patternToResourceName.get(parentPattern.get()));
       }
     }
 
+    return parentResources;
+  }
+
+  private static Optional<ResourceName> findParentResource(
```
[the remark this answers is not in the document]
lqiu96: Yeah it was, but I am only seeing a single reference for this, so I'm am considering removing this.

lqiu96:
Generated output for Java files via Hermetic Build: https://github.com/googleapis/google-cloud-java/pull/10952

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposerTest.java
lqiu96: Sounds good. I will attempt to re-create this.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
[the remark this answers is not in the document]
lqiu96: I removed this method.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
lqiu96: Updated.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
blakeli0: Thanks! Do you have an example of it?

lqiu96: Asset resource is defined as wildcard pattern: https://github.com/googleapis/googleapis/blob/573205a81b9b96ea6d4e0365373839b59f4b427e/google/cloud/asset/v1/assets.proto#L96-L99

so it could be something with this kind of resource reference: https://github.com/googleapis/googleapis/blob/573205a81b9b96ea6d4e0365373839b59f4b427e/google/cloud/asset/v1/asset_service.proto#L451-L456

The resource_reference for a field can be wildcard: https://github.com/googleapis/googleapis/blob/573205a81b9b96ea6d4e0365373839b59f4b427e/google/iam/v1/iam_policy.proto#L104-L106

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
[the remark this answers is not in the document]
lqiu96: > With this misconfiguration, the test would fail previously because we would generate setResource based on the resource pattern only

Currently, for this bindings not matching the resource pattern misconfiguration, that's true. It would fail the test and we would see it in the CI.

However, if we throw an exception on this case now, I *think* we may break for very niche cases. i.e. ServiceManagement comes to mind. It does not have any resources defined inside the library, but uses IAM mixins. The IAM mixin RPCs have resources defined in the httpbindings and the request messages have fields with resource_reference. It's not an RPC that the service team wrote, but is using. This would require them to add a resource for the `service/*` pattern (to be clear, I think it's probably something they *should* do, but currently don't).

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposerTest.java
lqiu96: I added a resource_name_testing.proto to generate a bunch of common test scenarios. I can't find an easy way to incorporate IAM mixins as the logic for golden files parses an individual service and not multiple services (i.e. Echo + IAM) to combine the RPCs.

I modelled the ServiceManagement situation in the `WildcardResourceReference` and `WildcardResourceReferenceStrictBindings` RPCs

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
```diff
@@ -249,11 +251,11 @@ private static Optional<ResourceName> findParentResource(
     for (String childPattern : childResource.patterns()) {
       Optional<String> parentPattern = ResourceReferenceUtils.parseParentPattern(childPattern);
       if (parentPattern.isPresent() && patternToResourceName.containsKey(parentPattern.get())) {
-        return Optional.of(patternToResourceName.get(parentPattern.get()));
+        parentResources.add(patternToResourceName.get(parentPattern.get()));
```
JoeWang1127, line 254: Since at most one parent resources is used, why not stop adding to the list once we find a resource?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
```diff
@@ -442,6 +449,87 @@ public static Expr createSimpleMessageBuilderValue(
...
+    // aren't used in the path. Additionally, this flag only impacts HttpJson callables as gRPC
+    // does not need to match the request the callable's URL. This method is used to generate
+    // default values for gRPC callables as well but gRPC's valuePatterns map will be empty and
+    // the logic will be skipped.
+    boolean isFieldUsedInPath = bindings != null && valuePatterns.containsKey(field.name());
+
+    // Holds the list of known resources (excluding the wildcard resource).
+    List<ResourceName> possibleResources =
+        resourceNames.values().stream()
+            .filter(x -> !x.isOnlyWildcard())
```
JoeWang1127, line 499: nit:
```suggestion
            .filter(resourceName -> !resourceName.isOnlyWildcard())
```

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
```diff
@@ -442,6 +449,87 @@ public static Expr createSimpleMessageBuilderValue(
...
+            .collect(Collectors.toList());
+
+    if (resourceReference.isOnlyWildcard()) {
+      if (isFieldUsedInPath) {
```
JoeWang1127, line 503: Why not combine these two if-statements?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
lqiu96: It is possible that the first parent resource found does not match the bindings.

For example:
Resource has these patterns:
- `parentResources/{parentResource}/resources/{resource}`
- `projects/{project}/locations/{location}`
- `examples/{example}/testing/{test}`

This would have these parent patterns:
- `parentResources/{parentResource}`
- `projects/{project}`
- `examples/{example}`

If an RPC has these bindings:
- `parent={parentResources/*}`
- `parent={projects/*}`

It may be possible that the first matched parent resource is `examples/*`, which does not match the bindings.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
lqiu96: There are *two* possible paths if resourceReference is wildcard:
- If used in path, then filter for the pattern that matches the bindings
- If not used in the path, then return all possible resources

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/defaultvalue/DefaultValueComposer.java
```diff
@@ -442,6 +449,87 @@ public static Expr createSimpleMessageBuilderValue(
...
+   * edge case where an RPC's HttpBindings defined in the proto may not match a message. This is
+   * possible for Mixin RPCs if service teams configures a custom HttpBindings in the service yaml
+   * file. For example, the GetIamPolicyRequest's resource field has a resource reference to match
+   * anything (`*`). This normally works because the GetIamPolicy RPC will take in any resource
+   * (HttpBindings is defined with `{resource=**}`). But if the custom path is expected to take in a
+   * resource with a more specific path (i.e. `{resource=service/*}), it won't work unless the
+   * resource has a `service/*` pattern. This logic will try to match a resource that has the
+   * pattern `service/*`.
+   */
+  static List<ResourceName> getMatchingResources(
```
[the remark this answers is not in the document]
lqiu96: Added `@VisibleForTesting` annotation

PR #2864, opened 2024-06-07
chore: Build all modules of sdk-platform-java in hermetic generation workflow. 

blakeli0:
This is to fix the issue in https://github.com/googleapis/sdk-platform-java/actions/runs/9418294714/job/25945455263?pr=2860.


JoeWang1127:
Could you please add the issue link to the pr body?

I think a better solution is not downloading/installing the generator if it's not used, generating proto-only libraries, for example.

That being said, install all modules can be a quick fix and we can improve this later.

JoeWang1127 (approved):

PR #2874, opened 2024-06-10
fix: Move the logic of getting systemProductName from static block to static method

blakeli0:
The static field `systemProductName` should only be used for testing purposes. In production environment, the logic of checking `isOnComputeEngine` should be exactly the same as before https://github.com/googleapis/sdk-platform-java/pull/2614.

Fixes b/346215860

### DIFF
--- a/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
+++ b/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
@@ -83,19 +83,7 @@ import org.threeten.bp.Duration;
  */
 public final class InstantiatingGrpcChannelProvider implements TransportChannelProvider {
 
-  static String systemProductName;
-
-  static {
-    try {
-      systemProductName =
-          Files.asCharSource(new File("/sys/class/dmi/id/product_name"), StandardCharsets.UTF_8)
-              .readFirstLine();
-    } catch (IOException e) {
-      // If not on Compute Engine, FileNotFoundException will be thrown. Use empty string
-      // as it won't match with the GCE_PRODUCTION_NAME constants
-      systemProductName = "";
-    }
-  }
+  private static String systemProductName;
 
   @VisibleForTesting
   static final Logger LOG = Logger.getLogger(InstantiatingGrpcChannelProvider.class.getName());
@@ -345,11 +333,27 @@ public final class InstantiatingGrpcChannelProvider implements TransportChannelP
   static boolean isOnComputeEngine() {
     String osName = System.getProperty("os.name");
     if ("Linux".equals(osName)) {
+      String systemProductName = getSystemProductName();
       // systemProductName will be empty string if not on Compute Engine
       return systemProductName.contains(GCE_PRODUCTION_NAME_PRIOR_2016)
           || systemProductName.contains(GCE_PRODUCTION_NAME_AFTER_2016);
     }
     return false;
+  }
+
+  private static String getSystemProductName() {
+    // The static field systemProductName should only be set in tests
+    if (systemProductName != null) {
+      return systemProductName;
+    }
+    try {
+      return Files.asCharSource(new File("/sys/class/dmi/id/product_name"), StandardCharsets.UTF_8)
+          .readFirstLine();
+    } catch (IOException e) {
+      // If not on Compute Engine, FileNotFoundException will be thrown. Use empty string
+      // as it won't match with the GCE_PRODUCTION_NAME constants
+      return "";
+    }
   }
 
   // Universe Domain configuration is currently only supported in the GDU


gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -345,13 +333,29 @@ boolean isCredentialDirectPathCompatible() {
   static boolean isOnComputeEngine() {
     String osName = System.getProperty("os.name");
     if ("Linux".equals(osName)) {
+      String systemProductName = getSystemProductName();
```
lqiu96, line 336: IIUC, this should be assigned to back static instance field, right? The value would cached and we shouldn't need to re-read the file on future channel creations.

```suggestion
     systemProductName = getSystemProductName();
```

Perhaps the systemProductName should also be volatile since it *may* (?) be possible that we may re-size a channel while the `isOnComputeEngine` runs?

lqiu96, line 336: IIUC, this should be assigned to back static instance field, right? The value would cached and we shouldn't need to re-read the file on future channel creations.

```suggestion
     systemProductName = getSystemProductName();
```

Perhaps the systemProductName should also be volatile since it *may* (?) be possible that we may re-size a channel while the `isOnComputeEngine` runs?

blakeli0: I did it on purpose to keep the behavior identical as before. `systemProductName` is a static field which could be accessed by all threads, we would have to make it `synchronized` if we want to cache it, which could make things more complicated. Since the intention of this field is to make testing easier, I would prefer to keep all other behaviors untouched.

lqiu96: Hmm, I might be confusing things or misunderstanding this. I thought the original purpose of having this static was that we wouldn't end up having to read the file every time a gRPC channel is created or resized or repooled (it would be done once when the Provider was first initialized). If we aren't caching this system property value, then we would be re-reading this file every time, right? It seems like there would be no point in having this static method or static variable.

cushon: One option might be to use something like guava's [`Suppliers.memoize`](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/base/Suppliers.html#memoize(com.google.common.base.Supplier)) to cache the result of this, but still only call the method once the first time it's used.

blakeli0: > If we aren't caching this system property value, then we would be re-reading this file every time isOnComputeEngine() is called, right? It seems like there would be no point in having this as a static method or as a static variable.

Yes, it would be called every time. But IIRC I think the attempt of caching it as a static value or changing it to a non-static method was to make the unit testing easier. So yes this PR would revert the caching part that was introduced in https://github.com/googleapis/sdk-platform-java/pull/2614, but still keep the field so that we can set it in the builder in unit tests.

lqiu96: I took a look at the previous version and it seems like `isOnComputeEngine()` was called on every channel creation every time DirectPath was enabled: https://github.com/googleapis/sdk-platform-java/blob/dadb6b99c50a7915b2f9f6cc713a81253519ff46/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L373-L376

Since we're just going to read the file on channel creation every time DirectPath is enabled, why don't we just change `isOnComputeEngine()` to be non-static (It's package-private so we should be fine) and the systemProductName to be non-static as well?

Unless I'm missing something, we can just make everything non-static and potentially look to cache this value in a future PR?

blakeli0: > Since we're just going to read the file on channel creation every time DirectPath is enabled, why don't we just change isOnComputeEngine() to be non-static (It's package-private so we should be fine) and the systemProductName to be non-static as well?

Potentially we could, but at this moment, I think it's better to keep everything as they are. Other considerations like change them to non-static or cache it can be future considerations. Also static methods are slightly performant than non-static methods, so if there is no obvious benefit, we can keep them as static.

lqiu96: Ok sounds good, I'll create an issue to investigate making it non-static/ non-static + caching and add it to the backlog.

lqiu96 (approved):

PR #2878, opened 2024-06-11
chore: initiate library generation if Dockerfile is changed

JoeWang1127:
In this PR:
- Initiate generation process if the Dockerfile is changed.
- Move `docker build` and `mvn install` in shell script to speed up the workflow run if no changes in the `generation_config.yaml` and `Dockerfile`.

Example run: https://github.com/JoeWang1127/sdk-platform-java/actions/runs/9473714657/job/26101875644?pr=8


.github/scripts/hermetic_library_generation.sh
```diff
@@ -73,21 +74,30 @@ fi
...
+if [[ ! ("${change_of_last_commit}" == *"${generation_config}"* || "${change_of_last_commit}" == *"${docker_file}"*) ]]; then
+    echo "The last commit doesn't contain any changes to the generation_config.yaml or Dockerfile, skipping the whole generation process." || true
     exit 0
 fi
 # copy generation configuration from target branch to current branch.
 git show "${target_branch}":"${generation_config}" > "${baseline_generation_config}"
 config_diff=$(diff "${generation_config}" "${baseline_generation_config}" || true)
 
+# install generator locally since we're using a SNAPSHOT version.
+mvn -V -B -ntp clean install -DskipTests
```
blakeli0, line 94: I'm not sure we want to move `mvn install` and `docker build` to the scripts here. It does provide better encapsulation, but people often want to run the scripts directly in local, doing this here would significantly slow things down.

JoeWang1127: This workflow will run on _every_ pull request and it'll take ~5 mins to build the image and install generator only to discovery that there's no config/Dockerfile change or no code change at last.

The trade-off here is to speed up checks in pull requests.

On the other hand, I don't think this script is good for local development since it will checkout branches and create pull request.

blakeli0: > This workflow will run on every pull request and it'll take ~5 mins to build the image and install generator only to discovery that there's no config/Dockerfile change or no code change at last.

That's a good point, agreed.

>  I don't think this script is good for local development

Looks like it does not. Can we add more info regarding local development in a follow up task? For example, what are the steps if I want to generate a random library with a SNAPSHOT version of generator and/or docker images? We have some good stuff already in [DEVELOPMENT.md](https://github.com/googleapis/sdk-platform-java/blob/main/library_generation/DEVELOPMENT.md), and we can add more on top of it.

PR #2890, opened 2024-06-13
feat: populate `.repo-metadata.json` from highest version

JoeWang1127:
In this PR:
- populate `.repo-metadata.json` from highest stable version of proto paths.

The `.repo-metadata.json` is only generated once per library and the first item in proto paths is used, so the proto paths with the highest stable version should be the first item after sorting the proto paths in the library.

The method used to compare two proto paths, `a` and `b`:
1. if both of `a` and `b` don't have a version, the one with lower depth is smaller.
2. if either `a` or `b` has a version, the one with the version is smaller.
3. if either `a` or `b` has a stable version, but not both, the one with the stable version is smaller, e.g., `google/spanner/v2` is smaller than `google/spanner/v3beta`.
4. if both `a` and `b` have a non-stable version, the one with higher version is smaller, e.g., `google/spanner/v2beta1` is smaller than `google/spanner/v2alpha2`.
5. if both `a` and `b` have a stable version, the one with lower depth is smaller, e.g., `google/spanner/v1` is smaller than `google/spanner/admin/v1`, `google/spanner/v1` is smaller than `google/spanner/admin/v2`
6. if both `a` and `b` have a stable version and the depth is the same, the one with higher version is smaller, e.g., `google/spanner/v2` is smaller than `google/spanner/v1`.

Test the change:
1. build the image locally:
```
docker build -f .cloudbuild/library_generation/library_generation.Dockerfile -t local:latest .
```
2. build the generator locally:
```
mvn -V -B -ntp clean install -DskipTests
```
3. run the image:
```
docker run --rm -u "$(id -u):$(id -g)" -v "$(pwd):/workspace" -v "$HOME/.m2:/home/.m2" local:latest
```

Result:
1. no change in common protos
2. `transport` changed to `both` in `java-iam/.repo-metadata.json`. This is expected since the `transport` in [`google/iam/v2/BUILD.bazel`](https://github.com/googleapis/googleapis/blob/master/google/iam/v2/BUILD.bazel#L94) is `grpc+rest`. Without this change, the `transport` is parsed from `google/iam/v2/BUILD.bazel`, which doesn't have a `java_gapic_library` target, thus the value is `grpc` by default.


library_generation/model/gapic_config.py
```diff
@@ -22,3 +26,44 @@ class GapicConfig:
...
+        if self_stable and (not other_stable):
+            return True
+        if (not self_stable) and other_stable:
+            return False
+        # if two configs have different depth in proto_path, the one with
+        # lower depth is smaller.
+        if len(self_dirs) != len(other_dirs):
+            return len(self_dirs) < len(other_dirs)
+        # otherwise, the one with higher version is smaller.
+        return self_version > other_version
```
diegomarquezp, line 65: I'm a bit ignorant to string comparison in python. Is `v2` greater than `v1` because "2" has a higher ASCII value than "1"?

diegomarquezp (commented):
QQ, Don't we have `.repo-metadata` golden files for our units? Maybe we can change one of these to test the proto_path sorting criteria

library_generation/model/gapic_config.py
JoeWang1127: > Is v2 greater than v1 because "2" has a higher ASCII value than "1"?

Yes.

I released that string comparison has a bug here: `v10` is smaller than `v2`. I'll change the algorithm.

library_generation/test/model/library_config_unit_tests.py
```diff
@@ -37,3 +38,29 @@ def test_get_library_returns_api_shortname(self):
             gapic_configs=list(),
         )
         self.assertEqual("secret", library.get_library_name())
+
+    def test_get_sorted_gapic_configs_returns_correct_order(self):
```
blakeli0, line 42: Can we create a test for each of the scenario? [v1alpha1, v1], [v1, v2] etc. Because 
1. It would be much more readable. For example, just by reading the tests, I don't know why `v1alpha1, v1, v2, admin_v2, non_versioned, v1beta1` is sorted to `v2, v1, admin_v2, v1beta1, v1alpha1, non_versioned`. 
2. We can also easily test more scenarios. e.g. scenarios of two non versioned libraries.

library_generation/model/gapic_config.py
```diff
@@ -22,3 +26,44 @@ class GapicConfig:
...
+
+    @staticmethod
```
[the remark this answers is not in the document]
JoeWang1127: Done.

library_generation/test/model/library_config_unit_tests.py
JoeWang1127: I added unit tests to verify comparison result of two `gapic_config`s.

library_generation/utils/proto_path_utils.py
```diff
@@ -45,3 +46,11 @@ def find_versioned_proto_path(proto_path: str) -> str:
             idx = proto_path.find(version)
             return proto_path[:idx] + version
     return proto_path
+
+
+def find_version_in_proto_path(proto_path: str) -> Optional[str]:
```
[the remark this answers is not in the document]
JoeWang1127: Done.

library_generation/model/gapic_config.py
JoeWang1127: Done.

JoeWang1127:
> QQ, Don't we have .repo-metadata golden files for our units? Maybe we can change one of these to test the proto_path sorting criteria

The golden files are used in testing `.repo-metadata.json` generation, `generate_prerequisite_files` to be specific.

If we want to test generating this file with a `library_config` (from which contains multiple `proto_path`), we need test `generate_composed_library`, which is not easy to do since this function aggregates many "small" functions and is tested in integration test.

blakeli0 (approved):

PR #2923, opened 2024-06-17
chore: adjust repo-metadata generation to fit existing files in HW repos

diegomarquezp:
This PR introduces a couple of new keys in the repo-metadata.json [model](https://github.com/googleapis/sdk-platform-java/blob/main/library_generation/model/library_config.py).
 * `min_java_version`, as seen in [pubsublite](https://github.com/googleapis/java-pubsublite/blob/11c6540e2dec642492f5bd324ee810ca037d5a86/.repo-metadata.json#L11), [spanner](https://github.com/googleapis/java-spanner/blob/9deb34846b57e238642b6aa13f3107b2204cf969/.repo-metadata.json#L10) and others.
 * `recommended_package`, as seen in [spanner](https://github.com/googleapis/java-spanner/blob/9deb34846b57e238642b6aa13f3107b2204cf969/.repo-metadata.json#L20), [storage](https://github.com/googleapis/java-storage/blob/87bf7371b6c4300b0f306ca36d1918d52adf721b/.repo-metadata.json#L19) and others

This constitutes an effort to fit our library generation to the existing hand-written libraries.

### DIFF
--- a/library_generation/test/utilities_unit_tests.py
+++ b/library_generation/test/utilities_unit_tests.py
@@ -40,8 +40,8 @@ library_1 = LibraryConfig(
     library_name="bare-metal-solution",
     rest_documentation="https://cloud.google.com/bare-metal/docs/reference/rest",
     rpc_documentation="https://cloud.google.com/bare-metal/docs/reference/rpc",
-    recommended_package='com.google.example',
-    min_java_version=8
+    recommended_package="com.google.example",
+    min_java_version=8,
 )
 library_2 = LibraryConfig(
     api_shortname="secretmanager",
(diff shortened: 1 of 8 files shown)


library_generation/test/utilities_unit_tests.py
```diff
@@ -40,6 +40,8 @@
     library_name="bare-metal-solution",
     rest_documentation="https://cloud.google.com/bare-metal/docs/reference/rest",
     rpc_documentation="https://cloud.google.com/bare-metal/docs/reference/rpc",
+    recommended_package="com.google.example",
```
blakeli0, line 43: I see that this change is applied to three golden files, is it because this test `LibraryConfig` is used in three tests?

diegomarquezp: @blakeli0 that's right. There are several tests using that one Library config, hence the changes in other golden files

JoeWang1127 (approved):
Can you add these two params in readme and specification?

library_generation/test/utilities_unit_tests.py
blakeli0: OK, ideally we should not reuse test data between unit tests, but let's get this in first and refactor them later.

blakeli0 (approved):

PR #2932, opened 2024-06-20
chore: add iam logging

JoeWang1127:
In this PR:
- Add logging module in `google/iam/v1`.
- Regenerate iam.


lqiu96:
Do you need to add in iam/logging to the proto_path for generation_config.yaml? https://github.com/googleapis/sdk-platform-java/blob/f58754178d16e5054326d78d1d372efe1408fa5e/generation_config.yaml#L40-L43

Or is that only for protos with services?

JoeWang1127:
> Do you need to add in iam/logging to the proto_path for generation_config.yaml?
> 
> https://github.com/googleapis/sdk-platform-java/blob/f58754178d16e5054326d78d1d372efe1408fa5e/generation_config.yaml#L40-L43
> 
> Or is that only for protos with services/ GAPICs. Not too familiar with hermetic build, going to default to Blake's review.

No need to add a proto_path because logging is a nested directory in `google/iam/v1` and all protos inside the proto_path should be generated by default. 

I removed the special case in `generate_library.sh` so that this falls to the default case.

java-iam/.repo-metadata.json
```diff
@@ -5,7 +5,7 @@
   "api_description": "Manages access control for Google Cloud Platform resources",
   "client_documentation": "https://cloud.google.com/java/docs/reference/proto-google-iam-v1/latest/overview",
   "release_level": "stable",
-  "transport": "grpc",
+  "transport": "both",
```
blakeli0, line 8: Do we know why the transport changed? Is it because this PR https://github.com/googleapis/sdk-platform-java/pull/2890?

blakeli0 (approved):

java-iam/.repo-metadata.json
JoeWang1127: Yes, now `iam/v2` is used to get the transport.

PR #3016, opened 2024-07-05
fix: `BaseApiTracer` to noop on attemptFailed via overloaded method call

diegomarquezp:
Fixes https://github.com/googleapis/sdk-platform-java/issues/3015
Fixes https://github.com/googleapis/sdk-platform-java/issues/3014

Gax tracing internally works with `attemptFailedDuration`, which defaults to a no-op. Downstream libraries use `attemptFailed`, which has a custom behavior. What happens when an attempt-failed event occurs is that `attemptFailedDuration` is called instead (in favor of using java.time methods internally). This fix makes `attemptFailedDuration`'s behavior to delegate the logic to `attemptFailed`.

The downstreams will keep failing because the repos haven't got adapted to the new change in gax. See the follow ups below.

### Fixes in `java-spanner`
image

### Fixes in `java-bigtable`
image

### Follow ups in `java-bigtable`
More failures in java-bigtable to be addressed in that repo:
```
Error:    BigtableTableAdminSettingsTest.testToString:175 expected to contain: totalTimeout=PT13H32M
but was            : BigtableTableAdminSettings{projectId=our-project-85, instanceId=our-instance-06, 
...
```

Fixed in https://github.com/googleapis/java-bigtable/pull/2274

### Follow ups in `java-spanner`
```
Error:  Failures: 
Error:    CompositeTracerTest.testMethodsOverrideMetricsTracer:238 Method not found in compositeTracerMethods: public void com.google.api.gax.tracing.MetricsTracer.attemptFailedDuration(java.lang.Throwable,java.time.Duration)
```
Fixed in https://github.com/googleapis/java-spanner/pull/3200

PR #3037, opened 2024-07-09
chore: move the `attemptFailed` logic from BaseApiTracer to ApiTracer

diegomarquezp:
Follow up of https://github.com/googleapis/sdk-platform-java/pull/3016

On `bigtable` and `spanner`:
image
image

Thanks @blakeli0 and @lqiu96 for the directions.

### DIFF
--- a/gax-java/gax/src/main/java/com/google/api/gax/tracing/ApiTracer.java
+++ b/gax-java/gax/src/main/java/com/google/api/gax/tracing/ApiTracer.java
@@ -120,15 +120,14 @@ public interface ApiTracer {
   /**
    * Adds an annotation that the attempt failed, but another attempt will be made after the delay.
    *
-   * <p>Defaults to do the same as {@link #attemptFailed(Throwable, org.threeten.bp.Duration)}. This
-   * is because customers may have older/legacy code that directly implements {@link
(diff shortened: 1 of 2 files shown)


gax-java/gax/src/main/java/com/google/api/gax/tracing/ApiTracer.java
```diff
@@ -121,7 +123,10 @@ default Scope inScope() {
    * @param error the transient error that caused the attempt to fail.
    * @param delay the amount of time to wait before the next attempt will start.
    */
-  default void attemptFailedDuration(Throwable error, java.time.Duration delay) {};
+  default void attemptFailedDuration(Throwable error, java.time.Duration delay) {
+    // defaults to do the same as attemptFailed(Throwable, org.threeten.bp.Duration)
```
lqiu96, line 127: nit: Can you add a blurb to reflect the `why`? 

i.e. Customers may have older/legacy code that directly implements ApiTracer's `attemptFailed()` method and their overriden logic should be invoked in gax.

diegomarquezp: Good suggestion, thank you. I added a similar blurb in the body comment.

lqiu96 (approved):

PR #3040, opened 2024-07-10
chore: make generator version an optional param

JoeWang1127:
In this PR:
- Change `gapic_generator_version` as an optional parameter.
- Set `gapic_generator_version` through environment variable if it's not specified in the generation config.


blakeli0 (approved):
I was hoping we can skip calling the generator if the version is not provided, maybe reusing the `proto_only` flag, but it probably requires more effort to populate the value and pass the flag around. I think this approach if good for now.

.github/scripts/hermetic_library_generation.sh
```diff
@@ -104,6 +107,7 @@ docker run \
   -u "$(id -u):$(id -g)" \
   -v "$(pwd):${workspace_name}" \
   -v "$HOME"/.m2:/home/.m2 \
+  -e generator_version="${generator_version}" \
```
diegomarquezp, line 110: From our [shell style guide](https://engdoc.corp.google.com/eng/doc/devguide/shell/styleguide.md?cl=head#s7.3-constants-and-environment-variable-names): _Constants and anything exported to the environment should be capitalized, separated with underscores, and declared at the top of the file._

The _top of the file_ part doesn't apply, but I think it's reasonable to make it uppercase for consistency with other env vars we set up in the Dockerfile

blakeli0: Makes sense, do you mind updating it in place of Joe?

diegomarquezp: Done

diegomarquezp (approved):

PR #3041, opened 2024-07-10
ci: Test with JDK 21

lqiu96:


lqiu96:
Error in gradle: `> BUG! exception in phase 'semantic analysis' in source unit '_BuildScript_' Unsupported class file major version 65`

Might need to bump the gradle version to support JDK 21.

PR #2707, opened 2024-04-30
chore: bake protoc into the hermetic build docker image

diegomarquezp:
### Does the config work with renovate bot?
Yes. After a local run using `LOG_LEVEL=debug npx --yes --package renovate -- renovate --platform=local &> renovate.out`, we obtained a dependency update in the logs:
```
           {
             "deps": [
               {     
                 "depName": "protocolbuffers/protobuf",
                 "currentValue": "25.2",
                 "datasource": "github-releases",
                 "extractVersion": "^v(?<version>.*)$",
                 "replaceString": "ARG PROTOC_VERSION=25.2\n",
                 "updates": [
                   {     
                     "bucket": "non-major",
                     "newVersion": "25.3",
                     "newValue": "25.3",
                     "releaseTimestamp": "2024-02-15T23:20:43.000Z",
                     "newMajor": 25,
                     "newMinor": 3,
                     "updateType": "minor",
                     "branchName": "renovate/protocolbuffers-protobuf-25.x"
                   },    
                   {     
                     "bucket": "major",
                     "newVersion": "26.1",
                     "newValue": "26.1",
                     "releaseTimestamp": "2024-03-27T20:28:47.000Z",
                     "newMajor": 26,
                     "newMinor": 1,
                     "updateType": "major",
                     "branchName": "renovate/protocolbuffers-protobuf-26.x"
                   }
                 ],
                 "packageName": "protocolbuffers/protobuf",
                 "versioning": "semver-coerced",
                 "warnings": [],
                 "sourceUrl": "https://github.com/protocolbuffers/protobuf",
                 "registryUrl": "https://github.com",
                 "currentVersion": "25.2",
                 "currentVersionTimestamp": "2024-01-09T23:52:55.000Z",
                 "isSingleVersion": true, 
                 "fixedVersion": "25.2"
               }
             ],

```

However, we disabled the setting for now. This log entry confirms it.
`DEBUG: Dependency: protocolbuffers/protobuf, is disabled (repository=local)`

PR #3045, opened 2024-07-11
chore: bake the grpc plugin in the hermetic library generation image

diegomarquezp:
This PR bakes the gRPC plugin in the hermetic library generation image.
The approach is very similar to https://github.com/googleapis/sdk-platform-java/pull/2707
 * The docker image will have a configurable `ARG` for the gRPC plugin
 * The docker image will download the gRPC plugin at the version specified in the previous point
 * The docker image will set the variables DOCKER_GRPC_VERSION and DOCKER_GRPC_LOCATION
 * `generate_library.sh` will now be able to avoid downloading the gRPC plugin if the computed plugin version is the same as DOCKER_GRPC_VERSION (all this via `utilities.sh`).


library_generation/utils/utilities.sh
```diff
@@ -282,7 +315,7 @@ get_proto_path_from_preprocessed_sources() {
   pushd "${sources}" > /dev/null
   local proto_library=$(find . -maxdepth 1 -type d -name 'proto-*' | sed 's/\.\///')
   local found_libraries=$(echo "${proto_library}" | wc -l)
-  if [ -z ${proto_library} ]; then
+  if [[ -z ${proto_library} ]]; then
```
diegomarquezp, line 318: I noticed that the shell unit tests were expecting the binary brackets because `proto_library` could be expanded to a string with spaces.

```
+ '[' -z proto-2 proto-1 ']'
/usr/local/google/home/diegomarquezp/google/sdk-platform-java/library_generation/test/../utils/utilities.sh: line 318: [: proto-2: binary operator expected
```

library_generation/utils/utilities.sh
```diff
@@ -147,7 +170,15 @@ download_tools() {
     export protoc_path=$(download_protoc "${protoc_version}" "${os_architecture}")
   fi
 
-  download_grpc_plugin "${grpc_version}" "${os_architecture}"
+  # similar case with grpc
+  if [[ "${DOCKER_GRPC_VERSION}" == "${grpc_version}" ]]; then
+    # if the specified protoc_version matches the one baked in the docker
```
blakeli0, line 175:
```suggestion
    # if the specified grpc_version matches the one baked in the docker
```

library_generation/utils/utilities.sh
```diff
@@ -198,13 +229,15 @@ download_protoc() {
...
     # download protoc-gen-grpc-java plugin from Google maven central mirror.
     download_from \
-    "https://maven-central.storage-download.googleapis.com/maven2/io/grpc/protoc-gen-grpc-java/${grpc_version}/protoc-gen-grpc-java-${grpc_version}-${os_architecture}.exe" \
-    "protoc-gen-grpc-java-${grpc_version}-${os_architecture}.exe"
-    chmod +x "protoc-gen-grpc-java-${grpc_version}-${os_architecture}.exe"
+    "https://maven-central.storage-download.googleapis.com/maven2/io/grpc/protoc-gen-grpc-java/${grpc_version}/${grpc_filename}" \
+    "${grpc_filename}"
+    chmod +x "${grpc_filename}"
   fi
+  echo "$(pwd)/${grpc_filename}"
```
blakeli0, line 240: Is this a debug log?

.cloudbuild/library_generation/library_generation.Dockerfile
```diff
@@ -32,11 +34,19 @@ COPY library_generation /src
...
 RUN source /src/utils/utilities.sh \
-	&& download_protoc "${PROTOC_VERSION}" "linux-x86_64"
+	&& download_protoc "${PROTOC_VERSION}" "${OS_ARCHITECTURE}"
 # we indicate protoc is available in the container via env vars
 ENV DOCKER_PROTOC_LOCATION=/protoc
 ENV DOCKER_PROTOC_VERSION="${PROTOC_VERSION}"
 
+# install grpc
+WORKDIR /grpc
+RUN source /src/utils/utilities.sh \
```
blakeli0, line 44: `source /src/utils/utilities.sh` is not needed anymore. We should probably split
```
RUN source /src/utils/utilities.sh \
	&& download_protoc "${PROTOC_VERSION}" "${OS_ARCHITECTURE}"
```
to 
```
RUN source /src/utils/utilities.sh
RUN download_protoc "${PROTOC_VERSION}" "${OS_ARCHITECTURE}"
```
Then here we can just do 
```
RUN download_grpc_plugin "${GRPC_VERSION}" "${OS_ARCHITECTURE}"
```

.cloudbuild/library_generation/library_generation.Dockerfile
```diff
@@ -19,7 +19,9 @@ SHELL [ "/bin/bash", "-c" ]
 
 ARG OWLBOT_CLI_COMMITTISH=ac84fa5c423a0069bbce3d2d869c9730c8fdf550
 ARG PROTOC_VERSION=25.3
+ARG GRPC_VERSION=1.62.2
```
blakeli0, line 22: Can we configure this with renovate bot, and group it with other grpc update? e.g. There is an existing [PR](https://github.com/googleapis/sdk-platform-java/pull/3034) that updates gRPC to the latest.

blakeli0 (commented):
We can still pass `grpc_version` from `generation_config.yaml`, I guess that would still override the baked in grpc version and the default grpc version from the pom?

library_generation/utils/utilities.sh
diegomarquezp: No, it's the return statement of the function. Many functions in `utilities.sh` are made this way. The way we use is `variable=$(my_function "arg1" "arg2")`. The only possible return values for `return` are integers from 0 to 255 (and consumed via `$?`) so it's not useful.

.cloudbuild/library_generation/library_generation.Dockerfile
diegomarquezp: This is something I considered initially when working on the Dockerfile. The effects of `source` are contained in the `RUN` statement only, because it is translated into `/bin/sh -c "source script.sh"` (see [SO](https://stackoverflow.com/questions/20635472/using-the-run-instruction-in-a-dockerfile-with-source-does-not-work))

diegomarquezp:
> We can still pass `grpc_version` from `generation_config.yaml`, I guess that would still override the baked in grpc version and the default grpc version from the pom?

@blakeli0 Yes, if `--grpc_version` is not passed via the config yaml, then it will infer it via `utilities.sh/get_protoc_version`.

https://github.com/googleapis/sdk-platform-java/blob/0a95554b4a4ecb2d2b5ca9c968aac82ba27e9887/library_generation/generate_library.sh#L89-L91

https://github.com/googleapis/sdk-platform-java/blob/0a95554b4a4ecb2d2b5ca9c968aac82ba27e9887/library_generation/generate_composed_library.py#L124-L133

https://github.com/googleapis/sdk-platform-java/blob/0a95554b4a4ecb2d2b5ca9c968aac82ba27e9887/library_generation/utils/utilities.py#L31-L39

.cloudbuild/library_generation/library_generation.Dockerfile
diegomarquezp: Done

https://github.com/googleapis/sdk-platform-java/blob/9795e23af1aeab77d6c9557000c311e21f972978/renovate.json#L55-L67

diegomarquezp: Done

.cloudbuild/library_generation/library_generation.Dockerfile
blakeli0: Cool, that's good to know!

blakeli0 (approved):

PR #3051, opened 2024-07-16
fix: Generator to skip generation for empty services.

### LINKED ISSUE #2750: Generator should skip generation for empty services
A scenario came up with https://github.com/googleapis/google-cloud-java/pull/10802 where https://github.com/googleapis/googleapis/blob/master/google/cloud/visionai/v1/prediction.proto contains an empty service (e.g. no RPCs). 

There tests fail when trying to compile the generated library containing this proto which means the generator is generating some classes for this empty service which are not compilable. The generator instead should skip generation for an empty service.

zhumin8:
fixes #2750
Skip parsing a service if no RPCs found for. 
In the scenario that only one service with no RPCs or all services have no RPCs, falls back to #2460

This pr also reverts a change brought by #985, and removes the relevant tests. For more context, this has been discussed  [here](https://github.com/googleapis/sdk-platform-java/issues/2750#issuecomment-2231737360).


gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -435,6 +435,17 @@ public static List<Service> parseService(
...
 
     return fileDescriptor.getServices().stream()
+        .filter(
+            serviceDescriptor -> {
+              List<MethodDescriptor> methodsList = serviceDescriptor.getMethods();
+              if (methodsList.isEmpty()) {
+                LOGGER.warning(
+                    String.format(
+                        "No rpc methods specified in %s, generation will be skipped.",
+                        serviceDescriptor.getName()));
```
lqiu96, line 445:
```suggestion
                        "Service %s has no RPC methods and will not be generated",
                        serviceDescriptor.getName()));
```

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpcrest/ServiceClientClassComposerTest.java
```diff
@@ -40,20 +40,6 @@ void generateServiceClasses() {
...
-    GapicContext context = GrpcRestTestProtoLoader.instance().parseShowcaseEcho();
-    Service echoProtoService = context.services().get(1);
-    GapicClass clazz = ServiceClientClassComposer.instance().generate(context, echoProtoService);
-
-    JavaWriterVisitor visitor = new JavaWriterVisitor();
-    clazz.classDefinition().accept(visitor);
-    GoldenFileWriter.saveCodegenToFile(this.getClass(), "EchoEmpty.golden", visitor.write());
-    Path goldenFilePath =
-        Paths.get(GoldenFileWriter.getGoldenDir(this.getClass()), "EchoEmpty.golden");
-    Assert.assertCodeEquals(goldenFilePath, visitor.write());
```
lqiu96, line 54: Do you think it would make sense to keep this test, but test that `EchoEmpty.golden` can't be found/ does not exist when running through the unit golden tests?

zhumin8: Problem is it won't go that far. This test would break at `Service echoProtoService = context.services().get(1);` because the second empty service is not parsed. So I don't think this test provides much additional value than the [ParserTest.java](https://github.com/googleapis/sdk-platform-java/pull/3051/files/dde16590263afe4bbe0e8923e98ec2e3a5e822ad#diff-366e75522a5cef0bc451f8b50c1f50a0924de66cf7d5affe11cdd9ad4ffc889c)

lqiu96: I see, since parseShowcaseEcho() calls `Parser.parseServices()` anyways. If this service isn't being used (https://github.com/googleapis/sdk-platform-java/blob/3b1a91551ab6bbaf6a46950e1677c15cdd70d2e9/gapic-generator-java/src/test/proto/echo_grpcrest.proto#L166-L169), perhaps we can remove it as well then? Assuming that nothing else breaks/ isn't being used anywhere else.

lqiu96 (approved):

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpcrest/ServiceClientClassComposerTest.java
zhumin8: Good point.

PR #3052, opened 2024-07-17
feat: add `transport` option to `generation_config.yaml`

diegomarquezp:
This PR adds a new optional entry "`transport`" to `generation_config.yaml`.

The `transport` entry will only have effect in the postprocessing output (i.e. generated libraries will still have its transport inferred via BUILD.bazel). The main output file affected by this is `.repo-metadata.json` and its derived files (e.g. README.md).

This addresses the [need to allow a custom transport for java-storage](https://github.com/googleapis/java-storage/pull/2619#discussion_r1662963790).

### DIFF
--- a/library_generation/README.md
+++ b/library_generation/README.md
@@ -105,29 +105,31 @@ The library level parameters define how to generate a (multi-versions) GAPIC
 library.
 They are shared by all GAPICs of a library.
 
-| Name                  | Required | Notes                                                                              |
-|:----------------------|:--------:|:-----------------------------------------------------------------------------------|
-| api_shortname         |   Yes    |                                                                                    |
-| api_description       |   Yes    |                                                                                    |
-| name_pretty           |   Yes    |                                                                                    |
-| product_docs          |   Yes    |                                                                                    |
-| library_type          |    No    | `GAPIC_AUTO` if not specified                                                      |
-| release_level         |    No    | `preview` if not specified                                                         |
-| api_id                |    No    | `{api_shortname}.googleapis.com` if not specified                                  |
-| api_reference         |    No    |                                                                                    |
-| codeowner_team        |    No    |                                                                                    |
-| client_documentation  |    No    |                                                                                    |
-| distribution_name     |    No    | `{group_id}:google-{cloud_prefix}{library_name}` if not specified                  |
-| excluded_poms         |    No    |                                                                                    |
-| excluded_dependencies |    No    |                                                                                    |
-| googleapis_commitish  |    No    | use repository level `googleapis_commitish` if not specified.                      |
-| group_id              |    No    | `com.google.cloud` if not specified                                                |
-| issue_tracker         |    No    |                                                                                    |
-| library_name          |    No    | `api_shortname` is not specified. This value should be unique among all libraries. |
-| rest_documentation    |    No    |                                                                                    |
-| rpc_documentation     |    No    |                                                                                    |
-| cloud_api             |    No    | `true` if not specified                                                            |
-| requires-billing      |    No    | `true` if not specified                                                            |
+| Name                  | Required | Notes                                                                                                           |
+|:----------------------|:--------:|:----------------------------------------------------------------------------------------------------------------|
+| api_shortname         |   Yes    |                                                                                                                 |
+| api_description       |   Yes    |                                                                                                                 |
+| name_pretty           |   Yes    |                                                                                                                 |
+| product_docs          |   Yes    |                                                                                                                 |
+| library_type          |    No    | `GAPIC_AUTO` if not specified                                                                                   |
+| release_level         |    No    | `preview` if not specified                                                                                      |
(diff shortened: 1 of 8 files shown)


library_generation/README.md
```diff
@@ -105,29 +105,31 @@ The library level parameters define how to generate a (multi-versions) GAPIC
...
+| issue_tracker         |    No    |                                                                                                                 |
+| library_name          |    No    | `api_shortname` is not specified. This value should be unique among all libraries.                              |
+| rest_documentation    |    No    |                                                                                                                 |
+| rpc_documentation     |    No    |                                                                                                                 |
+| cloud_api             |    No    | `true` if not specified                                                                                         |
+| requires-billing      |    No    | `true` if not specified                                                                                         |
+| transport             |    No    | must be one of `grpc`, `rest`, `http` (alias) or `both`. This will override the value obtained from BUILD.bazel |
```
blakeli0, line 131: I think we use `transport` from BUILD.bazel to determine if we should generate a grpc module or not, can you please confirm that? If yes, we should update the description to something like `This value would only be used for generating .repo-metadata.json and relevant sections in README`.

library_generation/utils/utilities.py
```diff
@@ -198,11 +199,14 @@ def generate_postprocessing_prerequisite_files(
     :param library: the library configuration
     :param proto_path: the path from the root of googleapis to the location of the service
     protos. If the path contains a version, it will be removed
-    :param transport: transport supported by the library
+    :param gapic_inputs: gapic_inputs obtained from the library's BUILD.bazel
     :param library_path: the path to which the generated file goes
     :param language: programming language of the library
     :return: None
     """
+    transport = (
```
blakeli0, line 207: Can we move this logic to `generate_composed_library.py`, right before calling the `utilities.py`, for simplicity?

library_generation/utils/utilities.py
```diff
@@ -224,7 +228,10 @@ def generate_postprocessing_prerequisite_files(
     # is one of grpc, http and both,
     if transport == "grpc":
         converted_transport = "grpc"
-    elif transport == "rest":
+    elif transport in [
```
blakeli0, line 231: I guess this is because we use `rest` in BUILD but `http` in `repo-metadata.json`? This is convenient but I feel we should just stick with `rest` for consistency and simplicity.

library_generation/utils/utilities.py
diegomarquezp: I initially did it that way, but then I found that it would be better for testing if we leave the `transport` inference logic to `generate_postprocessing_prerequisite_files`.

This test helper passes a hard-coded value:
https://github.com/googleapis/sdk-platform-java/blob/e46648f112a149f967783539d30b4c44474b39fe/library_generation/test/utilities_unit_tests.py#L279-L286

And we don't have unit tests for `generate_composed_library`, so the only way to have a realistic execution path (other than the integration test) is by moving this logic to this method.
Do you think it's enough to make an exception?

library_generation/README.md
diegomarquezp: Done

library_generation/utils/utilities.py
diegomarquezp: Done.

library_generation/generate_composed_library.py
```diff
@@ -72,13 +72,16 @@ def generate_composed_library(
...
+        # overriden by the config object if specified. Note that this override
+        # does not affect library generation but instead used only for
+        # generating postprocessing files such as README.
         util.generate_postprocessing_prerequisite_files(
             config=config,
             library=library,
             proto_path=util.remove_version_from(gapic.proto_path),
-            transport=gapic_inputs.transport,
             library_path=library_path,
+            gapic_inputs=gapic_inputs,
```
diegomarquezp, line 84: from @blakeli0: let's encapsulate this logic in a function of `model.LibraryConfig` so it's easier to test

library_generation/utils/utilities.py
diegomarquezp: Resolution from discussion with @blakeli0: let's move the logic to the model class

diegomarquezp: Done

library_generation/model/library_config.py
```diff
@@ -101,6 +104,13 @@ def get_artifact_id(self) -> str:
         """
         return self.get_maven_coordinate().split(MAVEN_COORDINATE_SEPARATOR)[-1]
 
+    def get_transport(self, gapic_inputs: GapicInputs) -> str:
```
blakeli0, line 107: Nit: Add a comment mentioning that the value returned from this method is only used for generating post-processing prerequisite files.

blakeli0 (approved):

library_generation/model/library_config.py
diegomarquezp: Done

PR #3056, opened 2024-07-23
feat: add full RetrySettings sample code to Settings classes

### LINKED ISSUE #2596: Generate snippets for RetrySettings
Currently the generate TotalTimeout snippets(e.g. in [Showcase](https://github.com/googleapis/sdk-platform-java/blob/main/showcase/gapic-showcase/src/main/java/com/google/showcase/v1beta1/ComplianceSettings.java#L71-L81)) is misleading for customers per b/328302469. We should generate snippets for the whole `RetrySettings` instead of just `TotalTimeout` field reduce confusion.

alicejli:
Fixes #2596 ☕️

Here's what the `SpeechSettings` c.g.c. snapshot looks like:
Screenshot from 2024-07-23 14-45-22


lqiu96:
I think the changes look good. You probably need to update showcase as well since it's complaining.

Additionally, is there a way to tell what kind of method the RPC is (specifically if it's an LRO RPC or not)? I think it may be beneficial to give the LROs different sample values. But if it's too hard to do that, then we can probably ignore it since it's not important.

alicejli:
> I think the changes look good. You probably need to update showcase as well since it's complaining.
>  ahh good point thank you - updated.
> Additionally, is there a way to tell what kind of method the RPC is (specifically if it's an LRO RPC or not)? I think it may be beneficial to give the LROs different sample values. But if it's too hard to do that, then we can probably ignore it since it's not important/ not that much benefit.

I think I can do this - do you have a set of sample Retry values for LROs that I can try to pop in?

lqiu96:
> > Additionally, is there a way to tell what kind of method the RPC is (specifically if it's an LRO RPC or not)? I think it may be beneficial to give the LROs different sample values. But if it's too hard to do that, then we can probably ignore it since it's not important/ not that much benefit.
> 
> I think I can do this - do you have a set of sample Retry values for LROs that I can try to pop in?

Let's try with these default values: https://github.com/googleapis/sdk-platform-java/blob/43de5b568dc0bbdaddf2419652e0cb16de77ddb6/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/RetrySettingsComposer.java#L68-L73

For LROs, we only need those four settings: https://github.com/googleapis/google-cloud-java/tree/main?tab=readme-ov-file#configuring-lro-timeouts

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/ApiVersionTestingStubSettings.golden
```diff
@@ -57,10 +59,21 @@ import javax.annotation.Generated;
  *             .echoWithVersionMethodSettings()
  *             .getRetrySettings()
  *             .toBuilder()
- *             .setTotalTimeout(Duration.ofSeconds(30))
+ *             .setInitialRetryDelayDuration(Duration.ofSeconds(1))
```
blakeli0, line 62: Good to see that we are using the new java.time methods! cc: @diegomarquezp

alicejli:
@lqiu96 I added logic to set different RetrySettings if the methods are LRO, but none of the goldens are using it, as there's currently logic in (`AbstractServiceSettingsComposer`)[https://github.com/googleapis/sdk-platform-java/blob/main/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceSettingsClassComposer.java#L136] to pick the first unary method as the example method.

I did update the unit tests so you can see what the LRO example would look like, but I'm guessing it would be very unlikely that there is a service that only has LROs.

lqiu96:
> @lqiu96 I added logic to set different RetrySettings if the methods are LRO, but none of the goldens are using it, as there's currently logic in (`AbstractServiceSettingsComposer`)[https://github.com/googleapis/sdk-platform-java/blob/main/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceSettingsClassComposer.java#L136] to pick the first unary method as the example method.
> 
> I did update the unit tests so you can see what the LRO example would look like, but I'm guessing it would be very unlikely that there is a service that only has LROs.

What I'm thinking is if we can find an LRO example and add the LRO equivalent of 
```
 * EchoStubSettings.Builder echoSettingsBuilder = EchoStubSettings.newBuilder();
 * echoSettingsBuilder
 *     .echoSettings()
 *     .setRetrySettings(
 *         echoSettingsBuilder
 *             .echoSettings()
 *             .getRetrySettings()
 *             .toBuilder()
 *             .setInitialRetryDelayDuration(Duration.ofSeconds(1))
 *             .setInitialRpcTimeoutDuration(Duration.ofSeconds(5))
 *             .setMaxAttempts(5)
 *             .setMaxRetryDelayDuration(Duration.ofSeconds(30))
 *             .setMaxRpcTimeoutDuration(Duration.ofSeconds(60))
 *             .setRetryDelayMultiplier(1.3)
 *             .setRpcTimeoutMultiplier(1.5)
 *             .setTotalTimeoutDuration(Duration.ofSeconds(300))
 *             .build());
 * EchoStubSettings echoSettings = echoSettingsBuilder.build();
```

as an additional sample. I don't know how much additional work it is, but in my mind it should be fairly simple. Might need a bit of modifications to be something like:
```
 * EchoStubSettings.Builder echoSettingsBuilder = EchoStubSettings.newBuilder();
 * echoSettingsBuilder
 *     .echoSettings()
 *     .setPollingAlgorithm(...)
```
from https://github.com/googleapis/google-cloud-java/tree/main?tab=readme-ov-file#configuring-lro-timeouts

If too much work, let's just create an issue and I'm fine with the above.

lqiu96 (approved):

PR #3067, opened 2024-07-24
feat: bake gapic-generator-java into the hermetic build docker image

diegomarquezp:
Makes the gapic-generator-java jar a prepared binary in the Docker image whose location is now assumed by the scripts. Developers need now to prepare this well-know location (specified in `library_generation/DEVELOPMENT.md`.


.cloudbuild/library_generation/library_generation.Dockerfile
```diff
@@ -12,11 +12,25 @@
...
+FROM gcr.io/cloud-devrel-public-resources/java21 AS ggj-build
+
+WORKDIR /sdk-platform-java
+COPY . .
+# {x-version-update-start:gapic-generator-java:current}
+ENV DOCKER_GAPIC_GENERATOR_VERSION="2.42.1-SNAPSHOT" 
+
+RUN mvn install -DskipTests -Dclirr.skip -Dcheckstyle.skip
+RUN ls "/root/.m2/repository/com/google/api/gapic-generator-java/" 
+RUN cp "/root/.m2/repository/com/google/api/gapic-generator-java/${DOCKER_GAPIC_GENERATOR_VERSION}/gapic-generator-java-${DOCKER_GAPIC_GENERATOR_VERSION}.jar" .
```
diegomarquezp, line 26: Only copying the jar allows us for a more precise test. We can use the env vars `DOCKER_GAPIC_GENERATOR_LOCATION` and `DOCKER_GAPIC_GENERATOR_VERSION` to confirm if the preconfigured generator was used

.github/scripts/hermetic_library_generation.sh
```diff
@@ -90,9 +90,6 @@ fi
 git show "${target_branch}":"${generation_config}" > "${baseline_generation_config}"
 config_diff=$(diff "${generation_config}" "${baseline_generation_config}" || true)
 
-generator_version=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout -pl gapic-generator-java)
```
diegomarquezp, line 93: there is a `docker build` call right after this deleted section. That image will contain the SNAPSHOT jar (and the corresponding env var)

.cloudbuild/library_generation/library_generation.Dockerfile
blakeli0: Does this jar have all the info we need? Can you test running the image without internet? Sometimes the dependencies of the jar are still downloaded.

.cloudbuild/library_generation/library_generation.Dockerfile
```diff
@@ -47,7 +61,16 @@ RUN source /src/utils/utilities.sh \
 ENV DOCKER_GRPC_LOCATION="/grpc/protoc-gen-grpc-java-${GRPC_VERSION}-${OS_ARCHITECTURE}.exe"
 ENV DOCKER_GRPC_VERSION="${GRPC_VERSION}"
 
-# use python 3.11 (the base image has several python versions; here we define the default one)
+
+# we transfer gapic-generator-java from the previous stage.
+# here we redeclare this env var since they are not preserved between stages
+ENV DOCKER_GAPIC_GENERATOR_VERSION="2.43.1-SNAPSHOT"
```
blakeli0, line 67: Can we remove this version? Since we know the location from the previous stage, can we just use whatever jar is there?

library_generation/utils/utilities.sh
```diff
@@ -179,6 +178,19 @@ download_tools() {
     export grpc_path=$(download_grpc_plugin "${grpc_version}" "${os_architecture}")
   fi
 
+  # similar case with gapic-generator-java
+  if [[ "${DOCKER_GAPIC_GENERATOR_VERSION}" == "${gapic_generator_version}" ]]; then
+    # if the specified gapic_generator_version matches the one baked in the docker
+    # container, we copy the generator jar into the output folder so it can be
+    # picked up by gapic-generator-java-wrapper
+    >&2 echo "Using gapic-generator-java version baked into the container: ${DOCKER_GAPIC_GENERATOR_VERSION}"
+    cp "${DOCKER_GAPIC_GENERATOR_LOCATION}" "${output_folder}"
```
blakeli0, line 187: I think this copy logic might already be covered [here](https://github.com/googleapis/sdk-platform-java/blob/6e20135d9777c221ee5c116b093a9c338ad24156/library_generation/utils/utilities.sh#L204-L205)?

library_generation/model/generation_config.py
```diff
@@ -25,7 +25,7 @@
 COMMON_PROTOS_LIBRARY_NAME = "common-protos"
 GAPIC_GENERATOR_VERSION = "gapic_generator_version"
 LIBRARIES_BOM_VERSION = "libraries_bom_version"
-GENERATOR_VERSION_ENV_KEY = "GENERATOR_VERSION"
+GENERATOR_VERSION_ENV_KEY = "DOCKER_GAPIC_GENERATOR_VERSION"
```
blakeli0, line 28: Now that we have a `DOCKER_GAPIC_GENERATOR_VERSION` specified in the docker file, I think we can remove the usage of `GENERATOR_VERSION` now.

library_generation/utils/utilities.sh
```diff
@@ -158,7 +158,6 @@ download_tools() {
   local grpc_version=$3
   local os_architecture=$4
   pushd "${output_folder}"
-  download_generator_artifact "${gapic_generator_version}" "gapic-generator-java-${gapic_generator_version}.jar"
```
blakeli0, line 161: Can you please update the comment above as well?

library_generation/utils/utilities.sh
diegomarquezp: I decided to not follow the `.m2` folder structure in favor of easier testing (https://github.com/googleapis/sdk-platform-java/pull/3067/files#r1698944411). I think it's ok to treat this baked-in jar as a special case outside of `.m2`?

.cloudbuild/library_generation/library_generation.Dockerfile
diegomarquezp: I ran `docker run --network none` and could get past `generate_library.sh` (i.e. no jars needed), but I found two things we may want to follow up with:

This function from the transferred synthtool fetches data from Maven:
https://github.com/googleapis/sdk-platform-java/blob/0ef66196744e21fc3b4e2553d817c0dea5efe127/library_generation/owlbot/synthtool/languages/java.py#L127

Our script pulls googleapis' proto definitions:
https://github.com/googleapis/sdk-platform-java/blob/0ef66196744e21fc3b4e2553d817c0dea5efe127/library_generation/utils/utilities.py#L175-L182


I thought there was a commit in synthtool that fetches libraries_bom from our config yaml?
https://github.com/googleapis/sdk-platform-java/blob/0ef66196744e21fc3b4e2553d817c0dea5efe127/library_generation/owlbot/synthtool/languages/java.py#L280-L287

For the googleapis download, I guess it's expected behavior

diegomarquezp: I did a temporary bypass (not committed) to see if the generation is successful. It is except for the `mvn fmt:format` part that needs to fetch data when processing a library. The data includes fetching the parent pom of the target library. Nothing about sdk-platform is fetched.

https://paste.googleplex.com/5092731556986880?raw

library_generation/model/generation_config.py
diegomarquezp: I thought I removed all of the references to the GENERATOR_VERSION env var but there was a missing one in `.github/scripts/hermetic_library_generation`. I also renamed the variable in line 26 to make it clear that it's a reference to the config yaml key

.cloudbuild/library_generation/library_generation.Dockerfile
diegomarquezp: https://github.com/googleapis/sdk-platform-java/blob/a5c18b2920bfd349119ed95a75da785d1fa3f23a/library_generation/utils/utilities.sh#L182-L187

library_generation/utils/utilities.sh
diegomarquezp: Thank you, done.

.cloudbuild/library_generation/library_generation.Dockerfile
blakeli0: Sounds good! Yes we should definitely but I don't think they should block this PR, since we already have all the jars in the docker image.

.cloudbuild/library_generation/library_generation.Dockerfile
blakeli0: I see. For protoc, we tried to make it more flexible since we do have the need to test different protoc versions frequently. I don't think we have this need for gapic-generator-java. Hence I think it's better to make the implementation simple. It could be a separate PR also.

.github/scripts/hermetic_library_generation.sh
```diff
@@ -107,7 +104,6 @@ docker run \
   -u "$(id -u):$(id -g)" \
   -v "$(pwd):${workspace_name}" \
   -v "$HOME"/.m2:/home/.m2 \
-  -e GENERATOR_VERSION="${generator_version}" \
```
blakeli0, line 110: I think we can remove the `.m2` volume above also?

library_generation/model/generation_config.py
blakeli0: Thanks, this is much clearer now! As I mentioned in another comment, maybe we can try to get rid of `GAPIC_GENERATOR_VERSION_CONFIG_KEY` all together.

library_generation/utils/utilities.sh
blakeli0: I see. Yes it's definitely OK to put it into a different location, as long as we don't use maven to retrieve it. Let's see if we can get rid of `gapic_generator_version` so that we don't need this whole block.

.cloudbuild/library_generation/library_generation.Dockerfile
diegomarquezp: ok, I removed gapic_generator_version from the scripts. We will now always assume `DOCKER_GAPIC_GENERATOR_LOCATION` has the jar we need. We don't care about the generator version anymore.

.github/scripts/hermetic_library_generation.sh
diegomarquezp: Done

library_generation/model/generation_config.py
diegomarquezp: Got rid of it

diegomarquezp:
@blakeli0 I removed `gapic_generator_version` altogether. We now only rely on `DOCKER_GAPIC_GENERATOR_LOCATION` that will point to the jar we need and not care about the location.

This however brought many changes in the test folder but most of them are simple.

.cloudbuild/library_generation/library_generation.Dockerfile
```diff
@@ -47,7 +63,13 @@ RUN source /src/utils/utilities.sh \
 ENV DOCKER_GRPC_LOCATION="/grpc/protoc-gen-grpc-java-${GRPC_VERSION}-${OS_ARCHITECTURE}.exe"
 ENV DOCKER_GRPC_VERSION="${GRPC_VERSION}"
 
-# use python 3.11 (the base image has several python versions; here we define the default one)
+
+# we transfer gapic-generator-java from the previous stage.
+ENV DOCKER_GAPIC_GENERATOR_LOCATION="/gapic-generator-java/gapic-generator-java.jar"
```
blakeli0, line 68: Do we need this location to be an env variable? Can this be inlined?

library_generation/utils/utilities.sh
```diff
@@ -179,30 +165,19 @@ download_tools() {
...
+  # We will assume the generator location environment variable will point to the
+  # location of the generator JAR and then we will simply copy it. Although
+  # these env vars are already validated upstream in the python scripts, we
+  # still validate them here for testing purposes
+  if [[ -z "${DOCKER_GAPIC_GENERATOR_LOCATION}" ]]; then
+    >&2 echo "Env var DOCKER_GAPIC_GENERATOR_LOCATION must be set and pointing to the generator jar"
+    exit 1
+  else
+    >&2 echo "Using gapic-generator-java version baked into the container"
+    cp "${DOCKER_GAPIC_GENERATOR_LOCATION}" "${output_folder}"
```
blakeli0, line 178: Can we just use the jar as it is without copying? Is it for `gapic-generator-java-wrapper`?

library_generation/test/resources/goldens/pr_description-golden.txt
```diff
@@ -2,9 +2,6 @@ This pull request is generated with proto changes between [googleapis/googleapis
 
 BEGIN_COMMIT_OVERRIDE
 BEGIN_NESTED_COMMIT
-fix(deps): update the Java code generator (gapic-generator-java) to 1.2.3
```
blakeli0, line 5: I think this is a feature we forgot, but we definitely still need it. There are three follow up changes that I can see

- Change [hermetic_library_generation.sh](https://github.com/googleapis/java-spanner/blob/main/.github/scripts/hermetic_library_generation.sh#L82) so that it does not parse from generation_config.yaml anymore.
- Change [update_generation_config.sh](https://github.com/googleapis/java-spanner/blob/main/.github/scripts/update_generation_config.sh#L97-L99) so that it automatically updates the docker image versions. 
- Change the commit generation logic so that it can detect the above version change and generate this commit message correctly. 

These changes can be in separate PRs but we need to make sure the existing automations are not breaking.

library_generation/utils/utilities.sh
diegomarquezp: The wrapper assumes the jar to be in CWD

https://github.com/googleapis/sdk-platform-java/blob/58da069a99a7240cbec18b7dbdb010fb74c4d551/library_generation/gapic-generator-java-wrapper#L6

I'm not sure if we can both avoid copying the jar and at the same time not have a `DOCKER_GAPIC_GENERATOR_LOCATION` variable. We need a way to know where the jar is and make it compatible with local development as well.

Maybe we can copy the jar directly into the home `.m2` folder? We would still be copying it just like we've been doing so far.

.cloudbuild/library_generation/library_generation.Dockerfile
diegomarquezp: We could set the jar into a well-known location. Developers would need to send their jar in there beforehand. 

I think this could solve the need for an env var and copying the jar?

In favor of the env var, I think it helps us separate the logic of the Dockerfile when it decides where to put the jar vs our scripts finding out where the jar is. But I'm not sure if this outweights the need to depend on an env var?

Copying the jar is more about keeping the changes minimal, without involving how the jar wrapper behaves. However, I think the reason we copy the jar into the output folder is because of the way we call the wrapper. 

https://github.com/googleapis/sdk-platform-java/blob/58da069a99a7240cbec18b7dbdb010fb74c4d551/library_generation/generate_library.sh#L209

We can configure the wrapper to expect the jar to be in the well-known location instead.

library_generation/test/resources/goldens/pr_description-golden.txt
blakeli0: Discussed offline that we are going to keep using `gapic_generator_version` in generation_config.yaml. The usage will be limited to parsing the version for ocker image and generate commit history, not used for generate code.

library_generation/test/generate_library_unit_tests.sh
```diff
@@ -5,7 +5,14 @@ set -xeo pipefail
 # Unit tests against ../utilities.sh
 script_dir=$(dirname "$(readlink -f "$0")")
 source "${script_dir}"/test_utilities.sh
-source "${script_dir}"/../utils/utilities.sh
+
+# we simulate a properly prepared environemnt (i.e. generator jar) in its
+# well-known location. Tests confirming the opposite case will make sure this
```
diegomarquezp, line 10:
```suggestion
# we simulate a properly prepared environment (i.e. generator jar in its
# well-known location). Tests confirming the opposite case will make sure this
```

diegomarquezp:
@JoeWang1127 @blakeli0 the integration tests now can't be locked to a specific generator version, so we will constantly need to update them. 

A couple ideas:
 * Add an `--update` flag to the integration tests
 * Keep a locked-in jar in the test resources folder

diegomarquezp:
[spring downstream check failing](https://github.com/googleapis/sdk-platform-java/actions/runs/10625302690/job/29455529366) because the release job still doesn't publish the maven artifacts

https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/3138#issuecomment-2318598130

image


cc: @mpeddada1

.cloudbuild/library_generation/library_generation.Dockerfile
diegomarquezp: We set the jar into a well-known location. No env vars used

library_generation/test/resources/goldens/pr_description-golden.txt
diegomarquezp: Done. Restored pr description generation.

.cloudbuild/library_generation/library_generation.Dockerfile
```diff
@@ -12,11 +12,26 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# install gapic-generator-java in a separate layer so we don't overload the image
+# with the transferred source code and jars
+FROM gcr.io/cloud-devrel-public-resources/java21 AS ggj-build
```
JoeWang1127, line 17: Could we use image in Artifact Registry?

For example, https://airlock.corp.goog/package?type=Docker&id=088032b3-3669-43c0-9e82-02871abf2f0e

.cloudbuild/library_generation/library_generation.Dockerfile
```diff
@@ -12,11 +12,26 @@
...
+# {x-version-update-start:gapic-generator-java:current}
+ENV DOCKER_GAPIC_GENERATOR_VERSION="2.44.1-SNAPSHOT" 
+# {x-version-update-end:gapic-generator-java:current}
+
+RUN mvn install -DskipTests -Dclirr.skip -Dcheckstyle.skip
+RUN cp "/root/.m2/repository/com/google/api/gapic-generator-java/${DOCKER_GAPIC_GENERATOR_VERSION}/gapic-generator-java-${DOCKER_GAPIC_GENERATOR_VERSION}.jar" \
+  "./gapic-generator-java.jar"
+
 # build from the root of this repo:
 FROM gcr.io/cloud-devrel-public-resources/python
```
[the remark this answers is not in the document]
JoeWang1127: Maybe as a follow-up.

library_generation/DEVELOPMENT.md
```diff
@@ -48,32 +48,36 @@ Although the scripts are designed to be run in a Docker container, you can also
...
 
-## Installing prerequisites
+## Assumptions made by the scripts
+### The Hermetic Build's well-known folder
+Located in `${HOME}/.library_generation`, this folder is assumed by the scripts
+to contain the generator JAR. Please note that this is a recent feature and only
+this jar is expected to be there. Soon enough, more binaries such as the gRPC
+and protoc plugins will also be stored here. Developers must make sure
+this folder is properly configured before running the scripts locally.
+Note that this relies on the `HOME` en var which is always
```
JoeWang1127, line 58: Do you mean `${HOME}/.library_generation` needs to manually created before running the image?

library_generation/DEVELOPMENT.md
```diff
@@ -48,32 +48,36 @@ Although the scripts are designed to be run in a Docker container, you can also
...
+Run `cd sdk-platform-java && mvn install -DskipTests -Dclirr.skip
+-Dcheckstyle.skip`. This will generate a jar located in
+`~/.m2/repository/com/google/api/gapic-generator-java/{version}/gapic-generator-java-{version}.jar`
 
-It requires python 3.x to be installed.
-You will need to specify a committish of the synthtool repo in order to have
-your generation results matching exactly what the docker image would produce.
-You can achieve this by inspecting `SYNTHTOOL_COMMITISH` in
-`.cloudbuild/library_generation/library_generation.Dockerfile`.
+Then `mv` the jar into the well-known location of the jar. The
```
JoeWang1127, line 68: When building an image, the generator jar is already put into the well-known location.

Why does a developer need to build and move the jar again?

library_generation/test/generate_library_unit_tests.sh
```diff
@@ -15,25 +22,21 @@ extract_folder_name_test() {
...
-  actual_version=$(get_grpc_version "2.24.0")
-  rm "gapic-generator-java-pom-parent-2.24.0.pom"
-  assertEquals "1.56.1" "${actual_version}"
-}
-
-get_grpc_version_failed_with_invalid_generator_version_test() {
-  local res=0
-  $(get_grpc_version "1.99.0") || res=$?
-  assertEquals 1 $((res))
+get_grpc_version_fails_with_no_env_var_test() {
```
JoeWang1127, line 25: Can we move the newly added shell test to `utilities_unit_tests.py` by using `util.sh_util()` method?

I think python unit tests are better maintained than shell tests. WDYT?

library_generation/utils/utilities.sh
```diff
@@ -149,16 +136,17 @@ get_protoc_version() {
 # DOCKER_GRPC_VERSION respectively, this function will instead set "protoc_path"
 # and "grpc_path" to DOCKER_PROTOC_PATH and DOCKER_GRPC_PATH respectively (no
 # download), since the docker image will have downloaded these tools beforehand.
+#
 # For the case of gapic-generator-java, no env var will be exported for the
 # upstream flow, but instead it will be assigned a default filename that will be
-# referenced by the file `library_generation/gapic-generator-java-wrapper`.
+# referenced by the file `library_generation/gapic-generator-java-wrapper`. Note
+# that we assume an env var DOCKER_GAPIC_GENERATOR_LOCATION to point to the jar
```
JoeWang1127, line 143: Could you remind me where to set `DOCKER_GAPIC_GENERATOR_LOCATION`?

I didn't find it in the development guide.

JoeWang1127:
> the integration tests now can't be locked to a specific generator version, so we will constantly need to update them

I think it's fine to test the snapshot version of the generator. 

Ideally, running the generation can produce the same file even if the generator updates, though some changes in the generator can produce different code and we can update the golden file accordingly, just like the showcase integration test.

At this point, I think it's fine. However, we need to improvement the integration test:
- Use golden file rather than branch in google-cloud-java
- Provide an easy way to update the golden file

@blakeli0 WDYT?

library_generation/DEVELOPMENT.md
diegomarquezp: They wouldn't need to prepare this jar when running the Docker image since the Dockerfile will take care of it. The thing is that this new section is meant for local development (it is under that section).

library_generation/DEVELOPMENT.md
diegomarquezp: No, this guide is part of the local development instructions. The Dockerfile will take care of preparing this aspect of the environment.

.cloudbuild/library_generation/library_generation.Dockerfile
diegomarquezp: I tried to scope this to baking the generator only. The next step in the milestone is to "cleanup" the Docker image, involving several improvements including using safer base images. Maybe we can leave it to the next task, WDYT?

.cloudbuild/library_generation/library_generation.Dockerfile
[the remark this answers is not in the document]
diegomarquezp: Yes, agreed it can be a follow up. I explained this is planned in our current milestone (https://github.com/googleapis/sdk-platform-java/pull/3067#discussion_r1739185627).

library_generation/test/generate_library_unit_tests.sh
diegomarquezp: I was under the impression that `genreate_library_units` is meant for `utilities.sh` and that `utilities_units.py` was for `utilities.py`. 

Could you please explain how are they `py` unit tests better maintained?

library_generation/utils/utilities.sh
diegomarquezp: Ah, this is wrong. I updated the comment to explain about the well-known location. Thanks for the catch!

diegomarquezp:
From discussion with @blakeli0 and @JoeWang1127, we can lock in a specific version string of the generator, download it in our integration test and then override the well-known location via volumes

.cloudbuild/library_generation/library_generation.Dockerfile
JoeWang1127: SGTM

diegomarquezp:
@JoeWang1127 I modified the integration test to lock-in a specific generator jar downloaded via `mvn`. PTAL when you have a chance :)

.kokoro/presubmit/downstream-compatibility-spring.sh
```diff
@@ -47,10 +47,16 @@ pushd spring-cloud-gcp/spring-cloud-generator
 
 
 # Generate showcase autoconfig
+pushd spring-cloud-generator
+# The script is not executable for non-owners. Here we manually chmod it.
+# TODO(diegomarquezp): remove this line after
+# https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/3183 is merged
+chmod 755 ./scripts/generate-showcase.sh
```
diegomarquezp, line 54: @blakeli0 @JoeWang1127  I fixed the spring check in this PR as well since I thought it's small enough. There is a TODO to be solved next week, once https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/3183 is merged

library_generation/DEVELOPMENT.md
JoeWang1127: Ah, got it.

JoeWang1127:
> @JoeWang1127 I modified the integration test to lock-in a specific generator jar downloaded via `mvn`. PTAL when you have a chance :)

Thanks, could you change the pr body since this is no longer a pending issue.

library_generation/test/generate_library_unit_tests.sh
JoeWang1127: > Could you please explain how are the py unit tests better maintained?

The shell script doesn't have (or we don't use) a testing framework so we need to implement our own testing utilities, e.g., assert, test counters, etc, in `test_utilities.sh`. Every time we want to add a test, we need to add the test function to the test list.

All these burden was introduced at the beginning stage of this project because we didn't use python at that time.

Right now we have a python language framework already setup and we should migrate all shell script (with tests) to python, although this is out of scope of this PR.

> I was under the impression that genreate_library_unit_tests.sh was meant for utilities.sh and that utilities_units.py was for utilities.py.

For newly added shell tests, I suggest to wrap the test with `util.sh_util()`.

I didn't notice the test is verifying `genreate_library.sh`. Create `genreate_library_unit_tests.py` and put the test there is better. WDYT?

library_generation/DEVELOPMENT.md
JoeWang1127, line 58: nit: one sentence per line.

library_generation/test/integration_tests.py
```diff
@@ -42,15 +44,24 @@
...
 baseline_config_name = "baseline_generation_config.yaml"
 current_config_name = "current_generation_config.yaml"
 
+# This variable is used to override the jar created by building the image
+# with our own downloaded jar in order to lock the integration test to use
+# a constant version specified in
+# library_generation/test/resources/integration/test_generator_coordinates
+# This allows us to decouple the generation workflow testing with what the
+# generator jar will actually generate.
+# See library_generation/DEVELOPMENT.MD
```
JoeWang1127, line 56:
```suggestion
# See library_generation/DEVELOPMENT.md
```

JoeWang1127: Could you specify which part of `development.md` you are refering?

diegomarquezp: Done.

library_generation/DEVELOPMENT.md
diegomarquezp: I found [this article](https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line) that confirms this is a good practice. I reformatted the file.

.kokoro/presubmit/downstream-compatibility-spring.sh
```diff
@@ -47,10 +47,16 @@ pushd spring-cloud-gcp/spring-cloud-generator
 
 
 # Generate showcase autoconfig
+pushd spring-cloud-generator
+# The script is not executable for non-owners. Here we manually chmod it.
+# TODO(diegomarquezp): remove this line after
+# https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/3183 is merged
```
diegomarquezp, line 53:
```suggestion
# https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/3183 is merged and released.
```

.cloudbuild/library_generation/library_generation.Dockerfile
```diff
@@ -47,7 +62,15 @@ RUN source /src/utils/utilities.sh \
 ENV DOCKER_GRPC_LOCATION="/grpc/protoc-gen-grpc-java-${GRPC_VERSION}-${OS_ARCHITECTURE}.exe"
 ENV DOCKER_GRPC_VERSION="${GRPC_VERSION}"
 
-# use python 3.11 (the base image has several python versions; here we define the default one)
+
+# Here we transfer gapic-generator-java from the previous stage.
+# Note that the destination is a well-known location that will be assumed at runtime
+# We hard-code the location string to avoid making it configurable (via ARG) as
+# well as to avoid it making it overridable at runtime (via ENV).
+COPY --from=ggj-build "/sdk-platform-java/gapic-generator-java.jar" "${HOME}/.library_generation/gapic-generator-java.jar"
```
JoeWang1127, line 70: A future improvement: only include necessary files in the final image.

diegomarquezp: Yes, we want to get rid of the source files (owlbot cli, sdk-platform, etc) by building in this stage then only copying the binaries. This is part of the image cleanup task.

library_generation/DEVELOPMENT.md
```diff
@@ -45,35 +44,43 @@ python -m unittest discover -s test/ -p "*unit_tests.py"
...
 (`library_generation/cli/entry_point.py`).
 
-## Installing prerequisites
+## Assumptions made by the scripts
+### The Hermetic Build's well-known folder
+Located in `${HOME}/.library_generation`, this folder is assumed by the scripts
+to contain the generator JAR. 
+Please note that this is a recent feature and only this jar is expected to be
+there. 
+Soon enough, more binaries such as the gRPC and protoc plugins will also be
```
blakeli0, line 57: I would prefer not mention this as this is not implemented yet. We may treat gRPC and protoc differently as well.

diegomarquezp: Got it, I removed this sentence.

library_generation/test/resources/integration/.gitignore
blakeli0, line 1: Can we move this to the `.gitignore` under the [root folder](https://github.com/googleapis/sdk-platform-java/blob/main/.gitignore#L22-L26)?

library_generation/test/integration_tests.py
```diff
@@ -247,6 +294,8 @@ def __run_entry_point_in_docker_container(
                 f"{repo_location}:/workspace/repo",
                 "-v",
                 f"{config_location}:/workspace/config",
+                "-v",
+                f"{config_dir}/{WELL_KNOWN_GENERATOR_JAR_FILENAME}:/home/.library_generation/{WELL_KNOWN_GENERATOR_JAR_FILENAME}",
```
blakeli0, line 298: I guess this would override the SNAPSHOT jar built in the previous step `docker build`?

diegomarquezp: Yes, exactly. This is just to ensure that the outcome of the test doesn't depend on changes to the generator code.

library_generation/test/resources/integration/.gitignore
diegomarquezp: Done

blakeli0 (approved):

JoeWang1127 (approved):

PR #3073, opened 2024-07-26
feat: update ErrorDetails to allow unpacking arbitrary messages

BenWhitehead:
Because the google.rpc.Status.details is a google.protobuf.Any, it can contain any message. ErrorDetails conveniently unpacks the standard status message types from google/rpc/error_details.proto but some services provide their own error details types.

This new method allows unpacking those custom messages without deserializing the entire google.rpc.Status another time.


blakeli0 (approved):
It makes sense, thanks Ben!

BenWhitehead:
Thanks for the review @blakeli0

PR #3075, opened 2024-07-30
fix: Generator callable generation is based on method type

lqiu96:
Fixes https://github.com/googleapis/sdk-platform-java/issues/3076


gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
```diff
@@ -460,59 +460,63 @@ private Map<String, VariableExpr> createCallableClassMembers(
...
+                    ConcreteReference.builder()
+                        .setClazz(OperationCallable.class)
+                        .setGenerics(
+                            Arrays.asList(
+                                protoMethod.inputType().reference(),
+                                protoMethod.lro().responseType().reference(),
+                                protoMethod.lro().metadataType().reference()))
+                        .build()))
+            .build());
+  }
```
lqiu96, line 518: Extract these out to helper methods to be used later in `createCallableInitExprs`. These were originally used in `createCallableClassMembers` to create a mapping to be used, but `createCallableInitExprs` not longer uses that output. 

I can't remove `createCallableClassMembers` method yet as that is still being used elsewhere and replacing it would be a much larger refactor/ effort

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
```diff
@@ -871,67 +884,116 @@ protected VariableExpr declareLongRunningClient() {
...
-  private Expr createCallableInitExpr(
+  // Can return multiple Exprs for a single RPC. Each of the Exprs will initialize a callable
+  // in the constructor. The possible combinations are Normal (Unary, Streaming, Batching) and
+  // either Operation or Paged (if needed). It is not possible to have three callable Exprs
+  // returned because LROs are not paged, so it will either be an additional LRO or paged callable.
+  private List<Expr> createCallableInitExprs(
       GapicContext context,
       Service service,
-      String callableVarName,
-      VariableExpr callableVarExpr,
```
lqiu96, line 878: `callableVarExpr` was the VariableExpr from `createCallableClassMembers` above. We no longer use the map from `createCallableClassMembers` and now generate the VariableExpr inside this method directly.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
zhumin8: I am a bit concerned that remaining `createCallableClassMembers` usages will cause us weird bugs in the future. Perhaps file a separate issue to backlog for refactor and move away from it?

blakeli0: I think `callableClassMemberVarExprs` itself is fine, since it already uses the `Method` to determine if the RPC is an LRO or not. The `Map<String, VariableExpr>` though, it is slightly concerning because if someone uses the key to determine the type of a callable again, then it would be a problem. Ideally we can have a `Callable` object to wrap both `name` and `type`, but I think that is OOS for now.

gapic-generator-java/src/test/proto/naming.proto
blakeli0, line 1: Can we rename this file to something more specific to the test case? Also can we have an RPC that is actually a LRO in the testing proto?

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/samples/servicesettings/stub/SyncGetApiOperation.golden
blakeli0, line 1: IIRC, this is something we have to include for every test since we have no way to tell the generator to not generate samples?

gapic-generator-java/src/test/resources/naming_grpc_service_config.json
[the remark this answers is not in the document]
lqiu96: Hmm odd, when I tried generating previously it was complaining that I was missing a service config json, so I coped one over. I just removed and it worked.

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/samples/servicesettings/stub/SyncGetApiOperation.golden
lqiu96: I don't think so for the AbstractTransportServiceStubClassComposer. I think this was an artifact from a previous run where I put this test in the ServiceStubSettings Composer. 

The region tag for this is goldensample_generated_NamingServiceStubSettings_GetApiOperation_sync
> NamingServiceStubSettings

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
lqiu96: > Perhaps file a separate issue to backlog for refactor and move away from it?

I'll create an issue in the backlog for this. Personally, I think should we move away from *any* usage of `Map<String, VariableExpr>` and just generate the VariableExpr whenever needed. From reading the logic, I think the reason the Map was added was to cache the VariableExpr as it's being used a few places.

> someone uses the key to determine the type of a callable again, then it would be a problem

Yep, the callable name shouldn't be the indicator of the RPC type. We don't know if there are any downstream usages of it, but I removed the constants and see no other usages. I think this was the only case where this type heuristic exists.

> Ideally we can have a Callable object to wrap both name and type, but I think that is OOS for now.

If we want to keep this mapping structure to cache the VariableExpr, I think `Map<Method, VariableExpr>` would work. Method already contains the name and type of the RPC.

If possible, I'd rather not have multiple functions take multiple `Map<String, VariableExpr>` parameters or the equivalent `Map<Method, VariableExpr>`:
1. https://github.com/googleapis/sdk-platform-java/blob/6913db5c64e35ec8925c25a04ebd6d529895c4fe/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java#L538-L540 
2. https://github.com/googleapis/sdk-platform-java/blob/6913db5c64e35ec8925c25a04ebd6d529895c4fe/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java#L648-L650

Refactoring something like this would be a much larger effort and much harder to get completely right. Perhaps multiple milestones/ steps could be made, but I don't know how feasible or time consuming it would be without a much thorough investigation.

lqiu96: > If we want to keep this mapping structure to cache the VariableExpr, I think Map<Method, VariableExpr> would work. Method already contains the name and type of the RPC.

Correction on this, it wouldn't work for callableClassMemberVarExprs as a Method could have multiple VariableExprs.

Looks like it would work for protoMethodNameToDescriptorVarExprs as it's a mapping of one Method -> one VariableExpr. And wouldn't matter for classMemberVarExprs as it turns out it's not a mapping of method -> variablexpr

blakeli0: > I think should we move away from any usage of Map<String, VariableExpr> and just generate the VariableExpr whenever needed

I agree, that's a good long term solution.

gapic-generator-java/src/test/proto/naming.proto
blakeli0: > I was looking to create a generic proto where I could throw in this and future tests cases for weird naming conventions

I think that's a good intention but that intention could easily get lost if the name is not clear. `naming.proto` is too generic just by looking at the name, I think we should at least name it something like `naming_conflict.proto`. 

In addition, unit tests should be quick and precise. Keep a running list is like adding test cases to a single test file, it works but not ideal. We may have different types of naming conflict in the future, but we don't know if it makes sense to put them in the same testing file. I would prefer to have a more specific proto in most cases. 

> The reason I didn't add an LRO RPC is because and LRO would always be generated as an LRO

Usually we want to test anything we touched. In this case, we should make sure the behavior is still the same as before for valid LROs. However, giving it more thought, that case is probably already tested somewhere else, probably in `echo.proto`. It is not ideal, as we should have a feature specific `LRO_operation_testing.proto` that is dedicated to LRO generation. But I think it's OK to not adding additional LRO testing for now.

lqiu96: Makes sense. I'll rename it to something like `naming_conflict.proto` since I can see how this can easily be overlooked/ get confusing in the future.

Agreed that unit tests should be simple and precise. I was looking for a way to only test `createCallableInitExprs`, but ending up having to recreate a service proto with new request messages and response messages + all the other configurations required to generate the java class and then link it to it set up to generate diffs, which is why I was looking to just have a single proto that could be re-used in every composer in the future.

LROs are indirectly tested with echo.proto and probably should have a more feature specific LRO.proto. I think this is the same case with Unary, Streaming, Paging, and Batching.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractTransportServiceStubClassComposer.java
```diff
@@ -871,67 +880,116 @@ protected VariableExpr declareLongRunningClient() {
     return null;
   }
 
-  private Expr createCallableInitExpr(
+  // Can return multiple Exprs for a single RPC. Each of the Exprs will initialize a callable
+  // in the constructor. The possible combinations are Normal (Unary, Streaming, Batching) and
+  // either Operation or Paged (if needed). It is not possible to have three callable Exprs
+  // returned because LROs are not paged, so it will either be an additional LRO or paged callable.
+  private List<Expr> createCallableInitExprs(
```
zhumin8, line 887: This comment provided helpful context for understanding. Thank you!

zhumin8 (approved):
LGTM.

blakeli0 (approved):

PR #3078, opened 2024-07-31
deps: add opentelemetry exporter-metrics and shared-resoucemapping to shared dependencies

JesseLovelace:
b/356675340


java-shared-dependencies/third-party-dependencies/pom.xml
```diff
@@ -37,6 +37,8 @@
     <perfmark-api.version>0.27.0</perfmark-api.version>
     <j2objc-annotations.version>3.0.0</j2objc-annotations.version>
     <google.cloud.opentelemetry.version>0.31.0</google.cloud.opentelemetry.version>
+    <opentelemetry-exporter-metrics.version>0.31.0</opentelemetry-exporter-metrics.version>
```
blakeli0, line 40: Do they just happen to have the same version or they will always have the same version? The `google.cloud.opentelemetry.version` above also has the same value 0.31.0. If they will always be the same, we don't have to introduce new properties.

lqiu96: @JesseLovelace bumping the above comment.

JesseLovelace: Fixed!

ldetmer (approved):

PR #3088, opened 2024-08-06
docs: Update the Javadoc of ObsoleteApi.java

blakeli0:
Update the Javadoc of ObsoleteApi.java. It's preferred to provide reasons and alternatives directly in the description, rather than links, as links are not stable.

### DIFF
--- a/api-common-java/src/main/java/com/google/api/core/ObsoleteApi.java
+++ b/api-common-java/src/main/java/com/google/api/core/ObsoleteApi.java
@@ -60,6 +60,6 @@ import java.lang.annotation.Target;
 })
 @Documented
 public @interface ObsoleteApi {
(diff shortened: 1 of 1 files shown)


zhumin8 (approved):
LGTM
vaguely related question, when do we use `@ObsoleteApi` vs. `@Deprecated`. Is `@ObsoleteApi` intended for APIs prior to deprecation, meaning it will be replaced with `@Deprecated` in the future?

lqiu96 (approved):

blakeli0:
> LGTM vaguely related question, when do we use `@ObsoleteApi` vs. `@Deprecated`. Is `@ObsoleteApi` intended for APIs prior to deprecation, meaning it will be replaced with `@Deprecated` in the future?

Yes, that's exactly the use case. In general, we don't want to introduce `@Deprecated` unless it is a major version update.

PR #3095, opened 2024-08-12
fix(gax): prevent truncation/overflow when converting time values

diegomarquezp:
https://github.com/googleapis/sdk-platform-java/pull/1872#discussion_r1695864350

### DIFF
--- a/gax-java/gax/src/test/java/com/google/api/gax/util/TimeConversionUtilsTest.java
+++ b/gax-java/gax/src/test/java/com/google/api/gax/util/TimeConversionUtilsTest.java
@@ -69,4 +69,44 @@ public class TimeConversionUtilsTest {
         jtInstant.toEpochMilli(), TimeConversionUtils.toThreetenInstant(jtInstant).toEpochMilli());
     assertNull(TimeConversionUtils.toThreetenInstant(null));
   }
+
+  @Test
+  void testToThreeteenInstant_bigInput_doesNotOverflow() {
+    // defaults to MAX_SECONDS plus the max value of long for the nanos part
+    java.time.Instant jtInstant = java.time.Instant.MAX;
+    org.threeten.bp.Instant ttInstant = TimeConversionUtils.toThreetenInstant(jtInstant);
+    assertEquals(jtInstant.getEpochSecond(), ttInstant.getEpochSecond());
+    assertEquals(jtInstant.getNano(), ttInstant.getNano());
+  }
+
+  @Test
+  void testToJavaTimeInstant_bigInput_doesNotOverflow() {
+    // defaults to MAX_SECONDS plus the max value of long for the nanos part
+    org.threeten.bp.Instant ttInstant = org.threeten.bp.Instant.MAX;
+    java.time.Instant jtInstant = TimeConversionUtils.toJavaTimeInstant(ttInstant);
+    assertEquals(jtInstant.getEpochSecond(), ttInstant.getEpochSecond());
+    assertEquals(jtInstant.getNano(), ttInstant.getNano());
+  }
+
+  @Test
+  void testToThreeteenDuration_bigInput_doesNotOverflow() {
+    // we use the max long value for the seconds part and an arbitrary int for the nanos part, so we
+    // can confirm
+    // that both components are preserved
+    java.time.Duration jtDuration = java.time.Duration.ofSeconds(Long.MAX_VALUE, 123);
+    org.threeten.bp.Duration ttDuration = TimeConversionUtils.toThreetenDuration(jtDuration);
+    assertEquals(jtDuration.getSeconds(), ttDuration.getSeconds());
+    assertEquals(jtDuration.getNano(), ttDuration.getNano());
+  }
+
+  @Test
+  void testToJavaTimeDuration_bigInput_doesNotOverflow() {
+    // we use the max long value for the seconds part and an arbitrary int for the nanos part, so we
+    // can confirm
+    // that both components are preserved
+    org.threeten.bp.Duration ttDuration = org.threeten.bp.Duration.ofSeconds(Long.MAX_VALUE, 123);
+    java.time.Duration jtDuration = TimeConversionUtils.toJavaTimeDuration(ttDuration);
+    assertEquals(jtDuration.getSeconds(), ttDuration.getSeconds());
+    assertEquals(jtDuration.getNano(), ttDuration.getNano());
+  }
 }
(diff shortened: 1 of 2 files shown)


gax-java/gax/src/test/java/com/google/api/gax/util/TimeConversionUtilsTest.java
```diff
@@ -69,4 +69,44 @@ void testToThreetenTimeInstant_validInput_succeeds() {
         jtInstant.toEpochMilli(), TimeConversionUtils.toThreetenInstant(jtInstant).toEpochMilli());
     assertNull(TimeConversionUtils.toThreetenInstant(null));
   }
+
+  @Test
+  void testToThreeteenInstant_bigInput_doesNotOverflow() {
+    // defaults to MAX_SECONDS plus the max value of long for the nanos part
```
lqiu96, line 75: I'm not too familiar with this. What do you mean by `plus the max value of long for nanos`?

I see that the MAX is `1000000000-12-31T23:59:59.999999999Z`. Does this mean that this is essentially where jtInstant.getNano() and jtInstant.getEpochSecond() are both Interger.MAX_VALUE?

gax-java/gax/src/test/java/com/google/api/gax/util/TimeConversionUtilsTest.java
```diff
@@ -69,4 +69,44 @@ void testToThreetenTimeInstant_validInput_succeeds() {
...
+    java.time.Instant jtInstant = TimeConversionUtils.toJavaTimeInstant(ttInstant);
+    assertEquals(jtInstant.getEpochSecond(), ttInstant.getEpochSecond());
+    assertEquals(jtInstant.getNano(), ttInstant.getNano());
+  }
+
+  @Test
+  void testToThreeteenDuration_bigInput_doesNotOverflow() {
+    // we use the max long value for the seconds part and an arbitrary int for the nanos part, so we
+    // can confirm
+    // that both components are preserved
```
lqiu96, line 95:
```suggestion
    // can confirm that both components are preserved
```

gax-java/gax/src/test/java/com/google/api/gax/util/TimeConversionUtilsTest.java
```diff
@@ -69,4 +69,44 @@ void testToThreetenTimeInstant_validInput_succeeds() {
...
+    org.threeten.bp.Duration ttDuration = TimeConversionUtils.toThreetenDuration(jtDuration);
+    assertEquals(jtDuration.getSeconds(), ttDuration.getSeconds());
+    assertEquals(jtDuration.getNano(), ttDuration.getNano());
+  }
+
+  @Test
+  void testToJavaTimeDuration_bigInput_doesNotOverflow() {
+    // we use the max long value for the seconds part and an arbitrary int for the nanos part, so we
+    // can confirm
+    // that both components are preserved
```
lqiu96, line 106:
```suggestion
    // can confirm that both components are preserved
```

gax-java/gax/src/test/java/com/google/api/gax/util/TimeConversionUtilsTest.java
diegomarquezp: Instants and Durations have two components: the seconds part and the nanos part, both being `long` for `Duration`, and `long` for seconds + `int` for nanos [in Instant](https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html).

> Does this mean that this is essentially where jtInstant.getNano() and jtInstant.getEpochSecond() are both Interger.MAX_VALUE?

Yes, but with `Long.MAX_VALUE` for the seconds part instead.

lqiu96: Ok gotcha. I couldn't make sense of where the overflow was originally coming from since MAX_VALUE itself shouldn't trigger an overflow. If I'm understanding this correctly, it was triggered from the old way of how we were converting.

The `toNanos()` method was limited to Long.MAX_VALUE which has a limit of 292 years. Since Instant.MAX_VALUE has Long.MAX_VALUE for both seconds and nanos, it wouldn't be able to fit into `toNanos()` into a Long.MAX_VALUE as adding all the seconds + nanos would easily overfill.

gax-java/gax/src/test/java/com/google/api/gax/util/TimeConversionUtilsTest.java
```diff
@@ -69,4 +69,42 @@ void testToThreetenTimeInstant_validInput_succeeds() {
...
+    java.time.Instant jtInstant = TimeConversionUtils.toJavaTimeInstant(ttInstant);
+    assertEquals(jtInstant.getEpochSecond(), ttInstant.getEpochSecond());
+    assertEquals(jtInstant.getNano(), ttInstant.getNano());
+  }
+
+  @Test
+  void testToThreeteenDuration_bigInput_doesNotOverflow() {
+    // we use the max long value for the seconds part and an arbitrary int for the nanos part, so we
+    // can confirm that both components are preserved
+    java.time.Duration jtDuration = java.time.Duration.ofSeconds(Long.MAX_VALUE, 123);
```
lqiu96, line 95: qq, any specific reason for an arbitrary value of 123 and not Long/Integer.MAX_VALUE? Perhaps we use MAX_VALUE for consistency?

lqiu96 (approved):
LGTM added a small question/comment

blakeli0 (approved):

gax-java/gax/src/test/java/com/google/api/gax/util/TimeConversionUtilsTest.java
diegomarquezp: No specific reason. I modified this to use the max possible nano value without overflowing the duration

gax-java/gax/src/test/java/com/google/api/gax/util/TimeConversionUtilsTest.java
diegomarquezp: > If I'm understanding this correctly, it was triggered from the old way of how we were converting.

Exactly.

> it wouldn't be able to fit into toNanos() into a Long.MAX_VALUE as adding all the seconds + nanos would easily overfill

Yes. I don't think we will ever deal with durations instants that last longer than a few hours but it's good to support the theoretical limits of these classes, even because a user may just try a max value and that may break the library.

PR #3129, opened 2024-08-21
Draft: Selective api poc

zhumin8:
This is a draft pr as prototype to changes for selective api generation feature.
The changes in this pr include api-common-java changes copied over manually for testing purposes.


gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -425,6 +427,66 @@ public static List<Service> parseService(
...
+      MethodDescriptor method, Optional<com.google.api.Service> serviceYamlProtoOpt) {
+    method.getInputType().getFullName();
+    // default to include all when no service yaml or no library setting section.
+    if (!serviceYamlProtoOpt.isPresent()
+        || serviceYamlProtoOpt.get().getPublishing().getLibrarySettingsCount() == 0) {
+      return true;
+    }
+    List<ClientLibrarySettings> librarySettingsList =
+        serviceYamlProtoOpt.get().getPublishing().getLibrarySettingsList();
+    // TODO: verify if get(0) is reliable. may need to use package version.
```
zhumin8, line 472: will add verify package matches with proto package.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -425,6 +427,66 @@ public static List<Service> parseService(
...
+    if (serviceConfigOpt.isPresent()
+        && serviceConfigOpt.get().getLanguageSettingsOpt().isPresent()) {
+      GapicLanguageSettings languageSettings =
+          serviceConfigOpt.get().getLanguageSettingsOpt().get();
+      overriddenServiceName =
+          languageSettings.getJavaServiceName(fileDescriptor.getPackage(), s.getName());
+    }
+    return overriddenServiceName;
+  }
+  private static boolean shouldIncludeMethod(
```
zhumin8, line 462: working on a bit cleanups.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -425,6 +434,47 @@ public static List<Service> parseService(
...
+  private static boolean shouldIncludeMethodInGeneration(
+      MethodDescriptor method, Optional<List<String>> optionalIncludeMethodsList) {
+    // default to include all when no service yaml or no library setting section.
+    if (!optionalIncludeMethodsList.isPresent()) {
+      return true;
+    }
+    // default to include all when nothing specified.
+    if (optionalIncludeMethodsList.get().isEmpty()) {
+      return true;
+    }
```
zhumin8, line 474: A corner case, but probably worth to check in cloudSDK level.

zhumin8: confirmed in sdk level proposal. Default to accept all is appropriate here.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -433,18 +483,27 @@ public static List<Service> parseService(
...
-
+    String protoPackage = fileDescriptor.getPackage();
+    Optional<List<String>> inclusionMethodListFromServiceYaml =
+        getInclusionMethodListFromServiceYaml(serviceYamlProtoOpt, protoPackage);
     return fileDescriptor.getServices().stream()
         .filter(
             serviceDescriptor -> {
               List<MethodDescriptor> methodsList = serviceDescriptor.getMethods();
-              if (methodsList.isEmpty()) {
+              List<MethodDescriptor> methodListSelected =
```
blakeli0, line 493: This scenario is in case that all methods from a service are excluded?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -425,6 +434,47 @@ public static List<Service> parseService(
...
+      Optional<com.google.api.Service> serviceYamlProtoOpt, String protoPackage) {
+    if (!serviceYamlProtoOpt.isPresent()
+        || serviceYamlProtoOpt.get().getPublishing().getLibrarySettingsCount() == 0) {
+      return Optional.empty();
+    }
+    List<ClientLibrarySettings> librarySettingsList =
+        serviceYamlProtoOpt.get().getPublishing().getLibrarySettingsList();
+    // verify library_settings.version matches with proto package name
+    // This should be validated in upstream. Give warnings if not match.
+    if (!librarySettingsList.get(0).getVersion().isEmpty()
```
blakeli0, line 447: This is nice but I think we should just ignore this case. If `v1.foo.bar` is configured and the proto package is `v2`, we will continue generate `foo.bar` because the full name does not match.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -425,6 +434,47 @@ public static List<Service> parseService(
         Transport.GRPC);
   }
 
+  private static Optional<List<String>> getInclusionMethodListFromServiceYaml(
+      Optional<com.google.api.Service> serviceYamlProtoOpt, String protoPackage) {
```
blakeli0, line 438: It is usually preferred to not passing around `Optional` if we can check it before calling the method.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -425,6 +434,47 @@ public static List<Service> parseService(
...
+        librarySettingsList
+            .get(0)
+            .getJavaSettings()
+            .getCommon()
+            .getSelectiveGapicGeneration()
+            .getMethodsList();
+    return Optional.of(includeMethodsList);
+  }
+
+  private static boolean shouldIncludeMethodInGeneration(
```
blakeli0, line 465: Can we combine this method with `getInclusionMethodListFromServiceYaml`? To something like `shouldIncludeMethod(String methodName, com.google.api.Service serviceYaml)`? This also removes the need of returning `Optional<List<String>>`, which does not seem to provide much benefit. Was it for performance concerns?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
zhumin8: Yes. And this is likely now e.g. in vertexai's requirement it only exposes methods from 2 services.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
zhumin8: As discussed off-line, combining the two methods.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
zhumin8: summary from off-line discussion: will keep this check for logging purpose. This logic won't change behavior of output (yes, will include all when not matching).

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
zhumin8: I want to follow-up on this from our discussion earlier and get @blakeli0 your opinion on this. (FYI, I've combined this method into `shouldIncludeMethodInGeneration()` )
For now, I am slightly inclined to keep this `Optional<com.google.api.Service> serviceYamlProtoOpt` as input for 2 reasons: 
- The pattern of passing in this `Optional` is repeated in this `Parser` class, a larger scale refactor is needed to remove these.
- Because this service yaml is an optional config file that only some libraries have, it sort of make sense to have this Optional representation. And  I find the current way can better encapsulate this piece of logic: `shouldIncludeMethodInGeneration()` has all the information to decide if this method should or not be included. I could check `serviceYamlProtoOpt` before calling the method, but the 2 occurrences would repeat the logic in a reverse way, and it might be less readable for debugging in the future. 

Does this makes sense to you? WDYT?

gapic-generator-java/src/test/java/com/google/api/generator/gapic/protoparser/ParserTest.java
```diff
@@ -705,6 +710,99 @@ void parseServiceWithNoMethodsTest() {
     assertEquals("EchoWithMethods", services.get(0).overriddenName());
   }
 
+  @Test
+  void selectiveGenerationTest_shouldExcludeUnusedResourceNames() {
```
zhumin8, line 714: This test verifies that unused resource names (associated with methods not in inclusion list) will not be populated to GapicContext.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -425,6 +434,39 @@ public static List<Service> parseService(
...
+    }
+    List<ClientLibrarySettings> librarySettingsList =
+        serviceYamlProtoOpt.get().getPublishing().getLibrarySettingsList();
+    // Validate for logging purpose, this should be validated upstream.
+    // If library_settings.version does not match with proto package name
+    // Give warnings and disregard this config. default to include all.
+    if (!librarySettingsList.get(0).getVersion().isEmpty()
+        && !protoPackage.equals(librarySettingsList.get(0).getVersion())) {
+      LOGGER.warning(
+          "Service yaml config may be misconfigured. "
```
blakeli0, line 454:
```suggestion
          "Service yaml is misconfigured. "
```

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -425,6 +434,39 @@ public static List<Service> parseService(
...
+              + "Version in publishing.library_settings does not match proto package."
```
blakeli0, line 455: Can we add both versions to the log?

gapic-generator-java/src/test/proto/selective_api_generation.proto
blakeli0, line 1: Can we make this proto much shorter so that it only includes necessary info?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -425,6 +434,39 @@ public static List<Service> parseService(
         Transport.GRPC);
   }
 
+  private static boolean shouldIncludeMethodInGeneration(
```
blakeli0, line 437: Can we make this a default scope method so we can test it directly? This would enabled us testing more scenarios like the non matching packages, it would also simplify the golden tests.

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/ComposerTest.java
```diff
@@ -172,6 +173,29 @@ void testComposePackageInfo_emptyGapicContext_returnsNull() {
     assertNull(Composer.composePackageInfo(GapicContext.EMPTY));
   }
 
+  @Test
+  void testComposeSelectively_serviceClientShouldOnlyContainSelectedMethods() {
```
blakeli0, line 177: If we are only verifying `GapicContext`, we should test it in `ParserTest` as there are no changes in `Composer`.

gapic-generator-java/src/test/resources/selective_api_generation_no_java_settings_v1beta1.yaml
blakeli0, line 1: We probably don't need this yaml and `selective_api_generation_no_publishing_v1beta1.yaml` as the scenarios should be tested in `ParserTest` directly.

zhumin8:
implemented in #3323. closing

PR #3131, opened 2024-08-21
chore: always generate a repository if not a monorepo

diegomarquezp:
Context in [this doc](https://docs.google.com/document/d/1wrpyBtphdenM3BNelcnpBKGADYrGJUo686HXvSA0h-0/edit?pli=1&tab=t.luhbzlh25mfs#bookmark=id.ruacqw13dxd2)

This PR makes the generation workflow to always get triggered when the target repository is not a monorepo (i.e. handwritten libraries). 

This will still respect other changes between baseline and current configurations when it comes to PR description generation, but if no config changes are detected then no PR description will be created, assuming that the PR description of the original PR that will trigger the generation will be informative enough.


library_generation/test/utils/generation_config_comparator_unit_tests.py
```diff
@@ -58,6 +61,24 @@ def test_compare_config_not_change(self):
         )
         self.assertTrue(len(result.change_to_libraries) == 0)
 
+    def test_compare_non_monorepo_configs__creates_default_change(self):
```
JoeWang1127, line 64: What happens if googleapis commit changed in a non-monorepo config? Does the description.txt get generated?

library_generation/test/generate_pr_description_unit_tests.py
```diff
@@ -66,7 +67,7 @@ def test_get_commit_messages_with_same_current_and_baseline_returns_empty_messag
             ),
         )
 
-    def test_generate_pr_description_with_no_change_in_config(self):
+    def test_generate_pr_description_with_no_change_in_config__creates_no_file(self):
```
JoeWang1127, line 70:
```suggestion
    def test_generate_pr_description_with_no_change_in_config_creates_no_file(self):
```

library_generation/test/generate_pr_description_unit_tests.py
```diff
@@ -87,6 +88,29 @@ def test_generate_pr_description_with_no_change_in_config(self):
         )
         self.assertFalse(os.path.isfile(f"{pr_description_path}/pr_description.txt"))
 
+    def test_generate_pr_description_on_non_monorepo_without_changes__creates_no_file(
```
JoeWang1127, line 91:
```suggestion
    def test_generate_pr_description_on_non_monorepo_without_changes_creates_no_file(
```

blakeli0 (commented):
In general, I feel this approach creates a tighter coupling between the commit history generation and library generation, which is a direction we don't want to go. 
Can we try to fork from an earlier place? For example, we can pass `none` [here](https://github.com/googleapis/sdk-platform-java/blob/main/library_generation/cli/entry_point.py#L135) if the config is not monorepo. Because we would [always generate a library](https://github.com/googleapis/sdk-platform-java/blob/main/library_generation/generate_repo.py#L38) if `target_library_names` is none.

library_generation/model/config_change.py
```diff
@@ -58,6 +58,35 @@ def __init__(self, changed_param: str, current_value: str, library_name: str = "
         self.library_name = library_name
 
 
+class NoOpConfigChange(LibraryChange):
```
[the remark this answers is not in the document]
diegomarquezp: I couldn't find a source of truth on whether it makes sense to have an abstract class / interface if there is only one concrete class. I thought it could simplify the logic downstream to only deal with changes that should no-op instead of handling the case-specific non-monorepo change, mainly because it would future-proof it in case we want to add other no-op changes.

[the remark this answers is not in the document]
diegomarquezp: That said, I think @blakeli0's comment may make these two classes unnecessary.

library_generation/test/utils/generation_config_comparator_unit_tests.py
diegomarquezp: It was a great test to add, but I changed the approach following @blakeli0's suggestion. PR description generation won't be modified anymore.

library_generation/test/generate_pr_description_unit_tests.py
diegomarquezp: I removed these tests but renamed the ones in `test/cli` according to this suggestion (thanks!). I found [this article](https://g3doc.corp.google.com/eng/doc/devguide/testing/unit/best-practices.md?cl=head&polyglot=python#naming) as a confirmation for the right pattern in python.

diegomarquezp:
> In general, I feel this approach creates a tighter coupling between the commit history generation and library generation, which is a direction we don't want to go. Can we try to fork from an earlier place? For example, we can pass `none` [here](https://github.com/googleapis/sdk-platform-java/blob/main/library_generation/cli/entry_point.py#L135) if the config is not monorepo. Because we would [always generate a library](https://github.com/googleapis/sdk-platform-java/blob/main/library_generation/generate_repo.py#L38) if `target_library_names` is none.

@blakeli0 this approach is simpler, and I agree that it keeps separation of concerns. PTAL again when you have a chance!

library_generation/cli/entry_point.py
```diff
@@ -88,6 +88,20 @@ def generate(
     The commit history, if generated, will be available in
     repository_path/pr_description.txt.
     """
+    generate_impl(
```
JoeWang1127, line 91: Why create another wrapper function?

diegomarquezp: Good question. Although the [click documentation](https://click.palletsprojects.com/en/8.1.x/testing/#basic-testing) suggests to use its special `runner`, it would call the production functions `generate_from_yaml` (and others) all the way to produce a full generation, which is a use case I was avoiding.

My approach was to mock the generation functions and only confirm they were called with the expected arguments. However, click throws an error when you try to call a function annotated with `@main.command` directly. That's why I converted the function into a wrapper.

library_generation/cli/entry_point.py
```diff
@@ -88,6 +88,20 @@ def generate(
...
+    )
+
+
+def generate_impl(
+    baseline_generation_config_path: str,
+    current_generation_config_path: str,
+    repository_path: str,
+):
+    """
+    implementation method for generate()
```
diegomarquezp, line 102:
```suggestion
    implementation method for generate()
    this decouples the exposed interface from the implementation
    in order to ease testing
```

library_generation/cli/entry_point.py
```diff
@@ -88,6 +88,20 @@ def generate(
     The commit history, if generated, will be available in
     repository_path/pr_description.txt.
     """
+    generate_impl(
+        baseline_generation_config_path, current_generation_config_path, repository_path
+    )
+
+
+def generate_impl(
```
blakeli0, line 96: Prefix this method with `__` to indicate this is a private method?

library_generation/cli/entry_point.py
blakeli0: Maybe add additional info that this function is exposed for testing only and should not be called directly?

library_generation/test/resources/test-config/generation_config_library_modified.yaml
blakeli0, line 1: Can this file be simplified to contain minimal information?

diegomarquezp: Good point. I simplified it (only required params)

library_generation/cli/entry_point.py
diegomarquezp: Yes, it's indeed a private method. Done, and renamed the method as discussed internally.

library_generation/cli/entry_point.py
diegomarquezp: Good point. I improved the comment.

JoeWang1127 (approved):
Could you fix the unit tests?

blakeli0 (approved):

PR #3137, opened 2024-08-28
feat(gax): add API key authentication to ClientSettings

ldetmer:
Allow gax client libraries to authenticate using API key via setApiKey method exposed from ClientSettings.  Also added deduping to GRPC calls for api key headers.

Tested using LanguageServiceSettings

cc @westarle


gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
```diff
@@ -175,9 +176,19 @@ public static ClientContext create(StubSettings settings) throws IOException {
     // A valid EndpointContext should have been created in the StubSettings
     EndpointContext endpointContext = settings.getEndpointContext();
     String endpoint = endpointContext.resolvedEndpoint();
+    String apiKey = settings.getApiKey();
+    Credentials credentials = settings.getCredentialsProvider().getCredentials();
+    if (apiKey != null && credentials != null) {
```
ldetmer, line 181: I added this check here instead of in the build() method on ClientSettings, as it made the code more readable on line 187 to me - we're not overriding client credentials, as they would have to be null.

gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
```diff
@@ -187,22 +198,7 @@ public static ClientContext create(StubSettings settings) throws IOException {
...
-      }
-
-      URI gdchAudienceUri;
-      try {
-        gdchAudienceUri = URI.create(audienceString);
-      } catch (IllegalArgumentException ex) { // thrown when passing a malformed uri string
-        throw new IllegalArgumentException("The GDC-H API audience string is not a valid URI", ex);
-      }
-      credentials = ((GdchCredentials) credentials).createWithGdchAudience(gdchAudienceUri);
+      credentials = getGdchCredentials(settingsGdchApiAudience, endpoint, credentials);
```
ldetmer, line 201: created helper method for readability, no change to logic

gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
```diff
@@ -291,6 +284,31 @@ public static ClientContext create(StubSettings settings) throws IOException {
         .build();
   }
 
+  /**
+   * Constructs a new {@link com.google.auth.Credentials} object based on credentials provided with
+   * a GDC-H audience
+   */
+  private static Credentials getGdchCredentials(
```
ldetmer, line 291: no changes here, just pulled out to separate method to consolidate logic

gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
```diff
@@ -175,37 +176,29 @@ public static ClientContext create(StubSettings settings) throws IOException {
...
-        throw new IllegalArgumentException("The GDC-H API audience string is not a valid URI", ex);
+    String apiKey = settings.getApiKey();
+    Credentials credentials;
+    if (apiKey != null) {
+      // if API key exists it becomes the default credential
+      credentials = ApiKeyCredentials.create(settings.getApiKey());
+    } else {
+      credentials = settings.getCredentialsProvider().getCredentials();
+      // check if need to adjust credentials/endpoint/endpointContext for GDC-H
+      String settingsGdchApiAudience = settings.getGdchApiAudience();
```
ldetmer, line 187: no change in logic from here down, just nested inside else statement

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiCredentials.java
```diff
@@ -0,0 +1,149 @@
...
+import java.util.concurrent.TimeUnit;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Test suite to confirm a client can be instantiated with API key credentials and correct
+ * authorization sent via headers to back end
+ */
+class ITApiCredentials {
```
blakeli0, line 46: Maybe rename to `ITApiKeyCredentials`?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiCredentials.java
blakeli0, line 1: Can we add another test case if ApiKey is set through both `ClientSettings` and `headerProvider`? Also wondering have we done any manual test for this scenario?

gax-java/gax/src/main/java/com/google/api/gax/rpc/StubSettings.java
```diff
@@ -234,6 +236,14 @@ public final String getGdchApiAudience() {
     return gdchApiAudience;
   }
 
+  /**
+   * Gets the ApiKey that should be used for authentication. If an empty string was provided it will
+   * return null
+   */
+  public final String getApiKey() {
+    return (apiKey == null || apiKey.isEmpty()) ? null : apiKey;
```
blakeli0, line 244: If a customer explicitly sets apiKey to empty, I think it is a mistake and we should error out instead of silently ignoring it, hence we should just return whatever is set directly.

gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientSettings.java
```diff
@@ -208,6 +219,7 @@ public B setBackgroundExecutorProvider(ExecutorProvider executorProvider) {
 
     /** Sets the CredentialsProvider to use for getting the credentials to make calls with. */
     public B setCredentialsProvider(CredentialsProvider credentialsProvider) {
+      explicitCredentialsProvided = true;
```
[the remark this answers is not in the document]
ldetmer: unfortunately no because that could have been set as the default CredentialProvider.  [The requirements state in that scenario](https://docs.google.com/document/d/13qovLzgsk_Rb5nizYnjDDvJ6jL_mIG7MUIadvlV7SqM/edit?resourcekey=0-rg7PSjFwP5PrFpJLsmBS8A&tab=t.0#bookmark=id.h83h02o5ksek) we should just ignore the default credential provider.  I can leave a comment in the code, but lets also discuss in weekly meeting to ensure we're on same page

gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientSettings.java
```diff
@@ -58,6 +58,10 @@ public abstract class ClientSettings<SettingsT extends ClientSettings<SettingsT>
 
   /** Constructs an instance of ClientSettings. */
   protected ClientSettings(Builder builder) throws IOException {
+    if (builder.stubSettings.getApiKey() != null && builder.explicitCredentialsProvided) {
```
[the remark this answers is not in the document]
ldetmer: I would also like to discuss this in our meeting. The base[ build operations are abstract](https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientSettings.java#L392).  So I'm not sure there is a good place to add to the existing build flow.  Alternatively I could add this logic to [when we actually build the credential object in ClientContext](https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java#L180)

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiCredentials.java
ldetmer: good call, unfortunately it adds 2 separate x-goog-api-key headers and even if they are the same, the backend service fails saying "INVALID_ARGUMENT: API key not valid. Please pass a valid API key."

I will add a check that fails if both are provided, but will wait until we decide where validation check should live.

ldetmer: Additional debugging:

this is where the headers are actually getting [merged](https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/internal/MetadataApplierImpl.java#L70)

This is where the [credential headers supplied by getRequestMetadata](https://github.com/grpc/grpc-java/blob/master/auth/src/main/java/io/grpc/auth/GoogleAuthLibraryCallCredentials.java#L120) are getting read.

At this point in the flow we have already started a stream with io.grpc.netty and do not receive control back until we receive a failure from the back end.

I can de-dup by requesting the credential metadata before handing off, but I don't believe this is the proper solution.  I think it might be best to include the logic as part of [GoogleAuthLibraryCallCredentials](https://github.com/grpc/grpc-java/blob/master/auth/src/main/java/io/grpc/auth/GoogleAuthLibraryCallCredentials.java#L120) but would have to work with grpc team.  also this would be a different solution for http.

If possible, due to the migrated timeline, lets keep deduping cred headers out of scope for this release

gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientSettings.java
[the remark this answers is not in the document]
ldetmer: a couple of options:

1) Most thorough: we add tracking to generated ServiceStubSettings classes to include tracking the defaultCredentialProvider object and check to see if current credentialProvider object matches and if so ignore.  Only negative is this will add a bit of generated code which is only necessary for this validation.

2) Best compromise: track if ClientSettings.setCredential was called and if not check if credentialProvider is NOT GoogleCredentialsProvider.  If either of these is true and api key is set, throw error.
While this will not catch a user using StubSettings to set the credential provider, that is not the published [java docs samples of usage](https://github.com/googleapis/google-cloud-java/blob/main/java-language/google-cloud-language/src/main/java/com/google/cloud/language/v2/LanguageServiceClient.java#L186)

3) Simplest: treat any GoogleCredentialsProvider as default credentials.

gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
```diff
@@ -175,37 +176,29 @@ public static ClientContext create(StubSettings settings) throws IOException {
...
-      } else {
-        throw new IllegalArgumentException("Could not infer GDCH api audience from settings");
-      }
-
-      URI gdchAudienceUri;
-      try {
-        gdchAudienceUri = URI.create(audienceString);
-      } catch (IllegalArgumentException ex) { // thrown when passing a malformed uri string
-        throw new IllegalArgumentException("The GDC-H API audience string is not a valid URI", ex);
+    String apiKey = settings.getApiKey();
```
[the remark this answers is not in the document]
ldetmer: unfortunately there is some logic that updates Endpoint metadata that is also used in transport provider.  Without a more indepth refactor I tried to consolidate as much as possible

gax-java/gax/src/main/java/com/google/api/gax/rpc/StubSettings.java
ldetmer: I removed from now.  Waiting to see how we perform validation in the end as to where we will check for nulls

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITApiCredentials.java
ldetmer: added dedup logic for credential headers in GRPC.  HTTP already was deduping

gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientSettings.java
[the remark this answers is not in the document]
ldetmer: removed validation for now

gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientSettings.java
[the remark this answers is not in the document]
ldetmer: removed validation for now.

gax-java/gax-grpc/src/test/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProviderTest.java
```diff
@@ -877,6 +879,43 @@ public void canUseDirectPath_nonGDUUniverseDomain() {
     Truth.assertThat(provider.canUseDirectPath()).isFalse();
   }
 
+  @Test
```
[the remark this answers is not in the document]
ldetmer: Right I could expose the headers from GrpcHeaderInterceptor, but GrpcHeaderInterceptor is only exposed from a grpc package private ManagedChannelOrphanWrapper, so I would also have to make GrpcHeaderInterceptor a member variable and expose it from InstantiatingGrpcChannelProvider, but that seemed too specific for unit testing so instead I chose to cover it in the IT tests.  Let me know if you think it would be better to expose the variable and cover it in unit

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -496,6 +499,20 @@ private ManagedChannel createSingleChannel() throws IOException {
...
   }
 
+  // dedup any headers explicitly set with headers provided by credential, with preference to
+  // credential headers
+  private Map<String, String> mergeHeadersWithCredentialHeaders() {
+    Map<String, String> userHeaders = new HashMap<>(headerProvider.getHeaders());
+    if (credentials != null) {
+      try {
+        userHeaders.keySet().removeAll(credentials.getRequestMetadata().keySet());
+      } catch (Exception e) {
```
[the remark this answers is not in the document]
ldetmer: you're correct, updated it to catch IO, but I left as a try catch because I didn't think we wanted to fail the whole call in case of an error requesting the credentials here.  In that case if they couldn't request them when actually making the call it would still fail, but otherwise if they can get retrieve it during the actual grpc call, it would succeed.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -407,8 +409,10 @@ ChannelCredentials createMtlsChannelCredentials() throws IOException, GeneralSec
   }
 
   private ManagedChannel createSingleChannel() throws IOException {
+    Map<String, String> mergedHeaders = mergeHeadersWithCredentialHeaders();
```
lqiu96, line 412: Off the top of my head I am a bit worried about how this may impact channel pooling. Our channel pools may constant recreate channels at some frequency and this may end up calling `getRequestMetadata()` over and over.

IIRC, some downstream teams have some latency requirements and this may end up slowing down applications.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/util/GrpcCapturingClientInterceptor.java
```diff
@@ -0,0 +1,102 @@
...
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import com.google.api.gax.httpjson.ApiMethodDescriptor;
+import com.google.api.gax.httpjson.HttpJsonMetadata;
+import io.grpc.ForwardingClientCall;
+import io.grpc.Status;
+
+/** Implements a client interceptor to retrieve the metadata from a GRPC client request. */
+public class GrpcCapturingClientInterceptor implements ClientInterceptor {
```
lqiu96, line 48: Thanks! I like this. I think we had this in our backlog a while back: https://github.com/googleapis/sdk-platform-java/issues/2743

Probably can mark this as closed once this PR is in.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
ldetmer: thanks for bringing this up.  I re-work this to be part of the build() process

ldetmer: ok updated so headers are deduped as part of the transportprovider build process

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -496,6 +500,25 @@ private ManagedChannel createSingleChannel() throws IOException {
...
+   * {@link io.grpc.auth.GoogleAuthLibraryCallCredentials#applyRequestMetadata}.  GRPC does not dedup headers {@link https://github.com/grpc/grpc-java/blob/a140e1bb0cfa662bcdb7823d73320eb8d49046f1/api/src/main/java/io/grpc/Metadata.java#L504} so we must before initiating the call.
+   */
+  private void removeCredentialDuplicateHeaders() {
+    if (headerProvider != null) {
+      headersWithDuplicatesRemoved.putAll(headerProvider.getHeaders());
+    }
+    if (credentials != null) {
+      try {
+        Map<String, List<String>> credentialRequestMetatData = credentials.getRequestMetadata();
+        if (credentialRequestMetatData != null) {
```
lqiu96, line 513: The other concern I have with this is that this is still going to make a network call to retrieve the metadata. gRPC-Java is going to be doing this as well, so this ends up being 2 network calls, right?

Would it be possible just to remove anything with `API-Keys` or `Authorization` in the header so that we don't need to do the `credentials.getRequestMetadata()`?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -564,6 +587,14 @@ public boolean shouldAutoClose() {
     return true;
   }
 
+  /**
+   * @return list of provided headers that will be sent with GRPC call with any duplicates removed
+   *     see {@link #removeCredentialDuplicateHeaders()}
+   */
+  public Map<String, String> getHeadersWithDuplicatesRemoved() {
```
blakeli0, line 594: Is this only used in tests? If it is, we can make this package private and mark it as `@VisibleForTesting`? Or we don't have to expose another method, just make `removeCredentialDuplicateHeaders` package privare.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
blakeli0: Thanks Lawrence for bring up this concern! It is definitely a concern if some other types of credentials are making network calls, we don't know that behavior before and I thought it is part of the credentials object already. Do you have links to how other types of credentials would make network calls?

lqiu96: > Thanks Lawrence for bring up this concern! It is definitely a concern if some other types of credentials are making network calls, we don't know that behavior before and I thought it is part of the credentials object already. Do you have links to how other types of credentials would make network calls?

I could be wrong on this, but I believe almost every Credential that flows under OauthCredentials is making a network call to retrieve the Access Token. The exception that comes to mind is for SA Credentials using the SSJWT flow.

Oauth2Credential's `asyncFetch`: https://github.com/googleapis/google-auth-library-java/blob/afe4ff9f1f3c406c72ecb66fba50d7463f620467/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java#L182-L184

ldetmer: Lawrence is correct, the headers should be pretty static, @blakeli0 how do you feel about filtering out a static list of authorization headers? If we think this list is going to change in the future, as part of the fast follow I can include  exposing the auth header from the Credentials object.  Or we can remove this entirely until we have that solution?

ldetmer: as discussed we will only dedup for apikey credentials for this PR.  action item for @ldetmer to test how other credentials with duplicate headers behave

lqiu96 (approved):
LGTM

gax-java/gax/src/test/java/com/google/api/gax/rpc/ClientContextTest.java
```diff
@@ -1074,4 +1072,46 @@ public void testStreamWatchdogInterval_backportMethodsBehaveCorrectly() {
         ct -> ct.getStreamWatchdogCheckIntervalDuration(),
         ct -> ct.getStreamWatchdogCheckInterval());
   }
+
+  @Test
+  public void testSetApiKey_createsApiCredentials() throws IOException {
```
blakeli0, line 1077: nit: These test methods don't have to be `public` anymore after we migrated to junit 5.

ldetmer: ah thanks! done

blakeli0 (approved):

PR #3140, opened 2024-08-29
feat: add toString to futures returned by operations

igorbernstein2:
Sometimes an operation can get stuck indefinitely. The underlying reasons can vary significantly:
- the underlying attempt rpc can get stuck due to a bug in grpc (ie https://github.com/grpc/grpc-java/pull/11026)
- the operation can get stuck in layers above gax: https://github.com/googleapis/java-bigtable/pull/1939
- or it can get stuck in gax itself (dont have a pointer handy)

Guava futures provide some observability for ListenableFutures, but in creating the custom ApiFutures in gax, we lose that functionality. This PR sprinkles a few to toString to allow callers to inspect the internal state of the operation. For example with these changes, the toString() of the future returned from bigtableDataClient.mutateRows() changes from

> TransformFuture@652ce654[status=PENDING, info=[inputFuture=[com.google.api.core.ApiFutureToListenableFuture@522ba524], function=[com.google.api.core.ApiFutures$ApiFunctionToGuavaFunction@29c5ee1d]]]

to
> ListenableFutureToApiFuture{delegate=TransformFuture@7ac9af2a[status=PENDING, info=[inputFuture=[ApiFutureToListenableFuture{apiFuture=CallbackChainRetryingFuture{super=com.google.api.gax.retrying.CallbackChainRetryingFuture@7bb004b8[status=PENDING], latestCompletedAttemptResult=null, attemptResult=null, attemptSettings=TimedAttemptSettings{globalSettings=RetrySettings{totalTimeout=PT10M, initialRetryDelay=PT0.01S, retryDelayMultiplier=2.0, maxRetryDelay=PT1M, maxAttempts=0, jittered=true, initialRpcTimeout=PT1M, rpcTimeoutMultiplier=1.0, maxRpcTimeout=PT1M}, retryDelay=PT0S, rpcTimeout=PT1M, randomizedRetryDelay=PT0S, attemptCount=0, overallAttemptCount=0, firstAttemptStartTimeNanos=635709620001791}}}], function=[com.google.api.core.ApiFutures$ApiFunctionToGuavaFunction@652ce654]]]}

This allows us to reason about whats stuck. I'm working another PR that will add a close(timeout) to the Batcher that will use this functionality to identify why batcher.close() timed out

Thank you for opening a Pull Request! Before submitting your PR, please read our [contributing guidelines](https://github.com/googleapis/gapic-generator-java/blob/main/CONTRIBUTING.md). 

There are a few things you can do to make sure it goes smoothly:
[...]

PR #3141, opened 2024-08-29
feat: add Batcher#close(timeout) and Batcher#cancelOutstanding

igorbernstein2:
There have been reports of batcher.close() hanging every once in awhile. Currently it is impossible to debug because we dont expose any internal state to analyze.

This PR adds 2 additional methods that should help in diagnosing issues:
1. close(timeout) will try to close the batcher, but if any of the underlying batch operations fail, the exception message will contain a wealth of information describing the underlying state of operations as provided by #3140
2. cancelOutstanding this allows for remediation for close(timeout) throwing an exception.

The intended usecase is dataflow connector's FinishBundle:

```java
try {
  batcher.close(Duration.ofMinutes(1));
} catch(TimeoutException e) {
   // log details why the batch failed to close with the help of #3140
   logger.error(e);
   batcher.cancelOutstanding();
  batcher.close(Duration.ofMinutes(1));
}
```

Example exception message:

> Exception in thread "main" com.google.api.gax.batching.BatchingException: Timed out trying to close batcher after PT1S. Batch request prototype: com.google.cloud.bigtable.data.v2.models.BulkMutation@2bac9ba. Outstanding batches: Batch{operation=CallbackChainRetryingFuture{super=null, latestCompletedAttemptResult=ImmediateFailedFuture@6a9d5dff[status=FAILURE, cause=[com.google.cloud.bigtable.data.v2.models.MutateRowsException: Some mutations failed to apply]], attemptResult=null, attemptSettings=TimedAttemptSettings{globalSettings=RetrySettings{totalTimeout=PT10M, initialRetryDelay=PT0.01S, retryDelayMultiplier=2.0, maxRetryDelay=PT1M, maxAttempts=0, jittered=true, initialRpcTimeout=PT1M, rpcTimeoutMultiplier=1.0, maxRpcTimeout=PT1M}, retryDelay=PT1.28S, rpcTimeout=PT1M, randomizedRetryDelay=PT0.877S, attemptCount=8, overallAttemptCount=8, firstAttemptStartTimeNanos=646922035424541}}, elements=com.google.cloud.bigtable.data.v2.models.RowMutationEntry@7a344b65}


igorbernstein2:
api compatibility is failing with:
>Run REPOS_UNDER_TEST="java-datastore" ./.kokoro/presubmit/downstream-compatibility.sh
Setup maven mirror
Installing this repo's modules to local maven.
Error:  [7](https://github.com/googleapis/sdk-platform-java/actions/runs/10622278689/job/29446168617?pr=3141#step:7:8)012: com.google.api.gax.batching.Batcher: Method 'public void cancelOutstanding()' has been added to an interface
Error:  7012: com.google.api.gax.batching.Batcher: Method 'public void close(org.threeten.bp.Duration)' has been added to an interface
Error:  Failed to execute goal org.codehaus.mojo:clirr-maven-plugin:2.[8](https://github.com/googleapis/sdk-platform-java/actions/runs/10622278689/job/29446168617?pr=3141#step:7:9):check (default) on project gax: There were 2 errors. -> [Help 1]

But it is wrong because the Batcher interface is marked as `@InternalExtensionOnly` so adding new methods should be allowed. I tried ignoring it using a clirr ignore, but that doesnt seem to work. Whats the correct way to get it to start passing?

gax-java/gax/clirr-ignored-differences.xml
```diff
@@ -105,4 +105,10 @@
     <className>com/google/api/gax/tracing/MetricsTracer</className>
     <field>*</field>
   </difference>
+  <!-- Ignore method additions to an InternalExtensionOnly interface-->
+  <difference>
+    <differenceType>7012</differenceType>
+    <className>com.google.api.gax.batching.Batcher</className>
```
blakeli0, line 111: The package name should be split by `/` not `.`
```suggestion
    <className>com/google/api/gax/batching/Batcher</className>
```

gax-java/gax/src/main/java/com/google/api/gax/batching/BatcherImpl.java
```diff
@@ -374,6 +398,17 @@ public void close() throws InterruptedException {
...
+      StringJoiner batchesStr = new StringJoiner(",");
+      for (Batch<ElementT, ElementResultT, RequestT, ResponseT> batch :
+          outstandingBatches.keySet()) {
+        batchesStr.add(batch.toString());
+      }
+      String msg = "Timed out trying to close batcher after " + timeout + ".";
+      msg += " Batch request prototype: " + prototype + ".";
+      msg += " Outstanding batches: " + batchesStr;
+
+      throw new BatchingException(msg);
```
igorbernstein2, line 411: should I make this a more specific exception? ie. `BatcherCloseException`

blakeli0: I think it depends on how do you plan to handle it in java-bigtable. If you want to rely on it to do some additional logics, then it makes sense to create a specific exception. If not, I don't think it is necessary.

Separately, depends on how to plan to handle this exception, you may want to logic the exception as well.

gax-java/gax/src/main/java/com/google/api/gax/batching/BatcherImpl.java
```diff
@@ -457,6 +494,17 @@ void add(
       totalThrottledTimeMs += throttledTimeMs;
     }
 
+    void setOperation(@Nonnull ApiFuture<ResponseT> operation) {
```
blakeli0, line 497: I feel `operation` might be a too generic name, can we rename it to something more specific? Maybe something like `responseFuture`?

gax-java/gax/src/main/java/com/google/api/gax/batching/Batcher.java
```diff
@@ -77,13 +79,25 @@ public interface Batcher<ElementT, ElementResultT> extends AutoCloseable {
...
   void close() throws InterruptedException;
 
+  /**
+   * Closes this Batcher by preventing new elements from being added, then flushing the existing
+   * elements and waiting for all the outstanding work to be resolved. If all of the outstanding
+   * work has not been resolved, then a {@link BatchingException} will be thrown with details of the
+   * remaining work. The batcher will remain in a closed state and will not allow additional
+   * elements to be added.
+   */
+  void close(@Nullable Duration timeout) throws InterruptedException;
```
blakeli0, line 99: We recently introduced java.time to gax in [this PR](https://github.com/googleapis/sdk-platform-java/pull/1872), which introduced alternative java.time methods for all threeten methods. Going forward, we will deprecate all threeten methods. Can you please change this to `java.time.Duration` instead?

gax-java/gax/src/main/java/com/google/api/gax/batching/BatcherImpl.java
```diff
@@ -86,7 +91,8 @@ public class BatcherImpl<ElementT, ElementResultT, RequestT, ResponseT>
   private final BatcherReference currentBatcherReference;
 
   private Batch<ElementT, ElementResultT, RequestT, ResponseT> currentOpenBatch;
-  private final AtomicInteger numOfOutstandingBatches = new AtomicInteger(0);
+  private final ConcurrentMap<Batch<ElementT, ElementResultT, RequestT, ResponseT>, Boolean>
```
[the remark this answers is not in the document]
igorbernstein2: I was hoping to avoid additional locking. afaik there are no lock free sets which is why I'm using the concurrent map as a set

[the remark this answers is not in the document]
blakeli0: Ah, I forgot there is no `ConcurrentHashSet` in Java, this makes sense then.

gax-java/gax/src/main/java/com/google/api/gax/batching/BatcherImpl.java
igorbernstein2: I updated the intended usage in the description. I would prefer to avoid logging & throwing to avoid noise

gax-java/gax/src/main/java/com/google/api/gax/batching/BatcherImpl.java
```diff
@@ -374,6 +404,17 @@ public void close() throws InterruptedException {
...
+      StringJoiner batchesStr = new StringJoiner(",");
+      for (Batch<ElementT, ElementResultT, RequestT, ResponseT> batch :
+          outstandingBatches.keySet()) {
+        batchesStr.add(batch.toString());
+      }
+      String msg = "Timed out trying to close batcher after " + timeout + ".";
+      msg += " Batch request prototype: " + prototype + ".";
+      msg += " Outstanding batches: " + batchesStr;
+
+      throw new TimeoutException(msg);
```
blakeli0, line 417: Do you want to wrap the original exception as well to keep the stacktrace? Or only the message is good enough?

igorbernstein2: I dont think the original stacktrace will be useful

blakeli0 (approved):

PR #3146, opened 2024-08-30
chore: re-generate when config or Dockerfile is changed

JoeWang1127:
In this PR:
- Re-generate common protos and iam if generation config or Dockerfile is changed.

I use #3145 to verify the change.


.github/scripts/hermetic_library_generation.sh
```diff
@@ -99,22 +94,18 @@ docker run \
   -v "$(pwd):${workspace_name}" \
   -v "$HOME"/.m2:/home/.m2 \
   -e GENERATOR_VERSION="${generator_version}" \
-  gcr.io/cloud-devrel-public-resources/java-library-generation:"${image_tag}" \
-  --baseline-generation-config-path="${workspace_name}/${baseline_generation_config}" \
-  --current-generation-config-path="${workspace_name}/${generation_config}"
+  gcr.io/cloud-devrel-public-resources/java-library-generation:"${image_tag}"
```
blakeli0, line 97: Is the commit history going to be populated correctly if we don't pass the baseline generation config into the image?

.github/scripts/hermetic_library_generation.sh
```diff
@@ -99,22 +94,18 @@ docker run \
...
 
 # commit the change to the pull request.
-rm -rdf output googleapis "${baseline_generation_config}"
+rm -rdf output googleapis
 git add --all -- ':!pr_description.txt'
 changed_files=$(git diff --cached --name-only)
 if [[ "${changed_files}" == "" ]]; then
-    echo "There is no generated code change with the generation config and Dockerfile change ${config_diff}."
```
blakeli0, line 111: We actually mentioned `Dockerfile change` in the message but it was not achieved before?

.github/scripts/hermetic_library_generation.sh
JoeWang1127: No, the commit history has to go through baseline/current config comparison.

However, in this repo, we don't have config changes other than googleapis commit since everything else is baked in the image.

.github/scripts/hermetic_library_generation.sh
JoeWang1127: Dockerfile is not part of generation config, so the previous `docker run` won't generate anything if Dockerfile is the only change.

diegomarquezp (approved):
LGTM. IIUC this is more generic than only allowing changes in the Dockerfile. I think this can allow modifications coming from anything that alters the output of the generation: postprocessing, gapic generator, Dockerfile, etc. And it's great to allow this btw, but maybe the title/description should be something like "re-generate when config or generation workflow has changed"?

diegomarquezp:
> LGTM. IIUC this is more generic than only allowing changes in the Dockerfile. I think this can allow modifications coming from anything that alters the output of the generation: postprocessing, gapic generator, Dockerfile, etc. And it's great to allow this btw, but maybe the title/description should be something like "re-generate when config or generation workflow has changed"?

This is wrong. There was a modification on what would allow the generation to start made in https://github.com/googleapis/sdk-platform-java/commit/74f20a253e65d9d4253541c4c8250d75abe8d0b8. So only the Dockerfile or the config yaml would trigger this.

PR #3161, opened 2024-09-03
chore: always generate regardless of what file changed

JoeWang1127:
In this PR:
- Always re-generate common protos and iam in every commit.
- Secure generation workflow (use environment variable to avoid script injection), inspired by https://github.com/googleapis/java-bigtable/pull/2317

### DIFF
--- a/.github/scripts/hermetic_library_generation.sh
+++ b/.github/scripts/hermetic_library_generation.sh
@@ -64,17 +64,14 @@ fi
 
 image_tag=local
 workspace_name="/workspace"
-docker_file="library_generation.Dockerfile"
+baseline_generation_config="baseline_generation_config.yaml"
 message="chore: generate libraries at $(date)"
 
+git checkout "${target_branch}"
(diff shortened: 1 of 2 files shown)


.github/scripts/hermetic_library_generation.sh
```diff
@@ -94,10 +91,12 @@ docker run \
   -v "$(pwd):${workspace_name}" \
   -v "$HOME"/.m2:/home/.m2 \
   -e GENERATOR_VERSION="${generator_version}" \
-  gcr.io/cloud-devrel-public-resources/java-library-generation:"${image_tag}"
+  gcr.io/cloud-devrel-public-resources/java-library-generation:"${image_tag}" \
+  --baseline-generation-config-path="${workspace_name}/${baseline_generation_config}" \
```
blakeli0, line 95: I thought the generation process is still going to be skipped inside the scripts if there is no diff between the two generation configs?

PR #3164, opened 2024-09-04
fix: Remove null check for paged Protobuf messages

lqiu96:
## Background 
This was caught from ErrorProne flagging an [ImpossibleNullComparison](https://errorprone.info/bugpattern/ImpossibleNullComparison) when upgrading to Protobuf 4.27.4 runtime. Protobuf messages should be non-null (with type specific default values) when they are serialized. Removing the null check that is being flagged from ErrorProne.

## Testing
We have some issues with testing this inside Showcase. Protobuf serializes messages with default values and does not allow for null (will throw a NPE). Since showcase is configured to receive and send Protobuf Messages, we cannot configure the response to be return null values (at best we can return something like an empty string or an empty map).

Instead, we have opted to create a sample Spring server to return a JSON payload with certain fields set to null. Create a GAPIC client and manually set the endpoint to point to the local server (i.e. `localhost:8080`). The Spring application will have a endpoint that matches the path of the RPC (i.e. `@GetMapping("/compute/v1/projects/{project}/zones/{zone}/instances")`) and returns the JSON. This proves that a null fields sent from the server is serialized properly into a Protobuf message and the message's fields are non-null.

```
@RestController
public class ComputeController {

  @GetMapping("/compute/v1/projects/{project}/zones/{zone}/instances")
  String listInstances(@PathVariable String project, String zone) {
    return "{\"items\": null, \"id\": 5}";
  }
}
```
In the example above, the `items` field is a list and the printing it out in the client library will result in `[]`.

Additionally, we have added some small, local tests against two repeated Protobuf messages:
1. BackendBucketList's items is a non-null List
```
JsonFormat.Parser parser = JsonFormat.parser().ignoringUnknownFields()
        .usingTypeRegistry(TypeRegistry.newBuilder().add(BackendBucketList.getDescriptor()).build());
BackendBucketList.Builder builder = BackendBucketList.newBuilder();
parser.merge("{\"items\": null}", builder);
BackendBucketList list = builder.build();
System.out.println(list.getItemsList());
```

2. PredictResponse's metadata is a non-null Map
```
JsonFormat.Parser parser = JsonFormat.parser().ignoringUnknownFields()
        .usingTypeRegistry(TypeRegistry.newBuilder().add(PredictResponse.getDescriptor()).build());;
PredictResponse.Builder builder = PredictResponse.newBuilder();
parser.merge("{\"recommendation_token\": \"343\"}", builder);
System.out.println(builder.build().getMetadataMap());
```


blakeli0 (approved):

blakeli0 (commented):
Do you mind adding more info regarding why we can not test it with showcase and how we tested with a local server?

lqiu96:
> Do you mind adding more info regarding why we can not test it with showcase and how we tested with a local server?

Added to the description.

PR #3170, opened 2024-09-05
feat: expose property in GrpcTransportChannel if it uses direct path.

### LINKED ISSUE #3134: canUseDirectPath returning false when DirectPath is enabled
Spanner Client library is trying to use [canUseDirectPath](https://github.com/googleapis/sdk-platform-java/blob/e6042e60411ff1326da70bdd42c364b9abae953a/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L507) from the [GapicSpannerRPC](https://github.com/googleapis/java-spanner/blob/main/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java#L393) class

 This method is always returning `false` irrespective of the DirectPath is enabled or not. 
 
It is because of the [needsCredentials](https://github.com/googleapis/sdk-platform-java/blob/e6042e60411ff1326da70bdd42c364b9abae953a/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L239) which always returns "true" as credentials are always null. 
 
 Same method is working for gax because the credentials are set in [ClientContext](https://github.com/googleapis/sdk-platform-java/blob/e6042e60411ff1326da70bdd42c364b9abae953a/gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java#L231) before the [transportChannel](https://github.com/googleapis/sdk-platform-java/blob/e6042e60411ff1326da70bdd42c364b9abae953a/gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java#L239) is created. 
 
 However for Spanner library there is no way to use the  `canUseDirectPath` method as the actual channel provider never gets updated.

zhumin8:
This should fix #3134.

Spanner needs to know if `GrpcSpannerStub`’s transportChannel uses DirectPath for tracing purposes.
`canUseDirectPath()` from `stubSettings.getTransportChannelProvider()` does not provide accurate information because credentials are set afterwards when creating the transportChannel for `GrpcSpannerStub` (`GrpcSpannerStub.create(stubSettings)`)

Proposed change:
Expose a property in `GrpcTransportChannel` and pass this information down from channel provider.
Spanner can access from `ClientContext` as shown below
```
SpannerStubSettings stubSettings = options
        .getSpannerStubSettings()
        .toBuilder()
        .setTransportChannelProvider(channelProvider)
        .setCredentialsProvider(credentialsProvider)
        .setStreamWatchdogProvider(watchdogProvider)
        .setTracerFactory(options.getApiTracerFactory())
        .build();
ClientContext clientContext = ClientContext.create(stubSettings);
// create GrpcSpannerStub from clientContext instead. 
// this is equivelant to GrpcSpannerStub.create(stubSettings)
this.spannerStub =
    GrpcSpannerStub.create(clientContext);
// access this property from transport channel.
((GrpcTransportChannel) clientContext.getTransportChannel()).isDirectPath();
```


zhumin8:
Hi @mohanli-ml, 
We are trying to expose information about whether direct path is used via transport channel (GrpcTransportChannel). And want to double check on what is considered a direct path. In particular, what is xds and is it one type of direct path?
Regarding logic [here](https://github.com/googleapis/sdk-platform-java/blob/fc8d80d11182259c0c15d36c6427db5ba316735d/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L428-L432), is it fair to say that there are 2 flavors of direct path, one “basic” direct path and one c2p direct path? And criteria for c2p direct path is `canUseDirectPath() && isDirectPathXdsEnabled()`?
cc. @blakeli0

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcTransportChannel.java
```diff
@@ -60,6 +60,12 @@ public GrpcCallContext getEmptyCallContext() {
   /** The channel in use. */
   abstract ManagedChannel getManagedChannel();
 
+  abstract boolean getUseDirectPath();
```
blakeli0, line 63: This is an `AutoValue` class, we don't have to add another getter for it, just `abstract boolean useDirectPath()` should be good enough. See the [guide for autovalue](go/autovalue#example_java).

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -260,7 +260,8 @@ public TransportChannel getTransportChannel() throws IOException {
   private TransportChannel createChannel() throws IOException {
     return GrpcTransportChannel.create(
```
blakeli0, line 261: Can we change this to use a builder instead of using`create()`?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcTransportChannel.java
```diff
@@ -100,17 +106,23 @@ public void close() {
...
   public static Builder newBuilder() {
-    return new AutoValue_GrpcTransportChannel.Builder();
+    return new AutoValue_GrpcTransportChannel.Builder().setUseDirectPath(false);
   }
 
   public static GrpcTransportChannel create(ManagedChannel channel) {
     return newBuilder().setManagedChannel(channel).build();
   }
 
+  public static GrpcTransportChannel create(ManagedChannel channel, boolean canUseDirectPath) {
```
blakeli0, line 116: Since we already have a buider, I don't think we need to create another helper method.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcTransportChannel.java
zhumin8: According to guide go/autovalue/builders-howto#beans, get prefix or not need to be used consistently. We already have `getManagedChannel()` here, thus follow the same pattern.
The reason for a separate `useDirectPath()` method below is because we need a public method, and adding to the abstract method breaks clirr.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcTransportChannel.java
zhumin8: good point! let me switch to builder.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcTransportChannel.java
blakeli0: CLIRR should be ignored for AutoValue classes. I agree we should use `get/is` prefix consistently, but since this is a boolean, we should not prefix it with `get`. It could be something like `isDirectPath`, `useDirectPath`, `canUseDirectPath`, `directPathEnabled`, `shouldUseDirectPath` etc. Also see relevant discussion [here](https://stackoverflow.com/questions/5322648/for-a-boolean-field-what-is-the-naming-convention-for-its-getter-setter).

In addition, we should only have one method. It was a mistake to have both `getChannel` and `getManagedChannel` IMO.

zhumin8: As discussed offline, I did not realize AutoValue supports "is" prefix for getters. AutoValue does enforce the consistency in using prefix in getters with "you must use these prefixes consistently or not at all." But it **does** allow for get/is prefixes to co-exist.

Updating to `isDirectPath()`/`setDirectPath()`

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcTransportChannel.java
```diff
@@ -60,6 +60,8 @@ public GrpcCallContext getEmptyCallContext() {
   /** The channel in use. */
   abstract ManagedChannel getManagedChannel();
 
+  public abstract boolean isDirectPath();
```
blakeli0, line 63:
```suggestion
  abstract boolean isDirectPath();
```

zhumin8: iiuc, this getter method needs to be accessible outside of the package by Spanner. It needs to be public.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcTransportChannel.java
```diff
@@ -100,7 +102,7 @@ public void close() {
   }
 
   public static Builder newBuilder() {
-    return new AutoValue_GrpcTransportChannel.Builder();
+    return new AutoValue_GrpcTransportChannel.Builder().setDirectPath(false);
```
blakeli0, line 105: Can we add a test case for the default value here?

blakeli0 (approved):

PR #3173, opened 2024-09-06
chore: add a reusable workflow for library generation

JoeWang1127:


.github/scripts/action.yaml
```diff
@@ -0,0 +1,11 @@
+name: Copy library generation script
+description: Copy library generation shell script to the calling workflow
+
+runs:
+  using: "composite"
```
JoeWang1127, line 5: I didn't use reusable workflow because, even though the reusable workflow defined here, the context of the runner is associated with the caller workflow.

When a reusable workflow is triggered by a caller workflow, the `github` context is always associated with the caller workflow. See https://docs.github.com/en/actions/sharing-automations/reusing-workflows#overview

.github/scripts/hermetic_library_generation.sh
```diff
@@ -73,31 +81,23 @@ git checkout "${current_branch}"
...
-  -t gcr.io/cloud-devrel-public-resources/java-library-generation:"${image_tag}" \
-  .
 # run hermetic code generation docker image.
 docker run \
   --rm \
   -u "$(id -u):$(id -g)" \
   -v "$(pwd):${workspace_name}" \
-  -v "$HOME"/.m2:/home/.m2 \
-  -e GENERATOR_VERSION="${generator_version}" \
+  -v "${m2_folder}":/home/.m2 \
```
blakeli0, line 92: Do we still need to volume m2 folder now that generator is baked in?

.github/workflows/hermetic_library_generation.yaml
```diff
@@ -24,24 +24,48 @@ jobs:
...
+      shell: bash
+      run: |
+        git checkout "${HEAD_REF}"
+        mvn install -B -ntp -DskipTests -Dclirr.skip -Dcheckstyle.skip
+      env:
+        HEAD_REF: ${{ github.head_ref }}
+    - name: Get local generator version
+      shell: bash
+      run: |
+        GENERATOR_VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout -pl gapic-generator-java)
```
blakeli0, line 48: Can this step be combined with the step below?

.github/scripts/action.yaml
blakeli0: So looks like we are just reusing the scripts not the workflow? If that's the case, maybe we just bake the scripts into the image?

.github/workflows/hermetic_library_generation.yaml
JoeWang1127: Done.

.github/scripts/hermetic_library_generation.sh
JoeWang1127: I'll try to remove it.

.github/scripts/action.yaml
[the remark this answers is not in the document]
JoeWang1127: The name has to be `action.yaml` or `action.yml`, see https://docs.github.com/en/actions/sharing-automations/creating-actions/metadata-syntax-for-github-actions#about-yaml-syntax-for-github-actions

JoeWang1127:
The following error occurs if I removed the local install.
```
Error: ] Some problems were encountered while processing the POMs:
Error:  Non-resolvable import POM: Could not find artifact com.google.cloud:third-party-dependencies:pom:3.35.1-SNAPSHOT in google-maven-central-copy (https://maven-central.storage-download.googleapis.com/maven2) @ line 61, column 19
 @ 
Error:  The build could not read 1 project -> [Help 1]
Error:    
Error:    The project com.google.api.grpc:google-common-protos-parent:2.44.1-SNAPSHOT (/workspace/java-common-protos/pom.xml) has 1 error
Error:      Non-resolvable import POM: Could not find artifact com.google.cloud:third-party-dependencies:pom:3.35.1-SNAPSHOT in google-maven-central-copy (https://maven-central.storage-download.googleapis.com/maven2) @ line 61, column 19 -> [Help 2]
Error:  
Error:  To see the full stack trace of the errors, re-run Maven with the -e switch.
Error:  Re-run Maven using the -X switch to enable full debug logging.
Error:  
Error:  For more information about the errors and possible solutions, please read the following articles:
Error:  [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/ProjectBuildingException
Error:  [Help 2] http://cwiki.apache.org/confluence/display/MAVEN/UnresolvableModelException
Library postprocessing failed
Error: Process completed with exit code 1.
```

.github/workflows/hermetic_library_generation.yaml
```diff
@@ -24,24 +24,45 @@ jobs:
...
+    - name: Build image
+      shell: bash
+      run: |
+        GENERATOR_VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout -pl gapic-generator-java)
+        echo "GENERATOR_VERSION=${GENERATOR_VERSION}" >> "$GITHUB_ENV"
+        docker build \
+          -f .cloudbuild/library_generation/library_generation.Dockerfile \
+          -t gcr.io/cloud-devrel-public-resources/java-library-generation:"${GENERATOR_VERSION}" \
+          .
     - name: Generate changed libraries
```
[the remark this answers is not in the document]
JoeWang1127: Done

.github/scripts/action.yaml
```diff
@@ -0,0 +1,51 @@
+name: Copy library generation script
+description: Copy library generation shell script to the calling workflow
```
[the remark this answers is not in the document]
JoeWang1127: I updated the name and description.

.github/scripts/action.yaml
```diff
@@ -0,0 +1,51 @@
...
+    description: head branch
+    required: true
+  token:
+    description: PAT
+    required: true
+
+runs:
+  using: "composite"
+  steps:
+  - name: Prerequisite
```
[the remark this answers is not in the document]
JoeWang1127: I updated the step name.

.github/scripts/action.yaml
```diff
@@ -0,0 +1,51 @@
...
+    shell: bash
+    run: |
+      if [[ "${GITHUB_REPOSITORY}" != "${REPO_FULL_NAME}" ]]; then
+        echo "This PR comes from a fork. Skip library generation."
+        exit 0
```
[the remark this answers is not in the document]
JoeWang1127: Do you think we should fail this step?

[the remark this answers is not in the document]
diegomarquezp: I think any way of stopping the workflow without throwing an error would work fine. I added this comment because I think this `exit 0` will be considered as a successful step and [the next step would then start](https://docs.github.com/en/actions/sharing-automations/creating-actions/setting-exit-codes-for-actions#about-exit-codes). Maybe using `GITHUB_OUTPUT` along with `if` in the following steps may do the job ([SO](https://stackoverflow.com/a/61832535/7435861))?

.github/scripts/action.yaml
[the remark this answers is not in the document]
blakeli0: SGTM. Do you mind adding some comment on top of this file regarding how this file is used?

.github/scripts/action.yaml
```diff
@@ -0,0 +1,40 @@
...
+    description: base branch
+    required: true
+  head_ref:
+    description: head branch
+    required: true
+  image_tag:
+    description: the tag of hermetic build image
+    required: false
+  token:
+    description: PAT
```
blakeli0, line 14: What is `PAT`?

.github/scripts/hermetic_library_generation.sh
blakeli0: So .m2 is still needed?

.github/scripts/action.yaml
JoeWang1127: Personal access token.

.github/scripts/hermetic_library_generation.sh
JoeWang1127: If I removed this line, the build failed, please see https://github.com/googleapis/sdk-platform-java/pull/3173#issuecomment-2341588555

diegomarquezp: Looks like this is a special case of sdk-platform-java because it depends on snapshot jars

.github/scripts/action.yaml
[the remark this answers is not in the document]
JoeWang1127: I added a separate job to determine whether to run the library generation.

If the pull request comes from a fork, the generation will be skipped.

.github/scripts/action.yaml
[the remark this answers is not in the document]
JoeWang1127: Done.

.github/scripts/action.yaml
blakeli0: Can we use the full name instead of abbreviations since it's not common? See [style guide](https://developers.google.com/style/abbreviations#creating-abbreviations).

diegomarquezp (approved):

.github/scripts/action.yaml
JoeWang1127: Changed to full name.

.github/scripts/hermetic_library_generation.sh
blakeli0: From offline discussion, we are dependent on [mvn fmt:format](https://github.com/googleapis/sdk-platform-java/blob/7996aabf90d1b5a35ad1ff32afab3c522572d959/library_generation/owlbot/bin/entrypoint.sh#L83), hence a local .m2 volumn is needed. This will be baked into the docker image soon.

.github/workflows/hermetic_library_generation.yaml
```diff
@@ -24,24 +24,41 @@ jobs:
   library_generation:
     runs-on: ubuntu-latest
     steps:
+    - name: Determine whether the pull request comes from a fork
+      run: |
+        if [[ "${GITHUB_REPOSITORY}" != "${REPO_FULL_NAME}" ]]; then
+          echo "This PR comes from a fork. Skip library generation."
+          echo "SHOULD_RUN=false" >> $GITHUB_ENV
+        else
+          echo "SHOULD_RUN=true" >> $GITHUB_ENV
```
blakeli0, line 33: Is there a way to exit the workflow early so we don't have to pass this env to all the subsequent steps?

blakeli0 (approved):

.github/workflows/hermetic_library_generation.yaml
JoeWang1127: Looks like another way of skipping steps is creating a separate job and make the generation job depends on that, just like what we've done in unit/integration tests.

I'll keep it as-is now.

PR #3186, opened 2024-09-10
feat(gax): append cred-type header for auth metrics

zhumin8:
This is POC change in gax-java for auth metrics requirements on token usage. See go/googleapis-auth-metric-design for context.

[Credentials](https://github.com/googleapis/google-auth-library-java/blob/main/credentials/java/com/google/auth/Credentials.java) will expose `getMetricsCredentialType()` method, this change appends it to existing `x-goog-api-client` header

Note: Currently implement in gax at client level. There are 2 edge cases not covered and will create followups for: if handwritten library overrides credentials at rpc level; If handwritten library does not build on gax. (ref: b/370039645, b/370038458)

related change in `google-auth-library`https://github.com/googleapis/google-auth-library-java/pull/1503 included in 1.28.0.


gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
```diff
@@ -178,6 +178,7 @@ public static ClientContext create(StubSettings settings) throws IOException {
 
     String settingsGdchApiAudience = settings.getGdchApiAudience();
     Credentials credentials = settings.getCredentialsProvider().getCredentials();
+    String credentialType = credentials.getCredentialType();
```
blakeli0, line 181: What are the possible credential types? cc: @ldetmer as it may affect the API key project.

zhumin8: current metric design includes 4 types: u - UserCredentials, sa - ServiceAccountCredentials, mds - ComputeEngineCredentials, imp - ImpersonatedCredentials

gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
```diff
@@ -322,6 +325,9 @@ private static Map<String, String> getHeadersFromSettings(StubSettings settings)
     effectiveHeaders.putAll(userHeaders);
     effectiveHeaders.putAll(conflictResolution);
 
+    effectiveHeaders.computeIfPresent(
```
blakeli0, line 328: I think handwritten libraries and/or customers can override credentials on request level, can we do more investigation? Maybe we can add this header on request level not client level.

zhumin8: Let me look into this.

zhumin8:
downstream check for java-bigquerystorage is failing because java-bigquerystorage has hardcoded versions for auth libraries. raising https://github.com/googleapis/java-bigquerystorage/pull/2679 to fix.

gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
zhumin8: Update on this: Consider this is a relatively niche case, to unblock this current change, I will look into this separately and create a followup if needed. Updated description to reflect.

gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
```diff
@@ -210,7 +211,12 @@ public static ClientContext create(StubSettings settings) throws IOException {
     if (transportChannelProvider.needsExecutor() && settings.getExecutorProvider() != null) {
       transportChannelProvider = transportChannelProvider.withExecutor(backgroundExecutor);
     }
-    Map<String, String> headers = getHeadersFromSettings(settings);
+    CredentialTypeForMetrics credentialTypeForMetrics =
+        credentials == null
+            ? CredentialTypeForMetrics.DO_NOT_SEND
+            : credentials.getMetricsCredentialType();
+    Map<String, String> headers =
+        getHeadersFromSettingsAndAppendCredentialType(settings, credentialTypeForMetrics);
```
blakeli0, line 219: This method can just be `getHeaders`, where we can pass `credentials` into the method, and encapsulate all credentials metric related logic into one method.

gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
```diff
@@ -346,6 +353,11 @@ private static Map<String, String> getHeadersFromSettings(StubSettings settings)
     effectiveHeaders.putAll(userHeaders);
     effectiveHeaders.putAll(conflictResolution);
 
+    if (credentialTypeForMetrics != CredentialTypeForMetrics.DO_NOT_SEND) {
+      effectiveHeaders.computeIfPresent(
+          ApiClientHeaderProvider.getDefaultApiClientHeaderKey(),
```
blakeli0, line 358: Is it decided to append `cred-type/` to `x-goog-api-client` header? It seems all other header tokens in `x-goog-api-client` are static values from the client, not from user input. I wonder why we didn't choose to use a separate header key for this purpose?

zhumin8: yes, this design is already implemented for other languages. See details in design [go/googleapis-auth-metric-design](http://goto.google.com/googleapis-auth-metric-design) section 2.3.2. I don't think I saw discussion on using a different header specifically though.

zhumin8: I see your point. 
Let me know if you see better suited places for additional docs.

blakeli0 (approved):

PR #3188, opened 2024-09-10
chore: Add Clirr exemptions for Protobuf 4.27.4 runtime

lqiu96:
Clirr is flagging a bunch of 7006 Method Return Type changed errors. The Protoc generated changes are source code compatible and should ignore these errors in our generated code.

See [this doc](https://docs.google.com/spreadsheets/d/1qEzUWLougtiqh0ocRaG4e8w12pr89o0T_tQvB6YDStE/edit?pli=1&gid=0#gid=0) for more information. After adding all the Clirr checks exemptions in this PR to the downstream handwritten libraries, we are no longer seeing an Clirr failures.

### DIFF
--- a/library_generation/owlbot/templates/clirr/clirr-ignored-differences.xml.j2
+++ b/library_generation/owlbot/templates/clirr/clirr-ignored-differences.xml.j2
@@ -15,5 +15,72 @@
     <differenceType>7012</differenceType>
     <className>{{proto_path}}/*OrBuilder</className>
     <method>boolean has*(*)</method>
+  </difference>
+  <!-- The following 7006 exemptions are related to Protobuf 4.27.4 upgrade -->
+  <difference>
+    <differenceType>7006</differenceType>
+    <className>{{proto_path}}/**</className>
+    <method>* getDefaultInstanceForType()</method>
+    <to>**</to>
+  </difference>
+  <difference>
+    <differenceType>7006</differenceType>
+    <className>{{proto_path}}/**</className>
+    <method>* addRepeatedField(*)</method>
+    <to>**</to>
+  </difference>
+  <difference>
+    <differenceType>7006</differenceType>
+    <className>{{proto_path}}/**</className>
+    <method>* clear()</method>
+    <to>**</to>
+  </difference>
+  <difference>
+    <differenceType>7006</differenceType>
+    <className>{{proto_path}}/**</className>
+    <method>* clear()</method>
+    <to>**</to>
+  </difference>
+  <difference>
+    <differenceType>7006</differenceType>
+    <className>{{proto_path}}/**</className>
+    <method>* clearField(*)</method>
+    <to>**</to>
+  </difference>
+  <difference>
+    <differenceType>7006</differenceType>
+    <className>{{proto_path}}/**</className>
+    <method>* clearOneof(*)</method>
+    <to>**</to>
+  </difference>
+  <difference>
+    <differenceType>7006</differenceType>
+    <className>{{proto_path}}/**</className>
+    <method>* clone()</method>
+    <to>**</to>
+  </difference>
+  <difference>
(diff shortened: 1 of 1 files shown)


blakeli0 (commented):
The content LGTM, but I'm not sure how/if this change is going to propagated to each library. For example, we are triggering a regeneration in this PR, but there is [no generation change detected](https://github.com/googleapis/sdk-platform-java/actions/runs/10816931316/job/30009186187?pr=3188#step:3:35806) for java-common-protos and java-iam. 

@diegomarquezp @JoeWang1127 Do you know if this template is used at all?

lqiu96:
> The content LGTM, but I'm not sure how/if this change is going to propagated to each library. For example, we are triggering a regeneration in this PR, but there is [no generation change detected](https://github.com/googleapis/sdk-platform-java/actions/runs/10816931316/job/30009186187?pr=3188#step:3:35806) for java-common-protos and java-iam.
> 
> @diegomarquezp @JoeWang1127 Do you know if this template is used at all?

Oh, I see. If these templates are not being used, then I can also manually make these changes in each handwritten library.

JoeWang1127:
Looking at code in `library_generation/owlbot/bin/write_clirr_ignore.sh`:
```
# restore default clirr-ignored-differences.xml for protos if the file does not exist
for dir in `ls -d proto-google-*`
do
  if [ ! -f "${dir}/clirr-ignored-differences.xml" ]
  then
    tmp_dir=$(mktemp -d -t ci-XXXXXXXXXX)
    pushd ${dir}
    pushd src/main/java
    find * -name *OrBuilder.java | xargs dirname | sort -u | jq -Rns ' (inputs | rtrimstr("\n") | split("\n") ) as $data | {proto_paths: $data}' > ${tmp_dir}/paths.json
    popd
    python3 "${scripts_root}/owlbot/src/gen-template.py" --data=${tmp_dir}/paths.json --folder=${templates_dir}
    popd
  fi
done
```

The clirr file will be generated if only didn't exist before.

lqiu96:
Hmm, thanks for the link above. I guess it does make sense to only generate if the clirr file doesn't exist. I guess it is possible for new changes to overwrite the existing clirr configurations that handwritten teams may have added.

@blakeli0 @JoeWang1127 @diegomarquezp Do you have any concerns if we merge this in so that new libraries won't have any Clirr issues when generated with Protoc 3.25.x and Runtime 4.27.4+. I can go through all our handwritten (and potentially existing gapic libraries) and add these new clirr exemptions?

JoeWang1127:
> Do you have any concerns if we merge this in so that new libraries won't have any Clirr issues when generated with Protoc 3.25.x and Runtime 4.27.4+.

LGTM.

However, do we need to revert the change when updating protoc to 4.x?

lqiu96:
I don't think so. IIUC, Clirr runs against the previous released versions and these exemptions are for runtime upgrade from 3 -> 4 and I think they can continue to stay. I think when we do update from protoc 3.x to 4.x, we may need to include a new set of clirr exemption rules though.

JoeWang1127 (approved):
Can you add a comment to remind us to revert the PR if, after protoc major version updates, the build succeeds without the change.

blakeli0:
> Do you have any concerns if we merge this in so that new libraries won't have any Clirr issues when generated with Protoc 3.25.x and Runtime 4.27.4+

SGTM. We definitely need it for new libraries in google-cloud-java. FYI, @JoeWang1127 is looking into maybe always generate this file in the future.

blakeli0 (approved):

lqiu96:
> Can you add a comment to remind us to revert the PR if, after protoc major version updates, the build succeeds without the change.

Do old Clirr exemptions need to be removed for clirr to run properly? Can this just live inside the template even if it's not relevant any more.

Or is this mostly just a cleanup for the clirr file?

PR #3190, opened 2024-09-11
chore: do not use renovate to update protobuf

JoeWang1127:
In this PR:
- Do not use renovate to update protobuf version in `gax-java/dependencies.properties`.

We need to make sure the protobuf version in `gax-java/dependency.properties` (used by the self-service client) is compatible with protoc/protobuf version defined in the generator. 

We'll manually change the version.

Verification:

```
export LOG_LEVEL=debug && npx --yes --package renovate -- renovate --platform=local

...
{
   "datasource": "maven",
   "depName": "com.google.protobuf:protobuf-bom",
   "currentValue": "3.25.4",
   "fileReplacePosition": 1523,
   "registryUrls": [
     "https://maven-central.storage-download.googleapis.com/maven2/",
     "https://maven-central.storage-download.googleapis.com/maven2",
     "https://repo1.maven.org/maven2",
     "https://repo.maven.apache.org/maven2"
   ],
   "depType": "import",
   "groupName": "protobuf.version",
   "editFile": "gapic-generator-java-pom-parent/pom.xml",
   "updates": [],
   "packageName": "com.google.protobuf:protobuf-bom",
   "skipReason": "disabled"
},
```

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -21,17 +21,6 @@
     {
       "customType": "regex",
       "fileMatch": [
-        "^gax-java/dependencies\\.properties$"
-      ],
-      "matchStrings": [
-        "version\\.com_google_protobuf=(?<currentValue>.+?)\\n"
-      ],
-      "depNameTemplate": "com.google.protobuf:protobuf-java",
-      "datasourceTemplate": "maven"
-    },
-    {
-      "customType": "regex",
-      "fileMatch": [
                "^\\.cloudbuild/library_generation/library_generation\\.Dockerfile$"
       ],
       "matchStrings": [
--- a/renovate.json
+++ b/renovate.json
@@ -21,18 +21,6 @@
     {
       "customType": "regex",
       "fileMatch": [
-               "^\\.cloudbuild/library_generation/library_generation\\.Dockerfile$"
-      ],
-      "matchStrings": [
-        "ARG PROTOC_VERSION=[\"']?(?<currentValue>.+?)[\"']?\\s+"
-      ],
-      "datasourceTemplate": "github-releases",
-      "depNameTemplate": "protocolbuffers/protobuf",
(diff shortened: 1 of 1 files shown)


lqiu96 (approved):

suztomo (changes requested):
This change seems to detach the version upgrade between the two files.

suztomo:
How about turning off Renovate for Protobuf?

JoeWang1127:
> This change seems to detach the version upgrade between the two files.

Do you mean `gax-java/dependencies.properties` and `gapic-generator-java-pom-parent/pom.xml`?

blakeli0:
> This change seems to detach the version upgrade between the two files.

I think we should detach the version upgrade between `dependencies.properties` and `Dockerfile`/`pom`. Because `dependencies.properties` is only used by self-service, and it should be in sync with the protoc version in WORKSPACE, which is not fully controlled by us. While the versions in `Dockerfile`/`pom` are used by full service(hermetic build), and it is fully controlled by us.

blakeli0:
> How about turning off Renovate for Protobuf?

I can see the benefit of still using renovate bot as a hint to us that there is new version for prorobuf, instead of us manually checking if there is a newer version for protobuf. But eventually it would need manual change to upgrade, so I think either way is fine as long as we have a regular schedule to upgrade protobuf.

JoeWang1127:
> using renovate bot as a hint to us that there is new version for prorobuf

The renovate pr can't bring code change since it comes from a fork, so I think it's fine to remove it from renovate.

suztomo (approved):

lqiu96 (approved):
LGTM. I think the title is probably outdated now, right?

JoeWang1127:
> LGTM. I think the title is probably outdated now, right?

Thanks. Changed the title.

PR #3196, opened 2024-09-12
chore: [wip] secure hermetic-build docker image

diegomarquezp:
Not ready for review


.cloudbuild/library_generation/library_generation.Dockerfile
```diff
@@ -11,107 +11,149 @@
...
+# actively scan filesystem interactions such as fs.readFileSync). This makes the
+# app to fetch a file at runtime that is not available in the bundle context.
+# This is why we remove this import and its usage from the entrypoint.
+RUN sed -i '/testWebhook/d' src/bin/owl-bot.ts
+
+# Bundle the source code and its dependencies into a single javascript file
+# with all its dependencies embedded.
+# This is because SEA (see below) cannot
+# resolve external modules in a multi-file project.
+# We use the esbuild tool see https://esbuild.github.io/
```
diegomarquezp, line 39:
```suggestion
# We use the esbuild tool. See https://esbuild.github.io/
```

.cloudbuild/library_generation/library_generation.Dockerfile
```diff
@@ -11,107 +11,197 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+# Creates the owl-bot binary (no node runtime needed)
 
-# install gapic-generator-java in a separate layer so we don't overload the image
-# with the transferred source code and jars
-FROM gcr.io/cloud-devrel-public-resources/java21 AS ggj-build
+# node:22.1-alpine
+FROM us-docker.pkg.dev/artifact-foundry-prod/docker-3p-trusted/node@sha256:487dc5d5122d578e13f2231aa4ac0f63068becd921099c4c677c850df93bede8 as owlbot-cli-build
```
JoeWang1127, line 17: Can we use a version number instead of sha?

.cloudbuild/library_generation/library_generation.Dockerfile
```diff
@@ -11,107 +11,197 @@
...
+# resolve external modules in a multi-file project.
+# We use the esbuild tool. See https://esbuild.github.io/
+COPY ./.cloudbuild/library_generation/image-configuration/owlbot-cli-build-config.js .
+RUN npm i esbuild
+RUN node owlbot-cli-build-config.js
+
+# Compile the bundled javascript file into a Linux executable
+# Create a Standalone Executable Application (SEA) configuration file.
+# See https://nodejs.org/api/single-executable-applications.html
+RUN echo '{ "main": "bundle.js", "output": "sea-prep.blob" }' > build/sea-config.json
```
JoeWang1127, line 49: What is the benefits of converting owlbot-cli to a SEA compared to copying source to image?

.cloudbuild/library_generation/library_generation.Dockerfile
```diff
@@ -11,107 +11,197 @@
...
+RUN sed -i '/testWebhook/d' src/bin/owl-bot.ts
+
+# Bundle the source code and its dependencies into a single javascript file
+# with all its dependencies embedded.
+# This is because SEA (see below) cannot
+# resolve external modules in a multi-file project.
+# We use the esbuild tool. See https://esbuild.github.io/
+COPY ./.cloudbuild/library_generation/image-configuration/owlbot-cli-build-config.js .
+RUN npm i esbuild
+RUN node owlbot-cli-build-config.js
```
JoeWang1127, line 44: In order to create a SEA, do we have to bundle source code into a single js file?

.cloudbuild/library_generation/library_generation.Dockerfile
```diff
@@ -11,107 +11,197 @@
...
 
-# build from the root of this repo:
-FROM gcr.io/cloud-devrel-public-resources/python
+RUN --mount=type=cache,target=/root/.m2 cp "/root/.m2/repository/com/google/api/gapic-generator-java/${DOCKER_GAPIC_GENERATOR_VERSION}/gapic-generator-java-${DOCKER_GAPIC_GENERATOR_VERSION}.jar" \
+  "/gapic-generator-java.jar"
 
-SHELL [ "/bin/bash", "-c" ]
+# Builds the python scripts in library_generation
+# python:3.11-alpine
+FROM us-docker.pkg.dev/artifact-foundry-prod/docker-3p-trusted/python@sha256:0b5ed25d3cc27cd35c7b0352bac8ef2ebc8dd3da72a0c03caaf4eb15d9ec827a as python-scripts-build
```
JoeWang1127, line 79: Same as above, can we use version num instead of sha?

.cloudbuild/library_generation/library_generation.Dockerfile
```diff
@@ -11,107 +11,197 @@
...
 
+# This will use GOOGLE_APPLICATION_CREDENTIALS if passed in docker build command.
+# If not passed will leave it unset to support GCE Metadata in CI builds
+ARG GOOGLE_APPLICATION_CREDENTIALS
+
+RUN apk add bash curl
+
+# Install gcloud to obtain the credentials to use the Airlock repostiory
+RUN curl -sSL https://sdk.cloud.google.com | bash -e
```
JoeWang1127, line 88: Which version of the gcloud?

library_generation/DEVELOPMENT.md
```diff
@@ -129,9 +129,23 @@ This is convenient in order to avoid installing the dependencies manually.
...
 > folder.
 
+## Prepare your gcloud Application-Default-Credentials
+This is necessary for the build context to access the Airlock repository of
+Python packages.
+To configure your credentials:
+
+```bash
+# creates or updates the credentials file in ~/.config/gcloud
+gcloud auth application-default-login
```
JoeWang1127, line 139:
```suggestion
gcloud auth application-default login 
```

JoeWang1127:
I think this PR has done multiple things, change base image and optimize image building.

Can we modify the base image first and create a new PR for optimization?

.cloudbuild/library_generation/library_generation.Dockerfile
diegomarquezp: Although using a version string is more convenient and readable, using the SHA is [recommended](https://g3doc.corp.google.com/net/proto2/g3doc/releasing/dev/docker/container_tips.md#do-use-sha-digest-based-container-image-paths-whenever-and-wherever-possible) since the tag can be manipulated and be made to point to a different version.
I added the version string above each stage for convenience as well.

.cloudbuild/library_generation/image-configuration/airlock-pypirc
```diff
@@ -0,0 +1,9 @@
+[distutils]
+index-servers = ah-3p-staging-python
+# TODO: use this index instead when `lxml` and `versions` are available in the `trusted` airlock registry
+# index-servers = python-3p-trusted
```
JoeWang1127, line 4: Could you point to which doc/webpage we should look at whether it's available?

.cloudbuild/library_generation/library_generation.Dockerfile
```diff
@@ -11,107 +11,197 @@
...
 # {x-version-update-start:gapic-generator-java:current}
 ENV DOCKER_GAPIC_GENERATOR_VERSION="2.45.1-SNAPSHOT" 
 # {x-version-update-end}
 
-RUN mvn install -B -ntp -DskipTests -Dclirr.skip -Dcheckstyle.skip
-RUN cp "/root/.m2/repository/com/google/api/gapic-generator-java/${DOCKER_GAPIC_GENERATOR_VERSION}/gapic-generator-java-${DOCKER_GAPIC_GENERATOR_VERSION}.jar" \
-  "./gapic-generator-java.jar"
+# use Docker Buildkit caching for faster local builds
+RUN --mount=type=cache,target=/root/.m2 mvn install -B -ntp -T 1.5C \
+    -DskipTests -Dclirr.skip -Dcheckstyle.skip -Djacoco.skip -Dmaven.test.skip \
```
JoeWang1127, line 71: Why we need both `-DskipTests` and `-Dmaven.test.skip`?

.cloudbuild/library_generation/library_generation.Dockerfile
```diff
@@ -11,107 +11,197 @@
...
+WORKDIR /src
+
+RUN --mount=type=secret,id=credentials python -m pip install --target /usr/local/lib/python3.11 -r requirements.txt
+RUN python -m pip install --target /usr/local/lib/python3.11 .
+
+# Final image. Installs the rest of the dependencies and gets the binaries
+# from the previous stages. We use the node base image for it to be compatible
+# with the standalone binary owl-bot compiled in the previous stage
+# node:22.1-alpine
+FROM us-docker.pkg.dev/artifact-foundry-prod/docker-3p-trusted/node@sha256:487dc5d5122d578e13f2231aa4ac0f63068becd921099c4c677c850df93bede8 as final
```
JoeWang1127, line 110: Why can't we use a python base image?

.cloudbuild/library_generation/library_generation.Dockerfile
diegomarquezp: Yes, bundling is necessary:
https://github.com/googleapis/sdk-platform-java/blob/fb98222ddd61aa5b0e5e061dd8e4ffcbc12fdea2/.cloudbuild/library_generation/library_generation.Dockerfile#L39-L40

From the [docs](https://nodejs.org/api/single-executable-applications.html): _The single executable application feature currently only supports running a single embedded script using the CommonJS module system._

> Can be combine the two steps?
Do you mean using a single `RUN` + `&&`?

.cloudbuild/library_generation/cloudbuild-library-generation-push.yaml
JoeWang1127, line 1: Can we change the image name in this file?

.cloudbuild/library_generation/library_generation.Dockerfile
diegomarquezp: Tracked in https://github.com/googleapis/sdk-platform-java/pull/3196#discussion_r1763373017

.cloudbuild/library_generation/image-configuration/airlock-pypirc
```diff
@@ -0,0 +1,9 @@
...
+
+[ah-3p-staging-python]
+repository: https://us-python.pkg.dev/artifact-foundry-prod/ah-3p-staging-python/
+# TODO: use this repository instead when `lxml` and `versions` are available in the `trusted` airlock registry
+repository: https://us-python.pkg.dev/artifact-foundry-prod/python-3p-trusted/
```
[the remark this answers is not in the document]
diegomarquezp: Yes, indeed, thanks. I commented it out.

.cloudbuild/library_generation/library_generation.Dockerfile
diegomarquezp: Actually `skipTests` is redundant since it only skips the tests execution, whereas `test.skip` skips both the compilation and execution. I removed it. Thanks for the catch.

.cloudbuild/library_generation/cloudbuild-library-generation-push.yaml
diegomarquezp: Is this about the `_IMAGE_ID` of the image? We can change it, but how can we improve its name?

.cloudbuild/library_generation/library_generation.Dockerfile
diegomarquezp: The standalone owlbot seems to need a few runtime libraries (`.so`) in the linux environment that are available in the node image by default (besides the node runtime). The python image doesn't have them. 
If we want to have a python image as base, I can try to use a python base and install node or whichever libraries the executable needs.

diegomarquezp: From discussion with @blakeli0 and @JoeWang1127, let's use a vanilla Alpine based image.

.cloudbuild/library_generation/library_generation.Dockerfile
JoeWang1127: I was referring whether we can build a SEA from multiple js source code, looks like it has to be bundled together first.

.cloudbuild/library_generation/cloudbuild-library-generation-push.yaml
JoeWang1127: > Is this about the `_IMAGE_ID` of the image?

Yes.

Can we create a image repo in AR?

.cloudbuild/library_generation/image-configuration/airlock-pypirc
diegomarquezp: Yes, I added a link to the package lookup tool.

.cloudbuild/library_generation/library_generation.Dockerfile
blakeli0, line 17: How do we plan to update these base images? It might be fine to not update this one, but for the Java and Python one, we may want to update the Maven/JDK/Python version regularly.

.cloudbuild/library_generation/library_generation.Dockerfile
```diff
@@ -11,107 +11,197 @@
...
 WORKDIR /sdk-platform-java
 COPY . .
 # {x-version-update-start:gapic-generator-java:current}
 ENV DOCKER_GAPIC_GENERATOR_VERSION="2.45.1-SNAPSHOT" 
 # {x-version-update-end}
 
-RUN mvn install -B -ntp -DskipTests -Dclirr.skip -Dcheckstyle.skip
-RUN cp "/root/.m2/repository/com/google/api/gapic-generator-java/${DOCKER_GAPIC_GENERATOR_VERSION}/gapic-generator-java-${DOCKER_GAPIC_GENERATOR_VERSION}.jar" \
-  "./gapic-generator-java.jar"
+# use Docker Buildkit caching for faster local builds
```
blakeli0, line 69: What does `Docker Buildkit caching` mean? Since the build does not take too much time(~1.5 minute), I think it's OK to download the maven dependencies every time. What I'm trying to avoid is that we build an image with stale dependencies.

blakeli0 (commented):
> We can add a secret in the repo or migrate this action to Cloud Build.

Agree that we should migrate them to Cloud Build, same as the integration tests. This gets us closer to not dependent on Github specific features.

.cloudbuild/library_generation/library_generation.Dockerfile
diegomarquezp: The Buildkit caching allows to reuse the specified folders in the specific step. This would not impact any CI pipeline as they don't preserve any kind of cache. The main purpose is to speed up local builds for development, and can be disabled via `docker build --no-cache`.

My perspective is that we probably won't work on modifying the java source code when working on the Docker image, so several image builds may benefit from caching the `mvn install` output.

.cloudbuild/library_generation/library_generation.Dockerfile
diegomarquezp: Good point, I'll figure this out in the follow up (enhancements) since it is now out of scope.

.cloudbuild/library_generation/cloudbuild-library-generation-push.yaml
diegomarquezp: What if we leave the changes related to the AR migration as a follow up? Should be a small PR.

PR #3203, opened 2024-09-16
chore: do not exit early if there's no code change

JoeWang1127:
In this PR:
- Do not exit if there's no code change.

Context: we found an issue in bigtable (https://github.com/googleapis/java-bigtable/pull/2336) such that PR description is overwritten by subsequent commit due to config updates. However, the generation associated with that [commit](https://github.com/googleapis/java-bigtable/pull/2336/commits/b6e7e1d566c4f2364e3eb8f766fec31c396ded57) didn't have code change (because the code change is already committed) and exit early so the PR description is not changed back.


blakeli0 (approved):

diegomarquezp:
IIUC https://github.com/googleapis/java-firestore/pull/1787 went through the same problem?

PR #3208, opened 2024-09-17
feat: generate a GAPIC library from api definition

JoeWang1127:
In this PR:
- Generate a GAPIC library from api definition (proto, service yaml), rather than from googleapis repository.

To generate a GAPIC library with hermetic build image, the user has to prepare the api definition to a directory and set `--api-definition-path` to this directory.

This feature allows the user to generate a GAPIC library from any protos, rather than checking in protos in googleapis repository.

Internal ticket: b/362705386


library_generation/test/integration_tests.py
```diff
@@ -40,21 +41,21 @@
 repo_prefix = "https://github.com/googleapis"
 output_dir = shell_call("get_output_folder")
 # this map tells which branch of each repo should we use for our diff tests
-committish_map = {
+commitish_map = {
```
diegomarquezp, line 44: Fun fact: for a long time I thought it _only_ was `committish` (two `t`s). 
I happens to be an alternative convenience term coming from `commit-ish`.


The [git glossary](https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefcommit-ishacommit-ishalsocommittish) offers either `commit-ish` or `committish`. I always thought this would be too much to start a campaign to update all config yamls and its definition.

diegomarquezp (approved):

library_generation/cli/entry_point.py
```diff
@@ -63,10 +63,21 @@ def main(ctx):
...
+    help="""
+    The path to which the api definition (proto and service yaml) resides.
+    If not specified, the path is the current working directory.
+    """,
+)
 def generate(
     baseline_generation_config_path: str,
     current_generation_config_path: str,
     repository_path: str,
+    api_definition_path: str,
```
blakeli0, line 80: I think this is more like a root folder of all api definitions. e.g. The [proto_path](https://github.com/googleapis/sdk-platform-java/blob/main/generation_config.yaml#L40) configure in generation_config.yaml is a relative path to this path right?

.github/scripts/hermetic_library_generation.sh
```diff
@@ -84,16 +84,29 @@ git show "${target_branch}":"${generation_config}" > "${baseline_generation_conf
...
+git checkout "${googleapis_commitish}"
+popd
+
 # run hermetic code generation docker image.
 docker run \
   --rm \
   -u "$(id -u):$(id -g)" \
   -v "$(pwd):${workspace_name}" \
   -v "${m2_folder}":/home/.m2 \
+  -v "${api_def_dir}:${workspace_name}/api" \
```
blakeli0, line 101: This is essentially googleapis folder, maybe we can just use `/googleapis` to make it easier to understand?

library_generation/README.md
```diff
@@ -48,6 +48,20 @@ right version for each library.
 Please refer [here](go/java-client-releasing#versionstxt-manifest) for more info
 of versions.txt.
 
+### Api definition path (`api_definition_path`), optional
+
+The path to where the api definition (proto, service yaml) resides.
+
+The default value is the current working directory when running the script.
+
+Note that you need not only the protos defined the service, but also the transitive
```
blakeli0, line 57: What does the error look like if the dependencies are missing?

library_generation/test/integration_tests.py
blakeli0: Thanks Diego! That's good to know! Yeah I agree we should follow the naming conventions. Maybe an enhancement in the backlog.

blakeli0:
Thanks Joe! I think this is a good first step, but it still requires us to have a generation_config.yaml and parsing googleapis committish from it. As a follow up, can we try to see how to generate a library with only a folder? Not relying on generation_config.yaml?

.github/scripts/hermetic_library_generation.sh
JoeWang1127: I try to use a generalized name here to specify that the script can generate from any protos, not necessarily from googleapis.

library_generation/README.md
JoeWang1127: This is the error message when I remove `google/api` folder in [this commit
](https://github.com/googleapis/sdk-platform-java/pull/3208/commits/69e23dfa069633788bc47fb620beee48d3e1cf9f)

```
google/api/field_behavior.proto: File not found.
google/api/resource.proto: File not found.
google/cloud/alloydb/v1/resources.proto:19:1: Import "google/api/field_behavior.proto" was not found or had errors.
google/cloud/alloydb/v1/resources.proto:20:1: Import "google/api/resource.proto" was not found or had errors.
```

JoeWang1127: Also updated the doc.

JoeWang1127:
> As a follow up, can we try to see how to generate a library with only a folder? Not relying on generation_config.yaml?

Sure, I'll create a follow up PR to remove `googleapis_commitish` in the generation config.

Generation config can be optional after all necessary parameters are read from service yaml.

.github/scripts/hermetic_library_generation.sh
blakeli0: The docker image can be used by any protos, but I think `hermetic_library_generation.sh` is only going to be used for generating googleapis. For generating other protos, the volumn name actually doesn't really matter, it's good as long as the volumn and api-definition-path match.
That being said, it's fine if you want to make it generic, but maybe change it to `apis` to imply that it could contain multiple.

library_generation/DEVELOPMENT.md
```diff
@@ -144,16 +152,21 @@ repo to this folder).
...
    yourself. This avoids folder ownership changes since it runs as root by
    default.
- * `-v/path/to/google-cloud-java:/workspace` maps the host machine's
-   google-cloud-java folder to the /workspace folder. The image is configured to
-   perform changes in this directory
- * `$(cat image-id)` obtains the image ID created in the build step
+ * `-v /path/to/google-cloud-java:/workspace` maps the host machine's
+   google-cloud-java folder to the /workspace folder. 
+   The image is configured to perform changes in this directory.
+   Note that this setup assumes the api definition resides in host machine's
```
blakeli0, line 167: Can we change this example to use the `api_definition_path` parameter?

.github/scripts/hermetic_library_generation.sh
JoeWang1127: > but I think hermetic_library_generation.sh is only going to be used for generating googleapis

Makes sense, I changed it to `googleapis`.

library_generation/DEVELOPMENT.md
JoeWang1127: Done.

library_generation/cli/entry_point.py
blakeli0: Maybe change this to `api_definitions_path` or `api_defs_path`?

JoeWang1127: Changed it to `api_definitions_path`.

blakeli0 (approved):

PR #3227, opened 2024-09-23
chore: migrate java tests

JoeWang1127:
In this PR:
- Migrate `test_language_java.py` and test resources from synthtool.
- Remove synthtool installation step in CI.
- Remove unused methods in `java.py`.
- Refactor code to align with the current coding style.


library_generation/owlbot/templates/java_library/.github/workflows/update_generation_config.yaml
```diff
@@ -18,7 +18,7 @@ on:
   schedule:
   - cron: '0 2 * * *'
   workflow_dispatch:
-
+{% raw %}
```
JoeWang1127, line 21: We don't want to render `${{ secrets.CLOUD_JAVA_BOT_TOKEN }}`, use `raw` to ignore the variable.

library_generation/test/resources/test-owlbot/testdata/multi-config.yaml
blakeli0, line 1: Is this file still used in unit tests? Same question for all other test files like `sponge-log-golden.xml`. I don't see them being used so they should've probably be removed from synthtool long time ago.  Can we only keep the ones used in unit tests?

JoeWang1127: Thanks for the suggestion.

I removed unused test resources.

blakeli0 (approved):

PR #3245, opened 2024-09-26
chore: Add OpenTelemetry incubating semantic conventions to shared dependencies

michaelpri10:
Follow up to https://github.com/googleapis/sdk-platform-java/pull/3020 but adding the `opentelemetry-semconv-incubating` dependency which lives in the [same repository](https://github.com/open-telemetry/semantic-conventions-java/tree/main/semconv-incubating) as the previously added `opentelemetry-semconv` dependency. Approval for that dependency is at b/351882514. This allows us to use the [`MessagingIncubatingAttributes`](https://github.com/open-telemetry/semantic-conventions-java/blob/main/semconv-incubating/src/main/java/io/opentelemetry/semconv/incubating/MessagingIncubatingAttributes.java) class which provides some GCP specific semantic conventions for OpenTelemetry.

### DIFF
--- a/java-shared-dependencies/third-party-dependencies/pom.xml
+++ b/java-shared-dependencies/third-party-dependencies/pom.xml
@@ -145,6 +145,11 @@
         <version>${opentelemetry-semconv.version}</version>
       </dependency>
       <dependency>
+        <groupId>io.opentelemetry.semconv</groupId>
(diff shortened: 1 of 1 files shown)


jinseopkim0:
Could you please create a new buganizer issue for approval?

PR #3271, opened 2024-10-03
chore: bake Java formatter into image

JoeWang1127:
In this PR:
- Bake Java formatter into image.


.cloudbuild/library_generation/library_generation.Dockerfile
```diff
@@ -35,6 +35,7 @@ SHELL [ "/bin/bash", "-c" ]
 ARG OWLBOT_CLI_COMMITTISH=ac84fa5c423a0069bbce3d2d869c9730c8fdf550
 ARG PROTOC_VERSION=25.5
 ARG GRPC_VERSION=1.67.1
+ARG JAVA_FORMAT_VERSION=1.7
```
JoeWang1127, line 38: I think we can upgrade the formatter, the latest version is 1.23.0.

It will cause code formatting so it's better in a separate PR.

JoeWang1127: This fomatter is not part of client library and we should be able use a new version of Java to run it.

library_generation/utils/utilities.sh
```diff
@@ -380,3 +375,17 @@ download_googleapis_files_and_folders() {
...
+get_java_formatter_location() {
+  echo "${JAVA_FORMATTER_LOCATION}"
+}
+
+error_if_not_exists() {
+  local required_tool=$1
+  if [[ ! -f "${required_tool}" ]]; then
+    >&2 echo "File ${required_tool} not found in the "
+    >&2 echo "filesystem. Please configure your environment and store the "
+    >&2 echo "generator jar in this location"
```
blakeli0, line 388: Since this logic is reused now, we should update the wording from "generator jar" to something more generic.

showcase/scripts/generate_showcase.sh
blakeli0, line 1: We probably don't need this sh anymore once we move to hermetic build image in showcase right? Or at least it should be much simpler, things like downloading jars would not be needed anymore.

.cloudbuild/library_generation/library_generation.Dockerfile
blakeli0: Since it is a jar, I think it still needs to be run with Java. I see that we are currently using [Java 21](https://github.com/googleapis/sdk-platform-java/blob/main/.cloudbuild/library_generation/library_generation.Dockerfile#L17) in our build file, but we probably  want to use a lower version to guarantee the compatibility since our client library still supports Java 8.

showcase/scripts/generate_showcase.sh
JoeWang1127: Yes, I'll cleanup these files when I'm working on b/325061217

library_generation/utils/utilities.sh
JoeWang1127: Done.

.cloudbuild/library_generation/library_generation.Dockerfile
JoeWang1127: I think this repo has a check to verify the compiled code is compatible with Java 8 (maven profile should make sure of that). Of course, we can change it to Java8 as well.

On the other hand, running generator/formatter jar only requires JRE which I think we can choose a different version from JDK. We can optimize the image to only include a JRE but not a JDK.

blakeli0 (approved):

PR #3272, opened 2024-10-03
fix: Disable automatically retrieving Universe Domain from Metadata Server

lqiu96:
See internal ticket b/349488459 for more info.

External info:
ComputeEngineCredentials in client libraries should not validate the universe domain. Validating the universe domain requires retrieving it from Metadata Server (MDS) and this will be temporarily disabled. 

For users that using client libraries, there will be no automatic call to MDS. For users that use the Credentials directly and manually call `ComputeEngineCredentials.getUniverseDomain()` and the universe domain is not explicitly set, it will make a call to MDS.

### DIFF
--- a/gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
+++ b/gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
@@ -153,7 +153,7 @@ public abstract class EndpointContext {
     }
     String credentialsUniverseDomain = Credentials.GOOGLE_DEFAULT_UNIVERSE;
     // If credentials is not NoCredentialsProvider, use the Universe Domain inside Credentials
-    if (credentials != null && (!(credentials instanceof ComputeEngineCredentials))) {
+    if (credentials != null) {
       credentialsUniverseDomain = credentials.getUniverseDomain();
     }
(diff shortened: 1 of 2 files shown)


gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -147,7 +148,8 @@ public void validateUniverseDomain(
     }
     String credentialsUniverseDomain = Credentials.GOOGLE_DEFAULT_UNIVERSE;
     // If credentials is not NoCredentialsProvider, use the Universe Domain inside Credentials
-    if (credentials != null) {
+    // (TODO: b/349488459) - Disable automatic requests to MDS until 01/2025
+    if (credentials != null && (!(credentials instanceof ComputeEngineCredentials))) {
```
blakeli0, line 152: Can we add a unit test for this scenario?

blakeli0 (approved):

blakeli0 (commented):
Can we update the title and description to reflect the change before merging? Also this PR probably should be a `fix`.

lqiu96:
> Can we update the title and description to reflect the change before merging? Also this PR probably should be a `fix`.

Updated and add some external info.

PR #3289, opened 2024-10-07
chore: remove release-please in generated README.md

JoeWang1127:
In this PR:
- Remove release-please settings in README.md template because the version is managed by hermetic build.

PR #3290, opened 2024-10-09
feat: selectively generate libraries

JoeWang1127:
In this PR:
- Allow selectively generate libraries by adding `--library-names` to CLI.

We have a request of partially releasing google-cloud-java (b/331628538), which requires only generating a selective list of libraries.

This change allows a user-defined list of libraries to be generated even if the corresponding library configs are the same.

### DIFF
--- a/library_generation/cli/entry_point.py
+++ b/library_generation/cli/entry_point.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 import os
 import sys
-
+from typing import Optional
 import click as click
 from library_generation.generate_pr_description import generate_pr_descriptions
 from library_generation.generate_repo import generate_from_yaml
@@ -52,6 +52,15 @@ def main(ctx):
     """,
 )
 @click.option(
+    "--includes",
+    type=str,
+    default=None,
+    show_default=True,
+    help="""
+    A list of library names that will be generated, separated by comma.
+    """,
+)
+@click.option(
     "--repository-path",
     type=str,
     default=".",
@@ -77,6 +86,7 @@ def main(ctx):
 def generate(
     baseline_generation_config_path: str,
     current_generation_config_path: str,
+    includes: Optional[str],
     repository_path: str,
     api_definitions_path: str,
 ):
@@ -84,10 +94,14 @@ def generate(
     Compare baseline generation config and current generation config and
     generate changed libraries based on current generation config with commit
     history.
+    If `includes` is specified, libraries whose name can be found in the current
+    generation config will *also* be generated.
 
     If baseline generation config is not specified but current generation
-    config is specified, generate all libraries based on current generation
-    config without commit history.
+    config is specified, generate all libraries if `includes` is not specified,
+    based on current generation config without commit history.
+    If `includes` is specified, only libraries whose name can be found in the
+    current generation config will be generated.
 
     If current generation config is not specified but baseline generation
     config is specified, raise FileNotFoundError because current generation
@@ -95,8 +109,10 @@ def generate(
 
     If both baseline generation config and current generation config are not
     specified, generate all libraries based on the default generation config,
-    which is generation_config.yaml in the current working directory. Raise
-    FileNotFoundError if the default config does not exist.
+    which is generation_config.yaml in the current working directory.
+    If `includes` is specified, only libraries whose name can be found in the
+    default generation config will be generated.
+    Raise FileNotFoundError if the default config does not exist.
 
     The commit history, if generated, will be available in
     repository_path/pr_description.txt.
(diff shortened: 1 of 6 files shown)


blakeli0:
Compare to selectively release libraries, I think this is not a hard blocker? Because we can comment out other libraries in the branch as a workaround?

JoeWang1127:
> Compare to selectively release libraries, I think this is not a hard blocker? Because we can comment out other libraries in the branch as a workaround?

Yes, it's not a blocker but it requires a error-prone manual step for the release manager. Besides, it's not complex to implement since we've already had `get_target_libraries` to filter library names. We only need to pass a list from CLI.

JoeWang1127:
Integration test failed:
```
Difference in README.md:
--- 

+++ 

@@ -40,6 +40,7 @@

 

 If you are using Maven without the BOM, add this to your dependencies:

 

+<!-- {x-version-update-start:google-cloud-alloydb:released} -->

 

 ```xml

 <dependency>

@@ -60,6 +61,7 @@

 ```Scala

 libraryDependencies += "com.google.cloud" % "google-cloud-alloydb" % "0.35.0"

 ```

+
```

We need to merge #3289 first.

library_generation/cli/entry_point.py
```diff
@@ -164,12 +185,17 @@ def __generate_repo_and_pr_description_impl(
...
+    # Pass None if we want to fully generate the repository.
     target_library_names = (
         config_change.get_changed_libraries()
         if not _needs_full_repo_generation(config_change=config_change)
         else None
     )
+    # Generate the union of two library lists if neither of the library list
+    # indicates a full generation.
+    # The deduplication is done in `generate_from_yaml`.
+    if (target_library_names is not None) and (include_libraries is not None):
```
blakeli0, line 197: I feel this making the logic more complicated, if `include_libraries` is provided, it should just override `target_library_names`. Especially we plan to remove  `target_library_names` in the future.

library_generation/cli/entry_point.py
```diff
@@ -191,6 +217,12 @@ def _needs_full_repo_generation(config_change: ConfigChange) -> bool:
     return not current_config.is_monorepo() or current_config.contains_common_protos()
 
 
+def _parse_library_name_from(includes: str) -> Optional[list[str]]:
+    if includes is None:
+        return None
+    return includes.split(",")
```
blakeli0, line 223: Maybe trim the library name?

library_generation/cli/entry_point.py
```diff
@@ -51,6 +51,15 @@ def main(ctx):
     metadata about library generation.
     """,
 )
+@click.option(
+    "--includes",
```
blakeli0, line 55: `includes` is a little ambiguous to me that the user may not know what is the info we are including. For example, is this `api_shortname` or `library_name` or something else?  Just `library_names` or `target_library_names`(we are already using it in scripts) sounds more reasonable to me.

library_generation/cli/entry_point.py
```diff
@@ -51,6 +51,15 @@ def main(ctx):
...
+    type=str,
+    default=None,
+    show_default=True,
+    help="""
+    A list of library names that will be generated, separated by comma.
```
blakeli0, line 60: Can we add more docs here? I think it is basically `library_name != null ? library_name : api_shortname` from `generation_config.yaml`?

library_generation/cli/entry_point.py
JoeWang1127: Changed to `library_names`.

library_generation/cli/entry_point.py
JoeWang1127: Yes, I also found it was a bit of confusing in this scenario.

I changed this to `library_names` takes preference if specified.

library_generation/cli/entry_point.py
JoeWang1127: Changed the comment.

library_generation/cli/entry_point.py

library_generation/cli/entry_point.py
```diff
@@ -191,6 +221,12 @@ def _needs_full_repo_generation(config_change: ConfigChange) -> bool:
     return not current_config.is_monorepo() or current_config.contains_common_protos()
 
 
+def _parse_library_name_from(includes: str) -> Optional[list[str]]:
+    if includes is None:
+        return None
+    return includes.strip().split(",")
```
blakeli0, line 227: Actually I meant trim after split.

blakeli0 (approved):

library_generation/cli/entry_point.py
JoeWang1127: I'll make the change before merging.

PR #3299, opened 2024-10-17
feat: Move release note generation to a sub module

JoeWang1127:
In this PR:
- Separate projects to three modules: common, library_generation and release_note_generation
- Add `generate_release_note.py` to separate PR description generation from `entrypoint.py`.
- Only install common and library_generation module in image.
- Remove PR description comparison in integration test.

Example generation workflow in downstream library: https://github.com/googleapis/java-bigtable/actions/runs/11442581326/job/31833887512

Note that `release_note_generation` module still depends on `library_generation` module because we didn't separate config change functions into `common` module. This will be in a follow up PR.


.github/workflows/verify_library_generation.yaml
```diff
@@ -80,23 +67,23 @@ jobs:
...
       run: |
         set -ex
-        pushd library_generation
-        pip install -r requirements.txt
-        pip install .
-        popd
+        pip install --upgrade pip
+        pip install hermetic_build/common
+        pip install hermetic_build/library_generation
+        pip install hermetic_build/release_note_generation
```
JoeWang1127, line 78: How about we keep the dependencies together.

JoeWang1127: Created `requirements.in` and `requirements.txt` in each module.

.github/scripts/action.yaml
```diff
@@ -37,11 +37,34 @@ inputs:
...
     shell: bash
     run: |
       cd ${{ github.action_path }}
-      cp hermetic_library_generation.sh $GITHUB_WORKSPACE
+      cp hermetic_library_generation.sh "${GITHUB_WORKSPACE}"
+  - name: Install python packages
+    shell: bash
+    run: |
+      cd "${GITHUB_WORKSPACE}"
+      pip install --upgrade pip
```
blakeli0, line 61: Do we have to upgrade `pip`? Can this be done in previous steps like `setup-python`?

.github/scripts/action.yaml
```diff
@@ -37,11 +37,34 @@ inputs:
...
+      pip install --require-hashes -r hermetic_build/common/requirements.txt
+      pip install hermetic_build/common
+      pip install --require-hashes -r hermetic_build/library_generation/requirements.txt
+      pip install hermetic_build/library_generation
```
blakeli0, line 65: I don't think we need to install `library_generation` anymore once we move `ConfigChange` out of this module?

JoeWang1127: Yes. We need to install `library_generation` because config change is still part of it. Once it moves to `common`, we don't need to install it.

.github/scripts/action.yaml
JoeWang1127: According to this [comment](https://github.com/actions/setup-python/issues/791#issuecomment-1909393984), pip is not updated in this action.

As for pip, we don't need it to be updated. I'll remove this line.

JoeWang1127: Done.

.github/workflows/verify_library_generation.yaml
```diff
@@ -46,32 +46,21 @@ jobs:
...
-        curl https://pyenv.run | bash
-        # setup environment
-        export PYENV_ROOT="$HOME/.pyenv"
-        export PATH="$PYENV_ROOT/bin:$PATH"
-        echo "PYENV_ROOT=${PYENV_ROOT}" >> $GITHUB_ENV
-        echo "PATH=${PATH}" >> $GITHUB_ENV
-
-        set +ex
+        python-version: 3.12
     - name: install python dependencies
```
diegomarquezp, line 50:
```suggestion
    - name: install python modules and dependencies
```

.github/workflows/verify_library_generation.yaml
```diff
@@ -80,23 +69,26 @@ jobs:
     - uses: actions/checkout@v4
     - uses: actions/setup-python@v5
       with:
-        python-version: 3.11
+        python-version: 3.12
     - name: install python dependencies
```
diegomarquezp, line 73:
```suggestion
    - name: install python modules and dependencies
```

hermetic_build/common/requirements.txt
```diff
@@ -0,0 +1,109 @@
+#
+# This file is autogenerated by pip-compile with Python 3.11
```
diegomarquezp, line 2: What are the implications of generating a reqs file with a different python version? Maybe we can create one with 3.12 instead?

hermetic_build/library_generation/DEVELOPMENT.md
```diff
@@ -21,9 +21,9 @@ code declared in a "golden branch" of the repo.
 It requires docker and python 3.x to be installed.
 
 ```
+python -m pip install --require-hashes -r requirements.txt
 python -m pip install .
```
diegomarquezp, line 25: The development guide shows commands to be run from the `sdk-platform-java/library_generation` folder. However this command now must be run from `sdk-platform-java/hermetic_build/library_generation` (or change `pip install .` to `pip install library_generation` and run it from `//hermetic_build`. We can modify [this note](https://github.com/googleapis/sdk-platform-java/blob/d5e74d9b3cf38f8aef78f1990d94778a255a5d59/library_generation/DEVELOPMENT.md?plain=1#L2) to point to `hermetic_build`. 

With this, then we could correct this to

```suggestion
python -m pip install library_generation
```

hermetic_build/library_generation/README.md
```diff
@@ -89,14 +89,6 @@ will be created/modified:
 | pom.xml (repo root dir)             | Always generated from inputs                                             |
 | versions.txt                        | New entries will be added if they don’t exist                            |
 
-### Change history
```
diegomarquezp, line 92: IIUC this feature is not being actually removed. Should we keep it?

hermetic_build/library_generation/requirements.txt
```diff
@@ -0,0 +1,417 @@
+#
+# This file is autogenerated by pip-compile with Python 3.11
```
diegomarquezp, line 2: same concern about python versions

diegomarquezp (commented):
Thanks for this PR. Looks good so far - just some minor comments.

.github/workflows/verify_library_generation.yaml
JoeWang1127: Done.

hermetic_build/library_generation/DEVELOPMENT.md
JoeWang1127: Done.

.github/workflows/verify_library_generation.yaml
JoeWang1127: Done.

hermetic_build/library_generation/README.md
JoeWang1127: Change history is not generated by `entry_point.py` so I think we don't need to include this part.

I plan to raise another PR to consolidate library generation and release note generation into a README in `hermetic_build` directory.

hermetic_build/common/requirements.txt
JoeWang1127: `pip-tools` (which contains `pip-compile`) is built with python 3.11.

You can confirm this by running `which pip-compile`.

diegomarquezp: Ah, looks like pip-compile is available only for 3.11, but it's installable via `pip install pip-tools` from 3.12 onwards. I managed to locally generate the hashes with python 3.12. Can you confirm this works?

```bash
pyenv install 3.12
pyenv local 3.12
python --version
pip install pip-tools
pip-compile --generate-hashes
```

hermetic_build/library_generation/README.md
diegomarquezp: Ah gotcha, I was missing the fact that this readme is in the `library_generation` folder.

hermetic_build/common/requirements.txt
JoeWang1127: Thanks.

I changed the `requirements.txt` using python 3.12.

diegomarquezp (approved):

blakeli0 (approved):

Issue #2503, opened 2024-02-26
HttpJson Implementation is recording Operation(Success|Fail) metric before retries are attempted


lqiu96:
Discovered in the Otel Draft PR: https://github.com/googleapis/sdk-platform-java/pull/2500

### Issue
For Otel, this is causing the number of attempts to always be recorded as 1 instead of the number of retries actually attempted. It does not impact retries being invoked, but will record incorrect metrics.

Steps for creating a UnaryCallable (1 is the inner most callable and 5 is the last callable invoked):
### HttpJson Echo Unary:
1. HttpJsonDirectCallable
2. HttpJsonUnaryRequestParamCallable (optional -- created if params exist)
3. TracedUnaryCallable
4. HttpJsonExceptionCallable
5. RetryingCallable

### gRPC Echo Unary:
1. GrpcDirectCallable
2. GrpcUnaryRequestParamCallable (optional -- created if params exist)
3. GrpcExceptionCallable
4. RetryingCallable
5. TracedUnaryCallable

### TracedUnaryCallable 
Inside the TracedUnaryCallable, the callback is invoked once in innerCallable has been invoked. The callback implementation is the [TraceFinisher](https://github.com/googleapis/sdk-platform-java/blob/7902a41c87240d607179d07c28cce462ea135c5f/gax-java/gax/src/main/java/com/google/api/gax/tracing/TraceFinisher.java). 

From the [implementation](https://github.com/googleapis/sdk-platform-java/blob/7902a41c87240d607179d07c28cce462ea135c5f/gax-java/gax/src/main/java/com/google/api/gax/tracing/TraceFinisher.java#L45-L57), it'll record an operation success or failure, which is intended to invoked after all the retries are exhausted (as the retries are attempts).

### Possible Solution
We should move creating the TracedUnaryCallable to the last step for HttpJson. This would require refactoring this file: https://github.com/googleapis/sdk-platform-java/blob/7902a41c87240d607179d07c28cce462ea135c5f/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonCallableFactory.java

### Concerns
Would this impact existing behavior? Do we need to do this for the other Callable types (StreamingCallables?)

PR #3300, opened 2024-10-21
fix: httpjson callables to trace attempts (started, failed)

diegomarquezp:
Fixes https://github.com/googleapis/sdk-platform-java/issues/2503 

### Approach
This PR uses the approach suggested by the same bug.
> We should move creating the TracedUnaryCallable to the last step for HttpJson. This would require refactoring this file: https://github.com/googleapis/sdk-platform-java/blob/7902a41c87240d607179d07c28cce462ea135c5f/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonCallableFactory.java

### Confirmation that it works
The confirmation that it works is in the modified tests [(e.g. `retry()`)](https://github.com/googleapis/sdk-platform-java/blob/1edf55754d1f602a3bf70b3a44d1c42689ae961b/gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/RetryingTest.java#L158-L159). When using the proposed test version against the version of `HttpJsonCallableFactory` from the main branch, it will fail because it will not record attempts started or failed whatsoever: 

image

The image above shows all tests having failed when using the production files from `main`, implying that no tracer attempts are recorded for any of the tests.


gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/RetryingTest.java
```diff
@@ -263,7 +297,12 @@ void retryOnUnexpectedException() {
     UnaryCallable<Integer, Integer> callable =
         HttpJsonCallableFactory.createUnaryCallable(
             callInt, callSettings, httpJsonCallSettings, clientContext);
-    ApiException exception = assertThrows(ApiException.class, () -> callable.call(initialRequest));
+    // Should an unexpected RuntimeException be received as an UnknownException?
+    UnknownException exception =
+        assertThrows(UnknownException.class, () -> callable.call(initialRequest));
```
lqiu96, line 302: Might be missing some context, but why does it need to be received as an UnknownException? Was there a runtime issue with receiving it as an ApiException?

lqiu96:
I think you can also remove a bunch of the `@Disabled` annotations in the showcase ITs as well. https://github.com/googleapis/sdk-platform-java/blob/7512cfa21cd0dd47c976be6b654fde656b1f0359/showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java#L323

I think they should pass with your PR.

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/RetryingTest.java
diegomarquezp: @diegomarquezp to share the full stack trace of both exceptions for further analysis (from discussion with @blakeli0 @lqiu96 and @zhumin8)

lqiu96: Hmm, I tried assertThrows locally with ApiException and it passed: `ApiException exception = assertThrows(ApiException.class, () -> callable.call(initialRequest));`

diegomarquezp: Hmm, yes that's true. I can't figure how that error came up then. I'll modify the description. Thank you for double checking this.

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/RetryingTest.java
```diff
@@ -396,4 +448,35 @@ public static UnaryCallSettings<Integer, Integer> createSettings(
...
+
+          @Override
+          public void attemptFailedRetriesExhausted(Throwable error) {
+            tracerFailedRetriesExhausted.set(true);
+            super.attemptFailedRetriesExhausted(error);
+          }
+        };
+    ApiTracerFactory tracerFactory = (parent, spanName, operationType) -> tracer;
+    return tracerFactory;
+  }
```
lqiu96, line 481: I wonder if this could be refactored into something like a TestTracerFactory + TestApiTracer that basically does this. Increments the value and exposes getters for the values.

Not something that needs to be done in this PR, but we can keep in mind in case we duplicate this in other unit tests

lqiu96 (commented):
~~Can you try and re-enable the Disabled ITOtelMetric tests and see if it works? LGTM otherwise~~

Edit: Spoke too soon.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
```diff
@@ -771,7 +763,7 @@ void testHttpJson_multipleFailedAttempts_successfulOperation() throws Exception
             .setSuccess(BlockResponse.newBuilder().setContent("httpjsonBlockResponse"))
             .build();
 
-    grpcClient.block(blockRequest);
+    httpClient.block(blockRequest);
```
diegomarquezp, line 766: @lqiu96 This test is the only one that's failing after re-enabling them. 

Just to confirm my understanding, this test sets up retry settings to perform 3 attempts in 1 second. Then we issue a `block` request to hold the response from being issued until 1 second later?

The one part that doesn't seem to be in order is this check:
https://github.com/googleapis/sdk-platform-java/blob/5c11e10d80256476a225ebcd8c9a0044ec83c864/showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java#L199-L204
That starts to get into Otel grounds I'm not familiar with. Is this a histogram of attempts per second?

Lastly, I was going to link the error logs (hoping it's the same as my local), but they [show](https://github.com/googleapis/sdk-platform-java/actions/runs/11508085713/job/32035590834?pr=3300) a java-iam related compilation error. I'll give this a retry.

cc: @blakeli0 @zhumin8

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/RetryingTest.java
diegomarquezp: May take a stab at it when implementing our yet-to-be made decision on the failing test https://github.com/googleapis/sdk-platform-java/pull/3300#discussion_r1815749326

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITOtelMetrics.java
lqiu96: I believe it should It should be testing that there were 3 attempts made. The retries should be configured so that it would retry at 0.5, 1, and 2 seconds. 

There error message doesn't seem to be super clear (which is room for improvement). I'm not sure why this is erroring out and not the gRPC test. I'm guessing some changes were made in the gRPC function and not ported over to the HttpJson side.

lqiu96: Ah, I think i figured out the issue. In the file L730-734 are modifying the settings for the Echo RPC and not the Block RPC:
```
    EchoStubSettings.Builder httpJsonEchoSettingsBuilder = EchoStubSettings.newHttpJsonBuilder();
    httpJsonEchoSettingsBuilder
        .echoSettings()
        .setRetrySettings(retrySettings)
        .setRetryableCodes(ImmutableSet.of(Code.DEADLINE_EXCEEDED));
```

Unfortunately, it's looks quite too easy to make a mistake when configuring retry settings.

diegomarquezp: Excellent, the test now passes after fixing this. +1 to easiness of mistakes - I double checked that the test was being setup for `block` instead of `echo`, but still missed this part!

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/RetryingTest.java
diegomarquezp: Done, now we have a separate test api tracer + factory.

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/RetryingTest.java
```diff
@@ -200,6 +213,9 @@ void retryMaxAttemptsExceeded() {
         HttpJsonCallableFactory.createUnaryCallable(
             callInt, callSettings, httpJsonCallSettings, clientContext);
     assertThrows(ApiException.class, () -> callable.call(initialRequest));
+    assertThat(tracerFactory.getTracerAttempts().get()).isEqualTo(2);
+    assertThat(tracerFactory.getTracerAttemptsFailed().get()).isEqualTo(1);
```
lqiu96, line 217: nit: Perhaps we can add TracerOperationFailed == true check?

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/RetryingTest.java
```diff
@@ -220,6 +236,9 @@ void retryWithinMaxAttempts() {
         HttpJsonCallableFactory.createUnaryCallable(
             callInt, callSettings, httpJsonCallSettings, clientContext);
     assertThat(callable.call(initialRequest)).isEqualTo(2);
+    assertThat(tracerFactory.getTracerAttempts().get()).isEqualTo(3);
+    assertThat(tracerFactory.getTracerAttemptsFailed().get()).isEqualTo(2);
+    assertThat(tracerFactory.getTracerOperationFailed().get()).isEqualTo(false);
```
lqiu96, line 241: nit: Perhaps we can add TracerFailedRetriesExhausted == false check?

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/RetryingTest.java
```diff
@@ -246,6 +265,9 @@ void retryOnStatusUnknown() {
         HttpJsonCallableFactory.createUnaryCallable(
             callInt, callSettings, httpJsonCallSettings, clientContext);
     assertThat(callable.call(initialRequest)).isEqualTo(2);
+    assertThat(tracerFactory.getTracerAttempts().get()).isEqualTo(4);
+    assertThat(tracerFactory.getTracerAttemptsFailed().get()).isEqualTo(3);
+    assertThat(tracerFactory.getTracerOperationFailed().get()).isEqualTo(false);
```
lqiu96, line 270: nit: Perhaps we can add TracerFailedRetriesExhausted == false check?

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/RetryingTest.java
```diff
@@ -264,6 +286,9 @@ void retryOnUnexpectedException() {
         HttpJsonCallableFactory.createUnaryCallable(
             callInt, callSettings, httpJsonCallSettings, clientContext);
     ApiException exception = assertThrows(ApiException.class, () -> callable.call(initialRequest));
+    assertThat(tracerFactory.getTracerAttempts().get()).isEqualTo(1);
+    assertThat(tracerFactory.getTracerAttemptsFailed().get()).isEqualTo(0);
+    assertThat(tracerFactory.getTracerOperationFailed().get()).isEqualTo(true);
```
lqiu96, line 291: Perhaps here too?

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/RetryingTest.java
```diff
@@ -293,6 +318,9 @@ void retryNoRecover() {
...
```
lqiu96, line 323: And here if it makes sense.

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/RetryingTest.java
```diff
@@ -319,6 +347,13 @@ void retryKeepFailing() {
 
     UncheckedExecutionException exception =
         assertThrows(UncheckedExecutionException.class, () -> Futures.getUnchecked(future));
+    // the number of attempts varies. Here we just make sure that all of them except the last are
+    // considered as failed
```
lqiu96, line 351: qq, isn't the last one also considered a failed attempt? Why is the tracerAttemptsFailed != tracerAttempts?

Also, can we add the operationfailed == true check here as well?

lqiu96 (commented):
LGTM. Added a few questions/ nits

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/RetryingTest.java
diegomarquezp: Done

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/RetryingTest.java
diegomarquezp: Done

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/RetryingTest.java
diegomarquezp: Done

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/RetryingTest.java
diegomarquezp: Done

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/RetryingTest.java
diegomarquezp: Done

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/RetryingTest.java
diegomarquezp: > Why is the tracerAttemptsFailed != tracerAttempts?

I'm not sure and I agree they should be the same value since this test just fetches an exception until it can't retry anymore (retries exhausted).

diegomarquezp: from discussion with @blakeli0, @lqiu96 and @zhumin8, let's double check `TestApiTracer`, confirm with showcase.

diegomarquezp: showcase has another tracer we can use to cross check `TestApiTracer`

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/testing/TestApiTracer.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+import org.threeten.bp.Duration;
+
+/**
+ * Test tracer that keeps count of different events. See {@link TestApiTracerFactory} for more
+ * details.
+ */
+public class TestApiTracer extends BaseApiTracer {
+
+  private final AtomicInteger tracerAttempts;
```
blakeli0, line 43: I think they can be called just `attempts`, same for the fields below.

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/testing/TestApiTracerFactory.java
```diff
@@ -0,0 +1,75 @@
...
+  private final AtomicBoolean tracerOperationFailed = new AtomicBoolean(false);
+  private final AtomicBoolean tracerFailedRetriesExhausted = new AtomicBoolean(false);
+  private final TestApiTracer instance =
+      new TestApiTracer(
+          tracerAttempts,
+          tracerAttemptsFailed,
+          tracerOperationFailed,
+          tracerFailedRetriesExhausted);
+
+  public AtomicInteger getTracerAttempts() {
```
blakeli0, line 55: I think it's better to get the tracer instance, and use the getters from the tracer in unit tests.

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/testing/TestApiTracerFactory.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Produces a {@link TestApiTracer}, which keeps count of the attempts made and attempts
+ * made-and-failed. It also keeps count of the operations failed and when the retries have been
+ * exhausted.
+ */
+public class TestApiTracerFactory implements ApiTracerFactory {
+  private final AtomicInteger tracerAttempts = new AtomicInteger();
```
blakeli0, line 44: These can be initialized within the tracer for better encapsulation.

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/testing/TestApiTracer.java
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  @Override
+  public void attemptFailed(Throwable error, Duration delay) {
+    tracerAttemptsFailed.incrementAndGet();
+    super.attemptFailed(error, delay);
+  }
+
+  @Override
+  public void attemptStarted(int attemptNumber) {
```
blakeli0, line 66: Use [attemptStarted(Object request, int attemptNumber)](https://github.com/googleapis/sdk-platform-java/blob/4d4c798feb6e9eee2c88fdbda1282d8c109b8eb5/gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracer.java#L141-L144) instead

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/testing/TestApiTracer.java
blakeli0, line 1: I like the idea of a test tracer in general. On the other hand, we should only implement what is needed for our use cases, so that we have lower chance to run into issues that are not related to our PR. For example, maybe we don't need to implement `attemptFailedRetriesExhausted`.

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/testing/TestApiTracer.java
```diff
@@ -0,0 +1,82 @@
...
+import com.google.api.gax.tracing.BaseApiTracer;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+import org.threeten.bp.Duration;
+
+/**
+ * Test tracer that keeps count of different events. See {@link TestApiTracerFactory} for more
+ * details.
+ */
+public class TestApiTracer extends BaseApiTracer {
```
blakeli0, line 41: We should use the interface `ApiTracer` instead.

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/testing/TestApiTracerFactory.java
diegomarquezp: I'd like to initialize them in the tracer itself. However the clients only expect tracer factories, so there is no way to have control over how the actual tracers are instantiated.
I'll double check on this.

https://github.com/googleapis/sdk-platform-java/blob/2df2e9f0b6183096005895f12530b1aa369d7ffd/gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/RetryingTest.java#L121-L128

diegomarquezp: Ah nvm, I realized that accessing the ApiTracer instance should be enough.

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/testing/TestApiTracer.java
diegomarquezp: renamed to `attemptsStarted`, removed the `tracer` prefix from all fields

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/testing/TestApiTracer.java
diegomarquezp: Done

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/testing/TestApiTracer.java
diegomarquezp: Done

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/testing/TestApiTracerFactory.java
diegomarquezp: Done

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/testing/TestApiTracerFactory.java
diegomarquezp: Done

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/testing/TestApiTracer.java
diegomarquezp: @lqiu96 figured that the retries exhausted event records the missing failed attempt in our tests:

> It looks like attempt_count is recorded when this is called: https://github.com/googleapis/sdk-platform-java/blob/4d4c798feb6e9eee2c88fdbda1282d8c109b8eb5/gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracer.java#L204. This seems like the number recorded in Otel should be correct.
I'm guessing your TestApiTracer would need to increment the attemptCount when attemptFailedRetriesExhausted is called as well to match behavior.


I'll keep this implementation and its retries exhausted flag.

lqiu96:
FYI @diegomarquezp 
CI error:
```
Error:  Failures: 
Error:    RetryingTest.retryKeepFailing:360 expected to be true
```

lqiu96 (approved):

diegomarquezp:
From @lqiu96 (thanks!):
> I have a suspicion as to why the race condition exists. I took a look into the code and I think I understand what you were explaining in the call.
> My suspicion of the cause of the race condition: https://github.com/googleapis/sdk-platform-java/blob/bb2a3be87291ae718ac0e8538025a0867a6b6ff6/gax-java/gax/src/main/java/com/google/api/gax/retrying/BasicRetryingFuture.java#L205
> This sets the result to the future which frees up the thread that is running the test. (basically .get() no longer blocks and if the thread schedules the test to run, it will). The tracerFinisher code which sets the operation failure status runs after this

This means that the tests removed in 065b01f7b1586f09edd9edf37f194d9e5f3db0c8 were actually expected behavior and is due to the way we set up the tests.

PR #3311, opened 2024-10-24
chore: migrate config change functions to `common` module

JoeWang1127:
In this PR:
- Migrate config change functions to `common` module.
- Library generation will accept library names directly.


hermetic_build/common/cli/changed_library_generation.py
```diff
@@ -0,0 +1,83 @@
...
+    """,
+)
+def create(
+    baseline_generation_config_path: str,
+    current_generation_config_path: str,
+) -> None:
+    """
+    Compare baseline generation config with current generation config and
+    generate changed library names (a comma separated string) based on current
+    generation config.
```
diegomarquezp, line 55:
```suggestion
    Compares baseline generation config with current generation config and
    generates changed library names (a comma separated string) based on current
    generation config.
```

hermetic_build/common/cli/changed_library_generation.py
```diff
@@ -0,0 +1,83 @@
...
+    if not os.path.isfile(baseline_generation_config_path):
+        raise FileNotFoundError(
+            f"{baseline_generation_config_path} does not exist. "
+            "A valid generation config has to be passed in as "
+            "baseline-generation-config-path."
+        )
+    current_generation_config_path = os.path.abspath(current_generation_config_path)
+    if not os.path.isfile(current_generation_config_path):
+        raise FileNotFoundError(
+            f"{current_generation_config_path} does not exist. "
```
diegomarquezp, line 67: nit: maybe abstract this duplicated exception message into a function/template?

hermetic_build/common/model/config_change.py
```diff
@@ -109,25 +106,22 @@ def get_qualified_commits(
...
-            )
-            if qualified_commit is not None:
-                qualified_commits.append(qualified_commit)
-            commit_parents = commit.parents
-            if len(commit_parents) == 0:
-                break
-            commit = commit_parents[0]
-        shutil.rmtree(tmp_dir, ignore_errors=True)
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            # we only need commit history, thus shadow clone is enough.
```
diegomarquezp, line 110:
```suggestion
            # we only need commit history, thus a shallow clone is enough.
```

diegomarquezp (approved):
LGTM with a few suggestions - hope they are useful.

hermetic_build/common/cli/changed_library_generation.py
JoeWang1127: Done.

hermetic_build/common/model/config_change.py
JoeWang1127: Done.

.github/scripts/hermetic_library_generation.sh
```diff
@@ -92,6 +92,12 @@ pushd "${api_def_dir}"
 git checkout "${googleapis_commitish}"
 popd
 
+# generate changed libraries
```
blakeli0, line 95: I think `generate changed libraries` is a little confusing, I thought we are actually generating libraries when I first read it, but we are actually comparing two configs and figuring out the libraries to generate.

hermetic_build/common/cli/changed_library_generation.py
blakeli0, line 1: I would avoid using the term `generation` in this file. `generate_repo` and `generate_pr_description` make sense because we are actually generating something as files. `generate_changed_library` though is not really generating something as files, it is an intermediate step to  `generate_repo` and `generate_pr_description`.  The file name could be just `get_changed_libraries`.

hermetic_build/common/cli/changed_library_generation.py
```diff
@@ -0,0 +1,83 @@
...
+# limitations under the License.
+import os
+
+import click as click
+
+from common.model.generation_config import from_yaml
+from common.utils.generation_config_comparator import compare_config
+
+
+@click.group(invoke_without_command=False)
```
blakeli0, line 22: Since this file is not supposed to be called directly, does it still have to be a cli? Maybe cli is still easier, but if not, can we call it as a Python function directly?

hermetic_build/library_generation/cli/entry_point.py
```diff
@@ -115,103 +101,63 @@ def generate(
...
+        or generation_config.contains_common_protos()
+    )
 
 
-def _parse_library_name_from(includes: str) -> Optional[list[str]]:
-    if includes is None:
+def _parse_library_name_from(
+    includes: str, generation_config: GenerationConfig
+) -> Optional[list[str]]:
+    if includes is None or _needs_full_repo_generation(generation_config):
```
blakeli0, line 160: Is this `_needs_full_repo_generation(generation_config)` a logic change?

hermetic_build/library_generation/cli/entry_point.py
```diff
@@ -30,18 +28,7 @@ def main(ctx):
...
-    type=str,
-    help="""
-    Absolute or relative path to a generation_config.yaml.
-    This config file is used for commit history generation, not library
-    generation.
-    """,
-)
-@click.option(
-    "--current-generation-config-path",
+    "--generation-config-path",
```
blakeli0, line 31: I think this means our library generation would not need internet access anymore?

hermetic_build/common/cli/changed_library_generation.py
JoeWang1127: We need to use this function in shell script, so I add a CLI to simplify the call.

The result is passed to `entry_point.py`.

.github/scripts/hermetic_library_generation.sh
JoeWang1127: Changed to `get changed library list`

hermetic_build/library_generation/cli/entry_point.py
JoeWang1127: I think so.

hermetic_build/common/cli/changed_library_generation.py
JoeWang1127: Done.

hermetic_build/library_generation/cli/entry_point.py
JoeWang1127: No, this is an existing logic.

hermetic_build/common/cli/changed_library_generation.py
JoeWang1127: This function is used in shell script so I add a CLI to simplify the call.

blakeli0 (approved):

hermetic_build/library_generation/cli/entry_point.py
JoeWang1127: This function determines whether we want a full generation, i.e., we want a full generation if it's a split repo or sdk-platform-java.

The logic is there before but I moved it here to help create the library name list. If the user doesn't specify `--library-names` or we do want a full generation, returns `None`, which means full generation.

Issue #3314, opened 2024-10-25
gax-2.56.0 IllegalArgumentException when determining the Protobuf bundle version within Spring boot app


MikeRocke:
#### Environment details

2. OS type and version: OSX and Linux
3. Java version: OpenJDK 21
4. artifact version(s): 2.56.0

#### Steps to reproduce

We have upgraded our spring boot app's dependency of 
`google-cloud-storage`
from version `2.43.2` to `2.44.0`
which has brought in gax-2.56.0, which now fails when our uber-jar attempts to start up

#### Code example

```java
        return StorageOptions.newBuilder()
                .build()
                .getService();
```

#### Stack trace
```
Caused by: java.lang.IllegalArgumentException: URI is not hierarchical
        at java.base/java.io.File.<init>(File.java:420)
        at com.google.api.gax.core.GaxProperties.getBundleVersion(GaxProperties.java:142)
        at com.google.api.gax.core.GaxProperties.<clinit>(GaxProperties.java:53)
        at com.google.api.gax.rpc.ApiClientHeaderProvider$Builder.<init>(ApiClientHeaderProvider.java:146)
        at com.google.api.gax.rpc.ApiClientHeaderProvider.newBuilder(ApiClientHeaderProvider.java:117)
        at com.google.cloud.http.HttpTransportOptions.getInternalHeaderProviderBuilder(HttpTransportOptions.java:216)
        at com.google.cloud.http.HttpTransportOptions.getHttpRequestInitializer(HttpTransportOptions.java:155)
        at com.google.cloud.storage.spi.v1.HttpStorageRpc.<init>(HttpStorageRpc.java:127)
        at com.google.cloud.storage.spi.v1.HttpStorageRpc.<init>(HttpStorageRpc.java:121)
        at com.google.cloud.storage.HttpStorageOptions$HttpStorageRpcFactory.create(HttpStorageOptions.java:392)
        at com.google.cloud.storage.HttpStorageOptions$HttpStorageRpcFactory.create(HttpStorageOptions.java:366)
        at com.google.cloud.ServiceOptions.getRpc(ServiceOptions.java:602)
        at com.google.cloud.storage.HttpStorageOptions.getStorageRpcV1(HttpStorageOptions.java:84)
        at com.google.cloud.storage.StorageImpl.<init>(StorageImpl.java:129)
        at com.google.cloud.storage.HttpStorageOptions$HttpStorageFactory.create(HttpStorageOptions.java:345)
[...]

PR #3315, opened 2024-10-25
fix: add additional potential exceptions when retrieving protobuf manifest file to get version

ldetmer:
#3314 3314


gax-java/gax/src/main/java/com/google/api/gax/core/GaxProperties.java
```diff
@@ -144,7 +144,11 @@ static Optional<String> getBundleVersion(Class<?> clazz) {
         Attributes attributes = jar.getManifest().getMainAttributes();
         return Optional.ofNullable(attributes.getValue("Bundle-Version"));
       }
-    } catch (URISyntaxException | IOException e) {
+    } catch (URISyntaxException
+        | IOException
+        | IllegalArgumentException
+        | SecurityException
+        | NullPointerException e) {
```
burkedavison, line 151: Can we just catch `Exception`? Is there any `Exception` that we would want to propagate from here?

burkedavison (approved):

burkedavison (approved):

gax-java/gax/src/main/java/com/google/api/gax/core/GaxProperties.java
ldetmer: no, we should safely ignore this as it doesn't block runtime.

blakeli0 (approved):

PR #3321, opened 2024-10-28
feat: Update Gapic-Generator to generate libraries using `java.time` methods

diegomarquezp:
This PR leverages the changes in https://github.com/googleapis/sdk-platform-java/pull/1872 so the generated code uses the new `java.time` methods.
All test-related changes are golden file/string changes.

#### SonarCloud complaints
The SonarCloud scanner complains about code duplication in the generator code, although after inspection I could not find a reason for stating so.
It also complains about showcase although the changes in this folder are strictly golden-oriented.

#### Downstream check failures
They are of the form "could not find remote branch `vX.X.X` to checkout". This seems not related to this PR as it is also present in https://github.com/googleapis/sdk-platform-java/pull/3318


lqiu96 (commented):
Changes LGTM. Can you re-run some of the flaky CI jobs? Also can you check to ensure that we can run and pass the downstream tests for the handwritten libraries?

diegomarquezp:
> Changes LGTM. Can you re-run some of the flaky CI jobs? Also can you check to ensure that we can run and pass the downstream tests for the handwritten libraries?

@lqiu96 the checks keep failing. I raised https://github.com/googleapis/sdk-platform-java/issues/3325 as this is also occurring in another PR.

lqiu96 (approved):
LGTM. Can we update the title to something like:
`feat: Update Gapic-Generator to generate libraries using java.time methods`.

This PR itself doesn't generate any library

PR #3322, opened 2024-10-29
update IT tests to ensure that a protobuf version is always getting sent

ldetmer:

### DIFF
--- a/showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITVersionHeaders.java
+++ b/showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/ITVersionHeaders.java
@@ -241,7 +241,7 @@ class ITVersionHeaders {
   @Test
   void testGrpcCall_sendsCorrectApiClientHeader() {
     Pattern defautlGrpcHeaderPattern =
-        Pattern.compile("gl-java/.* gapic/.*?--protobuf-.* gax/.* grpc/.* protobuf/.*");
+        Pattern.compile("gl-java/.* gapic/.*?--protobuf-\\d.* gax/.* grpc/.* protobuf/\\d.*");
     grpcClient.echo(EchoRequest.newBuilder().build());
     String headerValue = grpcInterceptor.metadata.get(API_CLIENT_HEADER_KEY);
     assertTrue(defautlGrpcHeaderPattern.matcher(headerValue).matches());
@@ -250,7 +250,7 @@ class ITVersionHeaders {
   @Test
   void testHttpJson_sendsCorrectApiClientHeader() {
     Pattern defautlHttpHeaderPattern =
-        Pattern.compile("gl-java/.* gapic/.*?--protobuf-.* gax/.* rest/ protobuf/.*");
+        Pattern.compile("gl-java/.* gapic/.*?--protobuf-\\d.* gax/.* rest/ protobuf/\\d.*");
     httpJsonClient.echo(EchoRequest.newBuilder().build());
     ArrayList<String> headerValues =
         (ArrayList<String>)


gax-java/gax/src/main/java/com/google/api/gax/util/ClassLoaderWrapper.java
```diff
@@ -0,0 +1,44 @@
...
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.google.api.gax.util;
+
+/* Wrapper class for reflection Class methods to enable unit testing. */
+public class ClassLoaderWrapper implements IClassLoaderWrapper {
```
burkedavison, line 33: `ClassLoader` has a very specific meaning in Java. `loadClass` also has a very specific meaning when talking about ClassLoaders. Let's not use those terms here if we're not using classloaders.

If we're going to create a wrapper around reflection calls for the purpose of testing, let's use the same labels as the underlying elements.

ClassLoaderWrapper --> ClassWrapper
loadClass --> forName

Also consider removing the interface is the only implementation is ever going to be this one. I'm not sure the indirection is benefiting us.

gax-java/gax/src/main/java/com/google/api/gax/core/GaxProperties.java
```diff
@@ -148,4 +150,31 @@ static Optional<String> getBundleVersion(Class<?> clazz) {
...
+  static String getProtobufVersion(IClassLoaderWrapper classLoader, Class clazz) {
+    try {
+      Class<?> protobufRuntimeVersionClass =
+          classLoader.loadClass("com.google.protobuf.RuntimeVersion");
+      return classLoader.getFieldValue(protobufRuntimeVersionClass, "MAJOR")
+          + "."
+          + classLoader.getFieldValue(protobufRuntimeVersionClass, "MINOR")
+          + "."
+          + classLoader.getFieldValue(protobufRuntimeVersionClass, "PATCH");
+    } catch (ClassNotFoundException | NoSuchFieldException | IllegalAccessException e) {
```
burkedavison, line 169: And SecurityException

gax-java/gax/src/main/java/com/google/api/gax/core/GaxProperties.java
```diff
@@ -148,4 +150,31 @@ static Optional<String> getBundleVersion(Class<?> clazz) {
...
+      Optional<String> protobufVersionFromManifest = getBundleVersion(clazz);
+      if (protobufVersionFromManifest.isPresent()) {
+        return protobufVersionFromManifest.get();
+      } else {
+        // If manifest file is not available default to protobuf generic version 3 as we know RuntimeVersion class is
+        // available in protobuf jar 4+.
+        return "3";
+      }
```
burkedavison, line 177:
```java
// If manifest file is not available default to protobuf generic version 3 as we know RuntimeVersion class is
// available in protobuf jar 4+.
return protobufVersionFromManifest.orElse("3");
```

gax-java/gax/src/main/java/com/google/api/gax/core/GaxProperties.java
```diff
@@ -148,4 +150,31 @@ static Optional<String> getBundleVersion(Class<?> clazz) {
...
     }
   }
+
+  /**
+   * Returns the Protobuf runtime version as reported by com.google.protobuf.RuntimeVersion,
+   * if class is available, otherwise by reading from MANIFEST file. If niether option is available defaults to
+   * protobuf version 3 as RuntimeVersion class is available in protobuf version 4+
+   */
+  @VisibleForTesting
+  static String getProtobufVersion(IClassLoaderWrapper classLoader, Class clazz) {
```
burkedavison, line 160: Please also make sure to rename this classloader argument, as this is not a classloader.

gax-java/gax/src/main/java/com/google/api/gax/util/IClassLoaderWrapper.java
```diff
@@ -0,0 +1,40 @@
...
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.google.api.gax.util;
+
+/* Interface that allows for unit testing reflection logic. */
+public interface IClassLoaderWrapper {
+  /* Wraps {@link java.lang.Class#loadClass} method  */
```
burkedavison, line 34: Please change. This is not wrapping loadClass.

gax-java/gax/src/test/java/com/google/api/gax/core/GaxPropertiesTest.java
```diff
@@ -175,6 +174,52 @@ public void testGetBundleVersion_noManifestFile() throws IOException {
     assertFalse(version.isPresent());
   }
 
+  @Test
+  void testGetProtobufVersion_success() throws Exception {
+    IClassLoaderWrapper mockClassLoader = mock(IClassLoaderWrapper.class);
```
burkedavison, line 179: Just reiterating while looking at this logic: The interface is not necessary to be able to mock the wrapper class.

gax-java/gax/src/main/java/com/google/api/gax/util/ClassLoaderWrapper.java
ldetmer: thanks for catching

gax-java/gax/src/main/java/com/google/api/gax/core/GaxProperties.java
ldetmer: nice!

burkedavison:
LGTM. Blake for approval.

gax-java/gax/src/main/java/com/google/api/gax/core/GaxProperties.java
```diff
@@ -148,4 +148,30 @@ static Optional<String> getBundleVersion(Class<?> clazz) {
...
+          + "."
+          + classWrapper.getFieldValue(protobufRuntimeVersionClass, "PATCH");
+    } catch (ClassNotFoundException
+        | NoSuchFieldException
+        | IllegalAccessException
+        | SecurityException e) {
+      // If manifest file is not available default to protobuf generic version 3 as we know
+      // RuntimeVersion class is
+      // available in protobuf jar 4+.
+      return getBundleVersion(clazz).orElse("3");
```
blakeli0, line 174: So we are going to have two different format of protobuf versions (3.x.x vs 3)? Is it going to cause issues when we try to aggregate the metrics?

gax-java/gax/src/main/java/com/google/api/gax/util/ClassWrapper.java
```diff
@@ -0,0 +1,45 @@
...
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.google.api.gax.util;
+
+/* Wrapper class for reflection {@link java.lang.Class} methods to enable unit testing. */
+public class ClassWrapper {
```
blakeli0, line 33: I don't think it's worth it to introduce a new public class just for unit testing in Gax. Can we move the wrapper methods to `GaxProperties` and make them package private?

ldetmer: i hope I understood what you meant.  I moved the class to package private (I need to keep the class so I can mock it, as I can't mock directly the class methods)

gax-java/gax/src/main/java/com/google/api/gax/core/GaxProperties.java
burkedavison: There might be 3.1.2, which provides precision. But there shouldn't be any 3.x.x.

We shouldn't be providing precision (significant digits) when we don't know them. We can fix any parsing logic as necessary.

gax-java/gax/src/main/java/com/google/api/gax/util/ClassWrapper.java
ldetmer: I refactored to remove the wrapper class/methods at all.  Let me know what you think!

blakeli0 (approved):

gax-java/gax/src/main/java/com/google/api/gax/util/ClassWrapper.java
blakeli0: Thanks, it looks good to me on the surface! 

> did you meant do not unit test this code

We should definitely do unit tests if possible. I see that the sonar coverage is complaining about the coverage now, can you please take a look at it and see if it makes sense?

ldetmer: thanks for calling out, I missed a test case! Fixed

PR #3330, opened 2024-10-31
feat: introduce `java.time` to java-core

diegomarquezp:
## In this PR:
 * Modify publicly exposed methods to offer a `java.time` alternative (suffixed with `Duration`). These methods will now hold the actual implementation, whereas the old signatures will encapsulate the new ones. Example: `retryDelay(org.threeten.bp.Duration)` encapsulates `retryDelayDuration(java.time.Duration)`

### _CLIRR_
```
[ERROR] 7012: com.google.cloud.testing.BaseEmulatorHelper$EmulatorRunner: Method 'public int waitForDuration(java.time.Duration)' has been added to an interface
```
This new interface method was added. However, we ignore this change alert because the method has a `default` implementation.

#### Problem
When using java.time functions, parsing a datetime string with offset produces a wrong time object (offset is not added to the effective epoch seconds).

#### Context
Full context in https://github.com/googleapis/sdk-platform-java/pull/3330/files#r1828424787

https://github.com/adoptium/jdk/commit/c6d209b505932b43d58502b119b86a6fa5b5338e was introduced as a fix (in Java 9) meant for [this issue](https://bugs.openjdk.org/browse/JDK-8066982).
In java 8, this means that the offset value is stored separately in a variable that is not respected by the parsing functions before this fix. The workaround is to use `appendZoneOrOffsetId()`, which stores the offset value in the `zone` variable of a parsing context, which is [respected as of java 8](https://github.com/adoptium/jdk8u/blob/31b88042fba46e87fba83e8bfd43ae0ecb5a9afd/jdk/src/share/classes/java/time/format/Parsed.java#L589-L591).

Additionally, under the consideration of this having unwanted side effects, we expanded the test suite to test for more edge and normal cases using an offset string. We also [searched](https://bugs.openjdk.org/browse/JDK-8202948?jql=affectedVersion%20%3D%20%228%22%20AND%20text%20~%20%22offset%22) the JDK's issue tracking database and found somewhat similar issues with parsing and offsets but no workaround that addressed our specific situation. This is also cause of no backports of the [fix](https://github.com/adoptium/jdk/commit/c6d209b505932b43d58502b119b86a6fa5b5338e) for Java 9 into Java 8.

#### Outcome
The behavior is kept the same as stated by our tests and downstream checks


diegomarquezp:
[Downstream checks](https://github.com/googleapis/sdk-platform-java/actions/runs/11619171012/job/32358202375?pr=3330) failing only for java-logging

```
/home/runner/work/sdk-platform-java/sdk-platform-java/java-shared-dependencies/target/java-datastore/google-cloud-datastore/src/main/java/com/google/cloud/datastore/testing/LocalDatastoreHelper.java:[49,7] error: LocalDatastoreHelper is not abstract and does not override abstract method stop(Duration) in BaseEmulatorHelper
```

java-core/google-cloud-core/clirr-ignored-differences.xml
```diff
@@ -12,4 +12,12 @@
...
         <method>long limit()</method>
     </difference>
+    <!-- These methods were switched to java.time and are internally used
+    in testing -->
+    <difference>
+        <differenceType>7005</differenceType>
+        <className>com/google/cloud/testing/*</className>
+        <method>* *(org.threeten.bp.Duration)</method>
+        <to>*</to>
+    </difference>
```
lqiu96, line 22: Do you know which methods were flagging this 7005 error code? This is method argument changed and I don't think that technically any method argument should have changed. Might be false positive and just want to double check.

java-core/google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
```diff
@@ -20,16 +20,16 @@
 
 import com.google.protobuf.util.Timestamps;
 import java.io.Serializable;
+import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeFormatterBuilder;
+import java.time.format.DateTimeParseException;
+import java.time.temporal.TemporalAccessor;
```
lqiu96, line 29: Did anything else in this file change? Why are the imports changing? Is this where the 7005 error codes are coming from?

java-core/google-cloud-core/src/main/java/com/google/cloud/testing/BaseEmulatorHelper.java
```diff
@@ -45,6 +45,7 @@
 import java.nio.file.Path;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
+import java.time.Duration;
```
lqiu96, line 48: Same with this.

java-core/google-cloud-core/clirr-ignored-differences.xml
diegomarquezp: It is an accurate flag. This exemption is for the `com/google/cloud/testing/*` classes, although the only one that applies is `BaseEmulatorHelper`.

My approach was to simply change the signature of these methods to `java.time` because they are only meant to be used in unit testing.

Maybe we still want to use an overload approach (`${method}Duration(java.time.Duration)`) so the CLIRR check flags "new methods" instead?

java-core/google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
diegomarquezp: I addressed the 7005 concern in https://github.com/googleapis/sdk-platform-java/pull/3330#discussion_r1826079247

Nothing else changed in this file. `TimeStamp` internally uses `java.time` objects but does not expose any of them via getters/setters. They are rather internal variables that allow convergence among different data types (e.g. `java.sql.Time`, proto classes, etc).

java-core/google-cloud-core/src/main/java/com/google/cloud/testing/BaseEmulatorHelper.java
diegomarquezp: This is what caused 7005 (https://github.com/googleapis/sdk-platform-java/pull/3330#discussion_r1826079247)

java-core/google-cloud-core/src/test/java/com/google/cloud/testing/BaseEmulatorHelperTest.java
```diff
@@ -28,12 +28,12 @@
 import java.net.URL;
 import java.net.URLConnection;
 import java.net.URLStreamHandler;
+import java.time.Duration;
```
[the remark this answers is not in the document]
diegomarquezp: The import change mainly affects the `TestEmulatorHelper` in this class. That class extends `BaseEmulatorHelper`, whose contract now demands to override methods that take `java.time.Duration` arguments.

diegomarquezp:
@lqiu96 I updated the PR description with the CLIRR output.

java-core/google-cloud-core/clirr-ignored-differences.xml
lqiu96: I see, I misread this. I didn't see that it was in the testing module and thought it was public methods used for testing.

java-core/google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
lqiu96: I see. These are indeed private so most should be fine. I think the one that does stand out is the DateTimeParseException: https://github.com/googleapis/sdk-platform-java/blob/4d4c798feb6e9eee2c88fdbda1282d8c109b8eb5/java-core/google-cloud-core/src/main/java/com/google/cloud/Timestamp.java#L199

I believe the exception type changed. Can you double check this? We may need to overload this one because of the different exception thrown.

java-core/google-cloud-core/clirr-ignored-differences.xml
lqiu96: Hmm, actually on some more thought I'm a bit on the fence about this. It is indeed in a testing package and marked with InternalApi, but I believe it is also packaged publicly in a JAR that anyone case use. I doubt anyone is using this anything other than testing purposes (i.e. BaseEmulatorHelperTest). Let's see if we can get some guidance/ another opinion.

java-core/google-cloud-core/clirr-ignored-differences.xml
```diff
@@ -12,4 +12,9 @@
         <className>com/google/cloud/ReadChannel</className>
         <method>long limit()</method>
     </difference>
+    <difference>
+        <differenceType>7012</differenceType>
+        <className>com/google/cloud/testing/BaseEmulatorHelper$EmulatorRunner</className>
+        <method>int waitForDuration(java.time.Duration)</method>
+    </difference>
```
diegomarquezp, line 19: This interface method has a `default` implementation

java-core/google-cloud-core/src/main/java/com/google/cloud/testing/BaseEmulatorHelper.java
```diff
@@ -180,10 +197,17 @@ public String getProjectId() {
...
+  @ObsoleteApi("Use stopDuration() instead")
+  public abstract void stop(org.threeten.bp.Duration timeout)
       throws IOException, InterruptedException, TimeoutException;
 
+  /** Stops the local emulator. */
+  public void stopDuration(java.time.Duration timeout)
+      throws IOException, InterruptedException, TimeoutException {
+    stop(toThreetenDuration(timeout));
+  }
+
```
diegomarquezp, line 210: `stop(org.threeten.bp.Duration)` will be the source-of-truth method so we don't break the contract of libraries downstream.

java-core/google-cloud-core/clirr-ignored-differences.xml
diegomarquezp: from @blakeli0 and @lqiu96, let's treat this as any other public surface: we only _add_ overloaded methods.

java-core/google-cloud-core/clirr-ignored-differences.xml
lqiu96: Can you add this as a comment to the file? i.e. something like `Default method added to interface`

java-core/google-cloud-core/src/main/java/com/google/cloud/testing/BaseEmulatorHelper.java
```diff
@@ -112,14 +115,21 @@ protected final void startProcess(String blockUntilOutput)
     }
   }
 
+  /** This method is obsolete. Use {@link #waitForProcessDuration(java.time.Duration)} instead */
+  @ObsoleteApi("Use waitForProcessDuration() instead")
+  protected final int waitForProcess(org.threeten.bp.Duration timeout)
+      throws IOException, InterruptedException, TimeoutException {
+    return waitForProcessDuration(toJavaTimeDuration(timeout));
+  }
```
lqiu96, line 123: Can we double check that this overload is needed? I believe it's protected and final and subclasses shouldn't be able to override this method. I *think* if we just migrate all internal uses cases in this repo to use java.time.Duration, we should be good.

lqiu96: I wrote this on the wrong one.

java-core/google-cloud-core/src/main/java/com/google/cloud/testing/BaseEmulatorHelper.java
```diff
@@ -112,14 +115,21 @@ protected final void startProcess(String blockUntilOutput)
...
    */
-  protected final int waitForProcess(Duration timeout)
+  protected final int waitForProcessDuration(java.time.Duration timeout)
       throws IOException, InterruptedException, TimeoutException {
     if (activeRunner != null) {
-      int exitCode = activeRunner.waitFor(timeout);
+      int exitCode = activeRunner.waitForDuration(timeout);
       activeRunner = null;
       return exitCode;
     }
```
lqiu96, line 135: Can we double check that this overload is needed? I believe it's protected and final and subclasses shouldn't be able to override this method. I think if we just migrate all internal uses cases in this repo to use java.time.Duration, we should be good.

java-core/google-cloud-core/src/main/java/com/google/cloud/testing/BaseEmulatorHelper.java
```diff
@@ -130,7 +140,14 @@ protected final int waitForProcess(Duration timeout)
...
 
-  private static int waitForProcess(final Process process, Duration timeout)
+  /** This method is obsolete. Use {@link #waitForProcessDuration(java.time.Duration)} instead */
+  @ObsoleteApi("Use waitForProcessDuration() instead")
+  private static int waitForProcess(final Process process, org.threeten.bp.Duration timeout)
+      throws InterruptedException, TimeoutException {
+    return waitForProcessDuration(process, toJavaTimeDuration(timeout));
+  }
+
+  private static int waitForProcessDuration(final Process process, java.time.Duration timeout)
```
lqiu96, line 150: I believe `private static` can be migrated without an overload, right?

java-core/google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
lqiu96: Bumping this comment.

java-core/google-cloud-core/src/main/java/com/google/cloud/testing/BaseEmulatorHelper.java
diegomarquezp: Thanks for checking this. Indeed there were a couple of unnecessary overloads. I switched this one directly to java.time. CLIRR still found it as a incompatible change, so I added a new clirr ignore entry.

https://github.com/googleapis/sdk-platform-java/blob/2d7950f445c93c77f0999caf190422635e50f1dd/java-core/google-cloud-core/clirr-ignored-differences.xml#L21-L27

java-core/google-cloud-core/clirr-ignored-differences.xml
diegomarquezp: Added a comment.

java-core/google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
diegomarquezp: Ah, great catch. I added an overload and now have the original method throwing a threeten exception

https://github.com/googleapis/sdk-platform-java/blob/2d7950f445c93c77f0999caf190422635e50f1dd/java-core/google-cloud-core/src/main/java/com/google/cloud/Timestamp.java#L196-L199

java-core/google-cloud-core/src/main/java/com/google/cloud/testing/BaseEmulatorHelper.java
diegomarquezp: Oh, looks like the datastore folks [use this method directly](https://github.com/googleapis/java-datastore/blob/92952c8e033aaeed7f0dcded1226bbaed577e140/google-cloud-datastore/src/main/java/com/google/cloud/datastore/testing/LocalDatastoreHelper.java#L322-L326), so me may want the overload for the time being? (ci [check failure](https://github.com/googleapis/sdk-platform-java/actions/runs/11671715727/job/32498740618?pr=3330))

diegomarquezp: I restored the overload

java-core/google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
```diff
@@ -189,6 +189,16 @@ public com.google.protobuf.Timestamp toProto() {
     return com.google.protobuf.Timestamp.newBuilder().setSeconds(seconds).setNanos(nanos).build();
   }
 
+  /** This method is obsolete. Use {@link #parseTimestampDuration(String)} instead */
+  public static Timestamp parseTimestamp(String timestamp) {
```
lqiu96, line 193: Add the @ObsoleteApi annotation

java-core/google-cloud-core/src/main/java/com/google/cloud/testing/BaseEmulatorHelper.java
```diff
@@ -112,14 +115,20 @@ protected final void startProcess(String blockUntilOutput)
     }
   }
 
+  /** This method is obsolete. Use {@link #waitForProcessDuration(java.time.Duration)} instead */
```
lqiu96, line 118: Add the @ObsoleteApi annotation

java-core/google-cloud-core/src/main/java/com/google/cloud/testing/BaseEmulatorHelper.java
lqiu96: Ok, right I believe that makes sense. They can't override this method but they can extend from BaseEmulatorHelper and use the parent implementation. Thanks for checking!

java-core/google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
```diff
@@ -54,7 +54,7 @@ public final class Timestamp implements Comparable<Timestamp>, Serializable {
       new DateTimeFormatterBuilder()
           .appendOptional(DateTimeFormatter.ISO_LOCAL_DATE_TIME)
           .optionalStart()
-          .appendOffsetId()
+          .appendZoneOrOffsetId()
```
lqiu96, line 57: qq, why did this change?

diegomarquezp: This is the method I found that could fix the tests. Honestly I did not dig deeper. The javadoc of the new method include:

>The offset ID parsing is equivalent to using appendOffset(String, String) using the arguments 'HH:MM:ss' and the no offset string '0'

PS: I did retry the tests with the original `appendOffsetId()` method and now they work. You will see that [this commit](https://github.com/googleapis/sdk-platform-java/pull/3330/commits/34bb09b9b7f3d821ddc9b7472afbe6b477b7d985) used `appendOffsetId()` and ran into offset parsing problems in the tests (the next commit introduces `asspendZoneAndOffsetId()`. I could also reproduce that problem locally by then.

I now restored it to `appendOffsetId()`

java-core/google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
diegomarquezp: My bad, done.

java-core/google-cloud-core/src/main/java/com/google/cloud/testing/BaseEmulatorHelper.java
diegomarquezp: Done

java-core/google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
diegomarquezp: After restoring, the [checks](https://github.com/googleapis/sdk-platform-java/actions/runs/11673610643/job/32504691386?pr=3330) fail like last time:

```
Error:  Failures: 
Error:    TimestampTest.parseTimestampWithTimeZoneOffset:245 value of: parseTimestampDuration(...)
expected: 2020-07-10T21:03:00Z
but was : 2020-07-10T14:03:00Z
```

diegomarquezp: I confirmed that this test passes in java 11. I could not find any release note between 8 and 11 talking about parsing behavior.

lqiu96: Hmm, it seems worrying that switching from threeten to java.time would cause a difference in behavior only for Java 8. Let's do some digging into `appendOffsetId`. Maybe it's a known regression in the JDK, but just want to check that `appendZoneOrOffsetId()` wouldn't break anything.

diegomarquezp: Ok, looks like it all boils down to https://github.com/adoptium/jdk/commit/c6d209b505932b43d58502b119b86a6fa5b5338e

That commit (in Java 9) was a fix meant for [this issue](https://bugs.openjdk.org/browse/JDK-8066982)

At least we now know that it's a bug. The affected line is `Timestamp.java:215`:
https://github.com/googleapis/sdk-platform-java/blob/685a3e91236f29c29cae42cc37ccdd1ed30a1e54/java-core/google-cloud-core/src/main/java/com/google/cloud/Timestamp.java#L213-L217

Next is to figure out why `appendZoneOrOffsetId` works:

What I know is that JDK 8 will [correctly use](https://github.com/adoptium/jdk8u/blob/31b88042fba46e87fba83e8bfd43ae0ecb5a9afd/jdk/src/share/classes/java/time/format/Parsed.java#L589-L591) the `zone` property to compute the epoch seconds. However, `zone` is only set when using [ZoneIdPrinterParser](https://github.com/adoptium/jdk8u/blob/31b88042fba46e87fba83e8bfd43ae0ecb5a9afd/jdk/src/share/classes/java/time/format/DateTimeFormatterBuilder.java#L3776) ([this](https://github.com/adoptium/jdk8u/blob/31b88042fba46e87fba83e8bfd43ae0ecb5a9afd/jdk/src/share/classes/java/time/format/DateTimeFormatterBuilder.java#L3908) is where the zone is set based on the offset via [`setParsed(ZoneId)`](https://github.com/adoptium/jdk8u/blob/31b88042fba46e87fba83e8bfd43ae0ecb5a9afd/jdk/src/share/classes/java/time/format/DateTimeParseContext.java#L417))

`appendZoneOrOffsetId` [computes](https://github.com/adoptium/jdk8u/blob/31b88042fba46e87fba83e8bfd43ae0ecb5a9afd/jdk/src/share/classes/java/time/format/DateTimeFormatterBuilder.java#L1114) offset strings (e.g. `-01:00`) into `ZoneOffset` classes in its [parsing function](https://github.com/adoptium/jdk8u/blob/31b88042fba46e87fba83e8bfd43ae0ecb5a9afd/jdk/src/share/classes/java/time/format/DateTimeFormatterBuilder.java#L3907) (i.e. computes a `zone` from the offset string). I think that's why: offsets are parsed into zones when using this `PrinterFormatter`, whereas the [`OffsetIdPrinterParser`](https://github.com/adoptium/jdk8u/blob/31b88042fba46e87fba83e8bfd43ae0ecb5a9afd/jdk/src/share/classes/java/time/format/DateTimeFormatterBuilder.java#L3312) used when calling `appendOffsetId()` [sets the offset in a separate field](https://github.com/adoptium/jdk8u/blob/31b88042fba46e87fba83e8bfd43ae0ecb5a9afd/jdk/src/share/classes/java/time/format/DateTimeFormatterBuilder.java#L3408) (which we see it's not respected as pointed in the bug itself).

diegomarquezp: BTW, the one thing that this parser does besides parsing offsets is to [also parse zone strings](https://github.com/adoptium/jdk8u/blob/31b88042fba46e87fba83e8bfd43ae0ecb5a9afd/jdk/src/share/classes/java/time/format/DateTimeFormatterBuilder.java#L1076-L1077) (i.e. `Canada/Toronto`). I believe that would not be harmful, but maybe a test using a textual zone string may suffice?

java-core/google-cloud-core/src/test/java/com/google/cloud/TimestampTest.java
```diff
@@ -237,14 +237,18 @@ void parseTimestampWithoutTimeZoneOffset() {
 
   @Test
   void parseTimestampWithTimeZoneOffset() {
```
diegomarquezp, line 239: From discussion with @blakeli0, let's increment this test to cover all possible cases where the `threeten` behavior differs from `java.time`

diegomarquezp: @blakeli0 I expanded the tests to include more edge cases. The tests using `threeten` behave in the same way (see https://github.com/googleapis/sdk-platform-java/pull/3339).
WDYT?

blakeli0: Thanks, it LGTM. IIUC, `appendZoneOrOffsetId` does add new things like `Zone` to the formatter, but we don't really use it in `Timestamp`. The parsed `Timestamp` in the end only includes  `seconds` and `nanos` which should not be affected by this change.

Can you please add a few test cases to this PR? For example, test parsing Strings with zone info?

diegomarquezp: Thank you.

> test parsing Strings with zone info?

I added a few in a90bf83b4faa8fd3bf9ff7b1513b6c600a0692e4

https://github.com/googleapis/sdk-platform-java/blob/b67db509d5f121a42b70eec1dd5824358053d973/java-core/google-cloud-core/src/test/java/com/google/cloud/TimestampTest.java#L291-L302

Although we don't expect zone IDs to be parsed, I added a few tests to confirm they behave like the docs explain it would.

java-core/google-cloud-core/src/main/java/com/google/cloud/Timestamp.java
diegomarquezp: A backport request from Java 9 to 8 to fix this is still open
https://bugs.openjdk.org/browse/JDK-8149540

lqiu96 (approved):
LGTM, thanks for all the work with confirming the Timestamp behavior!

Let's resolve the GraalVM issues and then merge it in.

PR #3335, opened 2024-11-05
fix: Fix flaky test ScheduledRetryingExecutorTest.testCancelOuterFutureAfterStart

blakeli0:
The test was flaky because the [Thread.sleep(150)](https://github.com/googleapis/sdk-platform-java/blob/b031b18a9bb35b77ca21d3665217aa4c219ced57/gax-java/gax/src/test/java/com/google/api/gax/retrying/ScheduledRetryingExecutorTest.java#L288) does not guarantee the main thread would wake up after exactly 150ms, it might take much longer if the resource is limited on the machine or we are running tests in parallel. 
Hence if the thread sleeps more than [1525ms](https://github.com/googleapis/sdk-platform-java/blob/0cddadb8ad3eddfffa356a479964d8a720937503/gax-java/gax/src/test/java/com/google/api/gax/retrying/ScheduledRetryingExecutorTest.java#L262), the future would be completed already, and the cancellation of the future would result in false. 
Update the total request duration to 32500ms now, it is still not bullet proof but we would have much less chance to run into flaky tests now.

fixes: https://github.com/googleapis/sdk-platform-java/issues/2669


lqiu96 (approved):
LGTM, let's try it and see if this helps reduce flakiness

PR #3338, opened 2024-11-06
fix: use BuildKit instead of legacy builder to build the Hermetic Build images

diegomarquezp:
Fixes the following error:

image

[Docker Buildkit](https://docs.docker.com/build/buildkit/) is a more advanced version and more actively maintained. We confirmed locally that the error occurs with an old version of the docker builder and that it is fixed when using `DOCKER_BUILDKIT=1`

### DIFF
--- a/.cloudbuild/library_generation/cloudbuild-library-generation-push-prod.yaml
+++ b/.cloudbuild/library_generation/cloudbuild-library-generation-push-prod.yaml
@@ -30,6 +30,8 @@ steps:
       "--file", ".cloudbuild/library_generation/library_generation.Dockerfile", "."]
     id: library-generation-build
     waitFor: ["-"]
+    env: 
+      - 'DOCKER_BUILDKIT=1'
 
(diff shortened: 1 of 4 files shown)


JoeWang1127 (approved):
Could you specify the error in the PR description?

blakeli0 (commented):
Now that it is not only used in local, but for Cloud Build too, can we update the [doc](https://github.com/googleapis/sdk-platform-java/blame/main/hermetic_build/DEVELOPMENT.md#L168-L172) again to reflect it? 

Also maybe we should change how we build the image in [integration tests](https://github.com/googleapis/sdk-platform-java/blob/main/hermetic_build/library_generation/tests/integration_tests.py#L199-L205) too. So that we have a  consistent way to build hermetic build images in all places.

blakeli0 (approved):

blakeli0:
Please fix the linting issue before merging.

PR #3341, opened 2024-11-06
chore: randomize temp output directory

JoeWang1127:
In this PR:
- Use `$RANDOM` (a shell variable that generates a random integer between 0 and 32767) to randomize temp output directory to support parallel generation.

### DIFF
--- a/.cloudbuild/library_generation/library_generation.Dockerfile
+++ b/.cloudbuild/library_generation/library_generation.Dockerfile
@@ -62,7 +62,7 @@ ENV HOME=/home
 ENV OS_ARCHITECTURE="linux-x86_64"
 
 # install OS tools
-RUN apk update && apk add unzip curl rsync openjdk11 jq bash nodejs npm git uuidgen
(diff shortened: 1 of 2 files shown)


.cloudbuild/library_generation/library_generation.Dockerfile
```diff
@@ -62,7 +62,7 @@ ENV HOME=/home
 ENV OS_ARCHITECTURE="linux-x86_64"
 
 # install OS tools
-RUN apk update && apk add unzip curl rsync openjdk11 jq bash nodejs npm git
+RUN apk update && apk add unzip curl rsync openjdk11 jq bash nodejs npm git uuidgen
```
blakeli0, line 65: Is there anyway to generate a random String without a new dependency? Every new dependency is subject to vulnerabilities.

JoeWang1127: Took sometime but I found `$RANDOM` should be sufficient.

blakeli0 (approved):

PR #3365, opened 2024-11-12
fix: instantiate GaxProperties at build time to ensure we get the protobuf version

ldetmer:
This should not vary from the runtime version for graalvm builds.

Tested: ran ITVersion tests in graalvm and printed out the header to see full value exists

### DIFF
--- a/gax-java/gax/src/main/resources/META-INF/native-image/com.google.api/gax/native-image.properties
+++ b/gax-java/gax/src/main/resources/META-INF/native-image/com.google.api/gax/native-image.properties
@@ -1,7 +1,9 @@
 Args = --enable-url-protocols=https,http \
 --initialize-at-build-time=org.conscrypt,\
   org.junit.platform.engine.TestTag,\
-  com.google.api.gax.core.GaxProperties \
+  com.google.api.gax.core.GaxProperties,\
+  com.google.common.base.Platform,\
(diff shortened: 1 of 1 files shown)


burkedavison (approved):

gax-java/gax/src/main/resources/META-INF/native-image/com.google.api/gax/native-image.properties
```diff
@@ -1,6 +1,9 @@
 Args = --enable-url-protocols=https,http \
 --initialize-at-build-time=org.conscrypt,\
-  org.junit.platform.engine.TestTag \
+  org.junit.platform.engine.TestTag,\
+  com.google.api.gax.core.GaxProperties,\
+  com.google.common.base.Platform,\
+  com.google.common.base.Platform$JdkPatternCompiler \
```
mpeddada1, line 6: Were these transitively initialized at build time due to GaxProperties?

mpeddada1 (approved):

gax-java/gax/src/main/resources/META-INF/native-image/com.google.api/gax/native-image.properties
ldetmer: yes via the Strings function

PR #3379, opened 2024-11-15
fix: return all library names if repo-level parameter changes

JoeWang1127:
In this PR:
- Return all library names as a comma-separated string if repo-level parameter changes.

Fix: #3381

### DIFF
--- a/hermetic_build/common/model/config_change.py
+++ b/hermetic_build/common/model/config_change.py
@@ -62,7 +62,6 @@ class QualifiedCommit:
 
 
 class ConfigChange:
-    ALL_LIBRARIES_CHANGED = "ALL_LIBRARIES"
 
     def __init__(
         self,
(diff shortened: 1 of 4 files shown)


hermetic_build/common/model/config_change.py
```diff
@@ -74,16 +74,17 @@ def __init__(
...
-        in the current_config will be generated.
+        A special list [ALL_LIBRARIES], will be returned if there is a
+        repository level change, which means all libraries in the current_config
+        will be generated.
 
         :return: library names of change libraries.
         """
         if ChangeType.REPO_LEVEL_CHANGE in self.change_to_libraries:
-            return ConfigChange.ALL_LIBRARIES_CHANGED
+            return [ConfigChange.ALL_LIBRARIES_CHANGED]
```
blakeli0, line 87: The previous behavior was that we return a list if there are changed libraries or a String `ConfigChange.ALL_LIBRARIES_CHANGED` if there is no changes?

blakeli0, line 87: Can we return all library names if `ChangeType.REPO_LEVEL_CHANGE`? We don't have to add a magic String just for all libraries changes.

JoeWang1127: The previous behavior:
- Return `None` if all libraries changed
- String of library names if some libraries changed
- Empty string if no library changed

The problem is `None` is interpreted as empty string in the shell script.

JoeWang1127: Done.

blakeli0 (approved):

PR #3385, opened 2024-11-19
chore(hermetic-build): update final base image and owlbot dependency

diegomarquezp:


.cloudbuild/library_generation/library_generation.Dockerfile
```diff
@@ -64,6 +62,9 @@ ENV OS_ARCHITECTURE="linux-x86_64"
 # install OS tools
 RUN apk update && apk add unzip curl rsync openjdk11 jq bash nodejs npm git
 
+# Remove unnecessary cross-spawn from npm to resolve CVE-2024-21538
+RUN rm -rf /usr/lib/node_modules/npm/node_modules/cross-spawn/
```
blakeli0, line 66: Does the latest npm includes a fix for it? If not, I think we should wait for it rather than this verify customized fix. Removing it could resolve the vulnerability, but based on the [description of the vulnerability](https://nvd.nist.gov/vuln/detail/CVE-2024-21538), I don't think it would affect hermetic build.

JoeWang1127: `cross-spawn` comes from npm CLI, the CLI updated the dependency in [v9.9.4](https://github.com/npm/cli/releases/tag/v9.9.4). However, the latest version of `npm` in [Alpine Package Keeper](http://dl-cdn.alpinelinux.org/alpine/v3.16/community/x86_64/) (apk) is still 8.10.0.

The waiting time is out of our control so how about I remove `npm` afterwards.

blakeli0: What is the SLO of fixing this issue?

.cloudbuild/library_generation/library_generation.Dockerfile
```diff
@@ -126,6 +124,7 @@ RUN git checkout "${OWLBOT_CLI_COMMITTISH}"
 RUN npm i && npm run compile && npm link
 RUN owl-bot copy-code --version
 RUN chmod o+rx $(which owl-bot)
+RUN apk del -r npm && apk cache clean
```
blakeli0, line 127: I like this approach! We don't need `npm` in our image as long as `owl-bot` package is already built.

.github/workflows/verify_library_generation.yaml
```diff
@@ -27,7 +27,7 @@ jobs:
           git checkout -b "${head_ref}" fork/${head_ref}
           changed_directories="$(git diff --name-only "fork/${head_ref}" "origin/${base_ref}")"
         fi
-        if [[ ${changed_directories} =~ "hermetic_build/" ]]; then
+        if [[ ${changed_directories} =~ "hermetic_build/" ]] || [[ ${changed_directories} =~ ".cloudbuild/library_generation/" ]]; then
```
blakeli0, line 30: Is this change related?

JoeWang1127: This change aims to run hermetic build test if Dockerfile is changed.

.cloudbuild/library_generation/library_generation.Dockerfile
JoeWang1127: Before 2025/1/29

blakeli0 (approved):

PR #3386, opened 2024-11-19
fix: plumb mtlsEndpoint to gRPC Channel provider instead of setting it in EndpointContext.

rmehta19:
Fix a problem where if DirectPath is to be used (`canUseDirectPath` is going to return true), but endpoint resolution determines conditions are met to use S2A, the endpoint incorrectly gets set to mtls endpoint. 

We only try S2A logic (e.g we only look at `useS2A`), if we have ruled out DirectPath as an option (`canUseDirectPathI()` returns false), but endpoint resolution logic doesn't know this, and will set endpoint to mtls endpoint if it determines S2A can be used, even if DirectPath can be used.

### DIFF
--- a/gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
+++ b/gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
@@ -272,10 +272,6 @@ public abstract class EndpointContext {
 
     /** Determines the fully resolved endpoint and universe domain values */
     private String determineEndpoint() throws IOException {
-      if (shouldUseS2A()) {
-        return mtlsEndpoint();
-      }
-
       MtlsProvider mtlsProvider = mtlsProvider() == null ? new MtlsProvider() : mtlsProvider();
       // TransportChannelProvider's endpoint will override the ClientSettings' endpoint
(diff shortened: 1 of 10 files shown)


rmehta19:
@lqiu96 @zhumin8 @blakeli0

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -272,10 +272,6 @@ private String determineUniverseDomain() {
 
     /** Determines the fully resolved endpoint and universe domain values */
     private String determineEndpoint() throws IOException {
-      if (shouldUseS2A()) {
```
rmehta19, line 275: WDYT of renaming this to `canUseS2A()`, since that's what the result of this function is really indicating. 

For example, it is possible that this function returns true and `canUseDirectPath` in InstantiatingGrpcChannelProvider` returns true, and we therefore end up not using S2A.

rmehta19:
Closing this and patching this into #3548

rmehta19:
Closing this and patching this into https://github.com/googleapis/sdk-platform-java/pull/3548

PR #3391, opened 2024-11-21
docs: hermetic build README syntax fix

burkedavison:
Providing the same `/workspace` mount point for two volumes produces an error. Expected second mount point is `/workspace/apis`


ldetmer (approved):

JoeWang1127 (approved):
Thank you.

PR #3397, opened 2024-11-25
chore: update action version through shell script

JoeWang1127:
In this PR:
- Update composite action version in `update_generation_config.sh` so that the action version is updated in the same PR with image update.
- Do not update the action version through renovate.


diegomarquezp (approved):
LGTM

PR #3402, opened 2024-11-27
deps: Update OpenTelemetry semantic convention packages in the shared dependencies

michaelpri10:
Updates a dependency ID that was added in #3020. See b/381286676.


diegomarquezp (approved):

java-shared-dependencies/third-party-dependencies/pom.xml
```diff
@@ -139,6 +139,11 @@
         <artifactId>opentelemetry-grpc-1.6</artifactId>
         <version>${opentelemetry-grpc-instrumentation.version}</version>
       </dependency>
+      <dependency>
+        <groupId>io.opentelemetry.semconv</groupId>
+        <artifactId>opentelemetry-semconv</artifactId>
+        <version>${opentelemetry-semconv.version}</version>
+      </dependency>
       <dependency>
```
lqiu96, line 147: Should the `io.opentelemetry:opentelemetry-semconv` deps below (L147-L150) be removed since it's now `io.opentelemetry.semconv:opentelemetry-semconv`?

michaelpri10: I had tried that originally, but saw some checks fail because the Pub/Sub library currently uses the old `io.opentelemetry:opentelemetry-semconv` dependency. I was planning to remove that old one in a follow-up PR after the dependency is removed in Pub/Sub.

lqiu96: Ah, makes sense. I believe our downstream tests use the latest released version of pubsub. In that case, would you mind either creating an issue in this repo or adding a comment in the ticket so we don't forget to remove the old deps?

lqiu96 (approved):

PR #3403, opened 2024-11-27
feat: add client side logging with slf4j

zhumin8:
Changes to introduce client logging debug feature to Gax. Guide on usage will be added separately to README file.

- Brings in `slf4j-api` as optional dependency to gax-java
- `LoggingUtils`, `LogData` and `LoggerProvider` are public because access are needed from gax-grpc and gax-httpjson packages
  - `LoggingUtils` handles logic to enable/disable from env var, and contains shared utility methods for record data for logging and record logs (abstracting away any need for SLF4J classes). 
  - `Slf4jUtils` any logic interacting with SLF4J classes.
  - `LoggerProvider` provides the SLF4J Logger. This is so that Logger interceptor classes do not declare Logger directly.
- This feature is guarded by env var GOOGLE_SDK_JAVA_LOGGING, only turned on if true. 
- By default it is off, user app should act as before (usual tests behaves as usual)

__Tests added:__
- Unit tests in Gax that need either GOOGLE_SDK_JAVA_LOGGING unset, or do not depend on env var
- LoggingEnabledTest: test for logger correctly setups when GOOGLE_SDK_JAVA_LOGGING = true. This is added to existing `envVarTest` profile
- Showcase tests:
  - ITLoggingDisabled: test no logging event should record when env var is turned off.
  - ITLogging: test logging event with slf4j2.x+logback
  - ITLogging1x: test logging event with slf4j1.x+logback
These tests may not compile depending on the logging dependency brought in. Set up profiles to control test compile and run:
    - Slf4j2_logback, slf4j1_logback, disabledLogging: brings in logging dependencies and setup compile config exclusions
    - Do not include `it/logging` folder for compile by default, or `enable-golden-tests`. `native` is also excluded for now.

__Notable changes since last reviewed:__
- Changes related to get slf4j-api deps as optional
- Logics to switch logging behavior based on slf4j major version
- Moved helper methods into Gax
- Revamped test setup
- use Protobuf utils to serialize message

__Context:__ [go/java-client-logging-design](http://goto.google.com/java-client-logging-design)

TODO:
- add rules for renovate bot to bypass some versions for testing


gax-java/gax/src/main/java/com/google/api/gax/logging/LoggingUtils.java
```diff
@@ -0,0 +1,467 @@
...
+      default:
+        logger.info(message);
+        // Default to INFO level
+    }
+
+    MDC.clear();
+  }
+
+  // JulWrapperLogger implementation
+  static class JulWrapperLogger implements Logger {
```
blakeli0, line 139: This JulWrapper seems complicated and we have to override a bunch of methods without implementing them. I think it's OK that we don't include this in our first release.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcLoggingInterceptor.java
```diff
@@ -0,0 +1,118 @@
...
+
+    // Initialize a JsonArray to hold all responses
+    JsonArray responsePayloads = new JsonArray();
+    return new ForwardingClientCall.SimpleForwardingClientCall<ReqT, RespT>(
+        next.newCall(method, callOptions)) {
+
+      @Override
+      public void start(Listener<RespT> responseListener, Metadata headers) {
+        if (logger.isInfoEnabled()) {
+          String requestId = UUID.randomUUID().toString();
```
blakeli0, line 35: I think `requestId` is for debugging purpose for now and would be removed later?

gax-java/gax/src/main/java/com/google/api/gax/logging/JsonContextMapHandler.java
```diff
@@ -0,0 +1,82 @@
...
+
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.logging.ConsoleHandler;
+import java.util.logging.Handler;
+import java.util.logging.LogRecord;
+
+public class JsonContextMapHandler extends Handler {
```
blakeli0, line 41: I think this class is only used in tests to format logs to Json?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcLoggingInterceptor.java
```diff
@@ -0,0 +1,118 @@
+package com.google.api.gax.grpc;
+
+import com.google.api.gax.logging.LoggingUtils;
+import com.google.gson.Gson;
+import com.google.gson.JsonArray;
+import com.google.gson.JsonObject;
+import io.grpc.*;
```
blakeli0, line 7: Avoid importing `*`.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcLoggingInterceptor.java
zhumin8: Yes for now. Later when we also log in tracer, we may need some requestID to track (not this one).

gax-java/gax/src/main/java/com/google/api/gax/logging/JsonContextMapHandler.java
zhumin8: This is to use with JUL fallback scenario, to format logs. We can provide it as a convenience to users.

gax-java/gax/src/main/java/com/google/api/gax/logging/LoggingUtils.java
zhumin8: As discussed offline, will not add fallback option for first release.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonLoggingInterceptor.java
```diff
@@ -0,0 +1,197 @@
...
+
+  private void recordRequestHeaders(HttpJsonMetadata headers, LogData.Builder logDataBuilder) {
+    try {
+      if (logger.isDebugEnabled()) {
+        JsonObject requestHeaders = new JsonObject();
+        headers
+            .getHeaders()
+            .forEach((key, value) -> requestHeaders.addProperty(key, value.toString()));
+        logDataBuilder.requestHeaders(gson.toJson(requestHeaders));
+        logDataBuilder.requestHeaders(gson.toJson(requestHeaders));
```
jskeet, line 136: Why are we calling this twice?

zhumin8: That's a mistake. Thanks for catching this!

zhumin8:
ci intergration are failing because cannot find slf4j classes.
```
external/com_google_api_gax_java/gax/src/main/java/com/google/api/gax/logging/LoggingUtils.java:39: error: symbol not found org.slf4j.Logger
import org.slf4j.Logger;
```
I added deps to dependencies.properties file: [172701d](https://github.com/googleapis/sdk-platform-java/pull/3403/commits/172701dfb642312366b693a4be37cdf8b68b5efb)
Is any where else I need to add these dependencies to? Or are the names malformed? @blakeli0

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcLoggingInterceptor.java
```diff
@@ -0,0 +1,197 @@
...
+import io.grpc.Status;
+import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.event.Level;
+
+@InternalApi
+public class GrpcLoggingInterceptor implements ClientInterceptor {
+
+  private static final Logger LOGGER = LoggingUtils.getLogger(GrpcLoggingInterceptor.class);
+  private static final Gson GSON = new Gson();
```
blakeli0, line 55: I don't think `gax-grpc` depends on `gson` as of now, we need to be careful introducing new dependencies. Can we try to not use `gson` if possible? If we have to use it, can we extract the common logic to `gax`?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcLoggingInterceptor.java
```diff
@@ -0,0 +1,197 @@
...
+      }
+    } catch (Exception e) {
+      LOGGER.error("Error recording response headers", e);
+    }
+  }
+
+  <RespT> void recordResponsePayload(RespT message, LogData.Builder logDataBuilder) {
+    try {
+      if (LOGGER.isDebugEnabled()) {
+        logDataBuilder.responsePayload(GSON.toJsonTree(message));
```
blakeli0, line 146: `message` is a Protobuf message, so we should use Protobuf utils to serialize it to Json if we have to. See how we serialized it in [ProtoRestSerializer](https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/ProtoRestSerializer.java#L82). Same comment for the request payload logging below.

gax-java/gax/src/main/java/com/google/api/gax/rpc/internal/SystemEnvironmentProvider.java
```diff
@@ -0,0 +1,50 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package com.google.api.gax.rpc.internal;
+
+import com.google.api.core.InternalApi;
+
+/** Represents the default system environment provider. */
+@InternalApi
+public class SystemEnvironmentProvider implements EnvironmentProvider {
```
blakeli0, line 37: Is this class needed? Can we do [System::getEnv](https://github.com/googleapis/sdk-platform-java/blob/d7b7dd98314de957b048c9ddb70ad3089f39fc63/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L626)?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcLoggingInterceptor.java
```diff
@@ -0,0 +1,197 @@
...
+  public <ReqT, RespT> ClientCall<ReqT, RespT> interceptCall(
+      MethodDescriptor<ReqT, RespT> method, CallOptions callOptions, Channel next) {
+
+    return new ForwardingClientCall.SimpleForwardingClientCall<ReqT, RespT>(
+        next.newCall(method, callOptions)) {
+      LogData.Builder logDataBuilder = LogData.builder();
+
+      @Override
+      public void start(Listener<RespT> responseListener, Metadata headers) {
+        logRequestInfo(method, logDataBuilder, LOGGER);
```
blakeli0, line 69: I think we should try not to build the LogData if `GOOGLE_SDK_JAVA_LOGGING` is not enabled, the performance impact could be high if the request/response payload is complex.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcLoggingInterceptor.java
```diff
@@ -0,0 +1,197 @@
...
+
+      @Override
+      public void sendMessage(ReqT message) {
+        logRequestDetails(message, logDataBuilder);
+        super.sendMessage(message);
+      }
+    };
+  }
+
+  // Helper methods for logging
```
blakeli0, line 103: Since we are exposing public util classes in `gax` already, I think it's better to move these to `gax` to prevent duplication. This can also help us remove the dependency of `gson` in `gax-grpc`, and `LogData` does not have to exposed publicly.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcLoggingInterceptor.java
```diff
@@ -0,0 +1,197 @@
...
+import io.grpc.ForwardingClientCallListener.SimpleForwardingClientCallListener;
+import io.grpc.Metadata;
+import io.grpc.MethodDescriptor;
+import io.grpc.Status;
+import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.event.Level;
+
+@InternalApi
+public class GrpcLoggingInterceptor implements ClientInterceptor {
```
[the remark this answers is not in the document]
zhumin8: With the current test setup, these needs to be public for testing purposes. I'll look into the test setup again

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcLoggingInterceptor.java
zhumin8: I was under impression that gax-grpc already depends on gson. I will try to extract to gax.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcLoggingInterceptor.java
zhumin8: Yes. I was inclined to do this as refactoring later, but I agree these are better of as utils methods from gax.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcLoggingInterceptor.java
zhumin8: Thanks for the pointer. Will look into.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcLoggingInterceptor.java
```diff
@@ -0,0 +1,197 @@
...
+      if (logger.isInfoEnabled()) {
+        logDataBuilder.serviceName(method.getServiceName()).rpcName(method.getFullMethodName());
+
+        if (!logger.isDebugEnabled()) {
+          LoggingUtils.logWithMDC(
+              logger, Level.INFO, logDataBuilder.build().toMapRequest(), "Sending gRPC request");
+        }
+      }
+    } catch (Exception e) {
+      logger.error("Error logging request info (and headers)", e);
```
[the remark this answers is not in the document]
zhumin8: These are catch-all safe guard to ensure logging logic never cause user application to crash. Especially due to original timeline, I was concerned that I would miss anything. I will definitely take another look at these logging logic.

That said, I still lean to keep these try-catch blocks as safety net. If we are concerned about flooding customers's log, maybe we can do nothing in the catch and let logging silently fail, which I think is acceptable.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcLoggingInterceptor.java
zhumin8: I don't think I fully get your concern here. Are you concerned about initializing of this LogData.Builder?
Otherwise, the actual building, e.g. adding request/response payload, are guarded with `LOGGER.isDebugEnabled()`.
When `GOOGLE_SDK_JAVA_LOGGING` is not enabled, `LOGGER.isDebugEnabled()` is false.

gax-java/gax/src/main/java/com/google/api/gax/rpc/internal/SystemEnvironmentProvider.java
zhumin8: This class is needed because of the current test setup, again was something of lower priority because it is not public and only for test setup. 
I will look into migrate tests to only use ci setups to control env vars, in that case, I will be able to get rid of this.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcLoggingInterceptor.java
blakeli0: Discussed offline that this is not a concern anymore. Because if `GOOGLE_SDK_JAVA_LOGGING` is not enabled, `LOGGER` would return a `NO_OP_LOGGER` that would return always return false for  `isDebugEnabled()`.

gax-java/pom.xml
```diff
@@ -205,6 +210,20 @@
       </exclusions>
       <scope>test</scope>
     </dependency>
+    <!--    Logback  for testing -->
+    <dependency>
```
blakeli0, line 214: It would be great if we don't have to add logback in unit tests. Maybe create our own SLF4J test logger, similar to the JUL [TestLogger](https://github.com/googleapis/sdk-platform-java/blob/c334a0c67acaad0d3de48100fd3b7cff78f4aef5/java-core/google-cloud-core/src/test/java/com/google/cloud/testing/LogRecorderTest.java#L74)

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcLoggingInterceptor.java
blakeli0, line 69: There is a chance that a call is started but failed before sending messages, so it would be better if we only record the log data (headers) here, and do the actual logging in `sendMessage`.

gax-java/gax/src/main/java/com/google/api/gax/logging/LoggingUtils.java
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  public static <RespT> void recordResponsePayload(
+      RespT message, LogData.Builder logDataBuilder, LoggerProvider loggerProvider) {
+    if (loggingEnabled) {
+      Slf4jUtils.recordResponsePayload(message, logDataBuilder, loggerProvider);
+    }
+  }
+
+  public static void logResponse(
```
lqiu96, line 74: nit: Is there a difference in behavior between method prefixes: `record...` vs `log...`? Mostly wondering if they could all share the same prefix `record` or `log`?

gax-java/gax/src/main/java/com/google/api/gax/logging/Slf4jUtils.java
```diff
@@ -0,0 +1,269 @@
...
+            logDataBuilder.responseStatus(status);
+          }
+          if (logger.isInfoEnabled() && !logger.isDebugEnabled()) {
+            Map<String, Object> responseData = logDataBuilder.build().toMapResponse();
+            Slf4jUtils.log(logger, Level.INFO, responseData, "Received Grpc response");
+          }
+          if (logger.isDebugEnabled()) {
+            Map<String, Object> responsedDetailsMap = logDataBuilder.build().toMapResponse();
+            Slf4jUtils.log(logger, Level.DEBUG, responsedDetailsMap, "Received Grpc response");
+          }
```
lqiu96, line 246: I can't see to find the difference between the two if blocks besides the log level. Is this the only difference?

gax-java/dependencies.properties
```diff
@@ -76,6 +76,8 @@ maven.com_google_http_client_google_http_client_gson=com.google.http-client:goog
 maven.org_codehaus_mojo_animal_sniffer_annotations=org.codehaus.mojo:animal-sniffer-annotations:1.24
 maven.javax_annotation_javax_annotation_api=javax.annotation:javax.annotation-api:1.3.2
 maven.org_graalvm_sdk=org.graalvm.sdk:nativeimage:24.1.2
+maven.org_slf4j_slf4j_api=org.slf4j:slf4j-api:2.0.16
```
lqiu96, line 79: Do you know `dependencies.properties` (or the bazel build) also have some form of optional dependency? Would this be something that we also need to add?

gax-java/pom.xml
```diff
@@ -215,6 +227,29 @@
...
+                <artifactId>java18</artifactId>
+                <version>1.0</version>
+              </signature>
+              <ignores>
+                <ignore>org.slf4j.*</ignore>
+              </ignores>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
```
lqiu96, line 252: can you add a note why this is needed

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcLoggingInterceptor.java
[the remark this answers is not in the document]
lqiu96: bumping this now again. Does this still need to be public with InternalApi?

showcase/gapic-showcase/pom.xml
```diff
@@ -215,5 +215,144 @@
...
+        <dependency>
+          <groupId>org.slf4j</groupId>
+          <artifactId>slf4j-api</artifactId>
+          <version>2.0.16</version>
+          <scope>test</scope>
+        </dependency>
+        <dependency>
+          <groupId>ch.qos.logback</groupId>
+          <artifactId>logback-classic</artifactId>
+          <!--   Latest version with java8 support   -->
```
lqiu96, line 241: does showcase run on java 8? Would it possible to use the a newer/ later version of logback to test with?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/logging/ITLogging.java
```diff
@@ -0,0 +1,207 @@
...
+    httpjsonClient.awaitTermination(
+        TestClientInitializer.AWAIT_TERMINATION_SECONDS, TimeUnit.SECONDS);
+  }
+
+  @Test
+  void testGrpc_receiveContent_logDebug() {
+    TestAppender testAppender = setupTestLogger(GrpcLoggingInterceptor.class);
+    assertThat(echoGrpc(ECHO_STRING)).isEqualTo(ECHO_STRING);
+
+    assertThat(true).isTrue();
```
lqiu96, line 98: is this a testing artifact left over?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/logging/ITLogging.java
```diff
@@ -0,0 +1,207 @@
...
+
+    assertThat(testAppender.events.size()).isEqualTo(2);
```
lqiu96, line 100: Can you leave a comment explaining a few of these magic numbers. Why is events.size() == 2 here and events.size() == 4 below?

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/logging/ITLogging.java
```diff
@@ -0,0 +1,207 @@
...
+        assertThat(payload.get("content")).isEqualTo(ECHO_STRING);
+        assertThat(payload.get("requestId")).isNotNull();
+        assertThat(payload.get("otherRequestId")).isNotNull();
+      }
+    }
+    testAppender.stop();
+  }
+
+  @Test
+  void testHttpJson_receiveContent_logDebug() {
```
lqiu96, line 148: could there also be `logInfo` equivalent test case well?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -681,6 +681,7 @@ private ManagedChannel createSingleChannel() throws IOException {
     builder =
         builder
             .intercept(new GrpcChannelUUIDInterceptor())
+            .intercept(new GrpcLoggingInterceptor())
```
lqiu96, line 684: just a thought (I don't know if there are any downsides or if this would work): Could we gate adding the logging interceptor here?

i.e. Check if the logging env var exists + is configured and only add an interceptor if so.

I think there was some mention about not using interceptors (potentially) so it may not work in the future.

gax-java/gax/src/main/java/com/google/api/gax/logging/LoggingUtils.java
zhumin8: `record...` just stores data, `log...` actually writes logs. 
I will add some comments to make this more clear.

gax-java/gax/src/main/java/com/google/api/gax/logging/Slf4jUtils.java
zhumin8: Yes, you are right. the `responseData` content passed in is different because of previous `record..` steps, but in this step, only difference is log level.  Reason to put map creation step inside if block, is to get it guarded so it does not create if info/debug is not enabled.

gax-java/dependencies.properties
zhumin8: I can look into this. But I don't think it really matters as bazel build here is only used for integration tests (which is irrelevant to logging feature), and we do not publish artifact with it.

showcase/gapic-showcase/pom.xml
zhumin8: Good point. My thought was that we would want to test for the oldest version and make sure it works. we test showcase on Java 11 onward. I will find newer version here.
Separately, this may fall into/be related to the project about dep upgrade strategy.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/logging/ITLogging.java
zhumin8: events.size()== 2 --> one for request and one for respond, each tested in below lines.
there is no events.size() == 4, I suppose you meant `assertThat(keyValuePairs.size()).isEqualTo(4);`?
similarly, values from keyValuePairs are checked in more details below that.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
zhumin8: I guess there is no harm in adding a gate here for now.
In the future, another thought is to allow setting enable logging via system properties in addition to env var, in that case, it could change at runtime and this would not work.

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/logging/ITLogging.java
zhumin8: good catch, thank you!

gax-java/dependencies.properties
lqiu96: > bazel build here is only used for integration tests

Do you know if self-service also uses the bazel build? If so, would it make sense for them to also have slfj be optional?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
lqiu96: I was just thinking that having it gate there might be a way to remove all the `if (loggingEnabled) {...}` checks in LoggingUtils.java. But it may not be possible if we know there are future changes that would make this invalid.

gax-java/pom.xml
zhumin8: this is outdated, and I removed in [20017cc](https://github.com/googleapis/sdk-platform-java/pull/3403/commits/20017cc941f9d4fc92212f38c054a9a748e14787)

showcase/gapic-showcase/pom.xml
zhumin8: upgrade done in https://github.com/googleapis/sdk-platform-java/pull/3403/commits/20017cc941f9d4fc92212f38c054a9a748e14787#diff-2786ec798ce7e963cf28c2b9ac79e3e7c7d0ac47114bfec84de71d09cfe93a20

showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/logging/ITLogging.java
zhumin8: added in [7c1ea76](https://github.com/googleapis/sdk-platform-java/pull/3403/commits/7c1ea762649703ceff780bc2b72b564d0788169c)

gax-java/dependencies.properties
zhumin8: Good question. Correct me if I am wrong, I am adding slf4j optional dependency to Gax. For self-service, if even if they use bazel build and build with slf4j (non-optional). Their published artifact depends on a **published** version of GAX? If so, we are good and do not need to alter the bazel process.
For the original question: from a quick search, bazel does not have a easy equivalent of maven's optional.
cc. @blakeli0 if you know any details on the self-service process?

showcase/gapic-showcase/pom.xml
```diff
@@ -215,5 +215,143 @@
...
+    </profile>
+    <profile>
+      <id>disabledLogging</id>
+      <!--   Only compile test to run when   GOOGLE_SDK_JAVA_LOGGING=!true -->
+      <activation>
+        <property>
+          <name>disable_logging</name>
+        </property>
+      </activation>
+      <dependencies>
```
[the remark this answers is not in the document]
zhumin8: I will add a note in comment, this profile tries to test that if logging is not enabled by env, even if logging deps are present, no log will execute. (Test run with this profile: https://github.com/googleapis/sdk-platform-java/blob/1641f5563e762fa1de2cb0189e0ade41474a9d76/showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/logging/ITLoggingDisabled.java#L67)

.github/workflows/ci.yaml
```diff
@@ -255,6 +258,34 @@ jobs:
             -P enable-integration-tests \
             --batch-mode \
             --no-transfer-progress
+      # The `slf4j2_logback` profile brings logging dependency and compiles logging tests, require env var to be set
+      - name: Showcase integration tests - Logging 2x
```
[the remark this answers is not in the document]
zhumin8: done [f611c74](https://github.com/googleapis/sdk-platform-java/pull/3403/commits/f611c74d1cedadbc20b19658724ca84c48feb1a3)

gax-java/pom.xml
```diff
@@ -165,8 +170,15 @@
     </dependencies>
   </dependencyManagement>
 
-  <!-- Test Dependencies -->
+
   <dependencies>
+    <!--    Logging dependency    -->
+    <dependency>
```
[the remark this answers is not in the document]
zhumin8: done [f611c74](https://github.com/googleapis/sdk-platform-java/pull/3403/commits/f611c74d1cedadbc20b19658724ca84c48feb1a3)

gax-java/gax/src/main/java/com/google/api/gax/logging/Slf4jUtils.java
```diff
@@ -0,0 +1,269 @@
...
+  private static boolean hasAddKeyValue;
+
+  static {
+    hasAddKeyValue = checkIfClazzAvailable("org.slf4j.event.KeyValuePair");
+  }
+
+  static boolean checkIfClazzAvailable(String clazzName) {
+    try {
+      Class.forName(clazzName);
+      return true; // SLF4j 2.x or later
```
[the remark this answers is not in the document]
zhumin8: done [539031b](https://github.com/googleapis/sdk-platform-java/pull/3403/commits/539031b0442d408ca407b740d0d5ae68a2765f24)

gax-java/gax/src/main/java/com/google/api/gax/logging/Slf4jUtils.java
```diff
@@ -0,0 +1,269 @@
...
+class Slf4jUtils {
+
+  private static final Logger NO_OP_LOGGER = org.slf4j.helpers.NOPLogger.NOP_LOGGER;
+  private static boolean loggingEnabled = LoggingUtils.isLoggingEnabled();
+  private static final Gson gson = new Gson();
+
+  private static boolean hasAddKeyValue;
+
+  static {
+    hasAddKeyValue = checkIfClazzAvailable("org.slf4j.event.KeyValuePair");
```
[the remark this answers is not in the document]
zhumin8: done in [539031b](https://github.com/googleapis/sdk-platform-java/pull/3403/commits/539031b0442d408ca407b740d0d5ae68a2765f24)

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcLoggingInterceptor.java
```diff
@@ -0,0 +1,134 @@
...
+
+  // Helper methods for logging
+  private static Map<String, String> metadataHeadersToMap(Metadata headers) {
+
+    Map<String, String> headersMap = new HashMap<>();
+    executeWithTryCatch(
+        () -> {
+          for (String key : headers.keys()) {
+
+            if (key.endsWith(Metadata.BINARY_HEADER_SUFFIX)) {
```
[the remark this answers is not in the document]
zhumin8: Unfortunately, there is no equivalent `Metadata.ASCII_STRING_SUFFIX`, only one for identifying binary header ([here](https://github.com/grpc/grpc-java/blob/57af63ad0a0203aa87b20bb29bc56cd9215fb2ef/api/src/main/java/io/grpc/Metadata.java#L61-L66))

However, we should be good since according to grpc's guide [here](https://grpc.io/docs/guides/metadata/#:~:text=while%20the%20values%20can%20be%20either%20ASCII%20strings%20or%20binary%20data.), values can only be either ASCII strings or binary data.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcLoggingInterceptor.java
zhumin8: obsolete

gax-java/pom.xml
zhumin8: already done

gax-java/gax/src/main/java/com/google/api/gax/rpc/internal/SystemEnvironmentProvider.java
zhumin8: outdated. This `SystemEnvironmentProvider` is removed in later changes and we do [System::getEnv](https://github.com/googleapis/sdk-platform-java/blob/d7b7dd98314de957b048c9ddb70ad3089f39fc63/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L626) directly now.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcLoggingInterceptor.java
zhumin8: done

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcLoggingInterceptor.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+  }
+
+  private static JsonObject mapHeadersToJsonObject(Metadata headers) {
+    JsonObject jsonHeaders = new JsonObject();
+    headers
+        .keys()
+        .forEach(
+            key -> {
+              Metadata.Key<String> metadataKey =
```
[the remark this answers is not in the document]
zhumin8: done

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcLoggingInterceptor.java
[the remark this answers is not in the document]
zhumin8: Unfortunately, as conversations diverged to other topics since then, I did not got the chance to revamp the testing. As of now, these still need to be public with InternalApi.

gax-java/dependencies.properties
```diff
@@ -76,6 +76,8 @@ maven.com_google_http_client_google_http_client_gson=com.google.http-client:goog
...
+maven.com_google_protobuf_protobuf_java_util=com.google.protobuf:protobuf-java-util:3.25.5
```
JoeWang1127, line 80: Why do we need this dependency?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcLoggingInterceptor.java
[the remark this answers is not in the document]
JoeWang1127: If this class has to be public due to testing purposes, would it make sense to add a `final`?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcLoggingInterceptor.java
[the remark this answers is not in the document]
JoeWang1127: Could you add a comment on why we need this `if` statement?

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonLoggingInterceptor.java
```diff
@@ -0,0 +1,116 @@
...
+        super.sendMessage(message);
+      }
+    };
+  }
+
+  // Helper methods for logging,
+
+  private static Map<String, String> httpJsonMetadataToMap(HttpJsonMetadata headers) {
+    Map<String, String> headersMap = new HashMap<>();
+    headers.getHeaders().forEach((key, value) -> headersMap.put(key, value.toString()));
```
JoeWang1127, line 113: Why we don't have if statement in this method as in GRPC interceptor?

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/InstantiatingHttpJsonChannelProviderTest.java
```diff
@@ -118,8 +118,12 @@ void managedChannelUsesDefaultChannelExecutor() throws IOException {
...
     ManagedHttpJsonInterceptorChannel interceptorChannel =
         (ManagedHttpJsonInterceptorChannel) httpJsonTransportChannel.getManagedChannel();
-    ManagedHttpJsonChannel managedHttpJsonChannel = interceptorChannel.getChannel();
-    assertThat(managedHttpJsonChannel.getExecutor()).isNotNull();
+    // call getChannel() twice because interceptors are chained in layers by recursive construction
+    // inside com.google.api.gax.httpjson.InstantiatingHttpJsonChannelProvider.createChannel
+    ManagedHttpJsonInterceptorChannel managedHttpJsonChannel =
+        (ManagedHttpJsonInterceptorChannel) interceptorChannel.getChannel();
+    ManagedHttpJsonChannel channel = managedHttpJsonChannel.getChannel();
+    assertThat(channel.getExecutor()).isNotNull();
```
JoeWang1127, line 126: Can we add a new test and leave the existing tests untouched?

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/InstantiatingHttpJsonChannelProvider.java
```diff
@@ -196,6 +196,7 @@ private HttpJsonTransportChannel createChannel() throws IOException, GeneralSecu
     HttpJsonClientInterceptor headerInterceptor =
         new HttpJsonHeaderInterceptor(headerProvider.getHeaders());
 
+    channel = new ManagedHttpJsonInterceptorChannel(channel, new HttpJsonLoggingInterceptor());
```
lqiu96, line 199: nit: Can this work similar to grpcChannel where this either takes in a chain of `.intercept(INTERCEPTORS)` from the builder or a list of interceptors in the constructor? I don't think we need to do `new ManagedHttpJsonInterceptorChannel` twice.

I see this as package-private so we can do whatever is easier and modify it later as well.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcLoggingInterceptor.java
```diff
@@ -0,0 +1,134 @@
...
+    return new ForwardingClientCall.SimpleForwardingClientCall<ReqT, RespT>(
+        next.newCall(method, callOptions)) {
+      LogData.Builder logDataBuilder = LogData.builder();
+
+      @Override
+      public void start(Listener<RespT> responseListener, Metadata headers) {
+        recordServiceRpcAndRequestHeaders(
+            method.getServiceName(),
+            method.getFullMethodName(),
+            null,
```
lqiu96, line 76: Can you add a small comment why this param is intentionally set to null here?

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonLoggingInterceptor.java
```diff
@@ -0,0 +1,116 @@
...
+
+    return new SimpleForwardingHttpJsonClientCall<ReqT, RespT>(next.newCall(method, callOptions)) {
+
+      LogData.Builder logDataBuilder = LogData.builder();
+
+      @Override
+      public void start(
+          HttpJsonClientCall.Listener<RespT> responseListener, HttpJsonMetadata headers) {
+        recordServiceRpcAndRequestHeaders(
+            null,
```
lqiu96, line 69: Can you add a small comment why this param is intentionally set to null here?

gax-java/gax/src/main/java/com/google/api/gax/logging/LogData.java
```diff
@@ -0,0 +1,153 @@
...
+import com.google.api.core.InternalApi;
+import com.google.auto.value.AutoValue;
+import com.google.gson.Gson;
+import java.util.HashMap;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+@InternalApi
+@AutoValue
+public abstract class LogData {
```
lqiu96, line 42: Do you think it would make sense to have something like LogRequestData (or RequestLogData or whatever name) that extends from LogData and the equivalent for Response? Request and Response classes would have fields specific them that aren't shared from LogData

gax-java/dependencies.properties
zhumin8: `slf4j-api` for logging, `protobuf-java-util` for parsing protobuf Message.

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/InstantiatingHttpJsonChannelProviderTest.java
zhumin8: no, the original test would fail.

gax-java/gax/src/main/java/com/google/api/gax/logging/LogData.java
zhumin8: Currently both request and respond logging extracts data from this shared LogData. The main benefit is less overhead in object creation, and both request/respond logging can access shared info(rpc name, service name).

We can create separate structure for request/respond data potentially, but I don't see it necessary at this point as there is no conflict/overrides between request and response data. Considering that we also want to limit classes exposed as public.

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonLoggingInterceptor.java
zhumin8: [091ab38](https://github.com/googleapis/sdk-platform-java/pull/3403/commits/091ab3890033231e2808c7cb5acf21a36b7ef66d)

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcLoggingInterceptor.java
zhumin8: [091ab38](https://github.com/googleapis/sdk-platform-java/pull/3403/commits/091ab3890033231e2808c7cb5acf21a36b7ef66d)

gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/InstantiatingHttpJsonChannelProvider.java
zhumin8: The current ManagedHttpJsonChannel.builder does not take `.intercept()`, `HttpJsonClientInterceptor` is a member added in `ManagedHttpJsonInterceptorChannel`. We can refactor later if we want to better parity between grpc and http setups.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcLoggingInterceptor.java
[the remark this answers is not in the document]
zhumin8: added comment in [091ab38](https://github.com/googleapis/sdk-platform-java/pull/3403/commits/091ab3890033231e2808c7cb5acf21a36b7ef66d)

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcLoggingInterceptor.java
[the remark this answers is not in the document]
zhumin8: unfortunately, adding `final` prevents mocking the class.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcLoggingInterceptor.java
zhumin8: updated to use protobuf utils. see [changes](https://github.com/googleapis/sdk-platform-java/blob/b3b328e770e48c0517c30aa95913a892670548fb/gax-java/gax/src/main/java/com/google/api/gax/logging/Slf4jUtils.java#L172-L176)

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcLoggingInterceptor.java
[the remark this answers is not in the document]
zhumin8: outdated. Offline discussion agreed on try-catch blocks to ensure silent failures for logging if any.

gax-java/gax-httpjson/src/test/java/com/google/api/gax/httpjson/InstantiatingHttpJsonChannelProviderTest.java
JoeWang1127: Could you add a comment about changes in the exising test cases?

gax-java/gax/src/main/java/com/google/api/gax/logging/Slf4jUtils.java
```diff
@@ -0,0 +1,268 @@
...
+  static <RespT> void recordResponsePayload(
+      RespT message, LogData.Builder logDataBuilder, LoggerProvider loggerProvider) {
+    LoggingUtils.executeWithTryCatch(
+        () -> {
+          Logger logger = loggerProvider.getLogger();
+          if (logger.isDebugEnabled()) {
+            Map<String, Object> messageToMapWithGson =
+                Slf4jUtils.messageToMapWithGson((Message) message);
+
+            logDataBuilder.responsePayload(messageToMapWithGson);
```
lqiu96, line 225: qq, how does this work for streaming calls where it's possible to receive multiple payloads per RPC call?

Would each payload be overwritten? Or are the payload chained together?

gax-java/gax/src/main/java/com/google/api/gax/logging/Slf4jUtils.java
```diff
@@ -0,0 +1,268 @@
...
+      String status, LogData.Builder logDataBuilder, LoggerProvider loggerProvider) {
+    LoggingUtils.executeWithTryCatch(
+        () -> {
+          Logger logger = loggerProvider.getLogger();
+          if (logger.isInfoEnabled()) {
+            logDataBuilder.responseStatus(status);
+          }
+          if (logger.isInfoEnabled() && !logger.isDebugEnabled()) {
+            Map<String, Object> responseData = logDataBuilder.build().toMapResponse();
+            Slf4jUtils.log(logger, Level.INFO, responseData, "Received Grpc response");
```
lqiu96, line 240: IIUC, `logResponse` is used by both HttpJson and gRPC. I don't think we should have `Grpc` in the message here.

gax-java/gax/src/main/java/com/google/api/gax/logging/Slf4jUtils.java
```diff
@@ -0,0 +1,269 @@
...
+  }
+
+  static <RespT> void recordResponsePayload(
+      RespT message, LogData.Builder logDataBuilder, LoggerProvider loggerProvider) {
+    LoggingUtils.executeWithTryCatch(
+        () -> {
+          Logger logger = loggerProvider.getLogger();
+          if (logger.isDebugEnabled()) {
+            Map<String, Object> messageToMapWithGson =
+                Slf4jUtils.messageToMapWithGson((Message) message);
```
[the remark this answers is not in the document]
zhumin8: I added [step]([c3db3e7](https://github.com/googleapis/sdk-platform-java/pull/3403/commits/c3db3e7961c27060f2336f571b92ef097eddd652)) to check for type before casting. Avoid generic RespT in this class is tempting, but I think I am inclined to keep logic here (inside of `logger.isDebugEnabled()` if statement) instead of casting before, so that we expose as little as possible when logging is not enabled for the level.

gax-java/gax/src/main/java/com/google/api/gax/logging/Slf4jUtils.java
zhumin8: thanks for catching!

gax-java/gax/src/main/java/com/google/api/gax/logging/Slf4jUtils.java
zhumin8: as discussed, will create a followup ticket to verify/test. Add showcase tests.

showcase/pom.xml
```diff
@@ -103,6 +113,51 @@
...
+            <configuration>
+              <!--   Do not compile by default (without logging deps)  -->
+              <testExcludes>
+                <testExclude>**/com/google/showcase/v1beta1/it/logging/*.java</testExclude>
+              </testExcludes>
+            </configuration>
+          </plugin>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-surefire-plugin</artifactId>
```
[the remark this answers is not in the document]
zhumin8: These are existing plugin set ups for `showcase` profile in current main branch (see [here](https://github.com/googleapis/sdk-platform-java/blob/ad26cf98548e325c99edb263baf8fe1a7696e634/showcase/pom.xml#L101-L137)).

These changes shows up because they are copied in `loggingTestBase` profile, and git compare puts them in awkward order.

gax-java/gax/src/main/java/com/google/api/gax/logging/Slf4jUtils.java
zhumin8: created https://github.com/googleapis/sdk-platform-java/issues/3645 and https://github.com/googleapis/sdk-platform-java/issues/3646

gax-java/gax/src/main/java/com/google/api/gax/logging/LoggingUtils.java
```diff
@@ -0,0 +1,141 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package com.google.api.gax.logging;
+
+import com.google.api.core.InternalApi;
+import java.util.Map;
+
+@InternalApi
+public class LoggingUtils {
```
zhumin8, line 37: From offline discussion, need to improve readability on these logging classes.
Ideas for renaming:
LoggingUtils -> SdkLogger
Slf4jLoggingHelpers -> Slf4jLogger

refactor ideas: create a SDKLogger for public access methods, hide all others

showcase/gapic-showcase/pom.xml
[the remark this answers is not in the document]
zhumin8: Discussed offline:
- first scenario is covered by default profile tests.
- This needs code change if we want more graceful failing. Currently it throws a ClassNotFoundException on org.slf4j.Logger.

blakeli0 (approved):

gax-java/gax/src/main/java/com/google/api/gax/logging/LoggingUtils.java
```diff
@@ -0,0 +1,141 @@
...
+    if (loggingEnabled) {
+      Slf4jLoggingHelpers.logRequest(message, logDataBuilder, loggerProvider);
+    }
+  }
+
+  public static void executeWithTryCatch(ThrowingRunnable action) {
+    try {
+      action.run();
+    } catch (Throwable t) {
+      // let logging exceptions fail silently
```
lqiu96, line 133: I wonder if it would make since to log this error at the lowest levels (maybe TRACE?) about any logging exceptions? 

It would be like using logging to debug logging errors... Which honestly doesn't sound like something people would be doing. Just some thoughts, probably doesn't make any sense.

gax-java/gax/src/main/java/com/google/api/gax/logging/Slf4jLoggingHelpers.java
```diff
@@ -0,0 +1,154 @@
...
+            addIfNotEmpty(logDataBuilder::httpUrl, endpoint);
+          }
+          if (logger.isDebugEnabled()) {
+            logDataBuilder.requestHeaders(requestHeaders);
+          }
+        });
+  }
+
+  private static void addIfNotEmpty(Consumer<String> setter, String value) {
+    if (value != null && !value.isEmpty()) {
```
lqiu96, line 78: nit: Guava's [Strings.isEmptyorNull](https://guava.dev/releases/20.0/api/docs/com/google/common/base/Strings.html#isNullOrEmpty-java.lang.String-)

gax-java/gax/src/main/java/com/google/api/gax/logging/LogData.java
lqiu96: Ok, I think I'm fine with this now, but I am just curious how this is going to work with Streaming calls. We can figure that out as part of the follow up tasks.

lqiu96 (approved):
LGTM. I think we have a few follow up tasks and clean up that we can tackle in future PRs.

PR #3405, opened 2024-12-03
ci: Add Downstream Protobuf-Java Source Compatibility Test

lqiu96:
Set up a nightly CI job that tests Protobuf compatibility with downstream libraries. This step of the CI job is to test source compatibility.

The nightly job will check the compatibility of these libraries with the latest in Protobuf-Java 3.x and and 4.x. This job can be manually invoked with the Protobuf-Java versions as a list.

Future steps may introduce testing for binary compatibility.

### DIFF
--- a/.github/workflows/downstream_protobuf_compatibility_check_nightly.yaml
+++ b/.github/workflows/downstream_protobuf_compatibility_check_nightly.yaml
@@ -37,7 +37,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-java@v4
         with:
-          java-version: 21
+          java-version: 17
           distribution: temurin
       - name: Print Protobuf-Java testing version
         run: echo "Testing with Protobuf-Java v${{ matrix.protobuf-version }}"
--- a/.github/workflows/downstream_protobuf_compatibility_check_nightly.yaml
+++ b/.github/workflows/downstream_protobuf_compatibility_check_nightly.yaml
@@ -15,7 +15,7 @@ name: Downstream Protobuf Compatibility Check Nightly
 jobs:
   downstream-protobuf-test:
     # Checks if PR comes from Release-Please branch or '' if invoked from nightly job
-    if: github.head_ref == 'release-please--branches--main' || github.head_ref == ''
+    if: github.head_ref == 'release-please--branches--main' || github.event_name == 'schedule'
     runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
--- a/.github/workflows/downstream_protobuf_compatibility_check_nightly.yaml
+++ b/.github/workflows/downstream_protobuf_compatibility_check_nightly.yaml
@@ -1,5 +1,4 @@
 on:
-  pull_request:
   workflow_dispatch:
     inputs:
       protobuf_versions:
--- a/.github/workflows/downstream_protobuf_compatibility_check_nightly.yaml
+++ b/.github/workflows/downstream_protobuf_compatibility_check_nightly.yaml
@@ -11,7 +11,7 @@ on:
   schedule:
     - cron: '0 1 * * *' # Nightly at 1am
 
-name: Downstream Compatibility Nightly
+name: Downstream Protobuf Compatibility Check Nightly
 jobs:
   downstream-protobuf-test:
     runs-on: ubuntu-22.04
--- a/.github/workflows/downstream_protobuf_compatibility_check_nightly.yaml
+++ b/.github/workflows/downstream_protobuf_compatibility_check_nightly.yaml
@@ -1,4 +1,5 @@
 on:
+  pull_request:
   workflow_dispatch:
     inputs:
       protobuf_versions:
--- a/.github/workflows/downstream_protobuf_compatibility_check_nightly.yaml
+++ b/.github/workflows/downstream_protobuf_compatibility_check_nightly.yaml
@@ -14,7 +14,7 @@ on:
 name: Downstream Protobuf Compatibility Check Nightly
 jobs:
   downstream-protobuf-test:
-    # Checks if PR comes from Release-Please branch or '' if invoked from nightly job
+    # Checks if PR comes from Release-Please branch or if invoked from nightly job
     if: github.head_ref == 'release-please--branches--main' || github.event_name == 'schedule'
     runs-on: ubuntu-22.04
     strategy:
--- a/.github/workflows/downstream_protobuf_compatibility_check_nightly.yaml
+++ b/.github/workflows/downstream_protobuf_compatibility_check_nightly.yaml
@@ -1,11 +1,11 @@
 on:
   pull_request:
-    # Updates on Release-Please PRs
+    # Runs on PRs targeting main, but will be filtered for Release PRs
     branches:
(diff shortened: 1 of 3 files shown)


.kokoro/presubmit/downstream-protobuf-source-compatibility.sh
blakeli0, line 1: Why do we put it under `.kokoro/presubmit`? It is actually not running from kokoro, is it because we plan to run it from kokoro alongside the integration tests in the future?

lqiu96: No particular reason. I just coped the implementation of ./downstream-compatibility.sh. I can move it to a nightly folder for now. 

I don't think the existing presubmits run on kokoro. Perhaps it can be re-named to something else?

lqiu96: I moved it over the nightly folder, but kept the kokoro name for now

.github/workflows/downstream_protobuf_compatibility_check_nightly.yaml
```diff
@@ -0,0 +1,48 @@
+on:
+  pull_request:
+    # Updates on Release-Please PRs
+    branches:
+      - 'release-please--branches--main'
```
blakeli0, line 5: I thought this means this workflow would run on any PRs targeting 'release-please--branches--main', not PRs created from 'release-please--branches--main', can you verify it?

.github/workflows/downstream_protobuf_compatibility_check_nightly.yaml
```diff
@@ -0,0 +1,48 @@
...
+    - cron: '0 1 * * *' # Nightly at 1am
+
+name: Downstream Protobuf Compatibility Check Nightly
+jobs:
+  downstream-protobuf-test:
+    runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        repo:
```
blakeli0, line 21: Are all of them dependent on `protobuf-java`? For example, I don't see `java-storage-nio` is dependent on `protobuf-java`. Or we are adding all handwritten libraries here in case they start depending on `protobuf-java` in the future?

.kokoro/nightly/downstream-protobuf-source-compatibility.sh
```diff
@@ -0,0 +1,51 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eo pipefail
+
+# Comma-delimited list of repos to test with the local java-shared-dependencies
+if [ -z "${REPOS_UNDER_TEST}" ]; then
+  echo "REPOS_UNDER_TEST must be set to run downstream-protobuf-source-compatibility.sh"
```
blakeli0, line 20: Can we provide more detailed error hint if they are not supplied? For example, what are the expected format of the input etc. Same comment for `PROTOBUF_RUNTIME_VERSION` below.

.kokoro/nightly/downstream-protobuf-source-compatibility.sh
```diff
@@ -0,0 +1,51 @@
...
+  echo "PROTOBUF_RUNTIME_VERSION must be set to run downstream-protobuf-source-compatibility.sh"
+  exit 1
+fi
+
+# Get the directory of the build script
+scriptDir=$(realpath "$(dirname "${BASH_SOURCE[0]}")")
+cd "${scriptDir}/../.." # cd to the root of this repo
+# common.sh lives in the presubmit folder
+source "$scriptDir/../presubmit/common.sh"
+setup_maven_mirror
```
blakeli0, line 35: Is this still needed now that we are overriding the property, not building sdk-platform-java anymore?

.github/workflows/downstream_protobuf_compatibility_check_nightly.yaml
```diff
@@ -0,0 +1,48 @@
+on:
+  pull_request:
+    # Updates on Release-Please PRs
+    branches:
+      - 'release-please--branches--main'
+  workflow_dispatch:
+    inputs:
+      protobuf_versions:
```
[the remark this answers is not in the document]
lqiu96: (As of now) I don't plan to re-use this job for Protoc testing. It'll most likely be a separate job that is manually invoked to test for future Protoc RC and future versions.

I'll rename this param to reference the protobuf runtime versions

.github/workflows/downstream_protobuf_compatibility_check_nightly.yaml
lqiu96: Oh, I see. Yeah, I think you're right. It's for the branch that it's targeting and not where it is raised from.

I think there are other (not built in) solutions for this.

.github/workflows/downstream_protobuf_compatibility_check_nightly.yaml
lqiu96: I took a quick look and don't think they're all dependent on protobuf-java. java-storage-nio seems to be the odd one out that doesn't have any protobuf-java references at all and a few others have some references only in tests. 

> adding all handwritten libraries here in case they start depending on protobuf-java in the future

I think it'll be safer to keep all of the downstream libraries in here. Just to be safe in case there are references of Protobuf-Java added in source for downstream libraries.

.kokoro/nightly/downstream-protobuf-source-compatibility.sh
lqiu96: Sure, I can add an example

.kokoro/nightly/downstream-protobuf-source-compatibility.sh
lqiu96: I don't think we need to call `setup_maven_mirror`, but I kept it here for consistency with the other jobs that need to pull deps from Maven Central

I think `setup_maven_mirror` can be moved to a common.sh file that doesn't solely exist in the presubmit folder, but I didn't want to make that part of this PR.

.github/workflows/downstream_protobuf_compatibility_check_nightly.yaml
lqiu96: Added a conditional that checks the head_ref (source branch) and the github.event type (schedule) so that it works for release please PRs and the nightly flow

.kokoro/nightly/downstream-protobuf-source-compatibility.sh
blakeli0: What is the difference if we don't call `setup_maven_mirror`?

lqiu96: I don't recall the original context of using the maven mirror, but I believe it was to help speed up the CI jobs by reducing the time it takes to download the JARs. I just did a search and it doesn't look like it's being used anymore (maybe this was removed recently?).

I'll remove it since it doesn't seem to be used anymore.

blakeli0 (approved):

PR #3406, opened 2024-12-03
chore: merge target branch before commit changes

JoeWang1127:
In this PR:
- Merge base branch to current branch before updating generation config.

Example run: https://github.com/googleapis/sdk-platform-java/actions/runs/12148752357/job/33877902009

### DIFF
--- a/.github/scripts/hermetic_library_generation.sh
+++ b/.github/scripts/hermetic_library_generation.sh
@@ -75,16 +75,8 @@ workspace_name="/workspace"
 baseline_generation_config="baseline_generation_config.yaml"
 message="chore: generate libraries at $(date)"
 
-# there's no local branch when running in GitHub action,
-# thus no need to use `git pull`.
 git checkout "${target_branch}"
 git checkout "${current_branch}"
-# only allow fast-forward merging; exit with non-zero result if there's merging
-# conflict.
-if git merge -m "chore: merge ${target_branch} into ${current_branch}" "${target_branch}" -ne 0; then
-  echo "Merge ${target_branch} into ${current_branch} has conflict, exit."
-  exit 1
-fi
 
 # copy generation configuration from target branch to current branch.
 git show "${target_branch}":"${generation_config}" > "${baseline_generation_config}"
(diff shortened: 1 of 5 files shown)


.github/scripts/hermetic_library_generation.sh
```diff
@@ -75,8 +75,16 @@ workspace_name="/workspace"
 baseline_generation_config="baseline_generation_config.yaml"
 message="chore: generate libraries at $(date)"
 
+# there's no local branch when running in GitHub action,
+# thus no need to use `git pull`.
 git checkout "${target_branch}"
 git checkout "${current_branch}"
+# only allow fast-forward merging; exit with non-zero result if there's merging
+# conflict.
+if [[ $(git merge -m "chore: merge ${target_branch} into ${current_branch}" "${target_branch}") -ne 0 ]]; then
```
blakeli0, line 84: Another way is to always update the branch to main before updating in https://github.com/googleapis/sdk-platform-java/blob/main/.github/scripts/update_googleapis_commit.sh, so that we don't have to deal with potential conflict.

JoeWang1127: Hmm, maybe change in `update_googleapis_commit.sh` is better because we only want to merge main in generation PR and other pull requests need manual merge anyway.

.github/scripts/update_googleapis_commit.sh
```diff
@@ -79,9 +84,16 @@ changed_files=$(git diff --cached --name-only)
 if [[ "${changed_files}" == "" ]]; then
     echo "The latest googleapis commit is not changed."
     echo "Skip committing to the pull request."
+else
+    git commit -m "${title}"
+fi
+
+unpushed_commit=$(git cherry -v "origin/${current_branch}" | wc -l)
```
diegomarquezp, line 91: What is this unpushed commit check meant to ensure? I'm not familiar enough with the cherry-pick subcommands

JoeWang1127: After this change, we may have up to two commits: merge commit and change commit.

We want to exit the script if no commit happens (otherwise this will be a infinite loop). `git cherry` is a way to find whether the local branch has commits that are not in the remote branch. If we find any such commit, push them to remote branch.

diegomarquezp: Makes sense, thanks. Maybe a small comment in this line may help others as well on how this ensures the remote is synced?

diegomarquezp (approved):
LGTM, just a small comment

PR #3409, opened 2024-12-04
chore: update base image to include tag

JoeWang1127:
In this PR:
- Include tag in base image.

Renovate Bot can update the image sha with a fixed tag ([doc](https://docs.renovatebot.com/docker/#digest-updating)). Without the tag, the updated image maybe not what we want.


diegomarquezp (approved):

diegomarquezp:
Nice, I wasn't aware we can both specify the hash and the tag. Thanks!

PR #3438, opened 2024-12-06
fix: graalvm missing build time class when using protobuf 4.x

ldetmer:
In case of protobuf 4+ need to include protobuf.Runtime in classes initialized at build time as that is used to determine protobuf version for graalvm.

Tested, temporarily bumped protobuf dependency to 4.28.1 without including Runtime in build time class list.  Received error:

```
Error: Classes that should be initialized at run time got initialized during image building:
 com.google.protobuf.RuntimeVersion was unintentionally initialized at build time. To see why com.google.protobuf.RuntimeVersion got initialized use --trace-class-initialization=com.google.protobuf.RuntimeVersion
```

Updated to include Runtime and received passing [native showcase tests](https://github.com/googleapis/sdk-platform-java/actions/runs/12205342463/job/34052473211?pr=3438)

### DIFF
--- a/gax-java/gax/src/main/resources/META-INF/native-image/com.google.api/gax/native-image.properties
+++ b/gax-java/gax/src/main/resources/META-INF/native-image/com.google.api/gax/native-image.properties
@@ -3,7 +3,7 @@ Args = --enable-url-protocols=https,http \
   org.junit.platform.engine.TestTag,\
   com.google.api.gax.core.GaxProperties,\
   com.google.common.base.Platform,\
-  com.google.common.base.Platform$JdkPatternCompiler, \
+  com.google.common.base.Platform$JdkPatternCompiler,\
   com.google.protobuf.RuntimeVersion \
   --features=com.google.api.gax.nativeimage.OpenCensusFeature,\
   com.google.api.gax.nativeimage.GoogleJsonClientFeature \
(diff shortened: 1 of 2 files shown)


JoeWang1127 (approved):

gax-java/gax/src/main/resources/META-INF/native-image/com.google.api/gax/native-image.properties
```diff
@@ -3,7 +3,8 @@ Args = --enable-url-protocols=https,http \
   org.junit.platform.engine.TestTag,\
   com.google.api.gax.core.GaxProperties,\
   com.google.common.base.Platform,\
-  com.google.common.base.Platform$JdkPatternCompiler \
+  com.google.common.base.Platform$JdkPatternCompiler,\
+  com.google.protobuf.RuntimeVersion \
```
blakeli0, line 7: Since this is a class only exist in protobuf 4.x, if a customer overrides protobuf to 3.x, I'm assuming it is going to fail for them?
If that's the case, we need to have a different solution, because our SDK is now guaranteed to be compatible with both protobuf 3.x and 4.x.

blakeli0: Answered offline that it is not an issue because the classes are not going to be loaded if they don't exist.

PR #3467, opened 2024-12-14
feat: add support for new setAllowHardBoundTokens field.

rmehta19:
Introduce new `setAllowHardBoundTokens` field.


rmehta19:
@lqiu96 , @blakeli0 , @zhumin8, please review, thanks!

rmehta19:
cc: @rockspore

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -126,6 +127,7 @@ public final class InstantiatingGrpcChannelProvider implements TransportChannelP
   @Nullable private final Boolean allowNonDefaultServiceAccount;
   @VisibleForTesting final ImmutableMap<String, ?> directPathServiceConfig;
   @Nullable private final MtlsProvider mtlsProvider;
+  @Nullable private final ArrayList<String> allowedValues;
```
rockspore, line 130: Should this variable name be more specific? Reading it alone doesn't tell you where these values will be used.

rockspore (commented):
Thanks for the PR, Riya! Not up to me but LGTM.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rmehta19: Thanks for calling this out! I updated this variable name to `allowedHardBoundTokenTypes`, WDYT?

rockspore: LGTM. Thanks.

rmehta19:
Friendly ping @lqiu96 , @blakeli0 , @zhumin8. Thanks!

blakeli0:
What is the usage of this new `allowHardBoundTokens` field? Do you mind providing more background info?

gax-java/gax/src/main/java/com/google/api/gax/rpc/TransportChannelProvider.java
```diff
@@ -97,6 +98,9 @@ public interface TransportChannelProvider {
    */
   TransportChannelProvider withEndpoint(String endpoint);
 
+  /** Sets the allowed hard bound token types. */
+  TransportChannelProvider setAllowHardBoundTokens(ArrayList<String> allowedValues);
```
blakeli0, line 102: I think this field is going to be only used by gRPC transport per the internal doc? And the teams that would use it are all handwritten libraries? If that's the case, I think they are likely going to initialize an `InstantiatingGrpcChannelProvider` in their repo directly and pass it to Gax, so can we only add this to `InstantiatingGrpcChannelProvider`?

rmehta19:
> What is the usage of this new `allowHardBoundTokens` field? Do you mind providing more background info?

I believe we discussed this in the chat offline. To summarize it'll be used by clients (libraries) to indicate if they want MTLS / Directpath bound tokens. Please let me know if there is anything we can make more clear.

gax-java/gax/src/main/java/com/google/api/gax/rpc/TransportChannelProvider.java
rmehta19: > I think this field is going to be only used by gRPC transport per the internal doc?

That is correct, hard bound tokens is only for gRPC.

> I think they are likely going to initialize an `InstantiatingGrpcChannelProvider` in their repo directly and pass it to Gax, so can we only add this to `InstantiatingGrpcChannelProvider`?

I think this will work since I believe the libraries separate their logic for HTTP and gRPC, so they don't end up using the interface `TransportChannelProvider`. Looking at GCS for example, [this is the case](https://github.com/googleapis/java-storage/blob/main/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java#L290)

@rockspore can you confirm that the places you want to use this setting are using the `InstantiatingGrpcChannelProvider` directly, not the interface `TransportChannelProvider`?

rockspore: For Cloud Spanner, I found [this](https://github.com/googleapis/java-spanner/blob/0777f9951e514a270aa44553a954415844b4fc0a/google-cloud-spanner/src/main/java/com/google/cloud/spanner/v1/SpannerSettings.java#L225) so it seems to be the case too, although I don't see where it gets called by default.

So yeah it should be good if we only do this in `InstantiatingGrpcChannelProvider`, as long as we have a way to set these default values eventually to all auto-generated clients.

rockspore: Thanks for the details! It makes perfect sense.

rmehta19: Thanks all for the discussion! Done in https://github.com/googleapis/sdk-platform-java/pull/3467/commits/60dafd0d6539960eaf6b4444468ae289a7df1ec8

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -126,6 +127,7 @@ public final class InstantiatingGrpcChannelProvider implements TransportChannelP
   @Nullable private final Boolean allowNonDefaultServiceAccount;
   @VisibleForTesting final ImmutableMap<String, ?> directPathServiceConfig;
   @Nullable private final MtlsProvider mtlsProvider;
+  @Nullable private final ArrayList<String> allowedHardBoundTokenTypes;
```
blakeli0, line 130: Please use `List` instead of `ArrayList`.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -225,6 +228,10 @@ public TransportChannelProvider withEndpoint(String endpoint) {
     return toBuilder().setEndpoint(endpoint).build();
   }
 
+  public TransportChannelProvider setAllowHardBoundTokens(ArrayList<String> allowedValues) {
```
blakeli0, line 231: Since this method is not on the interface anymore, I don't think we have to expose a separate setter for it. Exposing it on the builder should be good enough.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -700,6 +708,11 @@ public Builder setEndpoint(String endpoint) {
       return this;
     }
 
+    public Builder setAllowHardBoundTokenTypes(ArrayList<String> allowedValues) {
```
blakeli0, line 711: Do we expect customers to set this? Or only internal teams like Spanner/Storage to set it? If it is the latter, can we mark this method as `@InternalApi`?

rockspore: Right we don't want external customers to set it.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rmehta19: Done in https://github.com/googleapis/sdk-platform-java/pull/3467/commits/e193cc9412ae68b01ca5d1a21eb38baf35d68a71

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rmehta19: We only expect internal teams to set this field, marked as `@InternalApi` in https://github.com/googleapis/sdk-platform-java/pull/3467/commits/e193cc9412ae68b01ca5d1a21eb38baf35d68a71

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -700,6 +703,12 @@ public Builder setEndpoint(String endpoint) {
       return this;
     }
 
+    @InternalApi
```
blakeli0, line 706: Can you please add some Javadocs (non-confidential) to this public method? So that other developers don't have to look up "what is hardBoundTokenTypes?".

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -700,6 +703,12 @@ public Builder setEndpoint(String endpoint) {
...
+    public Builder setAllowHardBoundTokenTypes(List<String> allowedValues) {
```
blakeli0, line 707: What are the possible values of `allowedValues`? I think there are only two possible values so far per internal doc, if that's the case, can they be Enums so it's less error prone?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rmehta19: Done in https://github.com/googleapis/sdk-platform-java/pull/3467/commits/0e40f22636269b7db06281fe228cfb50dad2c0d5

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rmehta19: Yes we can capture the allowed values in an enum. Done in https://github.com/googleapis/sdk-platform-java/pull/3467/commits/0e40f22636269b7db06281fe228cfb50dad2c0d5

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -126,16 +126,32 @@ public final class InstantiatingGrpcChannelProvider implements TransportChannelP
...
   private final ApiFunction<ManagedChannelBuilder, ManagedChannelBuilder> channelConfigurator;
 
+  /*
+   * Experimental feature
+   *
+   * <p>{@link HardBoundTokenTypes} specifies if hard bound tokens should be used if DirectPath
+   * or S2A is used to estabilsh a connection to Google APIs.
+   *
+   */
+  public enum HardBoundTokenTypes {
```
zhumin8, line 142: Does it make sense to also mark this as internal?

zhumin8 (commented):
Can you please also add some test coverage for this change?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
blakeli0: Agree on making this internal as well. 
Thinking twice about it though, I see that it is an `Experimental feature`, is it that we will always set the tokens to certain values? Or it's just this feature is not stable yet, internal teams could still set this to different values? If it's the former, then we don't have to introduce another public enum since they would be obsolete soon.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -700,6 +717,30 @@ public Builder setEndpoint(String endpoint) {
...
+     * <p>allowedValues is {HardBoundTokenTypes.ALTS}: If DirectPath is used to create the channel,
+     * use hard ALTS-bound tokens for requests sent on that channel.
+     *
+     * <p>allowedValues is {HardBoundTokenTypes.MTLS_S2A}: If MTLS via S2A is used to create the
+     * channel, use hard MTLS-bound tokens for requests sent on that channel.
+     *
+     * <p>allowedValues is {HardBoundTokenTypes.ALTS, HardBoundTokenTypes.MTLS_S2A}: if DirectPath
+     * is used to create the channel, use hard ALTS-bound tokens for requests sent on that channel.
+     * If MTLS via S2A is used to create the channel, use hard MTLS-bound tokens for requests sent
+     * on that channel.
```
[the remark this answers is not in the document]
blakeli0: Agree that maybe the detailed explanation can be moved to the subsequent PR. The public doc for this method should be more like "What is HardBoundTokenTypes?", not "What would happen if we set HardBoundTokenTypes to different values?". But if listing the different behavior is the best way to explain "What is HardBoundTokenTypes?", I think this is OK too.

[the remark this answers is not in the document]
rmehta19: Thanks for calling this out! Looking at this again agreed that the behavior of setting the field to different values should be paired with the logic. Updated the javadoc to specify what the field means. Also made the enum javadoc a bit more specific. 

https://github.com/googleapis/sdk-platform-java/pull/3467/commits/90a32af5e30e009fda5f84f20bfcd5c46ee59ef3

rmehta19:
> Can you please also add some test coverage for this change?

Added to `testToBuilder`. Since we haven't added a `InstantiatingGrpcChannelProvider.Builder.getAllowHardBoundTokenTypes()`, I just added the setter to the test.

https://github.com/googleapis/sdk-platform-java/pull/3467/commits/a7f8384ec83e4cd09ba5218aa116a38b4db1fabb

rockspore:
> > Can you please also add some test coverage for this change?
> 
> Added to `testToBuilder`. Since we haven't added a `InstantiatingGrpcChannelProvider.Builder.getAllowHardBoundTokenTypes()`, I just added the setter to the test.
> 
> [a7f8384](https://github.com/googleapis/sdk-platform-java/commit/a7f8384ec83e4cd09ba5218aa116a38b4db1fabb)

JFYI, this will be more meaningfully tested in #3572 (not ready for review yet) and later on when MTLS logic is added.

blakeli0:
> > Can you please also add some test coverage for this change?
> 
> Added to `testToBuilder`. Since we haven't added a `InstantiatingGrpcChannelProvider.Builder.getAllowHardBoundTokenTypes()`, I just added the setter to the test.
> 
> [a7f8384](https://github.com/googleapis/sdk-platform-java/commit/a7f8384ec83e4cd09ba5218aa116a38b4db1fabb)

Since we don't expose a getter and we don't need one, I think this is good enough for now. I guess we are going to test in the subsequent PR that the channel would be created differently when passing different values?

rmehta19:
@blakeli0, One thing: when we are setting this field from the gapic generator, is it ok to have this enum? IIUC the gapic generator would then have to depend on gax-grpc in order to be able to set the field to {MTLS_S2A, ALTS} for example. Is this ok? If not, should we switch this back to a string?

rmehta19:
> > > Can you please also add some test coverage for this change?
> > 
> > 
> > Added to `testToBuilder`. Since we haven't added a `InstantiatingGrpcChannelProvider.Builder.getAllowHardBoundTokenTypes()`, I just added the setter to the test.
> > [a7f8384](https://github.com/googleapis/sdk-platform-java/commit/a7f8384ec83e4cd09ba5218aa116a38b4db1fabb)
> 
> Since we don't expose a getter and we don't need one, I think this is good enough for now. I guess we are going to test in the subsequent PR that the channel would be created differently when passing different values?

That's correct. @rockspore 's #3572 for example will likely introduce some tests that make sure that the correct credential (one that can fetch hard ALTS bound tokens) is created when this list includes ALTS and when DirectPath gets picked.

blakeli0:
> @blakeli0, One thing: when we are setting this field from the gapic generator, is it ok to have this enum? IIUC the gapic generator would then have to depend on gax-grpc in order to be able to set the field to {MTLS_S2A, ALTS} for example. Is this ok? If not, should we switch this back to a string?
> 
> cc: @lqiu96, @zhumin8 since I recall we discussed these changes to gapic generator in the internal doc.

Yes it does. Using java-language as an example, the generated code would likely be added to [this method](https://github.com/googleapis/google-cloud-java/blob/main/java-language/google-cloud-language/src/main/java/com/google/cloud/language/v1/stub/LanguageServiceStubSettings.java#L221-L224).

rmehta19:
> > @blakeli0, One thing: when we are setting this field from the gapic generator, is it ok to have this enum? IIUC the gapic generator would then have to depend on gax-grpc in order to be able to set the field to {MTLS_S2A, ALTS} for example. Is this ok? If not, should we switch this back to a string?
> > cc: @lqiu96, @zhumin8 since I recall we discussed these changes to gapic generator in the internal doc.
> 
> Yes it does. Using java-language as an example, the generated code would likely be added to [this method](https://github.com/googleapis/google-cloud-java/blob/main/java-language/google-cloud-language/src/main/java/com/google/cloud/language/v1/stub/LanguageServiceStubSettings.java#L221-L224).

Thanks for the code pointer!

blakeli0 (approved):
LGTM. Please change the PR title to follow the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).

zhumin8 (approved):
LGTM

rmehta19:
> LGTM. Please change the PR title to follow the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).

Done.

rmehta19:
@zhumin8 / @blakeli0 would you be able to merge (as I don't have write access)? Thanks!

Issue #3512, opened 2025-01-03
Nightly builds are failing due to copyright header update


mpeddada1:
The nightly builds are currently failing with the following error: 
```
Error:  Failures: 
Error:    ComposerTest.gapicClass_addApacheLicense_validInput_succeeds:81 Differences found: 
--- golden
+++ codegen
@@ -1,4 +1,4 @@
 /*
- * Copyright 2024 Google LLC
+ * Copyright 2025 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
```
Example: https://github.com/googleapis/sdk-platform-java/issues/3486

Nightly Failures: https://github.com/googleapis/sdk-platform-java/issues/3511

mpeddada1:
Fix is in progress at https://github.com/googleapis/sdk-platform-java/pull/3513

PR #3513, opened 2025-01-03
tests: Update the golden files' copyright to 2025

blakeli0:
Fixes https://github.com/googleapis/sdk-platform-java/issues/3512.

The copyright year for GAPIC generated files are automatically [set to current year](https://github.com/googleapis/sdk-platform-java/blob/9a65386dfc3146de4f9b36962caae2df0662ac98/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/comment/CommentComposer.java#L38). This creates a problem that all the tests are going to automatically fail when new year comes. Updating all the golden files' copyright to 2025 in this PR for now. 

There are golden test files need to be updated in spring-cloud-gcp as well since it is dependent on the generator.

In the future, we need to come up with a better strategy that the tests do not fail automatically, created an issue https://github.com/googleapis/sdk-platform-java/issues/3547 to track the future enhancement.

### DIFF
--- a/gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/ComposerTest.java
+++ b/gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/ComposerTest.java
@@ -61,6 +61,7 @@ class ComposerTest {
     assertTrue(context.containsServices());
   }
 
+  @Disabled
   @Test
   public void gapicClass_addApacheLicense_validInput_succeeds() {
     ClassDefinition classDef =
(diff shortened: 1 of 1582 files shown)


gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/ComposerTest.java
```diff
@@ -61,6 +62,7 @@ void initialSanityCheck() {
     assertTrue(context.containsServices());
   }
 
+  @Disabled
```
zhumin8, line 65: IIUC, this is a manual fix and this `@Disabled` is added temporarily and will be removed after this update is merged?

blakeli0: No, I was testing if adding `@Disabled` to this test is going to fix most of the issues or not. It proves not, we need to update integration tests, showcase tests, hermetic build tests and spring-cloud-gcp tests.  I'm going to update all the golden files in this repo for now to unblock the release, and create issues for improvements.

zhumin8 (approved):

lqiu96 (approved):

PR #3401, opened 2024-11-27
s2a fix: fix NPE.

rmehta19:

PR #3400, opened 2024-11-27
Revert "feat: Add experimental S2A integration in client libraries grpc transport"

lqiu96:
Reverts googleapis/sdk-platform-java#3326

Context: https://github.com/googleapis/sdk-platform-java/pull/3326#issuecomment-2494603518

s2a-gRPC is an Experimental library and shouldn't be added until it's stabilized.

PR #3326, opened 2024-10-29
feat: Add experimental S2A integration in client libraries grpc transport

rmehta19:
Modify the Client Libraries gRPC Channel builder to use mTLS via S2A if the experimental environment variable is set, S2A is available (We check this by using utility added in https://github.com/googleapis/google-auth-library-java/pull/1400), and a few more conditions (see `shouldUseS2A`).

Following https://google.aip.dev/auth/4115, Only attempt to use S2A after DirectPath and DCA (https://google.aip.dev/auth/4114) are ruled out as options. If conditions to use S2A are not met (env variable not set, or S2A is not running in environment, etc (`shouldUseS2A` returns false)), fall back to default TLS connection.

When we are creating S2A-enabled Grpc Channel Credentials, we first try to secure the connection between the client and the S2A via MTLS, using [MTLS-MDS](https://cloud.google.com/compute/docs/metadata/overview#https-mds) credentials. If MTLS-MDS credentials can't be loaded, then we fallback to a plaintext connection between the client and S2A. 

The parallel go implementation : googleapis/google-api-go-client#1874 (now lives here: https://github.com/googleapis/google-cloud-go/blob/main/auth/internal/transport/cba.go)

S2A Java client: https://github.com/grpc/grpc-java/tree/master/s2a

Resolving b/376258193 means that S2A.java is no longer experimental

PR #3548, opened 2025-01-07
feat: revert #3400: reintroduce experimental S2A integration in client libraries grpc transport

rmehta19:
**Revert #3400.**

**This PR re-introduces the S2A integration the Java Cloud SDK (initially introduced in #3326, and temporarily reverted in #3400).**

**This PR does this by reverting #3400 with the following patches:**
- load the S2A APIs via reflection. This allows us to merge the code while the [S2A API is still experimental in gRPC-Java](https://github.com/grpc/grpc-java/blob/master/s2a/src/main/java/io/grpc/s2a/S2AChannelCredentials.java) without introducing a diamond dependency conflict. Once the S2A APIs are stable, the reflection logic can be removed and the S2A API can be used directly (via a dependency on S2A API)
- fix NPE (#3401)
- use a different env var name for enabling the feature

**Below is the original description from #3326**

Modify the Client Libraries gRPC Channel builder to use mTLS via S2A if the experimental environment variable is set, S2A is available (We check this by using [SecureSessionAgent utility](https://github.com/googleapis/google-auth-library-java/blob/main/oauth2_http/java/com/google/auth/oauth2/SecureSessionAgent.java)), and a few more conditions (see `shouldUseS2A`).

Following https://google.aip.dev/auth/4115, Only attempt to use S2A after DirectPath and DCA (https://google.aip.dev/auth/4114) are ruled out as options. If conditions to use S2A are not met (env variable not set, or S2A is not running in environment, etc (`shouldUseS2A` returns false)), fall back to default TLS connection.

When we are creating S2A-enabled Grpc Channel Credentials, we first try to secure the connection between the client and the S2A via MTLS, using [MTLS-MDS](https://cloud.google.com/compute/docs/metadata/overview#https-mds) credentials. If MTLS-MDS credentials can't be loaded, then we fallback to a plaintext connection between the client and S2A. 

The parallel go implementation : googleapis/google-api-go-client#1874 (now lives here: https://github.com/googleapis/google-cloud-go/blob/main/auth/internal/transport/cba.go)

S2A Java client: https://github.com/grpc/grpc-java/tree/master/s2a

Resolving b/376258193 means that S2A.java is no longer experimental


gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/samples/bookshopclient/AsyncGetBook.golden
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 Google LLC
```
rmehta19, line 2: Looks like running `mvn fmt:format` resulted in all these changes. Perhaps these changes could be made in another PR and then can be removed from this PR.

rmehta19: Thanks! with the latest merge of main, these changes have been removed from this PR.

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -288,6 +306,37 @@ private String determineEndpoint() throws IOException {
...
+    @VisibleForTesting
+    boolean shouldUseS2A() {
+      // If mTLS endpoint is not available, skip S2A
+      if (Strings.isNullOrEmpty(mtlsEndpoint())) {
+        return false;
+      }
+
+      // If EXPERIMENTAL_GOOGLE_API_USE_S2A_JAVA is not set to true, skip S2A.
+      String s2AEnv;
+      s2AEnv = envProvider().getenv(S2A_ENV_ENABLE_USE_S2A);
```
xmenxk, line 319: nit: this can be one line.

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -288,6 +306,37 @@ private String determineEndpoint() throws IOException {
       return endpoint;
     }
 
+    /** Determine if S2A can be used */
+    @VisibleForTesting
+    boolean shouldUseS2A() {
```
xmenxk, line 311: There was a issue raised re: netty-tcnative dropping support for Windows and MacOS Intel platforms, do we want to add the runtime checks here an skip S2A if the code is running on an unsupported platforms?

We can also do that in a followup PR, but just to mention it here so we don't forget.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -449,13 +628,28 @@ private ManagedChannel createSingleChannel() throws IOException {
...
       } else {
+        // Could not create channel credentials via DCA. In accordance with
+        // https://google.aip.dev/auth/4115, if credentials not available through
+        // DCA, try mTLS with credentials held by the S2A (Secure Session Agent).
+        if (useS2A) {
+          channelCredentials = createS2ASecuredChannelCredentials();
+        }
+        if (channelCredentials != null) {
+          // Create the channel using S2A-secured channel credentials.
+          builder = Grpc.newChannelBuilder(mtlsEndpoint, channelCredentials);
```
xmenxk, line 648: I assume this is what we discussed re: only pick the mtls endpoint when we know s2a will be used and directpath will not be used, but let know if otherwise. thanks

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
rmehta19: Thanks for flagging this Kui, I think this is fine to get to in a followup PR. Sent you a ping with some details.

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
rmehta19: Done.

gax-java/gax/src/main/java/com/google/api/gax/rpc/TransportChannelProvider.java
```diff
@@ -90,13 +90,28 @@ public interface TransportChannelProvider {
...
    * <p>This method should only be called if {@link #needsEndpoint()} returns true.
    */
   TransportChannelProvider withEndpoint(String endpoint);
 
+  /**
+   * Sets the mtlsEndpoint to use when constructing a new {@link TransportChannel}.
+   *
+   * <p>This method should only be called if {@link #needsMtlsEndpoint()} returns true.
+   */
+  TransportChannelProvider withMtlsEndpoint(String mtlsEndpoint);
```
lqiu96, line 108: Let me think about this a bit more. I'm not the biggest fan of adding these two public methods in `needsMtlsEndpoint` + `withMtlsEndpoint` and would prefer not to if possible.

I know this is a limitation regarding DirectPath and how that's determined. If we can't find a reasonable alternative, then I think I'm fine with this.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -410,6 +461,134 @@ ChannelCredentials createMtlsChannelCredentials() throws IOException, GeneralSec
...
+      // Try to connect to S2A using mTLS.
+      ChannelCredentials mtlsToS2AChannelCredentials = null;
+      try {
+        mtlsToS2AChannelCredentials =
+            createMtlsToS2AChannelCredentials(rootFile, certKeyFile, certKeyFile);
+      } catch (IOException ignore) {
+        // Fallback to plaintext-to-S2A connection on error.
+        LOG.log(
+            Level.WARNING,
+            "Cannot establish an mTLS connection to S2A due to error creating MTLS to MDS TlsChannelCredentials credentials, falling back to plaintext connection to S2A: "
```
lqiu96, line 578: Why is this WARNING log level when the others are INFO?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -449,13 +628,28 @@ private ManagedChannel createSingleChannel() throws IOException {
...
+          channelCredentials = createS2ASecuredChannelCredentials();
+        }
+        if (channelCredentials != null) {
+          // Create the channel using S2A-secured channel credentials.
+          builder = Grpc.newChannelBuilder(mtlsEndpoint, channelCredentials);
+        } else {
+          // Use default if we cannot initialize channel credentials via DCA or S2A.
+          builder = ManagedChannelBuilder.forAddress(serviceAddress, port);
+        }
         builder = ManagedChannelBuilder.forAddress(serviceAddress, port);
```
lqiu96, line 653: I think we want to remove this last `builder = ManagedChannelBuilder.forAddress(serviceAddress, port);`  at L653. It looks like it would overwrite everything before it.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -410,6 +461,134 @@ ChannelCredentials createMtlsChannelCredentials() throws IOException, GeneralSec
...
+      Method newBuilder = s2aChannelCreds.getMethod("newBuilder", partypes);
+      Object arglist[] = new Object[2];
+      arglist[0] = s2aAddress;
+      arglist[1] = s2aChannelCredentials;
+      Object retObjBuilder = newBuilder.invoke(null, arglist);
+      Method build = s2aChannelCredsBuilder.getMethod("build", null);
+      Object retObjCreds = build.invoke(retObjBuilder, null);
+      return (ChannelCredentials) retObjCreds;
+    } catch (Throwable t) {
+      LOG.log(Level.INFO, "S2A APIs cannot be used: ", t.getMessage());
```
lqiu96, line 493: I think this should be an WARNING level message. If this is unable to be created, it would fallback to the normal flow? If so, can that also be added to the log statement?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rmehta19: IIRC, we made this a WARNING log level because it indicates something is wrong with using MTLS-MDS feature (this feature provides the credentials we use to do mTLS to S2A), not our feature. If we reach this log statement, we were able to find the MTLS-MDS credential files, however we failed to use them to build [TlsChannelCredentials](https://github.com/grpc/grpc-java/blob/master/api/src/main/java/io/grpc/TlsChannelCredentials.java)

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rmehta19: Thanks for catching this (result of copy paste typo)!! Removed in https://github.com/googleapis/sdk-platform-java/pull/3548/commits/95e70d321b713a26812c07d7377b07fa4c22c2cb.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rmehta19: > If this is unable to be created, it would fallback to the normal flow?
Yes, it would fallback to normal flow (TLS without S2A).

> I think this should be an WARNING level message...If so, can that also be added to the log statement?
Done.

https://github.com/googleapis/sdk-platform-java/pull/3548/commits/b80729413fd6002220874159159d05704e499dac

gax-java/gax/src/main/java/com/google/api/gax/rpc/TransportChannelProvider.java
lqiu96: Took a look at this and I think there are possible a few alternatives to what is in this PR. Not sure entirely what all the pros and cons are as of now, but just going to pose them as possibilities.

1. Remove the `.mtls` substring when using directpath flow. This is probably the simplest option, but is pretty much a workaround. Given that S2A may end up being used much more, I don't think I would like endpoint resolution to be based on string find and replace.
2. Create a default method `canUseDirectPath()` in the TransportChannelProvider interface
```
  default boolean canUseDirectPath() {
    return false;
  }
```
I believe this allows us to use it via the ClientContext to resolve the endpoint before we create the TransportChannel:
i.e.
```
    if (transportChannelProvider.needsEndpoint()) {
      if (transportChannelProvider.canUseDirectPath()) {
        transportChannelProvider =
            transportChannelProvider.withEndpoint(endpointContext.mtlsEndpoint());
      } else {
        transportChannelProvider = transportChannelProvider.withEndpoint(endpoint);
      }
    }
```

It shouldn't affect users who manually create `InstantiatingGrpcChannelProvider` since the logic still resides in there and is used during channel creation.

Let me talk with the team and see if there are any other potential concerns that I'm missing with this. I think option 2 might be a possibility and if not, then I think we can proceed with this.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
lqiu96: Hmm, actually I'm more on the fence about this now (and it's possible that it may just be that I don't fully know where to draw the line between INFO and WARNING).

I think perhaps these should all be WARNING level? Given that to even get into this flow users would explicitly do a bunch of manual configurations, we should warn them if any one of their configurations end up failing (i.e. env's root file/ pem file aren't place properly or mtlsendpoint isn't set). It would be like a warning that someting went wrong and we are falling back to use plaintext.

Not entirely sure though. @blakeli0 @zhumin8 who know more about logging

gax-java/gax/src/main/java/com/google/api/gax/rpc/TransportChannelProvider.java
rmehta19: As discussed offline, we will not plumb the mtls endpoint for now, and just set the mtls endpoint if S2A can be used in `EndpointContext` in https://github.com/googleapis/sdk-platform-java/pull/3548/commits/25445d3b803e6c522694a2f37313eca4ed2f2356

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -99,6 +104,15 @@ public final class InstantiatingGrpcChannelProvider implements TransportChannelP
   @VisibleForTesting
   static final String DIRECT_PATH_ENV_ENABLE_XDS = "GOOGLE_CLOUD_ENABLE_DIRECT_PATH_XDS";
 
+  // The public portion of the mTLS MDS root certificate is stored for performing
+  // cert verification when establishing an mTLS connection with the MDS. See
+  // https://cloud.google.com/compute/docs/metadata/overview#https-mds-root-certs
+  private static final String MTLS_MDS_ROOT_PATH = "/run/google-mds-mtls/root.crt";
+  // The mTLS MDS credentials are formatted as the concatenation of a PEM-encoded certificate chain
+  // followed by a PEM-encoded private key. See
+  // https://cloud.google.com/compute/docs/metadata/overview#https-mds-client-certs
```
lqiu96, line 113: nit
```suggestion
  // followed by a PEM-encoded private key. See {@link <a href=
  // "https://cloud.google.com/compute/docs/metadata/overview#https-mds-client-certs">this</a> for more information.} 
```

gax-java/gax/clirr-ignored-differences.xml
```diff
@@ -106,16 +106,26 @@
...
   <difference>
-    <differenceType>7002</differenceType>
-    <className>com/google/api/gax/rpc/FixedTransportChannelProvider</className>
-    <method>* withUseS2A(*)</method>
+    <differenceType>7013</differenceType>
+    <className>com/google/api/gax/rpc/EndpointContext</className>
+    <method>* useS2A()</method>
   </difference>
-  <!-- Ignore this as this was part of s2a-grpc ExperimentalApi revert -->
+  <!-- Ignore method addition to an TransportChannelProvider interface -->
```
lqiu96, line 115:
```suggestion
  <!-- Ignore method addition to TransportChannelProvider interface (InternalExtensionOnly) -->
```

gax-java/gax/clirr-ignored-differences.xml
```diff
@@ -106,16 +106,26 @@
     <className>com/google/api/gax/batching/Batcher</className>
     <method>*</method>
   </difference>
-  <!-- Ignore this as this was part of s2a-grpc ExperimentalApi revert -->
+  <!-- Ignore abstract method addition to an EndpointContext -->
```
lqiu96, line 109:
```suggestion
  <!-- Ignore abstract method addition to an EndpointContext (InternalApi) -->
```

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -285,9 +303,43 @@ private String determineEndpoint() throws IOException {
             "mTLS is not supported in any universe other than googleapis.com");
       }
 
+      if (shouldUseS2A()) {
+        return mtlsEndpoint();
+      }
```
lqiu96, line 308: If S2A is supposed to be used, we don't need to go through resolving logic, right? Perhaps we can move this up to the top?

lqiu96: Actually, on second though... I think let's keep it here. Universe domain support with mtls may require this additional logic. Let's keep it here.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -99,6 +104,15 @@ public final class InstantiatingGrpcChannelProvider implements TransportChannelP
   @VisibleForTesting
   static final String DIRECT_PATH_ENV_ENABLE_XDS = "GOOGLE_CLOUD_ENABLE_DIRECT_PATH_XDS";
 
+  // The public portion of the mTLS MDS root certificate is stored for performing
+  // cert verification when establishing an mTLS connection with the MDS. See
+  // https://cloud.google.com/compute/docs/metadata/overview#https-mds-root-certs
```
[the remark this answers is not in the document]
rmehta19: Done in https://github.com/googleapis/sdk-platform-java/pull/3548/commits/e92169620711c14e96d9c51fe5c0d057740a5685

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rmehta19: Done in https://github.com/googleapis/sdk-platform-java/pull/3548/commits/e92169620711c14e96d9c51fe5c0d057740a5685

gax-java/gax/clirr-ignored-differences.xml
rmehta19: Done in https://github.com/googleapis/sdk-platform-java/pull/3548/commits/e92169620711c14e96d9c51fe5c0d057740a5685

gax-java/gax/clirr-ignored-differences.xml
rmehta19: Done in https://github.com/googleapis/sdk-platform-java/pull/3548/commits/e92169620711c14e96d9c51fe5c0d057740a5685

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
rmehta19: I think this is fine to do while the feature is experimental. Long term, I think we do need to be plumbing the mtls endpoint to the channel creation logic. I pinged with more details. Done in https://github.com/googleapis/sdk-platform-java/pull/3548/commits/9fb0b49fd6c2afffc9b19f9f5b9b852730c26a7e

gax-java/gax/src/main/java/com/google/api/gax/rpc/TransportChannelProvider.java
```diff
@@ -97,6 +97,11 @@ public interface TransportChannelProvider {
    */
   TransportChannelProvider withEndpoint(String endpoint);
 
+  /** Sets whether to use S2A when constructing a new {@link TransportChannel}. */
+  default TransportChannelProvider withUseS2A(boolean useS2A) {
+    throw new UnsupportedOperationException("S2A is not supported");
+  }
```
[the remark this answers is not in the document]
rmehta19: Done in https://github.com/googleapis/sdk-platform-java/pull/3548/commits/fd6163a493b733e131376cad0c9da7e14acf3de2

gax-java/gax/clirr-ignored-differences.xml
```diff
@@ -106,16 +106,26 @@
...
+   <difference>
+    <differenceType>7012</differenceType>
+    <className>com/google/api/gax/rpc/TransportChannelProvider</className>
+    <method>* needsMtlsEndpoint()</method>
+  </difference>
+  <difference>
+    <differenceType>7012</differenceType>
+    <className>com/google/api/gax/rpc/TransportChannelProvider</className>
+    <method>* withMtlsEndpoint(*)</method>
+  </difference>
```
lqiu96, line 130: I think these can potentially be added in a future date (if needed).

rmehta19: Ah yes, thanks for catching. Done in https://github.com/googleapis/sdk-platform-java/pull/3548/commits/a814a5371f845b20ee0d9a23bbd01266833dccb6

gax-java/gax/src/main/java/com/google/api/gax/rpc/TransportChannelProvider.java
```diff
@@ -97,6 +97,13 @@ public interface TransportChannelProvider {
    */
   TransportChannelProvider withEndpoint(String endpoint);
 
+  /** Sets whether to use S2A when constructing a new {@link TransportChannel}. */
+  @BetaApi(
+      "The S2A feature is not stable yet and may change in the future. https://github.com/grpc/grpc-java/issues/11533.")
+  default TransportChannelProvider withUseS2A(boolean useS2A) {
+    throw new UnsupportedOperationException("S2A is not supported");
```
blakeli0, line 104: I know we decided to throw exception last time we reviewed it, but we discovered that this could break downstream libraries' or customers' tests if they have a mock `TransportChannelProvider`. To prevent such breaking changes, can we change this to `return this;` instead? Then we don't have to override this method in other implementations (`LocalChannelProvider`, `InstantiatingHttpJsonChannelProvider` etc.) of this interface as well.

rmehta19: Thanks for bringing this up! Changed the default behavior to `return this` and removed overriding of the method in all other implementations except gRPC.

https://github.com/googleapis/sdk-platform-java/pull/3548/commits/1703f51775763840028a7d3dc843cf0ccefbe68a

blakeli0 (approved):

rmehta19:
@blakeli0 I see there are 2 failing checks. Looking at the logs, the failures seem unrelated to the changes in this PR, WDYT?

zhumin8:
sonar check is not supported for forked branches. And you can ignore "Build with Airlock" check for now.

rmehta19:
@zhumin8 does this PR look ok to you? If so, are you able to merge (as I don't have write access)? Thank you!

blakeli0:
@rmehta19 Please resolve the conflict in `InstantiatingGrpcChannelProvider`, which is probably caused by https://github.com/googleapis/sdk-platform-java/pull/3467. We'll merge it once it is resolved.

rmehta19:
> @rmehta19 Please resolve the conflict in `InstantiatingGrpcChannelProvider`, which is probably caused by #3467. We'll merge it once it is resolved.

Thanks @blakeli0, the conflict is resolved.

rmehta19:
I was looking through resolved comments and I found [this one](https://github.com/googleapis/sdk-platform-java/pull/3548#discussion_r1918923727), I think mine and @lqiu96 's comments might have crossed at the same time.. 

I think the resolution is that we should be putting the 
```
if (shouldUseS2A()) {
        return mtlsEndpoint();
      }
```
at the end of `determineEndpoint()`. Right now it's at the beginning. WDYT @lqiu96, @blakeli0?

lqiu96:
> I was looking through resolved comments and I found [this one](https://github.com/googleapis/sdk-platform-java/pull/3548#discussion_r1918923727), I think mine and @lqiu96 's comments might have crossed at the same time..
> 
> I think the resolution is that we should be putting the
> 
> ```
> if (shouldUseS2A()) {
>         return mtlsEndpoint();
>       }
> ```
> 
> at the **end** of `determineEndpoint()` as "universe domain support with mtls may require this additional logic". Right now it's at the beginning. WDYT @lqiu96, @blakeli0?
> 
> I tested in an internal patch, having this at the beginning or end of `determineEndpoint()` works for our GCP use cases, and while S2A is experimental, `shouldUseS2A()` will anyways return false, so it won't impact customers (on or off GCP).

Took another look at this and I'm curious about the behavior regarding S2A and having the resolved endpoint be the non-mtls endpoint. `mtlsResolver` simply determines to use the mtls endpoint or the non-mtls endpoint. Based on the logic, I think it's possible that the resolved endpoint ends up being non-mtls endpoint, but S2A is true.

Should we guard against this or return an error? I think in this case, we may need to put this towards the end.

rmehta19:
> > I was looking through resolved comments and I found [this one](https://github.com/googleapis/sdk-platform-java/pull/3548#discussion_r1918923727), I think mine and @lqiu96 's comments might have crossed at the same time..
> > I think the resolution is that we should be putting the
> > ```
> > if (shouldUseS2A()) {
> >         return mtlsEndpoint();
> >       }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > at the **end** of `determineEndpoint()` as "universe domain support with mtls may require this additional logic". Right now it's at the beginning. WDYT @lqiu96, @blakeli0?
> > I tested in an internal patch, having this at the beginning or end of `determineEndpoint()` works for our GCP use cases, and while S2A is experimental, `shouldUseS2A()` will anyways return false, so it won't impact customers (on or off GCP).
> 
> Took another look at this and I'm curious about the behavior regarding S2A and having the resolved endpoint be the non-mtls endpoint. `mtlsResolver` simply determines to use the mtls endpoint or the non-mtls endpoint. Based on the logic, I think it's possible that the resolved endpoint ends up being non-mtls endpoint, but S2A is true.
> 
> Should we guard against this or return an error? I think in this case, we may need to put this towards the end.

It would be an error to use S2A with a non-MTLS endpoint, because when we start using the MTLS_S2A hard bound tokens on the connection, the requests would get rejected. To be clear: if S2A is used, the resolved endpoint needs to be an MTLS endpoint. 

> Based on the logic, I think it's possible that the resolved endpoint ends up being non-mtls endpoint, but S2A is true.

I'm not sure I am following this. With this check in the beginning or end of `determineEndpoint`, if  S2A(`shouldUseS2A`) is true, the resolved endpoint will be a MTLS endpoint. There is an early return however for GDCH, so I do think it makes sense to move the check to the end of `determineEndpoint`, because if GDCH is being used, a custom endpoint gets returned.

In https://github.com/googleapis/sdk-platform-java/pull/3548/commits/6f56cffd6263a06b28d9ed8e9780966dd1ce7a50 I moved the check to the end of `determineEndpoint`.

lqiu96:
> if S2A(shouldUseS2A) is true, the resolved endpoint will be a MTLS endpoint.

Oh sorry, what I wrote was confusing. What I meant is that it is possible that the `mtlsResolver` may return a non-mtls endpoint while `shouldUseS2A` is true (enabled). It would be confusing since `mtlsResolver` ends up checking the things that are required to enable MTLS (i.e. the required env vars and keystore, etc). So user may have set up for S2A, but the environment is accidentally configured to non-mtls. In this case, it would be that the environment can use non-mtls, but the user is set up for S2A (trying to mTLS), which seems to be a whole set of configuration issues/ mismatch. In this case, I think we would want to error out?

Hopefully that is clearer about what I was trying to point out.

rmehta19:
> > if S2A(shouldUseS2A) is true, the resolved endpoint will be a MTLS endpoint.
> 
> Oh sorry, what I wrote was confusing. What I meant is that it is possible that the `mtlsResolver` may return a non-mtls endpoint while `shouldUseS2A` is true (enabled). It would be confusing since `mtlsResolver` ends up checking the things that are required to enable MTLS (i.e. the required env vars and keystore, etc). So user may have set up for S2A, but the environment is accidentally configured to non-mtls. In this case, it would be that the environment can use non-mtls, but the user is set up for S2A (trying to mTLS), which seems to be a whole set of configuration issues/ mismatch. In this case, I think we would want to error out?
> 
> Hopefully that is clearer about what I was trying to point out.

Thanks for elaborating @lqiu96! I responded with detail on chat internally.

lqiu96:
Thanks for the explanation. The changes LGTM!

Issue #2470, opened 2024-02-13
showcase samples should be part of golden tests


zhumin8:
sample snippets [here](https://github.com/googleapis/sdk-platform-java/tree/main/showcase/gapic-showcase/samples) should be part of the golden testing for Showcase.

Steps to identify this:
```
rm -rf showcase/gapic-showcase/samples/
mvn clean install -DskipTests
cd showcase
```
then `mvn compile -P update`, expect the samples should be re-generated and added back. 
But this samples folder is not re-generated

JoeWang1127:
Probably because we didn't copy sample directory in https://github.com/googleapis/sdk-platform-java/blob/main/showcase/scripts/generate_components.sh

blakeli0:
This was done in the early phase of hermetic build when we didn't have the image. We should migrate the showcase module to use hermetic build image now, with a dedicated generation_config.yaml.

diegomarquezp:
https://github.com/googleapis/sdk-platform-java/pull/3568 uses the docker image and generates samples

PR #3568, opened 2025-01-14
feat: generate showcase using docker image

diegomarquezp:
Fixes https://github.com/googleapis/sdk-platform-java/issues/2470

#### In `showcase/`
 * Revamp the generation and verification scripts used by `-Penable-golden-tests` and `-Pupdate` to build and use the Hermetic Build Docker image

### Effects
 * New `samples/` folder added in `showcase`
 * Reformatted all of grpc, proto and gapic libraries in `showcase/`


.github/scripts/hermetic_library_generation.sh
```diff
@@ -95,6 +95,14 @@ changed_libraries=$(python hermetic_build/common/cli/get_changed_libraries.py cr
   --current-generation-config-path="${generation_config}")
 echo "Changed libraries are: ${changed_libraries:-"No changed library"}."
 
+# do not generate showcase automatically
```
blakeli0, line 98: We probably do want to generate showcase automatically? For example, if we update protoc version in the docker image, we do want the showcase module to be updated.

generation_config.yaml
```diff
@@ -1,8 +1,18 @@
 googleapis_commitish: 3cef095370b3176f67542e527ffde81e019df091
-# the libraries are ordered with respect to library name, which is
-# java-{library.library_name} or java-{library.api-shortname} when
-# library.library_name is not defined.
+# the libraries are ordered with respect to library name, which defaults to
+# {library.folder_name} or java-{library.library_name} or
+# java-{library.api-shortname} in sequence (i.e. if first is None, then try second and so on).
 libraries:
+- api_shortname: showcase
+  folder_name: showcase # prevents java-showcase
```
blakeli0, line 7: I feel it is probably too much effort to introduce a new argument `folder_name`, just to prevent `showcase` to be renamed to `java-showcase`. I think it is OK to rename it.

generation_config.yaml
```diff
@@ -14,18 +24,18 @@ libraries:
...
-  - proto_path: google/cloud/audit
-  - proto_path: google/cloud/location
-  - proto_path: google/geo/type
-  - proto_path: google/logging/type
-  - proto_path: google/longrunning
-  - proto_path: google/rpc
-  - proto_path: google/rpc/context
-  - proto_path: google/shopping/type
-  - proto_path: google/type
+    - proto_path: google/api
```
blakeli0, line 27: Is this change intended?

showcase/scripts/resources/BUILD.partial.bazel
```diff
@@ -0,0 +1,38 @@
...
+java_gapic_library(
+    name = "showcase_java_gapic",
+    srcs = [":showcase_proto_with_info"],
+    gapic_yaml = None,
+    grpc_service_config = "showcase_grpc_service_config.json",
+    # TODO(#1285): Enable rest_numeric_enums once https://github.com/googleapis/gapic-showcase/issues/1255 is
+    # fixed.
+    rest_numeric_enums = False,
+    service_yaml = "showcase_v1beta1.yaml",
+    test_deps = [
```
blakeli0, line 16: Can we make this file minimum? For example, I don't think we need any `test_deps` since we only parses `rest_numeric_enums`, `transport` and `include_samples`.

versions.txt
```diff
@@ -17,3 +17,6 @@ proto-google-iam-v2:1.45.1:1.45.1
 grpc-google-iam-v2:1.45.1:1.45.1
 google-cloud-core:2.49.1:2.49.1
 google-cloud-shared-dependencies:3.41.1:3.41.1
+gapic-showcase:0.0.0:0.0.1-SNAPSHOT
```
blakeli0, line 20: This is OK as long as release-please does not start bumping versions for showcase. Can you confirm it or maybe @suztomo can confirm?

showcase/scripts/generate_showcase.sh
```diff
@@ -1,129 +1,95 @@
...
+done
 
+if [ -z "${replace}" ]; then
+  replace="false"
+fi
+
+# download api definitions from googleapis repository
+googleapis_commitish=$(grep googleapis_commitish generation_config.yaml | cut -d ":" -f 2 | xargs)
+api_def_dir=$(mktemp -d)
+git clone https://github.com/googleapis/googleapis.git "${api_def_dir}"
```
blakeli0, line 31: Maybe I missed it but I don't see these repos being cleaned up after the generation is completed.

showcase/scripts/generate_showcase.sh
```diff
@@ -1,129 +1,95 @@
...
 get_version_from_pom() {
   target_pom="$1"
   key="$2"
   # prints the result to stdout
   grep -e "<${key}>" "${target_pom}" | cut -d'>' -f2 | cut -d'<' -f1
 }
+# looks at sdk-platform-java/showcase/gapic-showcase/pom.xml to extract the
+# version of gapic-showcase
+# see https://github.com/googleapis/gapic-showcase/releases
+showcase_version=$(get_version_from_pom \
```
blakeli0, line 52: Is this used? We should still use it to checkout showcase repo with a release tag.

showcase/scripts/generate_showcase.sh
```diff
@@ -1,129 +1,95 @@
...
-  --gapic_additional_protos "${gapic_additional_protos}" \
-  --rest_numeric_enums "${rest_numeric_enums}" \
-  --gapic_yaml "${gapic_yaml}" \
-  --service_config "${service_config}" \
-  --service_yaml "${service_yaml}" \
-  --include_samples "${include_samples}" \
-  --transport "${transport}" 
+echo "generating showcase"
+workspace_name="/workspace"
+docker run \
```
blakeli0, line 84: Since this is using docker now, I don't think we can run them from local Mac or Windows anymore, can you please update the README of the showcase module to indicate it?

hermetic_build/common/model/gapic_inputs.py
```diff
@@ -19,24 +19,32 @@
...
+# This pattern ignores api definitions that externally reference other repos.
+# For example: @googleapis_repo//google/cloud/...
+pattern_proto_library_to_additional_protos = {
+    r".*//google/cloud:common_resources_proto": "google/cloud/common_resources.proto",
+    r".*//google/cloud/location:location_proto": "google/cloud/location/locations.proto",
+    r".*//google/iam/v1:iam_policy_proto": "google/iam/v1/iam_policy.proto",
+}
+# This is a special case for showcase, where the additional protos
+# are not declared in the gapic library instead
+pattern_gapic_library_to_additional_protos = {
```
blakeli0, line 45: Do you mind elaborating on why we need this special case?

blakeli0 (commented):
Thanks Diego! This approach looks good in general, however, have we ever considered treating showcase a dedicate repo instead of a library in sdk-platform-java? Meaning that we create a separate generation_config.yaml in showcase module?

hermetic_build/common/model/gapic_inputs.py
diegomarquezp: From discussion, let's append the additional protos manually in generation_showcase.sh

diegomarquezp: Actually, we can just append a proto_library_with_info with the protos we need.

.github/scripts/hermetic_library_generation.sh
diegomarquezp: I'll be testing automatic generation of showcase

generation_config.yaml
diegomarquezp: I removed this new flag

generation_config.yaml
diegomarquezp: It's not. Removing.

showcase/scripts/resources/BUILD.partial.bazel
diegomarquezp: Yes, I kept it as slim as possible. Thanks.

showcase/scripts/generate_showcase.sh
diegomarquezp: Indeed they were not. I added a cleanup exit trap.

showcase/scripts/generate_showcase.sh
diegomarquezp: It should be used, exactly for checking out at the required tag. I'll fix this. Thanks

hermetic_build/common/model/gapic_inputs.py
diegomarquezp: Done

showcase/scripts/generate_showcase.sh
diegomarquezp: Good point. Done.

showcase/scripts/generate_showcase.sh
diegomarquezp: Done

showcase/scripts/generate_showcase.sh
diegomarquezp: Done

.github/scripts/hermetic_library_generation.sh
diegomarquezp: Confirmed automatic generation occurred in https://github.com/googleapis/sdk-platform-java/actions/runs/12901902434/job/35974671498

.cloudbuild/library_generation/library_generation.Dockerfile
```diff
@@ -125,6 +117,14 @@ RUN owl-bot copy-code --version
 RUN chmod o+rx $(which owl-bot)
 RUN apk del -r npm && apk cache clean
 
+# Here we transfer gapic-generator-java from the previous stage.
```
diegomarquezp, line 120: I moved the generator jar copy later in the final stage so the previous steps can run in parallel with `mvn install`. It saves dozens of seconds.

.github/scripts/hermetic_library_generation.sh
```diff
@@ -89,6 +89,10 @@ pushd "${api_def_dir}"
 git checkout "${googleapis_commitish}"
 popd
 
+# we also setup showcase
+source java-showcase/scripts/showcase_utilities.sh
+append_showcase_to_api_defs "${api_def_dir}"
```
blakeli0, line 94: Can this function be called from `generate_showcase.sh` instead of here?

diegomarquezp: It's called both from the automatic generation script (this one) and the manual update + golden test scripts (`generate_showcase.sh`). Since both scripts need to append the showcase protos to googleapis, I extracted this function into a utility file.

blakeli0: My concern is that this `hermetic_library_generation.sh` is reused by google-cloud-java and all handwritten libraries, they don't need to copy showcase protos to googleapis. So can we move this step to a sdk-platform-java specific place? For example, maybe in the [workflow](https://github.com/googleapis/sdk-platform-java/blob/main/.github/workflows/hermetic_library_generation.yaml)?

diegomarquezp:
Failing sonar checks are due to reformatting of showcase files

.github/scripts/hermetic_library_generation.sh
diegomarquezp: Ok, since the workflow also ends up using the common-usage `action.yaml`, I modified the scripts to accept a new flag `showcase_mode` that defaults to `false`. When `true`, it will download the api definitions. The download operation is still done in `hermetic_library_generation.sh`, but we prevent it from happening in repos that don't need these definitions.

blakeli0: I think only [action.yaml](https://github.com/googleapis/sdk-platform-java/blob/main/.github/scripts/action.yaml) is reused. The `hermetic_library_generation.yaml` though is different in each repo. This is the one in [sdk-platform-java](https://github.com/googleapis/sdk-platform-java/blob/main/.github/workflows/hermetic_library_generation.yaml) and this is the one in [google-cloud-java](https://github.com/googleapis/google-cloud-java/blob/main/.github/workflows/hermetic_library_generation.yaml). So we could add the logic that appends showcase protos to the `hermetic_library_generation.yaml` in sdk-platform-java only, and it would not affect other repos.

diegomarquezp: Done!

generation_config.yaml
```diff
@@ -3,6 +3,16 @@ googleapis_commitish: 27aa9d542f54ae1ea1277580550dd04aeb5b09f9
 # java-{library.library_name} or java-{library.api-shortname} when
 # library.library_name is not defined.
 libraries:
+- api_shortname: showcase
+  folder_name: showcase # prevents java-showcase
```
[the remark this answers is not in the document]
diegomarquezp: Removed

.github/scripts/hermetic_library_generation.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -e
+set -ex
```
JoeWang1127, line 2: Why do we want to add `-x` here?

JoeWang1127 (approved):

blakeli0 (approved):

PR #3572, opened 2025-01-17
feat: enable DirectPath bound token in InstantiatingGrpcChannelProvider

rockspore:
Prepares a ComputeEngineCredentials that fetches DirectPath bound tokens for the gRPC ChannelCredentials if applicable.


rockspore:
cc @rmehta19

rmehta19:
Auth dep has been bumped in https://github.com/googleapis/sdk-platform-java/pull/3577, so I think you'll be able to sync this PR and get the tests to pass.

rockspore:
This PR is now ready (thanks to @rmehta19!). @lqiu96 , @blakeli0 , @zhumin8 would you all be able to review it? Thanks!

I noticed some of the comments in #3591 would apply here too but wanted to wait till they got resolved there.

lqiu96:
> This PR is now ready (thanks to @rmehta19!). @lqiu96 , @blakeli0 , @zhumin8 would you all be able to review it? Thanks!
> 
> I noticed some of the comments in #3591 would apply here too but wanted to wait till they got resolved there.

Is there an order that these PRs should go in/ any dependency between the two? If not, we can try to get https://github.com/googleapis/sdk-platform-java/pull/3591 merged and then swing back to this.

rmehta19:
> > This PR is now ready (thanks to @rmehta19!). @lqiu96 , @blakeli0 , @zhumin8 would you all be able to review it? Thanks!
> > I noticed some of the comments in #3591 would apply here too but wanted to wait till they got resolved there.
> 
> Is there an order that these PRs should go in/ any dependency between the two? If not, we can try to get #3591 merged and then swing back to this.

I don't believe there's any dependency between the two. I think Luwei and I are creating similar helper functions though, so the one that merges second can make use of them.

rockspore:
> > > This PR is now ready (thanks to @rmehta19!). @lqiu96 , @blakeli0 , @zhumin8 would you all be able to review it? Thanks!
> > > I noticed some of the comments in #3591 would apply here too but wanted to wait till they got resolved there.
> > 
> > 
> > Is there an order that these PRs should go in/ any dependency between the two? If not, we can try to get #3591 merged and then swing back to this.
> 
> I don't believe there's any dependency between the two. I think Luwei and I are creating similar helper functions though, so the one that merges second can make use of them.

+1 on @rmehta19's comment. I am hoping this PR can make the upcoming release, but there is no required order. #3591 can go first.

rockspore:
@lqiu96 This PR should be ready now. I moved the `altsCallCredentials` creation into the builder. It largely follows #3591. Could you PTAL? Thanks.

rockspore:
> @lqiu96 This PR should be ready now. I moved the `altsCallCredentials` creation into the builder. It largely follows #3591. Could you PTAL? Thanks.

Friendly ping ;)

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -141,6 +141,7 @@ public final class InstantiatingGrpcChannelProvider implements TransportChannelP
   @Nullable private final Boolean keepAliveWithoutCalls;
   private final ChannelPoolSettings channelPoolSettings;
   @Nullable private final Credentials credentials;
+  @Nullable private final CallCredentials altsCallCredentials;
   @Nullable private final CallCredentials mtlsS2ACallCredentials;
```
lqiu96, line 145: Can altsCallCredentials and mtlsS2ACallCredentials share the same variable? Something like hardBoundCallCredentials?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -616,8 +618,13 @@ private ManagedChannel createSingleChannel() throws IOException {
     boolean useDirectPathXds = false;
     if (canUseDirectPath()) {
       CallCredentials callCreds = MoreCallCredentials.from(credentials);
+      // altsCallCredentials may be null and will be correctly ignored by
+      // GoogleDefaultChannelCredentials.
       ChannelCredentials channelCreds =
-          GoogleDefaultChannelCredentials.newBuilder().callCredentials(callCreds).build();
+          GoogleDefaultChannelCredentials.newBuilder()
+              .callCredentials(callCreds)
+              .altsCallCredentials(altsCallCredentials)
```
lqiu96, line 626: if altsCallCredentials is non-null, do we need/ should pass both CallCredentials over? Is that the intended behavior for GoogleDefaultChannelCredentials?

i.e.
check null altscallcredentials -> only send callCreds
check non-null altscallcredentials -> only send altsCallCredentials

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -1172,6 +1183,18 @@ boolean isMtlsS2AHardBoundTokensEnabled() {
           .anyMatch(val -> val.equals(HardBoundTokenTypes.MTLS_S2A));
     }
 
+    boolean isDirectPathBoundTokenEnabled() {
+      // If the list of allowed hard bound token types is empty or doesn't contain
+      // {@code HardBoundTokenTypes.ALTS}, the {@code credentials} are null or not of type
+      // {@code ComputeEngineCredentials} then DirectPath hard bound tokens should not be used.
+      // DirectPath hard bound tokens should only be used on ALTS channels.
+      if (allowedHardBoundTokenTypes == null
```
lqiu96, line 1191:
```suggestion
      if (allowedHardBoundTokenTypes.isEmpty()
```

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
lqiu96: Oh, from reading the code below, I'm assuming this is a limitation from directpath.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rockspore: Yes, GoogleDefaultChannelCredentials need the callCreds even when altsCallCreds is not null. Because gRPC internally allows DirectPath to CloudPath fallback. If the actual connection turns out to be CloudPath, the altsCallCreds cannot be used.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -1194,6 +1217,11 @@ public InstantiatingGrpcChannelProvider build() {
                 ComputeEngineCredentials.GoogleAuthTransport.MTLS,
                 ComputeEngineCredentials.BindingEnforcement.ON);
       }
+      if (isDirectPathBoundTokenEnabled()) {
+        this.altsCallCredentials =
+            createHardBoundTokensCallCredentials(
+                ComputeEngineCredentials.GoogleAuthTransport.ALTS, null);
+      }
```
lqiu96, line 1224: In the future if allowedHardBoundTokenTypes contains both MTLS and ALTS with S2A enabled, it's possible that both mtls and alts credentials would be non-null? I'm assuming this is becuase of directpath only being determined when we create the channel.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rockspore: Hmm, I guess we can at this point, because it's always one or the other. But does that provide any benefit?

If in the future, we want gRPC to use mTLS when falling back from DirectPath to CloudPath, we will need these two different creds again.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rockspore: Yes we don't want these two credentials to interfere with each other, even though it's mostly one or the other when the client makes the connection. So they can co-exist.

rockspore: cc @rmehta19

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rockspore: Thanks for the catch!

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
lqiu96: Is it always going to be one or the other? I'd rather just have one variable to keep track of the hard bound callcredential state -> mtls vs alts if possible. Adding more variables makes it harder to keep track of all the combinations of possible states (mtls can be null or non-null + alts can be null or non-null). 

> If in the future, we want gRPC to use mTLS when falling back from DirectPath to CloudPath

Is this something that is in the works and you know that it will require multiple types of callcredentials? Or is this just a hypothetical example?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -616,8 +618,13 @@ private ManagedChannel createSingleChannel() throws IOException {
     boolean useDirectPathXds = false;
     if (canUseDirectPath()) {
       CallCredentials callCreds = MoreCallCredentials.from(credentials);
+      // altsCallCredentials may be null and will be correctly ignored by
+      // GoogleDefaultChannelCredentials.
```
lqiu96, line 622: can you add something like may be null when X... I think from your comment below that it's for cloudpath scenarios?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
lqiu96: > even though it's mostly one or the other

Can you elaborate what you mean by mostly? If there is a scenario that both need to exist, then we'll need to have both and I think this looks fine

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rockspore: Let's consolidate the discussion around keeping two credentials to this thread.

> Is it always going to be one or the other?

I believe so in normal use cases although at this point you can see that DirectPath enablement is also controlled by an environment variable. So in theory, I suppose the decision could change in the runtime so two credentials need to exist independently to ensure there is no interference.

Even when that env var switch gets removed in the future, IMO we should still keep two credentials separate because we use a list of allowed bound token types to control the enablement of them independently. From that perspective, it should be fairly straightforward to track the null-ness of these two variables - they are enabled by the enum in the list.

> Is this something that is in the works and you know that it will require multiple types of callcredentials? Or is this just a hypothetical example?

Our team has talked about that a few times but there is no concrete plan as of now. So yes it's fair to be considered as a hypothetical example.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rockspore: Sorry I didn't explain very clearly. This may be null if we don't enable the bound token for DirectPath. And gRPC will use the original credentials for everything (DirectPath or CloudPath). Otherwise, gRPC will use it for DirectPath connections and the original credentials for CloudPath fallback.

I added a sentence for this.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rockspore: I tried to explain in the other comment. Let's resolve this there.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rmehta19: > DirectPath enablement is also controlled by an environment variable. So in theory, I suppose the decision could change in the runtime so two credentials need to exist independently to ensure there is no interference.

+1, the answer to "can DirectPath be used" could change from after this `transportChannelProvider` is built  (`.build()`) but before calling `transportChannelProvider.createSingleChannel()`.

In general, I think it may be simpler to keep these two creds separate. I think that combining them would complicate our `isMtlsS2AHardBoundTokensEnabled` and `isDirectPathBoundTokenEnabled` logic below. Since both can be true on GCE and DirectPath takes precedence over using S2A, I think we would end up needing to combine the logic from those functions if we wanted a single `hardBoundCallCredential`. While DirectPath and S2A are doing similar things when it comes to enabling the usage hard bound tokens, I think at this time it might be difficult to try and reason about them together like this 

Additionally, @lqiu96, from our discussion in https://github.com/googleapis/sdk-platform-java/pull/3591#discussion_r1939769759, It seems like creating 2 creds is ok, as they are created in the provider once, and are not exposed by the provider? That was a detailed discussion though, so I could be missing something.

lqiu96: I'm not entirely sure I'm following because I am still only see this as a limitation with how DirectPath is being resolved. 

Correct me if I'm wrong. For hard bound tokens, even if MTLS_S2A and ALTS are both in the hard-bound list, only one of them is going to to be used, right? Even if both are currently resolved to be non-null. The intended logic is roughly as follows:

```
If hardbound tokens list is non-empty:
- If DirectPath + ATLS token included -> Create the altsCallCredentials
- If Non-DirectPath + S2A enabled + S2A_MTLS  token included -> Create the mtlsCallCredentials

Else:
- Use the normal flow
```

Changing how DirectPath is done is out of scope for this PR and two different hard-bound tokens can live here for now. I'm just not piecing together why two separate hard bound tokens need to exist (besides for the directpath limitation), since it's going to be either ALTS or MTLS, right? Or is there some fallback logic that I may be forgetting about.

lqiu96 (approved):
LGTM. Any potential changes (regarding DirectPath) are out of scope for this PR.

rockspore:
> LGTM. Any potential changes (regarding DirectPath) are out of scope for this PR.

Thanks a lot for reviewing this @lqiu96! Could you help merge it once ready?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
lqiu96: Yes agreed. Really either way is fine logic-wise, but I just want to be cautious of adding in so many additional credential variables (as it went from 1 to 3). Each may include multiple states and this increases the mental load when figuring out how channels are created, which channels are created, and which credentials are used for each channel used.

PR #3591, opened 2025-01-27
feat: Enable MTLS_S2A bound token by default for gRPC S2A enabled flows

rmehta19:
Similar to implementation for DirectPath in https://github.com/googleapis/sdk-platform-java/pull/3572.

This is part of the experimental S2A feature (see #3400)


rmehta19:
@lqiu96 , @blakeli0 , @zhumin8 would you all be able to review this PR? Thanks!

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -592,6 +593,41 @@ ChannelCredentials createS2ASecuredChannelCredentials() {
     }
   }
 
+  boolean isMtlsS2AHardBoundTokensEnabled() {
```
lqiu96, line 596: just a thought (nothing that needs to be changed in this PR): With how many helper methods we have for S2A and hard bound tokens, I wonder if we can split these methods into a helper class in Gax-Grpc (something like S2AMtlsContext or something)

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -592,6 +593,41 @@ ChannelCredentials createS2ASecuredChannelCredentials() {
...
+    if (!useS2A) {
+      // If S2A cannot be used, {@code HardBoundTokenTypes.MTLS_S2A} hard bound tokens should not be
+      // used
+      return false;
+    }
+    if (allowedHardBoundTokenTypes == null
```
lqiu96, line 602: nit: I think this may have slipped in a previous review.

Is it possible to have allowedHardBoundTokenTypes default to a empty list? And we ensure that the setter can't set this value to be null? I think I would rather have something like an list.isEmpty() check rather than a null check

lqiu96, line 602: nit: Can we also merge these fail-fast if checks together?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -592,6 +593,41 @@ ChannelCredentials createS2ASecuredChannelCredentials() {
...
+        || credentials == null
+        || !(credentials instanceof ComputeEngineCredentials)) {
+      return false;
+    }
+    for (HardBoundTokenTypes boundTokenTypes : allowedHardBoundTokenTypes) {
+      if (boundTokenTypes == HardBoundTokenTypes.MTLS_S2A) {
+        return true;
+      }
+    }
+    return false;
```
lqiu96, line 612: Can we use [Stream.anyMatch](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#anyMatch-java.util.function.Predicate-) here?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -592,6 +593,41 @@ ChannelCredentials createS2ASecuredChannelCredentials() {
...
+    ComputeEngineCredentials.Builder credsBuilder =
+        ((ComputeEngineCredentials) credentials).toBuilder();
+    // We only set scopes and HTTP transport factory from the original credentials because
+    // only those are used in gRPC CallCredentials to fetch request metadata.
+    return MoreCallCredentials.from(
+        ComputeEngineCredentials.newBuilder()
+            .setScopes(credsBuilder.getScopes())
+            .setHttpTransportFactory(credsBuilder.getHttpTransportFactory())
+            .setGoogleAuthTransport(googleAuthTransport)
+            .setBindingEnforcement(bindingEnforcement)
```
lqiu96, line 627: Do we need to check if these are null before setting them in the builder? I know from below that we are hard-coding them to be MTLS and ON, but I'm not sure if this is something that may be optional/ customizable in the future.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rmehta19: Yes, we can, done in https://github.com/googleapis/sdk-platform-java/pull/3591/commits/7c20422a57882b0243b298ef8e6f33293e4aa61a

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rmehta19: Yes, done in https://github.com/googleapis/sdk-platform-java/pull/3591/commits/7c20422a57882b0243b298ef8e6f33293e4aa61a

rmehta19: Yes, we can have `allowedHardBoundTokenTypes` default to an empty list and make sure if someone passes null to setter, we set it to an empty list instead. Done in https://github.com/googleapis/sdk-platform-java/pull/3591/commits/7c20422a57882b0243b298ef8e6f33293e4aa61a

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rmehta19: We should be fine to not perform a null check. In [the auth library](https://github.com/googleapis/google-auth-library-java/blob/107a332eee18390f1b286906090fc7ada4874ad5/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java#L270-L275) we will only use the value if it is non-null.

> I'm not sure if this is something that may be optional/ customizable in the future.
This code isn't going to change in the future. The way that client libraries will opt out of the hard bound tokens feature is going to be by removing `HardBoundTokenTypes.MTLS_S2A` from the list of `allowedHardBoundTokenTypes`. This function (`createHardBoundTokensCallCredentials`) is only going to be called by the below code when `isMtlsS2AHardBoundTokensEnabled` evaluates to true.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rmehta19: Yes, I think this would be help to reduce the complexity in the `InstantiatingGrpcChannelProvider` file. I'm happy to do the cleanup of that in a followup CL.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -146,7 +148,7 @@ public final class InstantiatingGrpcChannelProvider implements TransportChannelP
   @VisibleForTesting final ImmutableMap<String, ?> directPathServiceConfig;
   @Nullable private final MtlsProvider mtlsProvider;
   @Nullable private final SecureSessionAgent s2aConfigProvider;
-  @Nullable private final List<HardBoundTokenTypes> allowedHardBoundTokenTypes;
```
lqiu96, line 149: Can you set this in the constructor below? And Remove the Nullable tag?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -914,7 +954,11 @@ Builder setUseS2A(boolean useS2A) {
      */
     @InternalApi
     public Builder setAllowHardBoundTokenTypes(List<HardBoundTokenTypes> allowedValues) {
-      this.allowedHardBoundTokenTypes = allowedValues;
+      if (allowedValues == null) {
+        this.allowedHardBoundTokenTypes = new ArrayList<>();
+      } else {
+        this.allowedHardBoundTokenTypes = allowedValues;
+      }
```
lqiu96, line 961: Can we use [Preconditions.checkNotNull](https://guava.dev/releases/19.0/api/docs/com/google/common/base/Preconditions.html#checkNotNull(T)) on this list?

And assign only if that passes

lqiu96: Also I think you need to add `this.allowedHardBoundTokenTypes = ...`  to the Builder's constructor.

rmehta19: Thanks for pointing these things out! Done in https://github.com/googleapis/sdk-platform-java/pull/3591/commits/d8b2959f1e05d4158570a2d9af93ff061ac8f310

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rmehta19: Done in https://github.com/googleapis/sdk-platform-java/pull/3591/commits/d8b2959f1e05d4158570a2d9af93ff061ac8f310

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -592,6 +597,35 @@ ChannelCredentials createS2ASecuredChannelCredentials() {
...
+    ComputeEngineCredentials.Builder credsBuilder =
+        ((ComputeEngineCredentials) credentials).toBuilder();
+    // We only set scopes and HTTP transport factory from the original credentials because
+    // only those are used in gRPC CallCredentials to fetch request metadata.
+    return MoreCallCredentials.from(
+        ComputeEngineCredentials.newBuilder()
+            .setScopes(credsBuilder.getScopes())
+            .setHttpTransportFactory(credsBuilder.getHttpTransportFactory())
+            .setGoogleAuthTransport(googleAuthTransport)
+            .setBindingEnforcement(bindingEnforcement)
```
lqiu96, line 625: qq, why create a new ComputeEngineCredentials builder rather than using the the one above? L616 and add the auth transprot and binding enforcement values?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -914,7 +958,10 @@ Builder setUseS2A(boolean useS2A) {
      */
     @InternalApi
     public Builder setAllowHardBoundTokenTypes(List<HardBoundTokenTypes> allowedValues) {
-      this.allowedHardBoundTokenTypes = allowedValues;
+      this.allowedHardBoundTokenTypes =
+          Preconditions.checkNotNull(
+              allowedValues, "Illegal Argument, allowedValues cannot be null");
```
lqiu96, line 963:
```suggestion
              allowedValues, "List of allowed HardBoundTokenTypes cannot be null");
```

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -592,6 +597,35 @@ ChannelCredentials createS2ASecuredChannelCredentials() {
     }
   }
 
+  boolean isMtlsS2AHardBoundTokensEnabled() {
+    if (!useS2A
+        // If S2A cannot be used, {@code HardBoundTokenTypes.MTLS_S2A} hard bound tokens should not
+        // be used
```
lqiu96, line 603: Can we move this comment to the method level comments and mention the the reason for the 4 conditions below?

gax-java/gax-grpc/src/test/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProviderTest.java
```diff
@@ -1103,6 +1103,82 @@ void createS2ASecuredChannelCredentials_returnsPlaintextToS2AS2AChannelCredentia
...
+            .setAllowHardBoundTokenTypes(
+                Collections.singletonList(
+                    InstantiatingGrpcChannelProvider.HardBoundTokenTypes.ALTS))
+            .setCredentials(computeEngineCredentials)
+            .build();
+    Truth.assertThat(provider.isMtlsS2AHardBoundTokensEnabled()).isFalse();
+  }
+
+  @Test
+  void isMtlsS2AHardBoundTokensEnabled_returnsTrue() {
```
lqiu96, line 1170: nit: `isMtlsS2AHardBoundTokensEnabled_returnsTrue`-> `isMtlsS2AHardBoundTokensEnabled_mtlsS2ATokenAllowedInList`

This test case could have the allowed list could have both MTLS_S2A and ALTS. As long as MTLS_S2A is included, it should pass.

lqiu96 (approved):
LGTM. Can the title be updated to be something like:

`feat: Enable MTLS_S2A bound token by default for gRPC S2A enabled flows`

So users who use gRPC don't need to worry.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rmehta19: This was done to avoid mutating the original credentials object (copied this logic from #3572) with something like:
```
MoreCallCredentials.from(
((ComputeEngineCredentials) credentials).toBuilder()
.setGoogleAuthTransport(googleAuthTransport)
.setBindingEnforcement(bindingEnforcement)
.build())
``` 

From testing internally with both options (mutate the original credential object vs. creating a new one), I found (via logging) that if we mutate the original credential object, only the first request on the channel triggers a token request to MDS with the desired request params. Subsequent requests on the channel don't trigger new MDS token requests. When we create a new credential object like in this PR, we get the expected behavior where on each request on the channel, there is a request to MDS with the desired request params. I tested this a few times to make sure it wasn't flakiness in logs, this does indeed seem to be the behavior. I'm not sure why mutating the credential results in this unexpected behavior.

Aside from this observation, I do think that creating a new credential object rather than mutating the original is a safer option, as I don't know how the original credential object is used after we use it. WDYT?

cc:@rockspore since he investigated the creation of new credential vs. mutating the credential more than I did.

rockspore: I also think the original credential object should keep its behavior unchanged.

Unrelated to this, why is a request made to MDS every RPC? I thought the token would be cached regardless of the binding?

gax-java/gax-grpc/src/test/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProviderTest.java
rmehta19: Done, added ALTS to the list as well, and verified it is passing. https://github.com/googleapis/sdk-platform-java/pull/3591/commits/8c060dc004f45634f9ce5c0d3a3f16a8111137c7.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rmehta19: Done in https://github.com/googleapis/sdk-platform-java/pull/3591/commits/8c060dc004f45634f9ce5c0d3a3f16a8111137c7

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rmehta19: Done in https://github.com/googleapis/sdk-platform-java/pull/3591/commits/8c060dc004f45634f9ce5c0d3a3f16a8111137c7

rmehta19:
Thanks @lqiu96 for the review!

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rockspore: I discussed with @rmehta19 offline and to clarify my question here, since one particular instance of ComputeEngineCredentials is only fetching one type of bound (or unbound) tokens, the caching logic should not differ at all.

lqiu96: Hmm, actually this does bring up a few interesting points:

1. Gax-Grpc has concept of Channel Pooling and allows multiple grpc-java channels to be created, shared, and re-created at fixed intervals. Pooling ends up calling `createSingleChannel()` many times over the span of a client library's lifespan. Since each of those calls ends up creating a new Credentials, each one would make a new MDS call since the token isn't cached (as a new object is created).
2. Creating a new Credentials would end up creating two sets of Credentials (1 that exists inside ClientContext without the S2A_MTLS values and 1 inside the ChannelProvider that has the true values). I'm worried that we may end up with a split/ multiple different Credentials values. ClientContext is public and we won't know all the user use cases that users may be doing with `ClientContext.getCredentials()`.
3. We also have FixedCredentialsProvider which is intended to supply the same Credentials object. If the Credentials is re-created in the Channel Provider, this may be problematic since it won't be the *same* object used.

lqiu96: No worries. I'm still taking a look. A few of these implications (re-building credentials and the corresponding access token behavior) were some stuff that were more so recently discovered and changed due to bugs filed. I think a few of the concerns (1 and 3) above can be mitigated by just changing when we re-create the CallCredentials.

I'm not sure concern 2 can be resolved in this PR. It was an oversight on my end and it was also more so a concern that came up when we were looking into resolving endpoints earlier last year.

rmehta19: > actually this does bring up a few interesting points:

Thanks for pointing these items out! I wasn't aware of these APIs in gax exposing the credentials. 

> I think a few of the concerns (1 and 3) above can be mitigated by just changing when we re-create the CallCredentials.

I agree that if we make the MTLS_S2A cred outside of `createSingleChannel` (and store it in a new var) we can avoid the cred being created each time `createSingleChannel` is called by pooling logic. The credential supplied via `FixedCredentialProvider` would be the un-modified cred. WDYT about setting `mtlsS2ACredentials` in the `InstantiatingGrpcChannelProvider` constructor, after `credentials` are set?

> I'm not sure concern 2 can be resolved in this PR

FWIW, I think it would be best if `ClientContext.getCredentials()` returns original cred (without the `MTLS_S2A` values). Because tokens obtained using `MTLS_S2A` cred should only be used on a channel established using MTLS via S2A.

lqiu96: > WDYT about setting mtlsS2ACredentials in the InstantiatingGrpcChannelProvider constructor, after credentials are set?

Yeah, I think this is an option. Or does this validation/ setting in the `Builder.build()` method. I think either is fine with me ATM.

> I think it would be best if ClientContext.getCredentials() / FixedCredentialsProvider returns original cred (without the MTLS_S2A values). Because tokens obtained using MTLS_S2A cred should only be used on a channel established using MTLS via S2A.

Yeah, I think my ideal situation would have been that the Credentials would be set/ modified prior to being passed to the ChannelProvider. However, since the flags/ options are only available in the ChannelProvider, I don't see an easy way to do this. And to be honest, I think `ClientContext.getCredentials()` never should have been public and thus eliminating this worry to begin with.

As of now, `ComputeEngineCredentials.toBuilder()` keeps the same access token. We have been thinking about trying to nullify the access token on that call because we want to tie an access token to the internal values of the token (and something like `toBuilder()` is intended to change those internal values). I think in the future, `toBuilder()` for any `*Credentials` class will nullify the access token and force a manual refresh for the access token. And something like FixedCredentialsProvider would end having multiple MDS calls for S2A_MTLS.

> Because tokens obtained using MTLS_S2A cred should only be used on a channel established using MTLS via S2A.

I think that does make sense. When S2A is enabled, MTLS_S2A is really the first option that is tried and when the channel is created, we still have no idea if MTLS_S2A can be done. If there are, we would fall back to the other option and use the old/ existing credentials.

rmehta19: > Yeah, I think this is an option. Or does this validation/ setting in the Builder.build() method. I think either is fine with me ATM.

SG. In <> I move the calls to`isMtlsS2AHardBoundTokensEnabled` and `createHardBoundTokensCallCredentials` to `Builder.build()` method. The implementation of `createHardBoundTokensCallCredentials`  is still the same, as discussed (create a new credential, do not mutate existing credential).

> I think that does make sense. When S2A is enabled, MTLS_S2A is really the first option that is tried and when the channel is created, we still have no idea if MTLS_S2A can be done. If there are, we would fall back to the other option and use the old/ existing credentials.

To be clear: if an MTLS_S2A token gets used on a channel that wasn't established using MTLS via S2A, the requests would fail; they should _only_ get used on MTLS via S2A channels.

Additionally, the sequence of decisions is:
- can we use S2A (check `useS2A`)?
     - if so: can we successfully create S2A-enabled `ChannelCredentials`? If we can't, fall back to default: TLS. If we can, should we use MTLS_S2A hard bound tokens (`mtlsS2ACallCredentials` != null)? If we should, create channel using `mtlsS2ACallCredentials` and S2A-enabled `ChannelCredentials` (`CompositeChannelCredentials`). If we shouldn't, create channel using just the S2A-enabled `ChannelCredentials`
     - if not: fall back to default: TLS

In other words, Once the channel using S2A+MTLS_S2A hard bound tokens is created, there is no possibility to fallback to using the old/existing credentials. This is different from DirectPath logic (in #3572), which _does_ have fallback behavior. 

> I think in the future, toBuilder() for any *Credentials class will nullify the access token and force a manual refresh for the access token. And something like FixedCredentialsProvider would end having multiple MDS calls for S2A_MTLS.

I think the nullify behavior you describe makes sense to implement. However, MTLS_S2A and ALTS credentials should stay scoped to the GrpcChannelProvider. For the reasons mentioned above, we only want to MTLS_S2A/ALTS credentials when we are in control of what channel they are being used on, since it would result in rejected Google API requests if tokens from those credentials are used on non-DirectPath / non MTLS via S2A channels.

lqiu96: > To be clear: if an MTLS_S2A token gets used on a channel that wasn't established using MTLS via S2A, the requests would fail; they should only get used on MTLS via S2A channels.

Yep, understood. My point above was that I wish that we could determine these configurations and create the proper Credentials prior to channel creation. This way we can pass the correct Credentials to the channel instead of having to modify the Credentials during channel creation. This is an existing limitation based on the way these classes were written previously and I don't think there is an easy way for us to change this without a major rewrite.

> In other words, Once the channel using S2A+MTLS_S2A hard bound tokens is created, there is no possibility to fallback to using the old/existing credentials.

It's not that I want the user to be able to choose which Credentials to use with the channel, but rather I'd want the user to be able to know which Credentials is being used. For example, if the Credentials is modified (adding binding_enforcement), the customer should able to figure that out if they do something like `getCredentials()`. I think this something better suited for something like logging tbh, but since we do expose a public `getCredentials()`, that's why I raised this concern.

> However, MTLS_S2A and ALTS credentials should stay scoped to the GrpcChannelProvider.

I don't disagree to the idea to try and narrow the credentials to where they are being used. I think this is something the old maintainers had done. IMO, I think this has actually made our lives so much harder given that we have a multitude of settings that are scattered among different modules and used during different phases of client initialization (gRPC Channel Provider in gax-grpc + stubsettings and client context in Gax). I'd almost wish that we just localized all the settings in once place so things can just fail when they're used incorrectly (i.e. MTLS_S2A and ALTS credentials just fail and throw an error message if used with HttpJsonChannelProvider).

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -1133,7 +1152,53 @@ public Builder setDirectPathServiceConfig(Map<String, ?> serviceConfig) {
...
+          ComputeEngineCredentials.newBuilder()
+              .setScopes(credsBuilder.getScopes())
+              .setHttpTransportFactory(credsBuilder.getHttpTransportFactory())
+              .setGoogleAuthTransport(googleAuthTransport)
+              .setBindingEnforcement(bindingEnforcement)
+              .build());
+    }
+
     public InstantiatingGrpcChannelProvider build() {
+      if (isMtlsS2AHardBoundTokensEnabled()) {
```
lqiu96, line 1194: Since the validation is done in the builder, can you set the default to be an empty list in the Builder's constructor? https://github.com/googleapis/sdk-platform-java/blob/c2d276857f6ca0e86c91e15af5e08a737bd1510b/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L826

And set the value from the Builder here as well: https://github.com/googleapis/sdk-platform-java/blob/c2d276857f6ca0e86c91e15af5e08a737bd1510b/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L829

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -1133,7 +1152,53 @@ public Builder setDirectPathServiceConfig(Map<String, ?> serviceConfig) {
...
+          ((ComputeEngineCredentials) this.credentials).toBuilder();
+      // We only set scopes and HTTP transport factory from the original credentials because
+      // only those are used in gRPC CallCredentials to fetch request metadata.
+      return MoreCallCredentials.from(
+          ComputeEngineCredentials.newBuilder()
+              .setScopes(credsBuilder.getScopes())
+              .setHttpTransportFactory(credsBuilder.getHttpTransportFactory())
+              .setGoogleAuthTransport(googleAuthTransport)
+              .setBindingEnforcement(bindingEnforcement)
+              .build());
```
lqiu96, line 1190: `toBuilder()` will create a new `ComputeEngineCredentials` object. I think you can just do `credentials.toBuilder().set(...).set(...).build()` here

rockspore: Thanks. That was my bad from #3572. I thought `toBuilder()` would keep references of many things of the original credentials object. But it looks like it only keeps the transport factory and scopes. So I think we only need to do

```
credentials.toBuilder()
              .setGoogleAuthTransport(googleAuthTransport)
              .setBindingEnforcement(bindingEnforcement)
              .build();
```

rmehta19: Thanks both! Done in https://github.com/googleapis/sdk-platform-java/pull/3591/commits/0fe5134ae4805d90e8ed734591b69832334288a1

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -175,7 +178,10 @@ private InstantiatingGrpcChannelProvider(Builder builder) {
     this.headerProvider = builder.headerProvider;
     this.useS2A = builder.useS2A;
     this.endpoint = builder.endpoint;
-    this.allowedHardBoundTokenTypes = builder.allowedHardBoundTokenTypes;
+    this.allowedHardBoundTokenTypes =
+        builder.allowedHardBoundTokenTypes == null
+            ? new ArrayList<>()
+            : builder.allowedHardBoundTokenTypes;
```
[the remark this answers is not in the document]
rmehta19: Done in https://github.com/googleapis/sdk-platform-java/pull/3591/commits/0fe5134ae4805d90e8ed734591b69832334288a1

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rmehta19: Done in https://github.com/googleapis/sdk-platform-java/pull/3591/commits/0fe5134ae4805d90e8ed734591b69832334288a1

lqiu96 (approved):
LGTM. In a future PR, I think we can try and see if we can clean up the S2A logic and move it into a helper/ util class.

rockspore:
@lqiu96 Could you merge this if it's ready to do so? Thanks. I will update my #3572 after that and let you know when that's ready for a look.

PR #3617, opened 2025-02-07
chore: adjust native-image config for spring-cloud-gcp

diegomarquezp:
Context: upgrading to GraalVM for JDK 23

Updating the image support in spring (https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/3536) caused errors at analysis time solved by https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/3536/commits/d3f6185f6f1cbca811ca22d8abe3b4149ea7877a

Gax seems to be a better place for this kind of configs.


diegomarquezp:
The config worked in spring
image

lqiu96 (approved):
LGTM. @mpeddada1 should be the final approval

mpeddada1 (approved):

diegomarquezp:
From discussion with @blakeli0, will investigate the root cause of the need for this change.

blakeli0 (approved):
This change is harmless so it is OK to merge it in if it is blocking GraalVM upgrade. But we need to understand why we have to make these changes and document them.

PR #3622, opened 2025-02-10
fix: Avoid creating message string prematurely for streaming calls

### LINKED ISSUE #3621: gax: Excessive byte[] allocations with large read stream requests
Originally opened here: https://github.com/googleapis/java-bigtable/issues/2480 then noticed it's not the right repo, so moving here for a fix.

Hi, recently we've been using the BigtableDataClient to send large requests using the `readRows` streaming method, we observed that with some large requests our CPUs increase to 100%, upon profiling it was mostly the GC running, running an allocation profile showed the following:
Image

Which is coming from: `com.google.api.gax.rpc.StateCheckingResponseObserver#onResponse`:
```java
public final void onResponse(V response) {
  Preconditions.checkState(!isClosed, getClass() + " received a response after being closed.");
  onResponseImpl(response);
}
```

The string concatenation is common to be optimized into a StringBuilder by the JVM, so there is a ton of allocation coming from this bit: `getClass() + " received a response after being closed."`, but the exception is not being thrown here.

Pre-computing the error message every time is an issue, we need to only do that if the exception will be thrown.

#### Environment details

1. Specify the API at the beginning of the title.
2. OS type and version: Ubuntu 22
3. Java version: OpenJDK17
4. bigtable version(s): 2.60.0

#### Steps to reproduce

  1. Create a BigtableDataClient
  2. In a loop send massive requests (1K+ keys)
  3. Profile the JVM observe the time and allocations

#### Code example

```java
BigtableDataSettings settings = BigtableDataSettings.newBuilder()
        .setMetricsProvider(NoopMetricsProvider.INSTANCE)
        .setProjectId("my-project")
        .setInstanceId("my-instance")
        .setAppProfileId("my-profile")
        .build();

BigtableDataClient client = BigtableDataClient.create(settings);
List<String> keys = List.of(); // 1K+ keys that exist
while (!Thread.interrupted()) {
    Query query = Query.create("my-table");
    keys.forEach(query::rowKey);
    client.readRows(query).stream().count(); // Just to consume the stream
}
```

#### Stack trace
See the screenshot above

#### External references such as API reference guides
N/A

#### Any additional information below
N/A

mina-asham:
- Creating the string message prematurely puts a lot of strain on the JVM and GC, switch to the format version of `Preconditions.checkState` which only constructs the string if we throw the exception
- Issue: https://github.com/googleapis/sdk-platform-java/issues/3621

Fixes #3621 ☕️

### DIFF
--- a/gax-java/gax/src/main/java/com/google/api/gax/rpc/StateCheckingResponseObserver.java
+++ b/gax-java/gax/src/main/java/com/google/api/gax/rpc/StateCheckingResponseObserver.java
@@ -43,7 +43,7 @@ public abstract class StateCheckingResponseObserver<V> implements ResponseObserv
    * ensuring consistent state.
    */
   public final void onStart(StreamController controller) {
-    Preconditions.checkState(!isStarted, getClass() + " is already started.");
+    Preconditions.checkState(!isStarted, "%s is already started.", getClass());
     isStarted = true;
 
     onStartImpl(controller);
@@ -56,7 +56,7 @@ public abstract class StateCheckingResponseObserver<V> implements ResponseObserv
    * consistent state.
    */
   public final void onResponse(V response) {
-    Preconditions.checkState(!isClosed, getClass() + " received a response after being closed.");
+    Preconditions.checkState(!isClosed, "%s received a response after being closed.", getClass());
     onResponseImpl(response);
   }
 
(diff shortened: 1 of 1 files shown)


lqiu96:
Thanks for raising the issue and providing a fix. This does seem like an issue and we'll review this.

Could you update the title to follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)? Perhaps something like `fix: Avoid creating message string prematurely for streaming calls".

mina-asham:
Thanks @lqiu96, updated the title!

zhumin8 (approved):
Looks reasonable to me.

Issue #2237, opened 2023-11-10
Handle `google.rpc.ErrorInfo` in HttpJson*Stub class


JoeWang1127:
Followup on #2229.

The generated HttpJsonEchoStub can't handle message type: `google.rpc.ErrorInfo` introduced by `echo.proto` (version `0.29.0`).
A easy integration test:
```
@Test
public void testEchoErrorDetails() {
  EchoErrorDetailsResponse response = httpjsonClient.echoErrorDetails(
      EchoErrorDetailsRequest
          .newBuilder()
          .setSingleDetailText("singleDetailText1774380934")
          .addAllMultiDetailText(new ArrayList<>())
          .build());
  assertThat(response.hasSingleDetail()).isEqualTo(true);
}
```
failed with the following error message:
```
Caused by: com.google.protobuf.InvalidProtocolBufferException: Cannot resolve type: type.googleapis.com/google.rpc.ErrorInfo
    at com.google.protobuf.util.JsonFormat$ParserImpl.mergeAny(JsonFormat.java:1511)
```

Possible solution:
Add the message type in 
https://github.com/googleapis/sdk-platform-java/blob/003b993f7ad7cae8ae8c101e0ff147e517dcd83e/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java#L1279-L1286

JoeWang1127:
In the possible solution part, the types are added recursively in `TypeRegistry.Builder.add()` method.

However, only two types serves as the "root" of this method, `WaitRequest` and `WaitResponse`, we need to understand the reason behind this before adding `ErrorInfo` into the registry.

lqiu96:
Took a look at some of the code in Gax and I believe it may be due to the ProtoOperationTransformers: https://github.com/googleapis/sdk-platform-java/blob/9d30ed9b05dd77348899d312117e3fb562e7e6bd/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ProtoOperationTransformers.java#L130-L140

ServiceStubSettings transforms it from Any to the classes: 
https://github.com/googleapis/sdk-platform-java/blob/9d30ed9b05dd77348899d312117e3fb562e7e6bd/showcase/gapic-showcase/src/main/java/com/google/showcase/v1beta1/stub/EchoStubSettings.java#L792-L795

I believe conversion from Any -> Class would require a TypeRegistry.

[...]

PR #3640, opened 2025-02-13
fix: NPE in parsing error status during HttpJson server streaming

baeminbo:
Fix NPE in https://github.com/googleapis/java-spanner/issues/3640. 

However, a separate type resolution error will still prevent successful error status parsing.  This requires a more substantial change, as discussed in [#2237 (comment).](https://github.com/googleapis/sdk-platform-java/issues/2237#issuecomment-2655697832).

----


baeminbo:
@blakeli0, could you please review this? Thanks.

blakeli0:
@baeminbo Thanks for reporting this issue and creating this PR! However, I would like to discuss the priority of this issue before we jumping to the solution. Please see my latest comment https://github.com/googleapis/java-spanner/issues/3640#issuecomment-2670018252. 
In terms of the current solution, I don't think it scales. Because it only fixes the issue for server streaming calls, while it could happen for any types of calls (unary, client streaming etc.).

baeminbo:
Unary calls do not throw the NullPointerException because a similar fix to this PR has already been applied. [HttpJsonDirectCallable](https://github.com/googleapis/sdk-platform-java/blob/v2.53.0/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonDirectCallable.java#L58-L69) sets `TypeRegistry` to the `CallOption` to avoid NPE. 

It appears that httpjson doesn't support client-streaming and bidi-streaming. 
1. [HttpJsonCallableFactory](https://github.com/googleapis/sdk-platform-java/blob/v2.53.0/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonCallableFactory.java), which is responsible to create callables, doesn't provide methods for client and bidi streaming types. 
2. It appears that auto-generated stub codes explicitly marks client and bidi streaming as unsupported. For example, bidi-streaming call [StreamingPull](https://github.com/googleapis/java-pubsub/blob/v1.137.0/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/stub/HttpJsonSubscriberStub.java#L1223-L1227) throws `UnsupportedOperationException`.
3. The lack of supporting those streaming types is confirmed in gapic generator code, specifically within [Model](https://github.com/googleapis/sdk-platform-java/blob/v2.53.0/gapic-generator-java/src/main/java/com/google/api/generator/gapic/model/Method.java#L121-L130). This restriction is applied in [HttpJsonServiceStubClassComposer](https://github.com/googleapis/sdk-platform-java/blob/v2.53.0/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/rest/HttpJsonServiceStubClassComposer.java#L1334-L1357) that generates the REST stub codes.

blakeli0:
Yes, you are right that unary callables already applied a similar fix, client streaming and bidi streaming are not supported in httpjson, they were poor examples. 
What I'm trying to get to is to come up with a generic problem, that is the current implementation `can not serialize a response that contains an Any message` (I _think_ even unary calls would have similar problems), because a [TypeRegistry](https://protobuf.dev/reference/java/api-docs/com/google/protobuf/TypeRegistry.html) with all the possible Message types is required. [Similar problems](https://stackoverflow.com/questions/57634281/google-protobuf-missing-descriptor-for-type-registry) have been encountered by others. In the Spanner case, BatchWriteResponse contains a [google.rpc.Status](https://github.com/googleapis/java-spanner/blob/6f90d847979a8ac35fc5ee03136a46f3bcda5128/proto-google-cloud-spanner-v1/src/main/proto/google/spanner/v1/spanner.proto#L1384), which contains [a list of Any messages](https://github.com/googleapis/googleapis/blob/2a8bb824f639c5e3b9567ea95a4e905d313fa6cf/google/rpc/status.proto#L48) that surfaced this problem. Even if we provide a default empty TypeRegistry, it still wouldn't work because we don't know an exhaustive list of possible types in this `Any` field. 

As I mentioned in https://github.com/googleapis/java-spanner/issues/3640#issuecomment-2672410302, I think the best way to move forward is to create a dedicated issue for it, we can discuss potential solutions and priority there.

PR #3650, opened 2025-02-20
ci: Use Linkage Checker to test Binary Compatibility

lqiu96:
Create a Binary Compatibility Check for Protobuf versions for all downstream libraries. Binary compatiblity is done via the [linkage checker tool](https://github.com/GoogleCloudPlatform/cloud-opensource-java/tree/master/dependencies).


.kokoro/nightly/downstream-protobuf-binary-compatibility.sh
```diff
@@ -0,0 +1,107 @@
...
+    build_artifact_list monorepo_handwritten_libraries
+  else
+    build_artifact_list downstream_handwritten_libraries
+  fi
+
+  # Linkage Checker /dependencies
+  popd
+
+  echo "Artifact List: ${artifact_list}"
+  # Only run Linkage Checker if the repo has any relevant artifacts to test for
```
blakeli0, line 97: I guess this is not a realistic scenario, it is in case someone passes in an invalid repo name?

.kokoro/nightly/downstream-protobuf-binary-compatibility.sh
```diff
@@ -0,0 +1,107 @@
...
+monorepo_handwritten_libraries["google-cloud-vertexai"]="com.google.cloud"
+monorepo_handwritten_libraries["google-cloud-resourcemanager"]="com.google.cloud"
+
+# 2. These are the mappings of all the downstream handwritten libraries' artifacts
+declare -A downstream_handwritten_libraries
+downstream_handwritten_libraries["google-cloud"]="com.google.cloud"
+
+# Builds a string output to `artifact_list`. It contains a comma separate list of Maven GAV coordinates. Parses
+# the `versions.txt` file by searching for the matching artifact_id_prefix to get the corresponding version.
+function build_artifact_list() {
```
blakeli0, line 46: For future enhancements: we could extract this function to a separate workflow, so that the artifact list can be passed to the binary check in a matrix, which can then be run in parallel. Similar to how we generate [Apiary libraries](https://github.com/googleapis/google-api-java-client-services/blob/main/.github/workflows/codegen.yaml#L9-L25).

.github/workflows/downstream_protobuf_compatibility_check_nightly.yaml
```diff
@@ -43,12 +43,17 @@ jobs:
         # which values to use and would resolve to ''.
         protobuf-version: ${{ fromJSON(format('[{0}]', inputs.protobuf_runtime_versions || '"3.25.5","4.28.3"')) }}
     steps:
-      - uses: actions/checkout@v4
+      - name: Checkout sdk-platform-java repo
+        uses: actions/checkout@v4
       - uses: actions/setup-java@v4
         with:
-          java-version: 17
+          # Use Java 11 for this as Linkage Checker is only compatible with Java 11 or below
```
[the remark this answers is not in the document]
lqiu96: Not sure yet. I suspect there may be older incompatible dependency versions, but there isn't much info given in the error message. https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/2395

.kokoro/nightly/downstream-protobuf-binary-compatibility.sh
```diff
@@ -0,0 +1,107 @@
...
+if [ -z "${PROTOBUF_RUNTIME_VERSION}" ]; then
+  echo "PROTOBUF_RUNTIME_VERSION must be set to run downstream-protobuf-binary-compatibility.sh"
+  echo "Expects a single Protobuf-Java runtime version i.e. PROTOBUF_RUNTIME_VERSION=\"4.28.3\""
+  exit 1
+fi
+
+# Create two mappings of possible API names (Key: Maven Artifact ID Prefix, Value: Maven Group ID)
+# for the libraries that should be tested.
+# 1. These are special handwritten libraries in google-cloud-java that should be tested
+declare -A monorepo_handwritten_libraries
```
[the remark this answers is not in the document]
lqiu96: Adding Translate. IIRC, DNS is apiary based (or wraps an Apiary library)

.kokoro/nightly/downstream-protobuf-binary-compatibility.sh
```diff
@@ -0,0 +1,107 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eo pipefail
+
+# Comma-delimited list of repos to test with the local java-shared-dependencies
```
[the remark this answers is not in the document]
lqiu96: Yeah, I can just remove this comment. It's copied from the source script

.kokoro/nightly/downstream-protobuf-binary-compatibility.sh
```diff
@@ -0,0 +1,107 @@
...
+downstream_handwritten_libraries["google-cloud"]="com.google.cloud"
+
+# Builds a string output to `artifact_list`. It contains a comma separate list of Maven GAV coordinates. Parses
+# the `versions.txt` file by searching for the matching artifact_id_prefix to get the corresponding version.
+function build_artifact_list() {
+  local -n api_maven_mapping=$1
+  for artifact_id_prefix in "${!api_maven_mapping[@]}"; do
+    group_id="${api_maven_mapping[${artifact_id_prefix}]}"
+
+    # Match all artifacts that start with the artifact_id_prefix to exclude any proto and grpc modules.
```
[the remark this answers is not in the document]
lqiu96: Sure, I can add grpc-* to be tested as well

.kokoro/nightly/downstream-protobuf-binary-compatibility.sh
lqiu96: This function only builds a string of packages for the linkage checker to test again. The artifactList corresponds to modules found in a repo and every repo is already built using a matrix. 

ie.. bigtable -> artifactList: google-cloud-bigtable
google-cloud-java -> artifactList: translate,resourcemanager,vertexai

blakeli0: Maybe I'm misunderstanding something, I thought there is still a [for loop](https://github.com/googleapis/sdk-platform-java/pull/3650/files#diff-561e09c5712d2899ae73c0d15235cfa6b1616cb571b3010e91b465c8d6c144a4R78) that runs each repo one by one?
Maybe I was not clear, when I mentioned "we could extract this function to a separate workflow", I meant that "we could extract the functionality of building a list of artifact for all repos to a separate workflow". 

i.e.  
```
google-cloud-java, java-bigtable, java-bigquery, java-bigquerystorage etc.
```
 -> 
```
google-cloud-bigtable, translate,resourcemanager,vertexai etc.
```

Then we put all of the artifacts to a Github matrix so that they can run in parallel.

.github/workflows/downstream_protobuf_compatibility_check_nightly.yaml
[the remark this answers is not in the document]
blakeli0: This is a slight concern when we drop Java 11, maybe it's worthwhile to at least put the issue into our backlog.

.kokoro/nightly/downstream-protobuf-binary-compatibility.sh
blakeli0: Is there a scenario that could cause "no artifacts found"?

.kokoro/nightly/downstream-protobuf-binary-compatibility.sh
lqiu96: This job is already running in parallel and each downstream repo has it's own job in the matrix. The for loop is only building a list of source artifacts for the linkage checker to use since each repo can have multiple relevant artifacts to test

.kokoro/nightly/downstream-protobuf-binary-compatibility.sh
lqiu96: Currently, no. But adding new downstream tests for repos relies on `versions.txt`'s accuracy. We have few hardcoded values in the script that always assume cloud/ google-cloud/ or com.google.cloud.

Adding something like auth to be tested would lead to `no artifacts found` as it no longer follows `google-cloud` prefix (rather google-auth).

I think it would be harder to figure out the cause if the error message from linkage checker is `Exception in thread "main" java.lang.IllegalArgumentException: Bad artifact coordinates`, rather than `Unable to find any matching artifacts to test in ...`

.kokoro/nightly/downstream-protobuf-binary-compatibility.sh
blakeli0: Discussed offline that we are going to maintain a hardcode list of `repo -> list of artifacts` mapping. This makes the shell script much easier to maintain.

blakeli0 (approved):

PR #3651, opened 2025-02-22
fix: S2A gRPC flow creates ComputeEngineCredentials via newBuilder.

rmehta19:
@rockspore pointed out that the credential should be created from scratch because when using [toBuilder](https://github.com/googleapis/google-auth-library-java/blob/main/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java#L648) the underlying [access token is copied](https://github.com/googleapis/google-auth-library-java/blob/37d228410e99799e4a7be8650fe472ea712c9b4d/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java#L657). 

This was confirmed to be a bug with local testing which:
- deployed a GAE app, the app performs the below two actions sequentially
- create Google API client ( `allowedHardBoundAccessTokens` empty in GrpcProvider) and then ping the API, logs show the bearer token is used, obtained from making call to MDS
- create a Google API client  ( `allowedHardBoundAccessTokens` contains `MTLS_S2A` in GrpcProvider) and then ping the API, logs show the bearer token is used. A call to MDS is **not** made. 

This is likely because the credential and channel have different lifetimes.

### DIFF
--- a/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
+++ b/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
@@ -1208,8 +1208,8 @@ public final class InstantiatingGrpcChannelProvider implements TransportChannelP
               ComputeEngineCredentials.newBuilder()
                   .setScopes(credsBuilder.getScopes())
                   .setHttpTransportFactory(credsBuilder.getHttpTransportFactory())
-                  .setGoogleAuthTransport(ComputeEngineCredentials.GoogleAuthTransport.MTLS)
-                  .setBindingEnforcement(ComputeEngineCredentials.BindingEnforcement.ON)
+                  .setGoogleAuthTransport(googleAuthTransport)
+                  .setBindingEnforcement(bindingEnforcement)
                   .build());
     }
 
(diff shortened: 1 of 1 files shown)


gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -1199,14 +1199,18 @@ boolean isDirectPathBoundTokenEnabled() {
     CallCredentials createHardBoundTokensCallCredentials(
         ComputeEngineCredentials.GoogleAuthTransport googleAuthTransport,
         ComputeEngineCredentials.BindingEnforcement bindingEnforcement) {
+      ComputeEngineCredentials.Builder credsBuilder =
+          ((ComputeEngineCredentials) credentials).toBuilder();
       // We only set scopes and HTTP transport factory from the original credentials because
```
rockspore, line 1204: For future readers, could you explain briefly why we create it from a newBuilder()?

rmehta19: Thanks! Done.

rmehta19:
@lqiu96 , please review, thanks!

lqiu96 (approved):

lqiu96:
Oh, can you also rename the title to follow conventional commits?

> fix: S2A gRPC flow creates ComputeEngineCredentials via newBuilder.

or some similar

rmehta19:
Thanks @lqiu96 . I also updated the branch with upstream main, I think you will have to rerun gcbrun?

Issue #3360, opened 2024-11-11
gax : GrpcSpannerStub create(ClientContext clientContext) not using `SpannerStubSettings` configuration


surbhigarg92:
Regarding the bug fix for [canUseDirectPath returning false when DirectPath is enabled](https://github.com/googleapis/sdk-platform-java/pull/3170), the recommendation was to pass `clientContext` in the `GrpcSpannerStub.create` instead of `SpannerStubSetting`

This approach works for standard requests, but it fails in cases where custom timeouts are applied. 

Further investigation showed that passing `clientContext` to `GrpcSpannerStub.create` results in a new `SpannerStubSettings.newBuilder().build()` instance being created rather than using the SpannerStubSettings from clientContext . See code [reference](https://github.com/googleapis/java-spanner/blob/main/google-cloud-spanner/src/main/java/com/google/cloud/spanner/v1/stub/GrpcSpannerStub.java#L247). 

Using the protected [constructor](https://github.com/googleapis/java-spanner/blob/main/google-cloud-spanner/src/main/java/com/google/cloud/spanner/v1/stub/GrpcSpannerStub.java#L260)  that accepts both stubSettings and clientContext directly resolves the issue. Refer [PR](https://github.com/googleapis/java-spanner/pull/3380) 

**Expected fix:** 
Expose a new public static method for accepting both stubSettings and clientContext

blakeli0:
Spanner chose a different approach of extending the GrpcSpannerStub. Closing as not planned.

PR #3670, opened 2025-03-01
fix: remove call credentials from call options if DirectPath

rockspore:
This PR eliminates the issue where call credentials get attached twice to a RPC in DirectPath cases. Particularly, when user credentials get used, the problem causes the RPC to fail due to the duplication of the quota project ID (internal-only context: [b/364288002](http://b/364288002)).

The approach is to strip the credentials in the callOptions returned by the `GrpcCallContext` if the `TransportChannel` is DirectPath. The side effect is that users won't be able to configure call credentials via the `ApiCallContext` if DirectPath is used.

We think this is acceptable because:

1. Users can still configure the credentials via `TransportChannelProvider.withCredentials()`. At a higher level, this is done by configuring the `CredentialsProvider` in the `StubSettings` or the `ServiceOptions`.
2. As of now, DirectPath has its own special authentication flow, in which the service account attached to the GCE VM or GKE Pod will be used. Although in some special cases, the call credentials will be used to authenticate the client's identity, the peculiar nature of DirectPath should justify us limiting the flexibility of how call credentials can be configured in this case.

## Tested DirectPath using Spanner 
Headers sent
```
[:authority: spanner.googleapis.com, :path: /google.spanner.v1.Spanner/BatchCreateSessions, :method: POST, :scheme: https, content-type: application/grpc, te: trailers, user-agent: spanner-java/6.86.0 grpc-java-netty/1.69.0, ..., grpc-accept-encoding: gzip, authorization: Bearer ya29.****, ..., authorization: Bearer 1234, grpc-timeout: 56962080u]
```
Bearer token is sent twice (first `ya29.***` value is valid and second `1234` is invalid). The second one was attached by customizing the ApiCallContext to send an invalid CallCredentials as part of the CallOptions. The call still succeeded as the first Bearer token in the Metadata is used.


rockspore:
@lqiu96 I've verified this would fix the DirectPath case. No idea about the obscure failure in java-firestore yet. Will add tests next week. I think this approach is promising.

lqiu96:
I think this does work for DirectPath. 

However, doesn't this issue also exist for S2A where both tokens are being passed? Also, this would override any user set CallOption value.

rockspore:
> However, doesn't this issue also exist for S2A where both tokens are being passed?

I was wondering if we could introduce a similar API as `isDirectPath()` for `TransportChannel` for S2A w/ bound tokens so that the same approach could be used there. 

> Also, this would override any user set CallOption value.

This is by far the simplest approach and doesn't even need to touch how channels are constructed in any other non-DirectPath case (S2A with bound token will also need to be handled later). And somewhat biased by this, I'm hoping we could be a bit opinionated in the DirectPath case to disregard users' call credentials in the call options, given we clearly document this behavior. Note in my other PR, we are overriding the call option value as well.

lqiu96:
> I was wondering if we could introduce a similar API as `isDirectPath()` for `TransportChannel` for S2A w/ bound tokens so that the same approach could be used there.

I would like this. I'm not able to easily map a way to determine this since I think it would require actually trying to create the CallCredentials and checking if they could be created. I guess that is a possibility we can also try. I was exploring dynamically setting the call options creds in https://github.com/googleapis/sdk-platform-java/pull/3671

> I'm hoping we could be a bit opinionated in the DirectPath case to disregard users' call credentials in the call options, given we clearly document this behavior.

I would like to be opinionated on this too. However, can you remind me the behavior if something like SACreds were used instead of UserCreds, would DP would work? Would it be possible to override the calloptions creds using DP?

> S2A with bound token will also need to be handled later

I'm not against this, though I would like to try and see if we can figure a generic solution to tackle sending creds twice, rather than adding one-off fixes for specific flows.

lqiu96:
>  ChannelCredentials are not required to always carry the CallCredentials

I see, so in the case of Mtls (non S2A) and just normal TLS, then the expectation is that CallCredentials should be attached via CallOptions? In that case, I think my assumption that always using ChannelCredentials instead of CallOptions is wrong.

> We only care about whether CallCredentials have been attached to the channel here.

Ok I see, I think I'm on the same page now. In this case only DP and MTLS_S2A would be the cases as you've mentioned above.

> we document that DirectPath does not take in any CallCredentials from the ApiCallContext and we should be good.

Yes I think if we take the opinionated route this what what we should do. The example that I was thinking above was that a user connected via establish a DP connection with some SA cred and overrides with a different SA cred (something like what Spanner has open: https://github.com/googleapis/java-spanner/blob/7a8a29be40258294cafd13b1df7df5ea349a675d/google-cloud-spanner/src/main/java/com/google/cloud/spanner/spi/v1/GapicSpannerRpc.java#L2037). Is this a behavior that works?

> I wonder if you are open to adding a more generic method to TransportChannel

Yeah, this is what the PR above was trying to explore. Adding a new method that determined if the CallCredentials was attached to the channel. When a GAPIC client is initialized, it will try to determine if this was a user set value or set by the client and it will ignore the client set value.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
```diff
@@ -535,7 +563,9 @@ public Channel getChannel() {
 
   /** The {@link CallOptions} set on this context. */
   public CallOptions getCallOptions() {
-    return callOptions;
+    if (!isDirectPath) return callOptions;
+    // Remove the CallCredentials attached to the callOptions if it's DirectPath.
+    return callOptions.withCallCredentials(null);
```
lqiu96, line 568: Can this logic reside inside the constructor? I think it would be better for the getter to just return value.

nit: Can the comment be updated to reflect the `why`. Perhaps something like (probably needs better wording): `CallCredentials is stripped from CallOptions because CallCredentials are attached to ChannelCredentials in DirectPath flows. Adding it again would duplicate the headers.`

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
```diff
@@ -602,7 +632,9 @@ public GrpcCallContext withChannel(Channel newChannel) {
         options,
         retrySettings,
         retryableCodes,
-        endpointContext);
+        endpointContext,
+        // Defaults to false again since we cannot tell whether the channel is DirectPath.
+        false);
```
lqiu96, line 637: Oh shoot, this is a valid public API. I don't see any usages of this in the googleapis org, but can't say for certain this isn't used.

Can we add a warning to the javadocs that passing in a DirectPath channel and specifying CallCreds may lead to issues?

lqiu96: > Can we add a warning to the javadocs that passing in a DirectPath channel and specifying CallCreds may lead to issues?

Actually, nvm on this. I think we confirmed this is a non-issue as the servers only use the channelcredential.

lqiu96:
I think the last thing is to update the GrpcCallContext's `withCredentials` and `withCallOptions` javadocs.

Thoughts on a small blurb like (re-phrase/ update): For certain flows like DirectPath, the channel is already created with a CallCredentials. Passing in an additional CallCredentials here will not override the ChannelCredential's CallCredential and may cause issues as the CallCredentials is duplicated. These flows will attempt to automatically strip the CallOption's CallCreds.

rmehta19:
(Not specific to this PR, but related to fixing this problem for MTLS_S2A)

>>I was wondering if we could introduce a similar API as isDirectPath() for TransportChannel for S2A w/ bound tokens so that the same approach could be used there.

> I would like this. I'm not able to easily map a way to determine this since I think it would require actually trying to create the CallCredentials and checking if they could be created.

I think it may be more than this. In order to set `isMtlsS2A()` (API similar to [`isDirectPath`](https://github.com/googleapis/sdk-platform-java/blob/1eda55fc5e4cecedba3f8c163db0ce259d180a66/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcTransportChannel.java#L63)), we'd need to check:
- could [CallCredentials](https://github.com/googleapis/sdk-platform-java/blob/1eda55fc5e4cecedba3f8c163db0ce259d180a66/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L196) be created?
- could [S2A channel credentials](https://github.com/googleapis/sdk-platform-java/blob/1eda55fc5e4cecedba3f8c163db0ce259d180a66/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L659-L662) be created?

This is due to the fact that for DirectPath: if you **can** use DirectPath then you **will** use DirectPath. However, this is not true for S2A, if you [**can**](https://github.com/googleapis/sdk-platform-java/blob/1eda55fc5e4cecedba3f8c163db0ce259d180a66/gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java#L318) use S2A, you **may** end up using S2A (you could end up using DirectPath since DP supersedes S2A or you could end up falling back to TLS if you fail to create S2A channel creds).

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
```diff
@@ -143,7 +146,8 @@ private GrpcCallContext(
       ApiCallContextOptions options,
       @Nullable RetrySettings retrySettings,
       @Nullable Set<StatusCode.Code> retryableCodes,
-      @Nullable EndpointContext endpointContext) {
+      @Nullable EndpointContext endpointContext,
+      boolean isDirectPath) {
```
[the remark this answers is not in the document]
rockspore: Got it. So I will keep this boolean name as is. We can follow up with @rmehta19's comment later.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
rockspore: Done. I also modify the merge logic to respect the isDirectPath of the coming context. Also several cases I take care of here:

1. If `withChannel` is subsequently called, we attach a CallCreds from the creds back to calloptions and revert `isDirectPath` to false.
2. If `withCallOptions` is subsequently called, we strip the CallCreds from it if DirectPath.
3. If `withCredentials` is subsequently called, we attach its corresponding CallCreds to the calloptions only if non-DirectPath

Please help check if I missed any corner case.

rockspore:
For posterity, we confirmed that `CallCredentials` attached to the `ChannelCredentials` will appear before the potentially same header from the `CallCredentials` in the `CallOptions` (with the current gRPC implementation) and the server only uses the first one. Example header list:

[:authority: [spanner.googleapis.com](http://spanner.googleapis.com/), :path: /google.spanner.v1.Spanner/BatchCreateSessions, :method: POST, :scheme: https, content-type: application/grpc, te: trailers, user-agent: spanner-java/6.86.0 grpc-java-netty/1.69.0, ..., grpc-accept-encoding: gzip, **authorization: Bearer ya29.******, ..., authorization: Bearer 1234, grpc-timeout: 56962080u]

The garbage token in the second authorization header didn't cause the call to fail the authn/z.

In general, we shouldn't rely on this gRPC implementation detail since it could in theory change the appending order and break everything. This PR will fix the duplication header issue so we are good.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
```diff
@@ -210,7 +216,20 @@ public GrpcCallContext withTransportChannel(TransportChannel inputChannel) {
           "Expected GrpcTransportChannel, got " + inputChannel.getClass().getName());
     }
     GrpcTransportChannel transportChannel = (GrpcTransportChannel) inputChannel;
-    return withChannel(transportChannel.getChannel());
+    return new GrpcCallContext(
+        transportChannel.getChannel(),
+        credentials,
+        transportChannel.isDirectPath() ? callOptions.withCallCredentials(null) : callOptions,
```
lqiu96, line 222: Sorry, I meant can we do this logic/ check inside the private constructor? This should only pass the calloptions.

Constructor contains the:
`this.callOptions =  isDirectPath() ? calloptions.withCallCredentials(null) : callOptions` logic

rockspore: Ah this makes so much sense! Sorry I don't know what I was thinking. Changed.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
```diff
@@ -96,6 +96,7 @@ public final class GrpcCallContext implements ApiCallContext {
   private final ImmutableMap<String, List<String>> extraHeaders;
   private final ApiCallContextOptions options;
   private final EndpointContext endpointContext;
+  private final boolean isDirectPath;
```
blakeli0, line 99: There was [an issue](https://github.com/googleapis/sdk-platform-java/issues/3360) reported that `isDirectPath` may not be correct populated in some cases. @lqiu96 is looking into it.

rockspore: Thanks for the info. I thought that referred to `canUseDirectPath()` which could be invoked multiple times before and after the credentials was set so might be no accurate, but by the time the ClientContext got this boolean from the channel, it should already be final.

lqiu96: Talked with @surbhigarg92 regarding this to get some more information about the issue. This issue from Spanner's POV is not so much that the `canUseDirectPath` value that is client is initialized with and used by the TransportChannel is incorrect, it's that Spanner doesn't have a way to get the `canUseDirectPath` value when initializing the client for Otel.

Spanner was using `GrpcSpannerStub.create(StubSettings)` and didn't have an easy way to access to TransportChannel's fields. Spanner could use [`GrpcSpannerStub.create(ClientContext)`](https://github.com/googleapis/java-spanner/blob/42cc9616fa74c765d5716fd948dc0823df0a07a6/google-cloud-spanner/src/main/java/com/google/cloud/spanner/v1/stub/GrpcSpannerStub.java#L246-L248), but that creates the Stub with a new StubSettings.

For them, the DirectPath transportchannel was always created correctly and the value the client uses is correct. They used this workaround to be able to access the field for their use case.

lqiu96 (approved):

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
```diff
@@ -602,7 +638,9 @@ public GrpcCallContext withChannel(Channel newChannel) {
         options,
         retrySettings,
         retryableCodes,
-        endpointContext);
+        endpointContext,
+        // Defaults to false again since we cannot tell whether the channel is DirectPath.
+        false);
```
blakeli0, line 643: This could lead to a potential bug. If `GrpcCallContext` contains a direct path channel, and it is re-created with another direct path channel using this method, then we would send duplicated headers again. 
Fortunately I don't see any usage of this method at this moment, but with the introduction of `isDirectPath` in `GrpcCallContext`, we should probably mark this method as `@ObsoleteApi`, and suggest everyone to use `withTransportChannel` instead. See an example [here](https://github.com/googleapis/sdk-platform-java/blob/3857e3f6dc633a9adaf29e9770a584419e2f2214/gax-java/gax/src/main/java/com/google/api/gax/rpc/ApiCallContext.java#L70-L71).

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
```diff
@@ -456,8 +486,11 @@ public ApiCallContext merge(ApiCallContext inputCallContext) {
       newDeadline = callOptions.getDeadline();
     }
 
+    boolean newIsDirectPath = grpcCallContext.isDirectPath;
+
     CallCredentials newCallCredentials = grpcCallContext.callOptions.getCredentials();
-    if (newCallCredentials == null) {
+    // If newIsDirectPath is true, then newChannel is guanrateed to be not null.
+    if (newCallCredentials == null && !newIsDirectPath) {
```
blakeli0, line 493: Is this change required? I think even if `newCallCredentials` are set, is would be overridden later in the constructor? If that's the case, I would prefer to not couple the direct path logic with a merge logic in this method. This method is simply merging values of two `ApiCallContext`, how we use those values are the responsibility of the constructors.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
```diff
@@ -228,7 +250,8 @@ public GrpcCallContext withEndpointContext(EndpointContext endpointContext) {
         options,
         retrySettings,
         retryableCodes,
-        endpointContext);
+        endpointContext,
```
blakeli0, line 253: Its unfortunate that we have to change all the places that use the constructor. Ideally, if we had a builder for `GrpcCallContext`, the code here would be simplified to `this.toBuilder().setEndpointContext(endpointContext).build()`, and we don't have to change the code here at all.

rockspore: Agreed. But I'd prefer to leave the refactoring with a builder out of this PR since it would look cleaner. I can open an issue for it. Let me know what you think.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
rockspore: No it's not. This was some overlooked left-over before I moved the logic to strip the call credentials from the call options inside the private constructor. Cleaned it up now.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
rockspore: Thanks. I checked that this API is only used in two test classes. So it should be safe to be marked as obsolete. Followed your example.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcCallContext.java
blakeli0: SG. Yes please create a separate issue and we can put it in our backlog.

rockspore: Thanks. Created #3681

blakeli0 (approved):

rockspore:
Thanks for all the reviews. No rush but just in case, please help merge it when you see fit since I don't have the permission.

PR #3673, opened 2025-03-04
fix: plumb mtls endpoint to TransportChannelProvider

rmehta19:
This PR plumbs the MTLS endpoint (separately from the resolved endpoint) from the `EndpointContext` to the `InstantiatingGrpcChannelProvider`.

Why not just set the MTLS endpoint in `EndpointContext` if S2A can be used?
- Although we can decide whether or not to try to use S2A in `EndpointContext`, we could fail to use S2A in `InstantiatingGrpcChannelProvider` (if autoconfig doesn't return an address), in which case we fall back to using a TLS connection
- DirectPath supersedes S2A in `InstantiatingGrpcChannelProvider`, and the decision to use DirectPath is made in `InstantiatingGrpcChannelProvider`, not `EndpointContext`. We may decide to use S2A in `EndpointContext`, but when we go to create the channel in `InstantiatingGrpcChannelProvider`, we may find that we should be using DirectPath, in which case we need to use the non-MTLS endpoint


gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -258,6 +260,11 @@ public boolean needsEndpoint() {
     return endpoint == null;
   }
 
+  @Override
+  public boolean needsMtlsEndpoint() {
+    return mtlsEndpoint == null;
```
rmehta19, line 265: If we want, I think we could make this:
```
return (mtlsEndpoint == null) && (useS2A == true)
```

That way, we can be sure we are only setting the `mtlsEndpoint` when we know:
- it is non-null
- it is non-empty
- it is needed by S2A

Note that the first two are checked in `shouldUseS2A` in EndpointContext.

rmehta19:
@lqiu96 , please review, thanks!

rmehta19:
Friendly ping @lqiu96 . Thanks!

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
lqiu96: I think what you have is fine.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -272,6 +279,21 @@ public TransportChannelProvider withEndpoint(String endpoint) {
...
+   *
+   * @param mtlsEndpoint The mtTLS endpoint to connect to
+   * @return A new {@link InstantiatingGrpcChannelProvider} with the specified mTLS endpoint
+   *     configured
+   */
+  @Override
+  public TransportChannelProvider withMtlsEndpoint(String mtlsEndpoint) {
+    validateEndpoint(mtlsEndpoint);
+    return toBuilder().setMtlsEndpoint(mtlsEndpoint).build();
+  }
```
lqiu96, line 295: Would you be able to add `@InternalApi` to all the new public methods? We have a few ways to configure endpoints (i.e. ClientSettings and TransportChannelProvider for endpoint and led us to have to [account for it in EndpointContext](https://github.com/googleapis/sdk-platform-java/blob/5192426dc1010b97c7bead505712cf995e9b1e31/gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java#L198-L207)). If we add these ways for users to configure the mtls value, I think we'll need to account for it in EndpointContext as well. 

Given that we can't make this package-private scope, we should let users know that they really shouldn't be tinkerining with these values since they're really used to help S2A (one flow for now).

I think a small note in the comment as well warning users `This public method is used by Gax to help configure S2A fallback and not intended for users` or something of that variant.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -666,7 +688,7 @@ private ManagedChannel createSingleChannel() throws IOException {
             channelCredentials =
                 CompositeChannelCredentials.create(channelCredentials, mtlsS2ACallCredentials);
           }
-          builder = Grpc.newChannelBuilder(endpoint, channelCredentials);
+          builder = Grpc.newChannelBuilder(mtlsEndpoint, channelCredentials);
```
lqiu96, line 691: nit: Small comment that the fallback uses mTLS not non-MTLS because of reason X

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rmehta19: Done in [7e36d1a](https://github.com/googleapis/sdk-platform-java/pull/3673/commits/7e36d1ac8667c5ccbe44659faf23bec94f27a027). Would you prefer making the change in EndpointContext in this PR, or a follow up?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rmehta19: Done.

rmehta19:
Thanks for the review @lqiu96 !

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
lqiu96: My thoughts are that we just leave EndpointContext as-is since we have publicly noted that these public methods have been marked with `@InternalApi` and shouldn't be set/customized by the user at all. If users do set these values via these new methods, they shouldn't expect support (as this is used to help to support S2A).

The [existing way](https://github.com/googleapis/sdk-platform-java/blob/3d06ab7a9ce74a37bf0bfffca3c52a3428e32ecc/gax-java/gax/src/main/java/com/google/api/gax/rpc/StubSettings.java#L535-L538) of setting the mtlsEndpoint in StubSettings is supported

lqiu96:
@rmehta19 FYI I think there are lint issues and you need to run `mvn fmt:format`

rmehta19:
Thanks @lqiu96 ! formatted.

rmehta19:
@lqiu96 , resolved merge conflict, could you run gcbrun again? Thanks!

lqiu96 (approved):
LGTM. This follows the approach that we had discussed in the previous S2A PR.

rmehta19:
Thanks @lqiu96 ! I see all the checks have passed, would you be able to merge once you get a chance?

lqiu96:
Adding @blakeli0 to take a quick glance as well in case there is anything missed.

rmehta19:
friendly ping, @blakeli0. Thanks!

gax-java/gax/src/main/java/com/google/api/gax/rpc/TransportChannelProvider.java
```diff
@@ -97,6 +97,20 @@ public interface TransportChannelProvider {
    */
   TransportChannelProvider withEndpoint(String endpoint);
 
+  /** True for gRPC transport provider which has no mtlsEndpoint set. */
+  default boolean needsMtlsEndpoint() {
```
blakeli0, line 101: I know we tried to avoid introducing these new public interfaces by encapsulating the logic to determine if we should use S2A in EndpointContext. Now that we need to have access to mtlsEndpoint in TransportChannelProvider  anyway, maybe we can move [shouldUseS2A](https://github.com/googleapis/sdk-platform-java/blob/f6bdadebb4286c53cbce725dac07bfd2198e4c18/gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java#L318-L344) to InstantiatingGrpcChannelProvider.java? And remove `withUseS2A` from this interface? This could be separate PR.

rmehta19: Agreed that by plumbing the MTLS endpoint separately, we no longer need the `shouldUseS2A` logic in EndpointContext. I'll take care of this in a followup PR.

blakeli0 (approved):

gax-java/gax/src/main/java/com/google/api/gax/rpc/TransportChannelProvider.java
rmehta19: @blakeli0 I started working on this and I realized that we should keep [`shouldUseS2A`](https://github.com/googleapis/sdk-platform-java/blob/f6bdadebb4286c53cbce725dac07bfd2198e4c18/gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java#L318-L344) in EndpointContext. The reason being is that we want to skip S2A if using GDCH, or if the client sets a custom endpoint. This info is only available in EndpointContext. 

Given this, we probably want to keep things as is. WDYT?

suztomo:
@rmehta19  There may be test failures related to this. Can you check? https://github.com/googleapis/java-bigtable/pull/2560#issuecomment-2816088021

PR #3679, opened 2025-03-07
ci: enforce Build with Airlock check

suztomo:
b/356854847

### DIFF
--- a/.github/sync-repo-settings.yaml
+++ b/.github/sync-repo-settings.yaml
@@ -37,6 +37,7 @@ branchProtectionRules:
       - should-run-library-generation-tests
       - showcase (11)
       - showcase (17)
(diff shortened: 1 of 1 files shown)


suztomo:
> Caused by: org.eclipse.aether.transfer.ArtifactNotFoundException: com.google.http-client:google-http-client-bom:pom:1.46.3 was not found in artifactregistry://us-maven.pkg.dev/artifact-foundry-prod/maven-3p-trusted during a previous attempt. This failure was cached in the local repository and resolution is not reattempted until the update interval of mirror-of-all has elapsed or updates are forced


https://github.com/googleapis/google-http-java-client/releases/tag/v1.46.3 was last week. Didn't it go to the 1P repo?

PR #3698, opened 2025-03-13
chore: restore showcase

JoeWang1127:
In this PR:
- Restore logging tests in showcase which was removed in #3568.
- Remove redundant samples in showcase


JoeWang1127:
slf4j 1.x failed:
```
Error:  /home/runner/work/sdk-platform-java/sdk-platform-java/java-showcase/gapic-showcase/src/test/java/com/google/showcase/v1beta1/it/logging/ITLogging1x.java:[59,45] cannot access org.slf4j.spi.LoggingEventAware
  class file for org.slf4j.spi.LoggingEventAware not found
```

java-showcase/gapic-showcase/pom.xml
```diff
@@ -281,7 +281,7 @@
         <dependency>
           <groupId>ch.qos.logback</groupId>
           <artifactId>logback-classic</artifactId>
-          <version>1.3.15</version>
+          <version>1.2.13</version>
```
JoeWang1127, line 284: Restore the version to fix compilation error.

diegomarquezp (approved):

java-showcase/gapic-showcase/pom.xml
zhumin8: curious what was the compilation error? I did not came across when changing to 1.3.x [here](https://github.com/googleapis/sdk-platform-java/pull/3686). 
Also, can you keep `logback-classic` and `logback-core` versions consistent. Perhaps add a property for them

zhumin8 (commented):
this could be a separate pr: on maintaining the testing versions for logback in gapic-showcase/pom.xml, I was thinking to setup renovate as in [this pr](https://github.com/googleapis/sdk-platform-java/pull/3703). Any better ideas?

java-showcase/gapic-showcase/pom.xml
JoeWang1127: The compilation error is https://github.com/googleapis/sdk-platform-java/pull/3698#issuecomment-2722682530.

> I did not came across when changing to 1.3.x [here](https://github.com/googleapis/sdk-platform-java/pull/3686).

I think the showcase test got removed before merging #3686.

JoeWang1127:
> this could be a separate pr: on maintaining the testing versions for logback in gapic-showcase/pom.xml, I was thinking to setup renovate as in [this pr](https://github.com/googleapis/sdk-platform-java/pull/3703). Any better ideas?

Maybe we can find a way to lock the logback version in slf4j 1.x testing but configure renovate to update others.

java-showcase/gapic-showcase/pom.xml
zhumin8: hmm. interesting, I might look into this error later. 
For now, I agree lets use 1.2.x versions for both

zhumin8:
> Maybe we can find a way to lock the logback version in slf4j 1.x testing but configure renovate to update others.

This is what I would prefer, but from quick search , I did not find renovate have feature to mark ignore.

zhumin8 (approved):

PR #3700, opened 2025-03-13
chore: allow generation-input as folder that supplies config and vers…

zhumin8:
…ions file.

This change adds `--generation-input` option to generation command, only intend to be used when `--generation-config-path` is not provided. If none of these 2 provided, then default behavior does not change.

Also added a clean up to delete the temp output folder created in generation process. This should only be an improvement to local dev workflow.

cc. @ldetmer


hermetic_build/library_generation/tests/cli/entry_point_unit_tests.py
```diff
@@ -122,6 +139,7 @@ def test_generate_non_monorepo_with_library_names_full_generation(
         # does special handling when a method is annotated with @main.command()
         generate_impl(
             generation_config_path=config_path,
+            generation_input=None,
```
zhumin8, line 142: this change is needed for the unit test, but not for real usage, as when not specified "click" would pass it in as None.

diegomarquezp (approved):
LGTM

PR #3706, opened 2025-03-17
chore: create logging module structure

JoeWang1127:
In this PR:
- Setup logging module structure.


pom.xml
```diff
@@ -24,6 +24,7 @@
     <module>gapic-generator-java-bom</module>
     <module>java-shared-dependencies</module>
     <module>sdk-platform-java-config</module>
+    <module>java-sdk-logging</module>
```
zhumin8, line 27: Is this included in releases?

JoeWang1127: No, the module is skipped during normal release.

java-sdk-logging/java-sdk-logging-bom/pom.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>com.google.api</groupId>
+    <artifactId>java-sdk-logging-parent</artifactId>
+    <version>0.1.0-SNAPSHOT</version>
```
zhumin8, line 9: I wonder if we want to keep the version name the same as the main project?

pom.xml
zhumin8: Great. Do you mind point me to where the normal release is setup?

java-sdk-logging/java-sdk-logging-bom/pom.xml
```diff
@@ -0,0 +1,41 @@
...
+    <relativePath>../pom.xml</relativePath>
+  </parent>
+  <packaging>pom</packaging>
+
+  <artifactId>java-sdk-logging-bom</artifactId>
```
zhumin8, line 14: Wonder if better be more specific like `java-sdk-logging-extensions-bom`? 
`java-sdk-logging-bom` might be a bit confusing as it sounds like containing all logging component, and I believe since we do not publish this regularly, this is not intended to host the core client debug logging logics.

JoeWang1127: Done.

java-sdk-logging/java-sdk-logging-bom/pom.xml
blakeli0, line 1: Why do we need another bom in this module? I don't think customers would need to import both logback and log4j, so I'm not sure a bom is needed.

java-sdk-logging/log4j-extension/pom.xml
blakeli0, line 1: It's better to use `log4j2` instead of `log4j` because log4j still being used.

JoeWang1127: Done.

java-sdk-logging/java-sdk-logging-bom/pom.xml
JoeWang1127: OK, `bom` may not be necessary in this case. I'll remove it.

pom.xml
JoeWang1127: `maven-deploy-plugin` in these `pom.xml`s are skipped.

This is the same setup in the root `pom.xml`.

java-sdk-logging/pom.xml
```diff
@@ -0,0 +1,54 @@
...
+    <version>2.55.1</version><!-- {x-version-update:gapic-generator-java:current} -->
+    <relativePath>../gapic-generator-java-pom-parent</relativePath>
+  </parent>
+  <modules>
+    <module>log4j2-extension</module>
+    <module>logback-extension</module>
+  </modules>
+
+  <properties>
+    <mockito.version>4.11.0</mockito.version>
```
blakeli0, line 21: We already have a mockito version in Gax, maybe we can extract that to a property and reuse it here? It could be a follow up PR as well.

pom.xml
blakeli0: Does it have to be added to the root pom?

JoeWang1127: No, it doesn't need to add to root pom.

Without this line, we need to specify the pom to build the module,
```
mvn clean install -f java-sdk-logging/pom.xml
```

I'll remove this line then.

java-sdk-logging/java-sdk-logging-bom/pom.xml
zhumin8: A followup to this, if we keep this standalone with its own versions, what version should we start with? 0.1.0 seems to indicate unstable, do we want that?

JoeWang1127: I think 0.1.0 is fine now since we are still developing it.

We can change the version to 1.0.0 when we want to release the library.

java-sdk-logging/pom.xml
JoeWang1127: Done

pom.xml
blakeli0: I would prefer the tests for this module are a separate CI, since it has no dependencies on other modules at all. Even if the tests fail, it should not block the release of other modules.

java-sdk-logging/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+    </dependencies>
+  </dependencyManagement>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-deploy-plugin</artifactId>
+        <configuration>
+          <skip>true</skip>
```
blakeli0, line 45: Since it is not part of the parent pom now, do we still have to skip this plugin?

.github/workflows/java_sdk_logging.yaml
```diff
@@ -0,0 +1,55 @@
...
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-java@v4
+      with:
+        java-version: 11
+        distribution: temurin
+    - name: Install parent module
+      run: |
+        mvn install -B -ntp -pl gapic-generator-java-pom-parent
+    - name: Java Linter
```
blakeli0, line 52: In addition to linter, can we add clirr check and enforcer check?

JoeWang1127: Sure, I'll add them in a follow up PR.

java-sdk-logging/pom.xml
JoeWang1127: Done.

PR #3709, opened 2025-03-18
chore: update logging mod

JoeWang1127:
A follow up of #3706:
- Change the group id to `com.google.cloud` as it is a customer-facing library.
- Add clirr and enforcer check in CI.

### DIFF
--- a/java-sdk-logging/license-checks.xml
+++ b/java-sdk-logging/license-checks.xml
@@ -5,6 +5,6 @@
 <module name="Checker">
   <module name="RegexpHeader">
     <property name="fileExtensions" value="java"/>
-    <property name="headerFile" value="./java-sdk-logging/java.header"/>
+    <property name="headerFile" value="./java.header"/>
   </module>
 </module>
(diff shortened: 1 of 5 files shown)


java-sdk-logging/license-checks.xml
```diff
@@ -5,6 +5,6 @@
 <module name="Checker">
   <module name="RegexpHeader">
     <property name="fileExtensions" value="java"/>
-    <property name="headerFile" value="./java-sdk-logging/java.header"/>
+    <property name="headerFile" value="./java.header"/>
```
blakeli0, line 8: Probably a separate topic: why do we have a `java.header` file in each submodule? I think we can have one under root, and all submodules reference it?

blakeli0 (approved):

java-sdk-logging/license-checks.xml
JoeWang1127: Maybe a follow up PR to create a `java.header` in root.

PR #3711, opened 2025-03-18
chore(hermetic-build): infer formatter version from java-shared-config property

diegomarquezp:
This ties the google-java-format.jar version to the one [declared in java-shared-config](https://github.com/googleapis/java-shared-config/blob/a315cf154ffb2dfd77ff7ce9892f8bb8a40ec7cc/java-shared-config/pom.xml#L31)

The purpose is to later upgrade this version to avoid the old Guava dependency.

I moved the download step to ggj-build so the final image only copies artifacts from "build related" stages.

### DIFF
--- a/.cloudbuild/library_generation/library_generation.Dockerfile
+++ b/.cloudbuild/library_generation/library_generation.Dockerfile
@@ -26,7 +26,7 @@ ENV DOCKER_GAPIC_GENERATOR_VERSION="2.55.2-SNAPSHOT"
 # Download the java formatter
 RUN mvn -pl gapic-generator-java-pom-parent help:evaluate -Dexpression='google-java-format.version' -q -DforceStdout > /java-formatter-version
 RUN cat /java-formatter-version
-RUN curl -o "/google-java-format.jar" "https://maven-central.storage-download.googleapis.com/maven2/com/google/googlejavaformat/google-java-format/$(cat /java-formatter-version)/google-java-format-$(cat /java-formatter-version)-all-deps.jar"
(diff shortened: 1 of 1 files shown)


.cloudbuild/library_generation/library_generation.Dockerfile
```diff
@@ -23,6 +23,12 @@ COPY . .
 ENV DOCKER_GAPIC_GENERATOR_VERSION="2.55.2-SNAPSHOT"
 # {x-version-update-end}
 
+# Download the java formatter
+RUN mvn -pl gapic-generator-java-pom-parent help:evaluate -Dexpression='google-java-format.version' -q -DforceStdout > /java-formatter-version
+RUN cat /java-formatter-version
+RUN curl -o "/google-java-format.jar" "https://maven-central.storage-download.googleapis.com/maven2/com/google/googlejavaformat/google-java-format/$(cat /java-formatter-version)/google-java-format-$(cat /java-formatter-version)-all-deps.jar"
```
suztomo, line 29:
```suggestion
RUN V=$(cat /java-formatter-version) curl -o "/google-java-format.jar" "https://maven-central.storage-download.googleapis.com/maven2/com/google/googlejavaformat/google-java-format/${V}/google-java-format-${V}-all-deps.jar"
```

suztomo (approved):

JoeWang1127 (approved):

.cloudbuild/library_generation/library_generation.Dockerfile
diegomarquezp: Nice

diegomarquezp: Seems like the form `X=$(command1) echo "${X}"` needs `&&` or will otherwise evaluate `${X}` as an empty string.
This should be `X=$(command1) && echo "${command1}"` (I found by trial and error).

PR #3716, opened 2025-03-19
ci: Use Java 8 for Linkage Checker

lqiu96:
Manual invocation: https://github.com/googleapis/sdk-platform-java/actions/runs/14038997003

### DIFF
--- a/.github/workflows/downstream_protobuf_compatibility_check_nightly.yaml
+++ b/.github/workflows/downstream_protobuf_compatibility_check_nightly.yaml
@@ -47,8 +47,8 @@ jobs:
         uses: actions/checkout@v4
       - uses: actions/setup-java@v4
         with:
-          # Use Java 11 for this as Linkage Checker is only compatible with Java 11 or below
-          java-version: 11
+          # TODO(https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/2395): Java 17+ support for Linkage Checker
+          java-version: 8
           distribution: temurin
       - name: Print Protobuf-Java testing version
         run: echo "Testing with Protobuf-Java v${{ matrix.protobuf-version }}"
--- a/.github/workflows/downstream_protobuf_compatibility_check_nightly.yaml
+++ b/.github/workflows/downstream_protobuf_compatibility_check_nightly.yaml
@@ -47,10 +47,11 @@ jobs:
         uses: actions/checkout@v4
       - uses: actions/setup-java@v4
         with:
-          # TODO(https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/2395): Java 17+ support for Linkage Checker
           java-version: 8
           distribution: temurin
       - run: echo "JAVA8_HOME=${JAVA_HOME}" >> $GITHUB_ENV
+      # Java Client Libraries are compiled with Java 11 and target Java 8. Java 11 is required because GraalVM
+      # minimum support is for Java 11.
       - uses: actions/setup-java@v4
         with:
           java-version: 11
--- a/.github/workflows/downstream_protobuf_compatibility_check_nightly.yaml
+++ b/.github/workflows/downstream_protobuf_compatibility_check_nightly.yaml
@@ -56,7 +56,6 @@ jobs:
         with:
           java-version: 11
           distribution: temurin
-      - run: echo "JAVA11_HOME=${JAVA_HOME}" >> $GITHUB_ENV
       - name: Print Protobuf-Java testing version
         run: echo "Testing with Protobuf-Java v${{ matrix.protobuf-version }}"
       - name: Perform downstream source compatibility testing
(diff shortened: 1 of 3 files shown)


lqiu96:
Two odd class file version issues:
1. Datatstore
```
Error: Exception in thread "main" java.lang.UnsupportedClassVersionError: com/google/cloud/datastore/emulator/CloudDatastore has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
```
2. Spanner
```
Error:  COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
Error:  /home/runner/work/sdk-platform-java/sdk-platform-java/java-spanner/google-cloud-spanner/src/main/java/com/google/cloud/spanner/nativeimage/SpannerFeature.java:[20,38] cannot access org.graalvm.nativeimage.hosted.Feature
  bad class file: /home/runner/.m2/repository/org/graalvm/sdk/nativeimage/24.1.2/nativeimage-24.1.2.jar(org/graalvm/nativeimage/hosted/Feature.class)
    class file has wrong version 55.0, should be 52.0
    Please remove or make sure it appears in the correct subdirectory of the classpath.
```

.github/workflows/downstream_protobuf_compatibility_check_nightly.yaml
```diff
@@ -47,8 +47,13 @@ jobs:
...
         with:
-          # Use Java 11 for this as Linkage Checker is only compatible with Java 11 or below
-          java-version: 11
+          # TODO(https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/2395): Java 17+ support for Linkage Checker
+          java-version: 8
+          distribution: temurin
+      - run: echo "JAVA8_HOME=${JAVA_HOME}" >> $GITHUB_ENV
+      - uses: actions/setup-java@v4
+        with:
+          java-version: 17
```
suztomo, line 56: Does this (using Java 17 to compile code) catch the Protobuf 4.30.1 Java 8 incompatibility issue?

suztomo (commented):
Please confirm using Java 17 in compilation can catch Protobuf 4.30.1's Java 8 incompatibility issue or not.

.github/workflows/downstream_protobuf_compatibility_check_nightly.yaml
lqiu96: Turns out it does not. Seems like it resolves the class file issue but wouldn't be able to test for the Protobuf v30.x issue.

.kokoro/nightly/downstream-protobuf-binary-compatibility.sh
```diff
@@ -65,7 +65,7 @@ function build_program_arguments() {
 # cloud-opensource-java contains the Linkage Checker tool
 git clone https://github.com/GoogleCloudPlatform/cloud-opensource-java.git
 pushd cloud-opensource-java
-mvn -B -ntp clean compile -T 1C
+mvn -B -ntp clean compile -T 1C -Djvm="${JAVA8_HOME}/bin/java"
```
lqiu96, line 68: Turns out the `-Djvm` flag is specific to Surefire not Maven. This still compiles with JDK 17.

lqiu96:
Ah, Downstream tests cannot use Java 8 for compilation: https://github.com/googleapis/sdk-platform-java/pull/3717. This is also documented go/cloud-java-native-image#class-file-has-wrong-version-550-should-be-530 (thanks @mpeddada1) 

I believe we must compile the client libraries with JDK 11+. Closing this PR. I think we will have to find a different solution to try to compile the client libraries using Protobuf 4.x and JDK 8.

 CC: @suztomo

lqiu96:
This seems a bit odd that the downstream tests failed for all client librarires when using Java 8, but it was able to compile and run unit tests and pass for most handwritten libraries using Java 8 in this PR.

Going to run another invocation to see

suztomo:
It passed.

.kokoro/nightly/downstream-protobuf-source-compatibility.sh
```diff
@@ -28,11 +28,18 @@ for repo in ${REPOS_UNDER_TEST//,/ }; do # Split on comma
   git clone "https://github.com/googleapis/$repo.git" --depth=1
   pushd "$repo"
 
+  # Compile with Java 11 and run the tests with Java 8
+  mvn clean compile -T 1C
+
+  export JAVA_HOME="${JAVA8_HOME}"
+  export PATH=$JAVA_HOME/bin:$PATH
+  java -version
```
suztomo, line 36: Maven's "-V" shows the version too.

lqiu96: Oh I left this in. I'll remove this version check since I've confirmed from the logs that it is switching versions properly.

Issue #3701, opened 2025-03-13
`update_generation_config.yaml` is not updated in downstream repos


diegomarquezp:
Downstream repos don't update this file. For example:
 - https://github.com/googleapis/java-datastore/commits/main/.github/workflows/update_generation_config.yaml
 - https://github.com/googleapis/java-bigtable/commits/main/.github/workflows/update_generation_config.yaml
 - https://github.com/googleapis/java-firestore/commits/main/.github/workflows/update_generation_config.yaml
 - https://github.com/googleapis/google-cloud-java/commits/main/.github/workflows/update_generation_config.yaml

This may extend to other files in the .github folder

diegomarquezp:
cc: @JoeWang1127 @suztomo
I noticed that https://github.com/googleapis/sdk-platform-java/blob/main/hermetic_build/library_generation/owlbot/templates/java_library/.kokoro/build.sh is very different from https://github.com/googleapis/java-logging/blob/main/.kokoro/build.sh. Same story with java-bigtable. The different `JOB_TYPE` handling is not present in this template.

Java-logging still [expects](https://github.com/googleapis/java-logging/blob/0b96d0cb5f1a8fe59b4e860038838cc003a9cbfd/owlbot.py#L90-L101) build.sh to be updated automatically.

diegomarquezp:
Fixed in https://github.com/googleapis/sdk-platform-java/pull/3723

PR #3723, opened 2025-03-26
chore(hermetic-build): include .kokoro and .github templates

diegomarquezp:
The templates come from https://github.com/googleapis/synthtool/tree/bf182cd41d9a7de56092cafcc7befe6b398332f6

The fix follows from https://setuptools.pypa.io/en/latest/userguide/datafiles.html:
```
Glob patterns do not automatically match dotfiles, i.e., directory or file names starting with a dot (.).
To include such files, you must explicitly start the pattern with a dot, e.g. .* to match .gitignore.
```

Fixes https://github.com/googleapis/sdk-platform-java/issues/3701 ☕


hermetic_build/library_generation/generate_library.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -eo pipefail
+set -exo pipefail
```
diegomarquezp, line 3: `xtrace` was the only way I could debug my way out in local development. This decision may have taken longer for someone less familiar with the codebase.

hermetic_build/DEVELOPMENT.md
```diff
@@ -78,7 +78,7 @@ as per [POSIX env var definition](https://pubs.opengroup.org/onlinepubs/96999197
 2. Move the jar into its well-known location.
 
    ```shell
-   mv /path/to/jar "${HOME}/.library_generation/gapic-generator-java.jar"
+   mv ~/.m2/repository/com/google/api/gapic-generator-java/{version}/gapic-generator-java-{version}.jar "${HOME}/.library_generation/gapic-generator-java.jar"
```
JoeWang1127, line 81: Is this too much detail? What if a user change the default maven dir?

hermetic_build/DEVELOPMENT.md
```diff
@@ -130,6 +130,9 @@ owl-bot copy-code --version
 The key step is `npm link`, which will make the command available in you current
 shell session.
 
+If you get a permission denied error when running the command `owl-bot`, try
+relinking owl-bot by running `npm unlink -g` and re-running the steps above.
```
JoeWang1127, line 134: Why would rerun the same command can solve the permission issue?

hermetic_build/library_generation/generate_library.sh
JoeWang1127: I didn't add `-x` because I think the extra logs may substantial and this level of logs is not recommended for production.

Do you think it's necessary to add these logs?

hermetic_build/library_generation/setup.py
```diff
@@ -24,6 +24,8 @@
             "owlbot/src/poms/*.py",
             "owlbot/templates/clirr/*.j2",
             "owlbot/templates/poms/*.j2",
+            "owlbot/templates/java_library/.github/**/*",
+            "owlbot/templates/java_library/.kokoro/**/*",
             "owlbot/templates/java_library/**/*",
```
JoeWang1127, line 29: Do we need this line?

hermetic_build/DEVELOPMENT.md
diegomarquezp: `npm link` has effect in the current shell session. My guess is that what's in an old session may get its permissions overriden.

hermetic_build/library_generation/setup.py
diegomarquezp: yes, there are files outside `.github` and `.kokoro` that this line would include.

hermetic_build/DEVELOPMENT.md
diegomarquezp: I thought it may be more convenient for copying the whole line and limit to making a replacement in `{version}`. I think a non-default maven setup is not a common use case, but still checks out, so I reverted this line.

hermetic_build/library_generation/generate_library.sh
diegomarquezp: I reverted this and explained how this would be useful in the dev guide.

hermetic_build/library_generation/utils/utilities.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -eo pipefail
+set -exo pipefail
```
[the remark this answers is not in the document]
diegomarquezp: I reverted this and explained how this would be useful in the dev guide.

diegomarquezp (commented):
Thanks for the review!

JoeWang1127 (approved):
Thanks for the fix.

suztomo:
I don't think CODEOWNERS can be templated any more.

suztomo:
Can you evaluate the templated, generated files are really needed? Seeing kokoro/nightly/java7.cfg, it seems you haven't checked that.

blakeli0 (approved):

hermetic_build/library_generation/owlbot/templates/java_library/.github/generated-files-bot.yml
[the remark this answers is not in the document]
suztomo: Having no downside effects doesn't mean it should be templated in the old style now. Here is the description of the bot https://github.com/googleapis/repo-automation-bots/tree/main/packages/generated-files-bot. This is the comment created by the generated-files bot:

image

The template relies on the old synth.metadata, which I don't think is updated any more.

If you use this generated-files-bot.yml correctly, we can use it to precisely warn people when they wrongly edit files that are maintained by the hermetic build process.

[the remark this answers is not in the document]
diegomarquezp: Thanks for pointing this out, Tomo. Looks like java-bigtable, for example, has a well-structured [synth.metadata](https://github.com/googleapis/java-bigtable/blob/main/synth.metadata) file that indicates the generated files and the templated files as well. This makes me think we need a more sophisticated logic to keep such a file up to date (e.g. use a base list of templated files + detect all generated files in the grpc, proto and gapic folders).

What if we leave this snippet bot config out of this PR and raise a follow up issue?

PR #3727, opened 2025-03-31
chore: remove logging module

JoeWang1127:
In this PR:
- Remove the logging module per go/debug-logging-proposal.


zhumin8 (approved):
basically reverting #3706

PR #3729, opened 2025-04-01
chore: Update gax-grpc's testlib refrences to use Message

lqiu96:


lqiu96:
Error in downstream CI logs:
```
/workspace/google-cloud-java/java-kms/google-cloud-kms/src/test/java/com/google/cloud/kms/v1/MockLocations.java:[40,3] method does not override or implement a method from a supertype
```

This error is expected as it requires downstream libraries to be updated with the generator changes. There are some breaking changes in the testlib.

PR #3730, opened 2025-04-02
feat: Selective gapic generation phase II

cindy-peng:
## Implement Selective GAPIC Generation (Phase II)

This PR implements Phase II of selective GAPIC generation within the `gapic-generator-java` project. This allows for finer control over the intended usage of generated client methods (public, internal, or hidden) by providing selective gapic generation configuration in service yaml.

#### 1. Model Updates
* Added a `isInternalApi` attribute to the internal representation of methods to track their intended visibility (e.g., public, internal).

#### 2. Parser Logic
* Introduced the `getMethodSelectiveGapicType()` method responsible for parsing the selective generation configuration for each method.
* Modified service filtering logic: Service classes will not be generated if the service definition contains no methods or includes only methods marked as **`HIDDEN`**.
* Enhanced `parseService()` to determine and assign the appropriate `SelectiveGapicType` to each service method and its corresponding generated variants (e.g., overloaded methods).

#### 3. Composer (Code Generation) Updates
* **Method Annotations:** For all method variants designated as `INTERNAL`, generate an `@InternalApi` annotation accompanied by a warning message discouraging external use.
* **Method Header Comments:** For methods marked as `INTERNAL`, generate a specific comment in the method's header indicating its intended internal-only usage.
* **Sample Generation:** Adjusted the logic for generating `package-info.java` samples to prevent the usage of any methods marked as `INTERNAL`.

#### 4. Tests
* Added **unit tests** covering the new parser logic and comment generation changes related to selective generation types.
* Added/updated **golden unit/integration tests** to verify the correct code output for various selective generation scenarios, including services with:
    * All public methods.
    * A mix of public, `INTERNAL`, and/or `HIDDEN` methods.
    * No public methods (verifying that the service class is not generated).


test/integration/goldens/iam/src/com/google/iam/v1/IAMPolicyClient.java
```diff
@@ -238,7 +238,7 @@ public IAMPolicyStub getStub() {
    * }</pre>
    *
    * @param request The request object containing all of the parameters for the API call.
-   * @throws com.google.api.gax.rpc.ApiException if the remote call fails
+   * @throws com.google.api.gax.rpc.ApiException if the remote call fails.
```
zhumin8, line 241: Thanks for trying to make this consistent with the @param. However, consider this PR is quite large on its own, do you mind reverting generator change for this so it is easier to focus during review?
Feel free to raise a separate PR for this change.

cindy-peng: Sounds good Min, that's a good suggestion! Actually I am also running into some issues updating showcase golden files with this small change. Will separate this into another PR.

zhumin8 (commented):
Looks good to me so far. Left a few small comments and I will take another pass.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -451,45 +457,48 @@ static boolean shouldIncludeMethodInGeneration(
...
+        selectiveGapicGenerationConfig.getGenerateOmittedAsInternal();
 
-    return includeMethodsList.contains(method.getFullName());
+    // Set method to PUBLIC if no SelectiveGapicGeneration Configuration is configured or the method
+    // is in the allow list.
+    // Otherwise, generate this method as INTERNAL or HIDDEN based on GenerateOmittedAsInternal
+    // flag.
+    if (includeMethodsList.isEmpty() && generateOmittedAsInternal == false
+        || includeMethodsList.contains(method.getFullName())) return SelectiveGapicType.PUBLIC;
+    else return generateOmittedAsInternal ? SelectiveGapicType.INTERNAL : SelectiveGapicType.HIDDEN;
```
zhumin8, line 479: Perhaps simplify a bit this if-else statement for readability?
Also, as per go/java-style#s4.1.1-braces-always-used, can you add braces to the return statements?

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/comment/SettingsCommentComposer.java
```diff
@@ -120,22 +120,22 @@ public CommentStatement getTransportProviderBuilderMethodComment() {
...
     return toSimpleComment(String.format(BUILDER_CLASS_DOC_PATTERN, outerClassName));
   }
 
   public static CommentStatement createCallSettingsBuilderGetterComment(
-      String javaMethodName, boolean isMethodDeprecated) {
+      String javaMethodName, boolean isMethodDeprecated, boolean isMethodInternal) {
     String methodComment = String.format(CALL_SETTINGS_BUILDER_METHOD_DOC_PATTERN, javaMethodName);
-    return isMethodDeprecated
-        ? toDeprecatedSimpleComment(methodComment)
+    return isMethodDeprecated || isMethodInternal
```
zhumin8, line 137: iiuc, `toDeprecatedInternalSimpleComment()` can also handle simple comment case, this conditional operator is redundant.
Also, perhaps rename to clarify and remove toSimpleComment() if not used anymore.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceStubSettingsClassComposer.java
```diff
@@ -140,6 +141,8 @@ public abstract class AbstractServiceStubSettingsClassComposer implements ClassC
   private static final String SETTINGS_LITERAL = "Settings";
 
   private static final String DOT = ".";
+  private static final String INTERNAL_API_WARNING =
```
zhumin8, line 144: nit: perhaps extract INTERNAL_API_WARNING to a somewhere to be accessible for each of these composes.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceSettingsClassComposer.java
```diff
@@ -270,40 +278,44 @@ private static List<MethodDefinition> createSettingsGetterMethods(
...
-                  SettingsCommentComposer.createCallSettingsGetterComment(
-                      getMethodNameFromSettingsVarName(javaMethodName), protoMethod.isDeprecated()))
-              .setAnnotations(
-                  protoMethod.isDeprecated()
-                      ? Arrays.asList(AnnotationNode.withType(TypeNode.DEPRECATED))
-                      : Collections.emptyList())
-              .build());
+      javaMethods.add(methodBuilderHelper(protoMethod, methodBuilder, javaMethodName));
       if (protoMethod.hasLro()) {
         javaMethodName = String.format("%sOperationSettings", javaStyleName);
```
zhumin8, line 286: nit: there is no need to update and reuse this string variable. It took me a second to realize the difference of L284 vs L289

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceClientClassComposer.java
```diff
@@ -802,11 +804,18 @@ private static List<MethodDefinition> createMethodVariants(
...
-            methodVariantBuilder.setAnnotations(
-                Arrays.asList(AnnotationNode.withType(TypeNode.DEPRECATED)));
+        annotations.add(AnnotationNode.withType(TypeNode.DEPRECATED));
+      }
+
+      if (method.isInternalApi()) {
+        annotations.add(
+            AnnotationNode.withTypeAndDescription(
+                typeStore.get("InternalApi"), INTERNAL_API_WARNING));
       }
```
zhumin8, line 816: similar logic is repeated a couple of times within this class. perhaps worth to extract to a helper method.

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/EchoServiceSelectiveGapicClient.golden
```diff
@@ -0,0 +1,639 @@
...
+   * // https://cloud.google.com/java/docs/setup#configure_endpoints_for_the_client_library
+   * try (EchoServiceShouldGeneratePartialPublicClient echoServiceShouldGeneratePartialPublicClient =
+   *     EchoServiceShouldGeneratePartialPublicClient.create()) {
+   *   EchoResponse response =
+   *       echoServiceShouldGeneratePartialPublicClient.chatShouldGenerateAsInternal();
+   * }
+   * }</pre>
+   *
+   * @param request The request object containing all of the parameters for the API call.
+   * @throws com.google.api.gax.rpc.ApiException if the remote call fails @InternalApi This method
```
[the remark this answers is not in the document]
cindy-peng: Yeah I noticed this earlier too. I am having a little trouble finding the issue. I think this is a new issue introduced by the new code. In fact, whenever there is new component e.g. `@deprecated` or `@return` after `@throws`, no `\n` will be attached. Originally I thought it's caused by the logic in [JavaDocComment](https://github.com/cindy-peng/sdk-platform-java/blob/f1f979d873868c0ef77b0a2d1f987aeb5ec5716f/gapic-generator-java/src/main/java/com/google/api/generator/engine/ast/JavaDocComment.java#L143), but the logic looks all right to me. I will keep investigating.

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/samples/echoserviceselectivegapicclient/AsyncChatAgainShouldGenerateAsPublic.golden
```diff
@@ -0,0 +1,58 @@
+/*
```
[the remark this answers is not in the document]
zhumin8: created https://github.com/googleapis/sdk-platform-java/issues/3740 to track.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceStubSettingsClassComposer.java
zhumin8: `composer/utils/CommonStrings.java` sounds good to me. I'd rather just include with this PR so we do not forget about it.

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpc/GrpcServiceStubClassComposerTest.java
```diff
@@ -93,6 +93,16 @@ void generateGrpcServiceStubClass_autopopulateField() {
     Assert.assertEmptySamples(clazz.samples());
   }
 
+  @Test
+  void generateGrpcServiceStubClass_selectiveGeneration() {
+    GapicContext context = GrpcTestProtoLoader.instance().parseSelectiveGenerationTesting();
+    Service service = context.services().get(1);
+    GapicClass clazz = GrpcServiceStubClassComposer.instance().generate(context, service);
+
+    Assert.assertGoldenClass(this.getClass(), clazz, "SelectiveGapicStub.golden");
```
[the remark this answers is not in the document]
zhumin8: Am I missing it or there is only test setup to verify -Client changes, but not -Stub, -StubSettings or -ClientSettings ?
For example, in below snippet of golden file, I expect `chatShouldGenerateAsInternalCallable()` also be marked as internal, and I believe changes in AbstractServiceStubClassComposer.java should do it. Do you have any test coverages for it?
https://github.com/googleapis/sdk-platform-java/blob/a87d3b0faf15ea767c0080df5cbea673b9aa0ab0/gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/EchoServiceSelectiveGapicClient.golden#L568-L571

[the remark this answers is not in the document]
cindy-peng: No I didn't add those tests earlier. I just updated the test files to include coverage for service settings, stub and stub settings. Please take a look:)

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/EchoServiceSelectiveGapicClient.golden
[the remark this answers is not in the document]
cindy-peng: @zhumin8 After some debugging, I found out the issue is actually with `google-java-format`. 

In [JavaWriterVisitor code](https://github.com/cindy-peng/sdk-platform-java/blob/f1f979d873868c0ef77b0a2d1f987aeb5ec5716f/gapic-generator-java/src/main/java/com/google/api/generator/engine/writer/JavaWriterVisitor.java#L1021, we parsed a well-formatted buffer string with line breaks before `@InternalApi`, somehow the line breaks were removed and put to somewhere else:
```java
String formattedClazz = JavaFormatter.format(buffer.toString());
```
I have created an issue for the team to track the issue: https://github.com/google/google-java-format/issues/1249.
Meanwhile, I added a fix to re-format the `@InternalApi` java doc comment in `JavaWriterVisitor`. Please take a look at the golden files and let me know if that works. Thanks!

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceStubSettingsClassComposer.java
cindy-peng: I just realized classes under `composer/utils/` are in a different package than all the composers under `composer/common/`. This will require common strings to be exposed as `Public` in order to be referenced in the composers, which could be a concern. 

Should I just create a `CommonStrings.java` file under [common](https://github.com/cindy-peng/sdk-platform-java/tree/6c301737751c6bbc7b681730a016be8d504c1911/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common) folder? Seems like it is mostly for the common abstract composer classes.

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/EchoServiceSelectiveGapicClient.golden
```diff
@@ -438,8 +438,8 @@ public class EchoServiceShouldGeneratePartialUsualClient implements BackgroundRe
    * }</pre>
    *
    * @param request The request object containing all of the parameters for the API call.
-   * @throws com.google.api.gax.rpc.ApiException if the remote call fails @InternalApi This method
-   *     is internal used only. Please do not use directly.
+   * @throws com.google.api.gax.rpc.ApiException if the remote call fails  
+   * @InternalApi This method is internal used only. Please do not use directly.
```
zhumin8, line 442: Sorry raising this a bit late. But I wonder if it makes more sense to add this note as part of the descriptive text, instead of as tag. Javadoc usually expects standard block tags as listed in [this guide](https://engdoc.corp.google.com/eng/doc/devguide/java/practices/javadoc.md?cl=head#at_clauses)

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/EchoServiceSelectiveGapicServiceSettings.golden
```diff
@@ -0,0 +1,260 @@
...
+
+    /**
+     * Returns the builder for the settings used for calls to
+     * chatShouldGenerateAsInternal. @InternalApi This method is internal used only. Please do not
+     * use directly.
+     */
+    public UnaryCallSettings.Builder<EchoRequest, EchoResponse>
+        chatShouldGenerateAsInternalSettings() {
+      return getStubSettingsBuilder().chatShouldGenerateAsInternalSettings();
+    }
```
zhumin8: Looks like these builder setter methods have comment added as internal api, but no `@InternalApi` annotation added.

gapic-generator-java/src/main/java/com/google/api/generator/engine/writer/JavaWriterVisitor.java
```diff
@@ -1020,6 +1020,15 @@ public void visit(ClassDefinition classDefinition) {
     if (!classDefinition.isNested()) {
       String formattedClazz = JavaFormatter.format(buffer.toString());
 
+      // fixing @InternalApi comment after formatting
+      // formatter removes the line break before @InternalApi comment.
+      // See https://github.com/google/google-java-format/issues/1249
```
zhumin8, line 1025: Great finding!
I think by convention, block tags should be starting with lowercase, changing it might make formatter recognize it correctly. 
One other thing I am not certain about, by using this custom tag, does the tools used in the publishing chain (e.g. https://github.com/googleapis/java-docfx-doclet) all be able to handle it correctly? From [this guide](https://www.baeldung.com/javadoc#3-custom-javadoc-tags), looks like some customizations are needed.

zhumin8: Perhaps it might be easier to include [this notice ](https://github.com/googleapis/sdk-platform-java/blob/6c301737751c6bbc7b681730a016be8d504c1911/gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/EchoServiceSelectiveGapicServiceSettings.golden#L110) as part of the descriptive text in Javadoc?

WDYT, @blakeli0 ?

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/EchoServiceSelectiveGapicServiceSettings.golden
cindy-peng: Yeah this should have been fixed in the latest commit already.

gapic-generator-java/src/main/java/com/google/api/generator/engine/writer/JavaWriterVisitor.java
cindy-peng: @zhumin8 You are absolutely right! After changing block tag `@InternalApi` to start with lowercase `@internalApi`, Java formatter was able to format everything correctly. I will make the change and remove the regex fix here.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceStubSettingsClassComposer.java
zhumin8: Either way is good with me. I am not super concerned about it being public because this is generator code and not client library code.

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/EchoServiceSelectiveGapicClient.golden
```diff
@@ -0,0 +1,637 @@
...
+   *   EchoResponse response =
+   *       echoServiceShouldGeneratePartialUsualClient.chatShouldGenerateAsInternal(name);
+   * }
+   * }</pre>
+   *
+   * @param name
+   * @throws com.google.api.gax.rpc.ApiException if the remote call fails
+   * @internalApi This method is internal used only. Please do not use directly.
+   */
+  @InternalApi("Internal API. This API is not intended for public consumption.")
```
blakeli0, line 472:
```suggestion
  @InternalApi("This API is not intended for public consumption.")
```
The leading "Internal API. " in the comment is slight redundant, WDYT?

cindy-peng: Done.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/protoparser/Parser.java
```diff
@@ -451,45 +460,53 @@ static boolean shouldIncludeMethodInGeneration(
...
+    List<String> includeMethodsList = selectiveGapicGenerationConfig.getMethodsList();
+
+    Boolean generateOmittedAsInternal =
+        selectiveGapicGenerationConfig.getGenerateOmittedAsInternal();
+
+    // Set method to PUBLIC if no SelectiveGapicGeneration Configuration is configured or the method
+    // is in the allow list.
+    // Otherwise, generate this method as INTERNAL or HIDDEN based on GenerateOmittedAsInternal
+    // flag.
+    if (includeMethodsList.isEmpty() && generateOmittedAsInternal == false
```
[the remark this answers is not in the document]
cindy-peng: I have refactored the logic and added unit tests in ParserTest to cover Internal, Hidden, Public. Please help take a look.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceStubSettingsClassComposer.java
cindy-peng: Sounds good. Added `composer/utils/CommonStrings.java`.

gapic-generator-java/src/main/java/com/google/api/generator/engine/writer/JavaWriterVisitor.java
cindy-peng: @zhumin8 I have changed the block tag into descriptive text. please help take a look. Thanks!

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/EchoServiceSelectiveGapicClient.golden
```diff
@@ -437,9 +437,10 @@ public class EchoServiceShouldGeneratePartialUsualClient implements BackgroundRe
    * }
    * }</pre>
    *
+   * <p><b>Warning: </b>This method is internal used only. Please do not use directly.
```
zhumin8, line 440: nit: "This method is for internal use only. Please do not use it directly."

zhumin8 (approved):
LGTM.

gapic-generator-java/src/test/java/com/google/api/generator/gapic/composer/grpc/goldens/EchoServiceSelectiveGapicClient.golden
cindy-peng: Done.

blakeli0 (approved):

PR #3735, opened 2025-04-08
chore: add jdk 8 verification to showcase CI

mpeddada1:
This PR adds showcase testing for JDK 8.
- Due to the issue discovered in https://github.com/googleapis/sdk-platform-java/pull/3735#issuecomment-2787212633, we had to compile the code with jDK 11 and run with JDK 8.
- Skipping SLF4J 2.x showcase run in JDK 8 due to:
```
Error:    bad class file: /home/runner/.m2/repository/ch/qos/logback/logback-classic/1.5.16/logback-classic-1.5.16.jar(ch/qos/logback/classic/Level.class)
Error:      class file has wrong version 55.0, should be 52.0
```
According to https://logback.qos.ch/dependencies.html 1.5.x targets at least JDK 11.


mpeddada1:
Adding `8` to the matrix results in:
```
Failed to execute goal com.spotify.fmt:fmt-maven-plugin:2.25:check (default) on project api-common: Execution default of goal com.spotify.fmt:fmt-maven-plugin:2.25:check failed: Unable to load the mojo 'check' in the plugin 'com.spotify.fmt:fmt-maven-plugin:2.25' due to an API incompatibility: org.codehaus.plexus.component.repository.exception.ComponentLookupException: com/spotify/fmt/Check has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
```

.github/workflows/ci.yaml
```diff
@@ -213,6 +213,69 @@ jobs:
...
+      - name: Install all modules using Java 11
+        shell: bash
+        run: |
+          mvn install -B -ntp -DskipTests -Dclirr.skip -Dcheckstyle.skip -Dfmt.skip
+      - uses: actions/setup-java@v3
+        with:
+          java-version: 8
+          distribution: temurin
+      - run: java -version
+      - name: Showcase golden tests
```
blakeli0, line 235: The golden tests verify the static generation results, which would not change with Java 8.

.github/workflows/ci.yaml
```diff
@@ -213,6 +213,69 @@ jobs:
...
+        working-directory: java-showcase
+        run: |
+            mvn test \
+            -P enable-golden-tests \
+            --batch-mode \
+            --no-transfer-progress
+      - name: Parse showcase version
+        working-directory: java-showcase/gapic-showcase
+        run: echo "SHOWCASE_VERSION=$(mvn help:evaluate -Dexpression=gapic-showcase.version -q -DforceStdout)" >> "$GITHUB_ENV"
+      - name: Install showcase server
```
blakeli0, line 245: Downloading showcase server should ideally only happen once, but I guess that may require this to be merged with the regular showcase tests. Is there a way for two CI jobs to share the same set up?

.github/workflows/ci.yaml
mpeddada1: Ah that's a good point, thank you

.github/workflows/ci.yaml
mpeddada1: Hm true, from looking at the logs, I think this step gets repeated even when we have a matrix. Haven't tried this out yet but we could have a separate job that does the download once and caches the showcase path. All the other showcase jobs would be dependent on this job. wdyt? 

However, this may lead to the "download" job showing on the CI.

blakeli0: > I think this step gets repeated even when we have a matrix

That's a good point, I think we can leave it as it is.

blakeli0 (approved):

PR #3767, opened 2025-04-28
fix: Only send mtlsEndpoint if it is non-null

lqiu96:
Temporary fix to only plumb the mtlsEndpoint to the gRPCChannelProviders if this has been set with a non-null value. Validation will still go through if it is set with bogus values.

Issue reported in https://github.com/googleapis/java-bigtable/pull/2565

This seems to due to how Bigtable wrappers the generated stubs. BigTable manually calls setEndpoint to for the StubSettings, but does not do the equivalent for setMtlsEndpoint. The EndpointContext is only aware of the TLS endpoint and passes a null value to the ChannelProviders. 

Potential longer term fix: BigTable's wrappers would need to set the mTLS Endpoint. This would be needed for BigTable to support S2A. This does not impact any of BigTable's existing functionality as the endpoint resolution for all non-S2A flows is done via `setEndpoint()`.

### DIFF
--- a/gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
+++ b/gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
@@ -223,7 +223,7 @@ public abstract class ClientContext {
       transportChannelProvider = transportChannelProvider.withEndpoint(endpoint);
     }
     transportChannelProvider = transportChannelProvider.withUseS2A(endpointContext.useS2A());
-    if (transportChannelProvider.needsMtlsEndpoint()) {
+    if (transportChannelProvider.needsMtlsEndpoint() && endpointContext.mtlsEndpoint() != null) {
       transportChannelProvider =
           transportChannelProvider.withMtlsEndpoint(endpointContext.mtlsEndpoint());
     }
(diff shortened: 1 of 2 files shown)


gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
```diff
@@ -223,7 +223,7 @@ public static ClientContext create(StubSettings settings) throws IOException {
       transportChannelProvider = transportChannelProvider.withEndpoint(endpoint);
     }
     transportChannelProvider = transportChannelProvider.withUseS2A(endpointContext.useS2A());
-    if (transportChannelProvider.needsMtlsEndpoint()) {
+    if (transportChannelProvider.needsMtlsEndpoint() && endpointContext.mtlsEndpoint() != null) {
```
blakeli0, line 226: Can we add a unit test for it? It would be great if we can mimic the Bigtable use case. Or maybe it is too complex to do so?

lqiu96: I think it may be possible and I can take a look. I was just planning to make this a temporary fix (revert in the next cycle or two).

blakeli0 (approved):

PR #3768, opened 2025-05-01
deps: Update maven-shared-utils to v3.2.1

lqiu96:
See b/392225607

### DIFF
--- a/java-shared-dependencies/dependency-convergence-check/pom.xml
+++ b/java-shared-dependencies/dependency-convergence-check/pom.xml
@@ -41,10 +41,6 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>org.apache.maven.shared</groupId>
-      <artifactId>maven-shared-utils</artifactId>
-    </dependency>
-    <dependency>
       <groupId>junit</groupId>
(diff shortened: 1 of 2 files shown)


suztomo (approved):

java-shared-dependencies/dependency-convergence-check/pom.xml
```diff
@@ -30,6 +40,10 @@
       <version>1.5.13</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.maven.shared</groupId>
+      <artifactId>maven-shared-utils</artifactId>
```
suztomo, line 45: (A minor thing) Did you find this line is needed? I think dependencyManagement section is enough.

suztomo (approved):

java-shared-dependencies/dependency-convergence-check/pom.xml
lqiu96: Ah, let me double check this. I think you're right.

lqiu96: Yep, removed. Thanks!

PR #3784, opened 2025-05-06
fix: S2A- Check if a default endpoint has been set

rmehta19:
Certain handwritten client libraries set this field to the (TLS) endpoint via [`GrpcTransportOptions.setUpChannelProvider`](https://github.com/googleapis/sdk-platform-java/blob/2a41393c0e148e96fd3a0a8e1505e0827aa4fd3a/java-core/google-cloud-core-grpc/src/main/java/com/google/cloud/grpc/GrpcTransportOptions.java#L165).

This isn't really a custom endpoint, so we should be able to use S2A when this is set.

Example with Cloud Logging: https://github.com/googleapis/java-logging/blob/644ebd756d44ffc02b58aa30d9a07c8afe700981/google-cloud-logging/src/main/java/com/google/cloud/logging/spi/v2/GrpcLoggingRpc.java#L137


rmehta19:
@lqiu96 , would you be able to review this? Thanks!

lqiu96:
> Certain handwritten client libraries set this field to the (TLS) endpoint via [GrpcTransportOptions.setUpChannelProvider](https://github.com/googleapis/sdk-platform-java/blob/2a41393c0e148e96fd3a0a8e1505e0827aa4fd3a/java-core/google-cloud-core-grpc/src/main/java/com/google/cloud/grpc/GrpcTransportOptions.java#L165).

I see, I wasn't aware of this use case. However, if a user does create their own TransportChannelProvider with a custom endpoint, then it wouldn't skip S2A. 

Can you remind me again what the original rationale was for skipping a custom endpoint?

rmehta19:
> Can you remind me again what the original rationale was for skipping a custom endpoint?

When a user is setting a custom endpoint, we can't be sure this is a MTLS endpoint, therefore we can't use S2A. We only want to use S2A when we can use the MTLS endpoint.

> However, if a user does create their own TransportChannelProvider with a custom endpoint, then it wouldn't skip S2A.

This is true, if a hw client library does this (say they set it to something foo.xyz.com) then S2A wouldn't get skipped, and S2A would be used to connect to MTLS endpoint of api (foo.mtls.googleapis.com) which isn't our desired behavior (we want to respect custom set endpoint and skip S2A in this case). Perhaps we could add some extra field (`GAXProvidedEndpoint`) to the TransportChannelProvider interface and [`GrpcTransportOptions.setUpChannelProvider`](https://github.com/googleapis/sdk-platform-java/blob/2a41393c0e148e96fd3a0a8e1505e0827aa4fd3a/java-core/google-cloud-core-grpc/src/main/java/com/google/cloud/grpc/GrpcTransportOptions.java#L165) can set that field instead? Then we could  keep the S2A logic in EndpointContext as is. WDYT @lqiu96?

rmehta19:
Friendly ping on this @lqiu96. Thank you!

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -327,8 +327,7 @@ boolean shouldUseS2A() {
       }
 
       // If a custom endpoint is being used, skip S2A.
-      if (!Strings.isNullOrEmpty(clientSettingsEndpoint())
-          || !Strings.isNullOrEmpty(transportChannelProviderEndpoint())) {
+      if (!Strings.isNullOrEmpty(clientSettingsEndpoint())) {
```
lqiu96, line 330: I'm worried that this may only end up potentially resolving the issue for the `GrpcTransportOptions.setUpChannelProvider` case. IIRC, `resolveEndpoint()`'s logic still revolves around if either clientSettings' endpoint or transportChannelProvider's endpoint is set, and if so, then that's the resolved endpoint. 

I'm worried that it would end up with s2a being enabled with the resolved endpoint still being the one configured in GrpcTransportOptions.setUpChannelProvider

rmehta19: > I'm worried that it would end up with s2a being enabled with the resolved endpoint still being the one configured in GrpcTransportOptions.setUpChannelProvider

I believe our logic in InstantiatingGrpcChannelProvider prevents this from happening. If S2A is enabled, we use the `mtlsEndpoint` (that is [plumbed separately](https://github.com/googleapis/sdk-platform-java/pull/3673) from the `endpoint`; `endpoint` == `resolvedEndpoint()`)

https://github.com/googleapis/sdk-platform-java/blob/49a7ae50071e75fe0d161a4eb9360a4fe4e6147b/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L686-L699

lqiu96: Oh right, there is the custom logic for S2A. What about the case if a user has a custom TransportChannelProvider with a non-mtls endpoint configured? IIUC above, the intention would be not use S2A, but our checks in EndpointContext would have shouldUseS2A as true.

rmehta19: > Oh right, there is the custom logic for S2A. What about the case if a user has a custom TransportChannelProvider with a non-mtls endpoint configured? IIUC above, the intention would be not use S2A, but our checks in EndpointContext would have shouldUseS2A as true.

It is correct that if the user is intentionally setting a custom endpoint via their own TransportChannelProvider, that we want to skip S2A, but with the change shouldUseS2A would evaluate to true.

> Gax and the generator no longer provide a default non-mtls endpoint for the client libraries. The "default" endpoint that is intended to be used is constructed inside the EndpointContext. If you want to check, you can theoretically just check if clientSettingsEndpoint or transportChannelEndpoint == buildEndpointTemplate(serviceName(), resolvedUniverseDomain())

Apologies, I missed this comment of yours. Thanks for bringing this up! I tested this out internally and this works. The latest commit contains this logic, WDYT @lqiu96

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -328,7 +328,9 @@ boolean shouldUseS2A() {
 
       // If a custom endpoint is being used, skip S2A.
       if (!Strings.isNullOrEmpty(clientSettingsEndpoint())
-          || !Strings.isNullOrEmpty(transportChannelProviderEndpoint())) {
+          || (!Strings.isNullOrEmpty(transportChannelProviderEndpoint())
+              && !buildEndpointTemplate(serviceName(), resolvedUniverseDomain())
+                  .contains(transportChannelProviderEndpoint()))) {
```
rmehta19, line 333: Adding this "contains" logic, since `buildEndpointTemplate` adds `443` (https) as default port, and [`GrpcTransportOptions.setUpChannelProvider`](https://github.com/googleapis/sdk-platform-java/blob/2a41393c0e148e96fd3a0a8e1505e0827aa4fd3a/java-core/google-cloud-core-grpc/src/main/java/com/google/cloud/grpc/GrpcTransportOptions.java#L165), for example, just sets the host.

In this way, we are sure to respect a custom endpoint set via `TransportChannelProvider`.

@lqiu96

lqiu96 (approved):
Code LGTM, added a small comment if you could take a look.

I think it should be fine, but we can try to address any additional things after more testing.

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
rmehta19: > Could this check for both clientSettings and transportChannelProvider just to be safe? Or is there a specific reason for only doing this for the transportchannelprovider?

I don't see an issue for doing this check for both `clientSettingsEndpoint` and `transportChannelProviderEndpoint`, the reason I initially did it for the latter was because I ran into it in the GrpcTransportOptions case :). I tested this internally and verified that doing this check on both is ok. Added this in the latest commit.

rmehta19:
> Code LGTM, added a small comment if you could take a look.
> 
> I think it should be fine, but we can try to address any additional things after more testing.

Thanks @lqiu96 ! Addressed your latest comment. PTAL when you get a chance.

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
lqiu96: FYI, Bigtable has custom code to setEndpoint via settings which I think would be an issue: https://github.com/googleapis/java-bigtable/blob/8d3dca43224179829829bcf91972610c666b130b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java#L807

May be an issue in the future as it auto sets a default. But you would prefer to handle those cases later on, then we can just proceed with this to unblock logging.

lqiu96:
Re-running the failed showcase test. Not sure if it's flaky or not and we can see after this run.

Also, can you update the title to follow conventioncommits with the update changes? i.e. `fix: S2A- Check if a default endpoint has been set` or however you see fit. Thanks!

rmehta19:
> Re-running the failed showcase test. Not sure if it's flaky or not and we can see after this run.
> 
> Also, can you update the title to follow conventioncommits with the update changes? i.e. `fix: S2A- Check if a default endpoint has been set` or however you see fit. Thanks!

Done, thanks @lqiu96 !

lqiu96 (approved):

rmehta19:
> Seeing some odd showcase errors. Can you try merging the latest into the branch and we can see if that resolves the CI issues?

Merged the latest changes from main @lqiu96

lqiu96:
> > Seeing some odd showcase errors. Can you try merging the latest into the branch and we can see if that resolves the CI issues?
> 
> Merged the latest changes from main @lqiu96

Thanks! Running the graalvm checks and should auto-merge after success

PR #3816, opened 2025-05-29
chore: temporarily stop using maven mirror

mpeddada1:
https://github.com/googleapis/sdk-platform-java/pull/3814#issuecomment-2919666879


zhumin8 (dismissed):

suztomo (dismissed):
Approving. Other script might reference the settings.xml. In that case commenting out the lines in the settings.xml is more robust.

suztomo (approved):

PR #3826, opened 2025-06-02
chore: migrate from `not(in(collection))` to `e -> !collection.contains(e)`.

eamonnmcmanus:
Guava's functional types have been obsoleted by the JDK's functional types.

(This reflects a change we are making throughout Google's internal source code.)


eamonnmcmanus:
> Hi @eamonnmcmanus, thanks for contributing! Is there an internal issue we can understand more about the background?

Yes, http://b/419373312. I or @kluever can supply more background if needed.

blakeli0:
> > Hi @eamonnmcmanus, thanks for contributing! Is there an internal issue we can understand more about the background?
> 
> Yes, http://b/419373312. I or @kluever can supply more background if needed.

Thanks! I know there are Javadoc to discourage people from using them already, but are they going to be deprecated as well so that it easily reminds us to not use them in the future?

PR #3829, opened 2025-06-06
feat: handle auto pagination for BigQuery v2

PhongChuong:

Fixes # b/360114900 & b/352331075

### DIFF
--- a/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceStubSettingsClassComposer.java
+++ b/gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceStubSettingsClassComposer.java
@@ -149,6 +149,23 @@ public abstract class AbstractServiceStubSettingsClassComposer implements ClassC
 
   private final TransportContext transportContext;
 
+  // Maps of package->methodName to types for pagination with max_results field.
+  private static final ImmutableMap<String, ImmutableMap<String, TypeNode>>
+      PAGINATE_MAX_RESULT_TYPES =
+      ImmutableMap.of(
+          "com.google.cloud.bigquery.v2",
+          ImmutableMap.of(
+              "ListDatasets",
+              TypeNode.UINT32VALUE,
+              "ListJobs",
+              TypeNode.INT32VALUE,
+              "ListModels",
+              TypeNode.UINT32VALUE,
+              "ListRoutines",
+              TypeNode.UINT32VALUE,
+              "ListTables",
+              TypeNode.UINT32VALUE));
+
   protected static final VariableExpr DEFAULT_SERVICE_SCOPES_VAR_EXPR =
       createDefaultServiceScopesVarExpr();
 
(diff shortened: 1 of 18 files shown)


PhongChuong:
I'm unsure which checks are relevant/needed. Please let me know,
Thank you,

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceStubSettingsClassComposer.java
```diff
@@ -149,6 +149,23 @@ public abstract class AbstractServiceStubSettingsClassComposer implements ClassC
 
   private final TransportContext transportContext;
 
+  // Maps of package->methodName to types for pagination with max_results field.
+  private static final ImmutableMap<String, ImmutableMap<String, TypeNode>>
+      PAGINATE_MAX_RESULT_TYPES =
+          ImmutableMap.of(
+              "com.google.cloud.bigquery.v2",
+              ImmutableMap.of(
```
blakeli0, line 157: I guess there are some legacy RPCs use `max_results` field and are not of wrapper types?

gapic-generator-java/src/test/proto/echo.proto
```diff
@@ -140,6 +140,16 @@ service Echo {
       body: "*"
     };
   }
+
+  // This is similar to the PagedExpand except that it uses
+  // max_results instead of page_size, as some legacy APIs still
+  // do. New APIs should NOT use this pattern.
+  rpc PagedExpandLegacy(PagedExpandLegacyRequest) returns (PagedExpandResponse) {
```
blakeli0, line 147: I don't think this test can prove much because the package is not bigquery, this is more like a negative test case which is already covered by other test cases. We need a positive golden test file that can demonstrate the changes.

gapic-generator-java/src/test/proto/bigquery_jobs.proto
blakeli0, line 1: Are golden tests for this file missing?

gapic-generator-java/src/test/java/com/google/api/generator/test/protoloader/TestProtoLoader.java
```diff
@@ -539,6 +540,44 @@ public GapicContext parseCallabeNameType() {
...
+  public GapicContext parseBigqueryService() {
+    FileDescriptor bigqueryFileDescriptor = JobProto.getDescriptor();
+    ServiceDescriptor biqqueryServiceDescriptor = bigqueryFileDescriptor.getServices().get(0);
+    assertEquals("JobService", biqqueryServiceDescriptor.getName());
+
+    String serviceYamlFilename = "bigquery_v2.yaml";
+    Path serviceYamlPath = Paths.get(testFilesDirectory, serviceYamlFilename);
+    Optional<com.google.api.Service> serviceYamlOpt =
+        ServiceYamlParser.parse(serviceYamlPath.toString());
+    assertTrue(serviceYamlOpt.isPresent());
```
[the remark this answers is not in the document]
PhongChuong: Deleted service.yaml but it seems like config json is still needed.

PhongChuong:
@blakeli0 , thank you for the review. I've updated the PR to include your suggestions. PTAL.

gapic-generator-java/src/test/proto/echo.proto
blakeli0: SGTM, thanks!

blakeli0 (approved):

Issue #3851, opened 2025-06-25
Implement X.509 WIF mTLS-transport support in Java SDK


andyrzhao:
We would like to implement X.509 WIF mTLS-transport support in Java Gax lib (to bring it to feature-parity with Golang and Python).

The 4-year old mTLS implementation in GAX only supports the legacy SecureConnect based mTLS. We've recently updated the Java Auth lib to support both the old SecureConnect based mTLS as well as the newer X.509 WIF based mTLS. The Gax library needs to be updated to call into the Java Auth Lib for mTLS provider, so it can use either implementation of mTLS based on availability.

Additional Call-outs:
1. The existing implementation of "SecureConnect based mTLS in the GAX" library will be considered deprecated and retained for backwards compatibility, but we are not aware of any direct usage of these old classes, since it is an internal feature.
2. There is an env var GOOGLE_API_USE_CLIENT_CERTIFICATE that gates the mTLS-usage in the transport layer (for both X.509 based mTLS and SecureConnect based mTLS). Currently, the default behavior when this env var is not set on the machine is that we will not use mTLS in the transport layer. This is actually different from the Golang SDK default behavior, where the default semantics is to enable mTLS unless the env var is set to "false". We would like to flip the switch to make mTLS support enabled by default in Java SDK at some point as well, but it will be done as a separate step, after we confirmed the refactoring support in GAX for X.509 based mTLS is stable.

PR #3852, opened 2025-06-25
feat(mtls): Add support for X.509-based mTLS-transport in Java GAX lib

andyrzhao:
Fixes https://github.com/googleapis/sdk-platform-java/issues/3851

- Refactors mTLS code path to use new CertificateBasedAccess class to determine mTLS behavior based on env vars.
- Refactors mTLS code path to use DefaultMtlsProviderFactory from Java auth lib for creating a default mTLS provider using either the legacy SecureConnect mtls provider or the newer X.509 mtls provider, depending on availability.

### DIFF
--- a/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
+++ b/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
@@ -153,7 +153,7 @@ public final class InstantiatingGrpcChannelProvider implements TransportChannelP
   @Nullable private final Boolean allowNonDefaultServiceAccount;
   @VisibleForTesting final ImmutableMap<String, ?> directPathServiceConfig;
   @Nullable private final MtlsProvider mtlsProvider;
-  @Nullable private final CertificateBasedAccess certificateBasedAccess;
+  private final CertificateBasedAccess certificateBasedAccess;
   @Nullable private final SecureSessionAgent s2aConfigProvider;
   private final List<HardBoundTokenTypes> allowedHardBoundTokenTypes;
   @VisibleForTesting final Map<String, String> headersWithDuplicatesRemoved = new HashMap<>();
@@ -489,7 +489,7 @@ public final class InstantiatingGrpcChannelProvider implements TransportChannelP
 
   @VisibleForTesting
   ChannelCredentials createMtlsChannelCredentials() throws IOException, GeneralSecurityException {
-    if (certificateBasedAccess == null || mtlsProvider == null) {
+    if (mtlsProvider == null) {
       return null;
     }
     if (certificateBasedAccess.useMtlsClientCertificate()) {
@@ -861,7 +861,7 @@ public final class InstantiatingGrpcChannelProvider implements TransportChannelP
     private boolean useS2A;
     private EnvironmentProvider envProvider;
     private SecureSessionAgent s2aConfigProvider = SecureSessionAgent.create();
-    private MtlsProvider mtlsProvider;
+    @Nullable private MtlsProvider mtlsProvider;
     private CertificateBasedAccess certificateBasedAccess;
     @Nullable private GrpcInterceptorProvider interceptorProvider;
     @Nullable private Integer maxInboundMessageSize;
--- a/gax-java/gax/src/test/java/com/google/api/gax/rpc/mtls/CertificateBasedAccessTest.java
+++ b/gax-java/gax/src/test/java/com/google/api/gax/rpc/mtls/CertificateBasedAccessTest.java
@@ -69,7 +69,7 @@ class CertificateBasedAccessTest {
   void testUseMtlsClientCertificateTrue() {
     CertificateBasedAccess cba =
         new CertificateBasedAccess(
-            name -> name.equals("GOOGLE_API_USE_MTLS_ENDPOINT") ? "auto" : "true");
+            name -> name.equals("GOOGLE_API_USE_CLIENT_CERTIFICATE") ? "true" : "auto");
     assertTrue(cba.useMtlsClientCertificate());
   }
 
@@ -77,7 +77,7 @@ class CertificateBasedAccessTest {
   void testUseMtlsClientCertificateFalse() {
     CertificateBasedAccess cba =
         new CertificateBasedAccess(
-            name -> name.equals("GOOGLE_API_USE_MTLS_ENDPOINT") ? "auto" : "false");
+            name -> name.equals("GOOGLE_API_USE_CLIENT_CERTIFICATE") ? "false" : "auto");
     assertFalse(cba.useMtlsClientCertificate());
   }
 }
--- a/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/InstantiatingHttpJsonChannelProvider.java
+++ b/gax-java/gax-httpjson/src/main/java/com/google/api/gax/httpjson/InstantiatingHttpJsonChannelProvider.java
@@ -50,6 +50,7 @@ import java.util.concurrent.Executor;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import javax.annotation.Nullable;
 
 /**
  * InstantiatingHttpJsonChannelProvider is a TransportChannelProvider which constructs a {@link
@@ -74,7 +75,7 @@ public final class InstantiatingHttpJsonChannelProvider implements TransportChan
   private final HttpJsonInterceptorProvider interceptorProvider;
   private final String endpoint;
   private final HttpTransport httpTransport;
-  private final MtlsProvider mtlsProvider;
+  @Nullable private final MtlsProvider mtlsProvider;
   private final CertificateBasedAccess certificateBasedAccess;
 
   private InstantiatingHttpJsonChannelProvider(
@@ -184,7 +185,7 @@ public final class InstantiatingHttpJsonChannelProvider implements TransportChan
   }
 
   HttpTransport createHttpTransport() throws IOException, GeneralSecurityException {
-    if (certificateBasedAccess == null || mtlsProvider == null) {
+    if (mtlsProvider == null) {
       return null;
     }
     if (certificateBasedAccess.useMtlsClientCertificate()) {


gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -1269,6 +1284,24 @@ CallCredentials createHardBoundTokensCallCredentials(
...
+      if (certificateBasedAccess == null) {
+        certificateBasedAccess = CertificateBasedAccess.createWithSystemEnv();
+      }
+      if (certificateBasedAccess.useMtlsClientCertificate()) {
+        if (mtlsProvider == null) {
+          // Attempt to create default MtlsProvider from environment.
+          try {
+            mtlsProvider = DefaultMtlsProviderFactory.create();
+          } catch (CertificateSourceUnavailableException e) {
+            // This is okay. Leave mtlsProvider as null;
```
lqiu96, line 1296: qq, can you elaborate why we can leave mtlsProvider as null here? What happens if leave this as null?

gax-java/gax-grpc/src/test/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProviderTest.java
```diff
@@ -92,6 +93,7 @@ class InstantiatingGrpcChannelProviderTest extends AbstractMtlsTransportChannelT
   private static final String API_KEY_AUTH_HEADER_KEY = "x-goog-api-key";
   private static String originalOSName;
   private ComputeEngineCredentials computeEngineCredentials;
+  private CertificateBasedAccess cba;
```
lqiu96, line 96: nit: can you spell out the field variable name

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -150,6 +153,7 @@ public final class InstantiatingGrpcChannelProvider implements TransportChannelP
   @Nullable private final Boolean allowNonDefaultServiceAccount;
   @VisibleForTesting final ImmutableMap<String, ?> directPathServiceConfig;
   @Nullable private final MtlsProvider mtlsProvider;
+  @Nullable private final CertificateBasedAccess certificateBasedAccess;
```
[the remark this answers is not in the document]
andyrzhao: So the "CertificateBasedAccess" is essentially a couple of helper functions that was moved out of the original MtlsProvider implementation, and since the previous MtlsProvider was Nullable, this was also marked Nullable to retain the same semantics - my hunch is that we should keep it Nullable for maximum compatbilitly/flexibility.

gax-java/gax-grpc/src/test/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProviderTest.java
andyrzhao: will do!

gax-java/gax/src/main/java/com/google/api/gax/rpc/mtls/CertificateBasedAccess.java
```diff
@@ -0,0 +1,75 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package com.google.api.gax.rpc.mtls;
+
+import com.google.api.gax.rpc.internal.EnvironmentProvider;
+
+/** Utility class for handling certificate-based access configurations. */
+public class CertificateBasedAccess {
```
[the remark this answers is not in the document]
andyrzhao: Will add documentation and reference to AIP 4114 here. But this class cannot be made package-private AFAIK, since it is being widely used in unit tests (and across packages).

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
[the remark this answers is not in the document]
lqiu96: I see. The old MtlsProvider was public in name, but we didn't allow users to set it into the client. I don't think we have direct use cases of users using it and I don't think we need to maintain compatibility for it.

If possible, I would rather have this be non-null so we don't need the null checks.

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -117,6 +124,9 @@ public static EndpointContext getDefaultInstance() {
   @Nullable
   public abstract MtlsProvider mtlsProvider();
 
+  @Nullable
+  public abstract CertificateBasedAccess certificateBasedAccess();
```
lqiu96, line 128: I believe the others getters are public, but they probably shouldn't have been.

Since this is new, can we make this package-private?

gax-java/gax/src/main/java/com/google/api/gax/rpc/mtls/CertificateBasedAccess.java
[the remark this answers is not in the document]
lqiu96: Ok, since this looks to be used in tests only, can you add `@InternalApi` up top to denote that this is not meant to be used by users?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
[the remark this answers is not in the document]
lqiu96: Oh wait, I think the MtlsProvider must remain Nullable given the possibility of `CertificateSourceUnavailableException`.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -1269,6 +1285,25 @@ CallCredentials createHardBoundTokensCallCredentials(
...
+          } catch (CertificateSourceUnavailableException e) {
+            // This is okay. Leave mtlsProvider as null so that we will not auto-upgrade
+            // to mTLS endpoints. See https://google.aip.dev/auth/4114.
+          } catch (IOException e) {
+            LOG.log(
+                Level.WARNING,
+                "DefaultMtlsProviderFactory encountered unexpected IOException: " + e.getMessage());
+          }
+        }
+      }
```
lqiu96, line 1305: qq, I noticed that this rough logic also exists here and in the EndpointContext. What's the reason that it needs to exist in the gRPC Channel Provider as well as EndpointContext?

gax-java/gax/src/main/java/com/google/api/gax/rpc/EndpointContext.java
```diff
@@ -212,6 +222,9 @@ public abstract static class Builder {
 
     public abstract Builder setMtlsProvider(MtlsProvider mtlsProvider);
 
+    public abstract Builder setCertificateBasedAccess(
+        CertificateBasedAccess certificateBasedAccess);
```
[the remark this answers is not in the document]
andyrzhao: yup

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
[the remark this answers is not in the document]
lqiu96: I think MtlsProvider must remain Nullable and CertificateBasedAccess can be non-Nullable. Preference for CertificateBasedAccess to be non-nullable so that we don't need so many null checks in the logic below and in other files.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
lqiu96: > In the other 2 locations (gRPC/HTTP channel provider), they are used for configuring the TLS settings of the Channel themselves.

Sorry, can you elaborate on this point? I wasn't aware there was settings that MtlsProvider itself was configuring anything. Can you link me to how it's configure TLS settings on the channel? 

My assumption was that it was only using the MtlsProvider logic (i.e. checking for env var) to see if Mtls was to be enabled and not touching anything on the channel.

gax-java/gax/src/test/java/com/google/api/gax/rpc/testing/FakeMtlsProvider.java
```diff
@@ -76,6 +58,11 @@ public KeyStore getKeyStore() throws IOException {
     return keyStore;
   }
 
+  @Override
+  public boolean isAvailable() throws IOException {
+    return true;
+  }
```
lqiu96, line 64: qq, what is this used for? Was this added in by mistake?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
andyrzhao: Let me clarify, for gRPC/HTTP channel provider, the "mtlsKeyStore" is used for creating the mTLS-enabled transport as seen in the code snippet below:

HttpTransport createHttpTransport() throws IOException, GeneralSecurityException {
    if (certificateBasedAccess == null || mtlsProvider == null) {
      return null;
    }
    if (certificateBasedAccess.useMtlsClientCertificate()) {
      KeyStore mtlsKeyStore = mtlsProvider.getKeyStore();
      if (mtlsKeyStore != null) {
        return new NetHttpTransport.Builder().trustCertificates(null, mtlsKeyStore, "").build();
      }
    }
    return null;
  }

The channel providers have no reference to the EndpointContext, so need to independently calculate and bootstrap the mTLS provider.

gax-java/gax/src/test/java/com/google/api/gax/rpc/testing/FakeMtlsProvider.java
andyrzhao: The updated MtlsProvider interface declares an "isAvailable()" method which is a new concept, but for the sake of testing and being backwards compatible with existing tests, the "FakeMtlsProvider" is "always available".

lqiu96: Ah right. Thanks for clarifying.

gax-java/gax/src/test/java/com/google/api/gax/rpc/mtls/CertificateBasedAccessTest.java
```diff
@@ -0,0 +1,83 @@
...
+        new CertificateBasedAccess(
+            name -> name.equals("GOOGLE_API_USE_MTLS_ENDPOINT") ? "auto" : "true");
+    assertTrue(cba.useMtlsClientCertificate());
+  }
+
+  @Test
+  void testUseMtlsClientCertificateFalse() {
+    CertificateBasedAccess cba =
+        new CertificateBasedAccess(
+            name -> name.equals("GOOGLE_API_USE_MTLS_ENDPOINT") ? "auto" : "false");
```
lqiu96, line 80: same as above `GOOGLE_API_USE_CLIENT_CERTIFICATE ? false : true`?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
lqiu96: I see, thanks for the explanation. 

In this case, I think what we have previously configured is probably a mistake (i.e. configuring multiple separate MtlsProviders). I think realistically the flow should be 
1. EndpointContext creates the MtlsProvider and CertificateBasedAccess class that is used to compute the endpoint
2. gRPC and HttpJson channel providers get the MtlsProvider and CertificateBasedAccess classes that were created in the endpointcontext

However, I think in order for that to be done, we'll need to create public methods inside TransportChannelProvider to allow access to them. It is possible with adding InternalApi and we have done previously to access mtlsEndpoint.

Let me think about the options a bit more. I think what you have is based on the existing code and should work, but I think it previously wasn't configured the best/ correctly and would love to try and fix it if possible.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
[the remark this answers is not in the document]
andyrzhao: I updated the latest revision to mark MtlsProvider Nullable and CertificateBasedAccess NonNullable in the HTTP/gRPC channel providers for consistency. However, CertificateBasedAccess is left as @Nullable in EndpointContext, since removing it broke compilation for a ton of tests... (will revisit if needed.)

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
lqiu96: Thanks @andyrzhao, taking a look at this and I don't think there is a clear best way to do this. 

Since it looks like these implementations (EndpointContext, GrpcChannelProvider, HttpJsonChannelProvider) are private implementations and not accessible to customers, I think I'm fine with keep it as-is. Adding public methods to clean this up can be done in a future time and perhaps a better implementation can also be found. The new implementation follows the old one and I don't think this would be a regression.

Would you mind creating an issue in our repo to track that future enhancement?

lqiu96 (approved):

PR #3874, opened 2025-07-31
feat: Set sampledToLocalTracing flag to true for all gRPC MethodDescriptors

lqiu96:
See b/416626151 for more information.

Problem: When the sampledToLocalTracing is set to false or not set, the gRPC specific metrics have the `grpc.method` attribute defined as `other` instead of the methodDescriptor's full name. This may be confusing to customers who may expect to see the full gRPC method name.

High Level: This flag enables captures for specific method names to help provide more detailed metrics. We do not expect any noticeable performance impact from this change.

This does not impact the method name that is already properly captured for Gax Metrics (e.g. Attempt Count, Operation Latency, etc). Specifically, the use must be using grpc-java telemetry module and be using [GrpcOpenTelemetry](https://grpc.github.io/grpc-java/javadoc/io/grpc/opentelemetry/GrpcOpenTelemetry.html) instead of [OpenTelemetry](https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/OpenTelemetry.java).

### DIFF
--- a/java-showcase/gapic-showcase/pom.xml
+++ b/java-showcase/gapic-showcase/pom.xml
@@ -96,12 +96,6 @@
       <groupId>com.google.protobuf</groupId>
       <artifactId>protobuf-java</artifactId>
     </dependency>
-    <!-- Add the grpc-java opentelemetry module to test for clients using GrpcOpenTelemetry -->
-    <dependency>
-      <groupId>io.grpc</groupId>
-      <artifactId>grpc-opentelemetry</artifactId>
-      <scope>test</scope>
-    </dependency>
     <dependency>
       <groupId>com.google.api.grpc</groupId>
       <artifactId>proto-google-common-protos</artifactId>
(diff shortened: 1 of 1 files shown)


lqiu96:
Looks like sonatype is complaining about inadequate coverage for the generated showcase files. This should be a non-issue.

java-showcase/gapic-showcase/pom.xml
```diff
@@ -96,6 +96,11 @@
       <groupId>com.google.protobuf</groupId>
       <artifactId>protobuf-java</artifactId>
     </dependency>
+    <!-- Add the grpc-java opentelemetry module to test for clients using GrpcOpenTelemetry -->
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-opentelemetry</artifactId>
+    </dependency>
```
blakeli0, line 103: Can this be a `test` scope dependency?

lqiu96: sounds good, will do

java-showcase/gapic-showcase/pom.xml
```diff
@@ -96,6 +96,12 @@
       <groupId>com.google.protobuf</groupId>
       <artifactId>protobuf-java</artifactId>
     </dependency>
+    <!-- Add the grpc-java opentelemetry module to test for clients using GrpcOpenTelemetry -->
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-opentelemetry</artifactId>
+      <scope>test</scope>
```
blakeli0, line 103: There is a section for `Otel testing libraries` in below sections, do you mind move it there?

PR #3886, opened 2025-08-19
chore: Update the upper bound dependencies file

lqiu96:
Updates to the following upper bound dependency versions:
```
j2objc-annotations=3.1
threetenbp=1.7.2
google.api-client=2.8.1
opentelemetry-semconv=1.34.0
```


lqiu96:
> It might also be worth adding a summary in the PR description on which dependencies have been updated in the dependency compatibility job.

done

PR #3926, opened 2025-10-07
feat(librariangen): scaffold Java language container for Librarian

meltsufin:
This change introduces the initial scaffolding for `librariangen`, a containerized Go application that will serve as the Java-specific code generator  within the Librarian pipeline.

The primary goal of this change is to establish the foundational components for the Java language container, including:
* A new `librariangen` CLI application: Written in Go, this will be the entry point for Java code generation tasks invoked by Librarian. It currently includes basic command dispatching and support for `--version`.
* A multi-stage `Dockerfile`: This creates a minimal container image for librariangen.
* Unit tests and CI: A new GitHub Actions workflow has been added to build and test the container, ensuring that the librariangen executable is correctly built and responds to commands.

This scaffolding is the first step in implementing the full code generation logic within the Librarian system, eventually replacing the Java hermetic code generation system.

The code is largely adapted from the Go language container at https://github.com/googleapis/google-cloud-go/tree/main/internal/librariangen.

Fixes: https://github.com/googleapis/librarian/issues/2500

### DIFF
--- a/internal/librariangen/main_test.go
+++ b/internal/librariangen/main_test.go
@@ -1,0 +1,94 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
+	"testing"
+)
+
+func TestRun(t *testing.T) {
+	ctx := context.Background()
+	tests := []struct {
+		name    string
+		args    []string
+		wantErr bool
+	}{
+		{
+			name:    "no args",
+			args:    []string{},
+			wantErr: true,
+		},
+		{
+			name:    "version flag",
+			args:    []string{"--version"},
+			wantErr: false,
+		},
+		{
+			name:    "flag as command",
+			args:    []string{"--foo"},
+			wantErr: true,
+		},
+		{
+			name:    "unknown command",
+			args:    []string{"foo"},
+			wantErr: true,
+		},
+		{
+			name:    "build command no flags",
+			args:    []string{"build"},
+			wantErr: false,
+		},
+		{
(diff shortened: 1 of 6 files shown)


suztomo (changes requested):
The "internal" folder having only librariangen folder is strange. It's because we are not going to move other _internal_ things (such as "hermetic_build") to the folder.

Shall we remove the "internal" folder?

For example, Python team's Dockerfile is at https://github.com/googleapis/google-cloud-python/tree/main/.generator.

internal/librariangen/main_test.go
```diff
@@ -0,0 +1,94 @@
...
+// limitations under the License.
+
+package main
+
+import (
+	"context"
+	"testing"
+)
+
+func TestRun(t *testing.T) {
```
suztomo, line 22: The argument parsing logic is to be in the shared module, so that each language container doesn't need to implement the argument parsing logic and its tests. https://github.com/googleapis/librarian/pull/2489/files#diff-704975122c152b21ff033c97bc92e9f3dc765402bf278620a3f88df835a9dd43R41

suztomo: So you don't have to write many test cases for argument parsing.

meltsufin:
> The "internal" folder having only librariangen folder is strange. It's because we are not going to move other _internal_ things (such as "hermetic_build") to the folder.
> 
> Shall we remove the "internal" folder?
> 
> For example, Python team's Dockerfile is at https://github.com/googleapis/google-cloud-python/tree/main/.generator.

This is a Go convention for non-exported packages that is enforced by the compiler and I think it's appropriate for our use-case: https://go.dev/doc/go1.4#internalpackages.

internal/librariangen/main_test.go
meltsufin: That's great, but let's integrate that in when it's ready.

suztomo:
@meltsufin Thank you for the info. The "internal" folder or its parent doesn't have go.mod. I'm not sure the "internal" is recognized as a package. Would you be willing to experiment the "internal" folder is indeed following https://go.dev/doc/go1.4#internalpackages ? I think it's a matter of whether a symbol `var Hello = "World"` in `internal/librariangen/main.go` is visible by another program or not.

(In another go program's `go.sum` you would declare `require https://github.com/googleapis/sdk-platform-java/internal/librariangen a407d733df5650041d4bfe01b4dd49f78fe175cf` and try to reference the symbol `librariangen.Hello`)

suztomo (approved):
(Approving this to unblock the work) I still want to know the "internal" experiment above.

PR #3942, opened 2025-10-15
deps: Bump grpc-java to v1.76.0

lqiu96:
This PR bumps the gRPC gen code and runtime versions from `v1.71.0` to `v1.76.0`.
- Notable in this version bump, the V2 gen code unary call API from gRPC now throws a new CheckedException `StatusException` that users will need to catch. This was part of gRPC 1.74.0 (https://github.com/grpc/grpc-java/pull/12126)

## Changes
- Bump the gRPC runtime in the pom to v1.76.0
- Bump the gRPC gen code version in the hermetic build docker image to v1.76.0
- Add new GraalVM configs for some JCTool classes. These classes are pulled in from gRPC -> shaded netty -> shaded JCTools. This came from gRPC-Java upgrading the Netty version to v4.1.124. Netty v4.1.115 uses a new method in JCTools (https://github.com/netty/netty/pull/14426)
- Update the WORKSPACE file to download `bazel_jar_jar` as part of this PR (https://github.com/grpc/grpc-java/pull/12243) in gRPC-Java v1.75.0

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -97,8 +97,8 @@ load("@io_grpc_grpc_java//:repositories.bzl", "grpc_java_repositories")
 
 grpc_java_repositories()
 
-// gRPC-Java uses jar_jar for bazel: https://github.com/grpc/grpc-java/pull/12243
-// The following lines are from jar_jar's README: https://github.com/bazeltools/bazel_jar_jar?tab=readme-ov-file#how-to-add-to-bazel-via-workspace
+# gRPC-Java uses jar_jar for bazel: https://github.com/grpc/grpc-java/pull/12243
+# The following lines are from jar_jar's README: https://github.com/bazeltools/bazel_jar_jar?tab=readme-ov-file#how-to-add-to-bazel-via-workspace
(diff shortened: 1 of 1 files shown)


mpeddada1 (dismissed):

mpeddada1 (dismissed):

mpeddada1 (approved):

WORKSPACE
```diff
@@ -97,9 +97,16 @@ load("@io_grpc_grpc_java//:repositories.bzl", "grpc_java_repositories")
 
 grpc_java_repositories()
 
-load("@envoy_api//bazel:repositories.bzl", "api_dependencies")
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
```
diegomarquezp: nit: would it make sense to add a quick explainer on how grpc needs this and maybe reference https://github.com/grpc/grpc-java/pull/12243?

blakeli0: I think the WORKSPACE file in googleapis needs this change as well, can we follow up on that once this change is released?

blakeli0 (approved):

PR #3952, opened 2025-10-17
feat(librariangen): add generate package

meltsufin:
Based on https://github.com/googleapis/google-cloud-go/tree/main/internal/librariangen/generate with adaptation for Java.

Currently it's just the scaffolding and more work is needed to generate a usable GAPIC library.

The `generate` package contains the core logic for the generation process, including:
- Reading and parsing the `generate-request.json` from librarian.
- Parsing `BUILD.bazel` files in the googleapis repository to extract GAPIC configuration.
- Building and executing `protoc` with the `gapic-generator-java` plugin.
- Unzipping and restructuring the generated files into the final library layout.

A `run-generate-library.sh` script is included for local development and end-to-end testing of the generation process.

Additionally, a `go.work` file has been added to the root of the repository to support the multi-module workspace structure.

### DIFF
--- a/internal/librariangen/generate/generator_test.go
+++ b/internal/librariangen/generate/generator_test.go
@@ -418,12 +418,14 @@ func TestUnzip(t *testing.T) {
 		if err != nil {
 			t.Fatalf("failed to open zip file: %v", err)
 		}
+		defer f.Close() // Ensure file is closed
 		zipWriter := zip.NewWriter(f)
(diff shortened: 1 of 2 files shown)


go.work
```diff
@@ -0,0 +1,3 @@
+go 1.24.7
+
+use ./internal/librariangen
```
suztomo: I don't think you need workspace. Just use "cd internal/librarian".

PR #3970, opened 2025-10-23
chore: add caffeine to third-party-dependencies

agrawal-siddharth:
b/454701734


agrawal-siddharth:
Used Guava instead of Caffeine to build an expiring cache.

PR #3984, opened 2025-11-06
fix: add per-service api_version to gapic_metadata

noahdietz:
Populates the `api_version` field on a per-`service` basis to the generated `gapic_metadata.json` using the value from the `service`-level annotation `google.api.api_version` already parsed for use in the `X-Goog-Api-Version`/`$apiVersion` system parameter (as of #2671).

Updates the `googleapis.commit` used in internal dependency artifact generation to capture change to `gapic_metadata.proto`. Chose the most recent as of writing this. This also has the effect of updating some unit test goldens and unit test fixture dependencies that are based on protos at that commit.

The Showcase `echo.proto` already has the `google.api.api_version` set on `service Echo`, so the `java-showcase` package is appropriately updated to account for inclusion of `apiVersion` in the `gapic_metadata.json` output. It is not in use by any API yet. 

It's presence in `gapic_metadata.json` will allow parsers to correlate `service` client to API version.

Internal bug http://b/452365074.


lqiu96 (approved):
Changes LGTM. @blakeli0 can you also take a quick look?

gapic-generator-java/pom.xml
```diff
@@ -14,7 +14,7 @@
     <clirr.skip>true</clirr.skip>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <!-- Should be in sync with repositories.bzl -->
-    <googleapis.commit>44d6bef0ca6db8bba3fb324c8186e694bcc4829c</googleapis.commit>
+    <googleapis.commit>9fcfbea0aa5b50fa22e190faceb073d74504172b</googleapis.commit>
```
blakeli0: This is probably a tech debt on our end. `gapic_metadata.proto` should be updated automatically along side other "config protos" like [service.proto](https://github.com/googleapis/sdk-platform-java/blob/main/java-common-protos/proto-google-common-protos/src/main/proto/google/api/service.proto).

blakeli0 (approved):

gapic-generator-java/pom.xml
noahdietz: Yep, agreed. I will leave this here for now, but LMK if you want me to file a tracking issue with some details.

gapic-generator-java/src/main/java/com/google/api/generator/gapic/composer/common/AbstractServiceClientClassComposer.java
```diff
@@ -1898,12 +1898,15 @@ private static void updateGapicMetadata(
...
 
-    metadataBuilder =
-        metadataBuilder.putServices(
-            service.name(),
-            GapicMetadata.ServiceForTransport.newBuilder()
-                .putClients("grpc", serviceClientProtoBuilder.build())
-                .build());
+    GapicMetadata.ServiceForTransport.Builder grpcServiceClient =
+        GapicMetadata.ServiceForTransport.newBuilder()
+            .putClients("grpc", serviceClientProtoBuilder.build());
```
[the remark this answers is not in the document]
noahdietz: > do you have any concerns of not having rest in gapic_metadata.json?

ATM, i have no concerns. In my survey of other languages, Java is not the only one. We've gone X years without publishing that metadata with no complaints, so I'm not sure it's all that urgent. I can file an idea bug to capture my findings across languages and at least document that there is a gap in the metadata for `rest`.

PR #3989, opened 2025-11-14
fix: Create a single S2AChannelCredentials per application

rmehta19:


rmehta19:
@rockspore @xmenxk @gtcooke94

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -161,6 +164,8 @@ public final class InstantiatingGrpcChannelProvider implements TransportChannelP
   @Nullable
   private final ApiFunction<ManagedChannelBuilder, ManagedChannelBuilder> channelConfigurator;
 
+  private static volatile ChannelCredentials s2aChannelCredentialsObject;
```
xmenxk: nit: maybe add a comment here that we are using this single credentials object?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -595,43 +606,57 @@ ChannelCredentials createPlaintextToS2AChannelCredentials(String plaintextAddres
...
-      } catch (IOException ignore) {
-        // Fallback to plaintext-to-S2A connection on error.
-        LOG.log(
-            Level.WARNING,
-            "Cannot establish an mTLS connection to S2A due to error creating MTLS to MDS TlsChannelCredentials credentials, falling back to plaintext connection to S2A: "
-                + ignore.getMessage());
-        return createPlaintextToS2AChannelCredentials(plaintextAddress);
+    if (s2aChannelCredentialsObject == null) {
+      synchronized (InstantiatingGrpcChannelProvider.class) {
+        if (s2aChannelCredentialsObject == null) {
```
xmenxk: curious why doing the null check on both line 609 and 611?

xmenxk (commented):
Thanks!

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rmehta19: Thanks for reviewing! This is part of the double checked locking pattern, WDYT?

rockspore (approved):
Thanks for the PR. LGTM.

xmenxk (approved):

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -556,13 +561,19 @@ ChannelCredentials buildS2AChannelCredentials(
...
-      File trustBundle, File privateKey, File certChain) throws IOException {
+      File trustBundle, File privateKey, File certChain)
+      throws IOException, GeneralSecurityException {
     if (trustBundle == null || privateKey == null || certChain == null) {
       return null;
     }
+    AdvancedTlsX509KeyManager keyManager = new AdvancedTlsX509KeyManager();
+    keyManager.updateIdentityCredentials(certChain, privateKey);
+    AdvancedTlsX509TrustManager trustManager = AdvancedTlsX509TrustManager.newBuilder().build();
+    ScheduledExecutorService executor = Executors.newSingleThreadScheduledExecutor();
```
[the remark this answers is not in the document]
ejona86: While the AI isn't smart enough to realize there will only be one executor created, and thus it doesn't benefit from being stored as a static, it is correct that this really needs to be a daemon thread to avoid keeping the JVM alive.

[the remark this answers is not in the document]
rmehta19: Done.

rmehta19:
I split the AdvancedTls changes into https://github.com/googleapis/sdk-platform-java/pull/4004

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -595,43 +599,57 @@ ChannelCredentials createPlaintextToS2AChannelCredentials(String plaintextAddres
...
-            createMtlsToS2AChannelCredentials(rootFile, certKeyFile, certKeyFile);
-      } catch (IOException ignore) {
-        // Fallback to plaintext-to-S2A connection on error.
-        LOG.log(
-            Level.WARNING,
-            "Cannot establish an mTLS connection to S2A due to error creating MTLS to MDS TlsChannelCredentials credentials, falling back to plaintext connection to S2A: "
-                + ignore.getMessage());
-        return createPlaintextToS2AChannelCredentials(plaintextAddress);
+    if (s2aChannelCredentialsObject == null) {
+      synchronized (InstantiatingGrpcChannelProvider.class) {
```
lqiu96: A few questions:

- Why opt to synchronize on on the class here? Can you add that as a comment?
- Can the synchronized scope be to be as narrow as possible? IIRC, `s2aConfigProvider.getConfig()` itself makes a network call. Is this to try and prevent multiple network calls going out simultaneously?

rmehta19: > Why opt to synchronize on on the class here? Can you add that as a comment?

We synchronize on the class here because we want all instances to use a single `s2aChannelCredentialsObject`. If we synchronize on the instance (`this`), I believe we could run into a race condition when initializing `s2aChannelCredentialsObject`: at startup, an application creates two threads which create a channel provider and create a channel, both threads reach this code and see `s2aChannelCredentialsObject` as null and create it.

Added a comment.

> Can the synchronized scope be to be as narrow as possible? IIRC, s2aConfigProvider.getConfig() itself makes a network call. Is this to try and prevent multiple network calls going out simultaneously?

We are more concerned with applications creating a new channel to S2A (and associated resources: thread pools, executors) each time they need to create a new channel which uses S2A for handshaking. (Shared doc internally with you on this) When an application creates a new API client for each request and then destroys it after completing the request (instead of reusing a single API client for all requests), it causes the underlying channel to S2A to be created and destroyed at a high rate. We found in pilot this caused runtime exceptions in certain applications which had high traffic (see shared doc for more details). 

The address and credentials to use for S2A are the same for the lifetime of the application (the credentials can be rotated, which https://github.com/googleapis/sdk-platform-java/pull/4004 takes care of), so this PR improves application resource usage when using S2A by reusing S2A related resources across separate instances of API clients. WDYT?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -161,6 +161,10 @@ public final class InstantiatingGrpcChannelProvider implements TransportChannelP
   @Nullable
   private final ApiFunction<ManagedChannelBuilder, ManagedChannelBuilder> channelConfigurator;
 
+  // This is initialized once for the lifetime of the application. This enables re-using
+  // channels to S2A.
+  private static volatile ChannelCredentials s2aChannelCredentialsObject;
```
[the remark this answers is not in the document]
rmehta19: I think you and I are describing the same thing. One note I mentioned is that the setter should be thread-safe. And, we should expose a reset function to restore the original value that is modified during tests, also thread-safe.

Agreed that it's not ideal to expose setters like this. Happy to implement this if it's preferred over double checked locking. Will update this thread once I have made the changes in a new commit.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -161,6 +161,23 @@ public final class InstantiatingGrpcChannelProvider implements TransportChannelP
...
+
+  /**
+   * Resets the s2aChannelCredentialsObject of the {@link InstantiatingGrpcChannelProvider} class
+   * for testing purposes.
+   *
+   * <p>This should only be called from tests.
+   */
+  @VisibleForTesting
+  public static void resetS2AChannelCredentialsObjectForTests() {
```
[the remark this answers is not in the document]
lqiu96: nit: `resetS2AChannelCredentials()`

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -161,6 +161,23 @@ public final class InstantiatingGrpcChannelProvider implements TransportChannelP
   @Nullable
   private final ApiFunction<ManagedChannelBuilder, ManagedChannelBuilder> channelConfigurator;
 
+  // This is initialized once for the lifetime of the application. This enables re-using
+  // channels to S2A.
+  private static volatile ChannelCredentials s2aChannelCredentialsObject;
```
lqiu96: nit: can this just be called `s2aChannelCredentials`?

gax-java/gax-grpc/src/test/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProviderTest.java
```diff
@@ -1155,6 +1155,7 @@ void createS2ASecuredChannelCredentials_bothS2AAddressesNull_returnsNull() {
             .setS2AConfigProvider(s2aConfigProvider)
             .build();
     assertThat(provider.createS2ASecuredChannelCredentials()).isNull();
+    InstantiatingGrpcChannelProvider.resetS2AChannelCredentials();
```
lqiu96: Can you add a unit test that creates two `InstantiatingGrpcChannelProvider` objects that both use S2A and assert that the s2aChannelCredential object is the same?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
rmehta19: done

lqiu96 (approved):

PR #4027, opened 2025-12-22
chore: remove build badges in README.md template

suztomo:
Removes old badge references.

In the monorepo generation test, I had to disable the README comparison check with google-cloud-java, because google-cloud-java hasn't received the updates from this change yet.

b/468377909


suztomo:
All checks passed.

PR #4030, opened 2025-12-30
fix: provide API to share the same background executor for channel po…

mutianf:
…ol resizing and refresh

Before this change ChannelPool creates a new SingleThreadScheduledExecutor every time getTransportChannel is called. We can reuse the same background executor from the clientSettings and can reduce the number of threads when the user creates many instances of the client.

### DIFF
--- a/gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
+++ b/gax-java/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java
@@ -209,7 +209,8 @@ public abstract class ClientContext {
     // executor can only be set from TransportChannelProvider#withExecutor directly, and a provider
     // will have a default executor if it needs one.
     if (transportChannelProvider.needsExecutor() && settings.getExecutorProvider() != null) {
-      transportChannelProvider = transportChannelProvider.withExecutor(backgroundExecutor);
+      transportChannelProvider =
+          transportChannelProvider.withExecutor(settings.getExecutorProvider().getExecutor());
     }
     if (transportChannelProvider.needsBackgroundExecutor()) {
       transportChannelProvider =


gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
```diff
@@ -173,7 +197,7 @@ public boolean isShutdown() {
         return false;
       }
     }
-    return executor == null || executor.isShutdown();
+    return !shouldShutdownExecutor || backgroundExecutor.isShutdown();
```
igorbernstein2: I dont think you need to account for the executor any more

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
```diff
@@ -216,9 +247,9 @@ public boolean awaitTermination(long timeout, TimeUnit unit) throws InterruptedE
       }
       entry.channel.awaitTermination(awaitTimeNanos, TimeUnit.NANOSECONDS);
     }
-    if (executor != null) {
+    if (shouldShutdownExecutor) {
       long awaitTimeNanos = endTimeNanos - System.nanoTime();
-      executor.awaitTermination(awaitTimeNanos, TimeUnit.NANOSECONDS);
+      backgroundExecutor.awaitTermination(awaitTimeNanos, TimeUnit.NANOSECONDS);
```
igorbernstein2: I dont think you care about awaiting executor. Or if want to continue to include the including the executor state in isShutdown, then you need to do separate accounting:
- when the scheduled tasks start, in a sync block check if the pool is shutting down and exit early, otherwise increment a refcount
- when the scheduled task is complete, decrement the refcount
- in isShutdown, set the shutdown flag and then wait until ref count is zero

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -950,6 +961,19 @@ public Builder setExecutorProvider(ExecutorProvider executorProvider) {
...
+     * executor should be managed by the caller.
+     *
+     * <p>This is optional. The background executor is used for channel refresh and channel resize
+     * on {@link ChannelPool}. This allows us to reuse the same executor for other long running
+     * operations.
+     */
+    public Builder setBackgroundExecutor(ScheduledExecutorService executor) {
+      this.backgroundExecutor = executor;
+      return this;
+    }
```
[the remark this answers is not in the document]
mutianf: It will be set from our client for example cause the channel provider doesn't have access to the background executor

blakeli0:
I think this is generally a good change. However, what are the use cases that customers have to create many instances of clients? 

In general we recommend to use only one client. Creating many clients have other implications, for example, the effort of loading credentials and creating channels would be duplicated. Customers can always pass their own CredentialProvider and ChannelProvider but it may easily misconfigured.

mutianf:
@blakeli0 They have an application that targets multiple bigtable instances with different retry settings.

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/ChannelPool.java
```diff
@@ -153,14 +171,22 @@ Channel getChannel(int affinity) {
   public ManagedChannel shutdown() {
     LOG.fine("Initiating graceful shutdown due to explicit request");
 
+    if (resizeFuture != null) {
+      resizeFuture.cancel(true);
```
igorbernstein2: are you sure you want to interrupt the tasks?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -839,6 +853,11 @@ public ChannelPoolSettings getChannelPoolSettings() {
     return channelPoolSettings;
   }
 
+  /** Gets the background executor for channel refresh and resize. */
+  public ScheduledExecutorService getBackgroundExecutor() {
```
igorbernstein2: can this be package private?

gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java
```diff
@@ -245,6 +247,16 @@ public TransportChannelProvider withExecutor(Executor executor) {
     return toBuilder().setExecutor(executor).build();
   }
 
+  @Override
+  public boolean needsBackgroundExecutor() {
+    return backgroundExecutor == null;
+  }
+
+  @Override
+  public TransportChannelProvider withBackgroundExecutor(ScheduledExecutorService executor) {
```
blakeli0: How do we plan to use it? Do we expect customers to create an `InstantiatingGrpcChannelProvider` with their own executor? 
I see that the same executor is being passed to both executor and backgroundExecutor in ClientContext, if that's the expected use case, we don't have to create new setter methods in TransportChannel.

mutianf: ChannelProvider can have 2 executors: background executor and executor. Executor is used for handling rpc callbakcs. Background executor is used for all other background tasks. In the GrpcChannelProvider case, it doesn't need an executor because Grpc ManagedChannel provides a default executor that's optimized for performance. And the background executor is used for managing channels in the channel pool. Normally we don't want to use the same background executor for callbacks and background tasks because it impacts the performances. The settings you see in ClientContext is a bit confusing and it's from an old fix where we deprecated overriding the default grpc executor on the managed channel: https://github.com/googleapis/sdk-platform-java/commit/95c4c7b4e143bfabf7a1e7fd0fa9703c0ae19ed0#diff-4e2f6e463b9b7d89de68e3f1a87765080045880c8dad018750f269e311f2471f. I don't think we actually go into the if statement by default
```
    if (transportChannelProvider.needsExecutor() && settings.getExecutorProvider() != null) {
      transportChannelProvider = transportChannelProvider.withExecutor(backgroundExecutor);
    }
```
because `settings.getExecutorProvider()` is null. Maybe it'll be easier to understand if the code is `transportChannelProvider.withExecutor(settings.getExecutorProvider().getExecutor());` ?

blakeli0: Awesome, that's much easier to understand. Thanks!

blakeli0 (approved):

PR #4044, opened 2026-01-07
feat: Add Metrics feature flag

diegomarquezp:
This PR introduces a feature flag to control the collection of metrics in GAX. Metrics are now disabled by default and can be enabled via an environment variable or system property.

#### Changes:

*   **Feature Flag Implementation**:
    *   Added `MetricsUtils.java` to handle the feature flag logic. It checks for the `GOOGLE_CLOUD_ENABLE_METRICS` environment variable or system property.
    *   Updated `MetricsTracerFactory.java` to respect the feature flag. If metrics are disabled, `newTracer` now returns a `BaseApiTracer` (no-op) instead of a `MetricsTracer`.
*   **Discovery and Warning**:
    *   Updated `MetricsTracer.java` to log an `INFO` message once per JVM instance if it is instantiated while metrics are disabled. This helps users who have explicitly configured a `MetricsTracerFactory` discover that they also need to enable the global feature flag.
*   **Testing**:
    *   **Unit Tests**: Added `MetricsUtilsTest.java` to verify the feature flag logic. Updated `MetricsTracerFactoryTest.java` to ensure it correctly returns a no-op tracer when metrics are disabled.
    *   **Integration Tests**: Updated `ITOtelMetrics.java` in the showcase module to include a new test case `testMetricsFeatureFlag()`, which verifies the end-to-end behavior of the flag.
*   **Maintenance**:
    *   Updated copyright years in modified files.
    *   Improved test setup in `MetricsTracerFactoryTest.java` using `@ExtendWith(MockitoExtension.class)`.

#### How to enable metrics:
To enable metrics collection, set the following environment variable or system property to true:
`GOOGLE_CLOUD_ENABLE_METRICS=true`


gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsTracerFactory.java
```diff
@@ -66,6 +66,9 @@ public MetricsTracerFactory(MetricsRecorder metricsRecorder, Map<String, String>
 
   @Override
   public ApiTracer newTracer(ApiTracer parent, SpanName spanName, OperationType operationType) {
+    if (!MetricsUtils.isMetricsEnabled()) {
```
westarle: I don't want to change how existing metrics works, the flag is intended to protect *new* metric instrumentation from people taking dependencies on it while it stabilizes.

gax-java/gax/src/main/java/com/google/api/gax/tracing/MetricsUtils.java
```diff
@@ -0,0 +1,59 @@
...
+  private MetricsUtils() {}
+
+  /**
+   * Returns true if metrics are enabled via the GOOGLE_CLOUD_ENABLE_METRICS environment variable or
+   * system property.
+   */
+  public static boolean isMetricsEnabled() {
+    String enableMetrics = System.getProperty(GOOGLE_CLOUD_ENABLE_METRICS);
+    if (enableMetrics == null) {
+      enableMetrics = System.getenv(GOOGLE_CLOUD_ENABLE_METRICS);
```
westarle: Do we need to cache the result, or can we put a note on the method that it should be cached by clients (I'd like to avoid a read to env or properties on every RPC).

PR #4055, opened 2026-01-12
docs(hermetic-build): Document process for updating golden files

diegomarquezp:
Added instructions for updating golden files using the chore/test-hermetic-build branch.


diegomarquezp:
The integration test looks good after following these instructions.

